PR #1, opened 2022-11-26
Github workflow to run code checks and pytest

reuterbal:
Opening this PR to test the workflow and eventually enable Github Actions.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,2 @@
+# When modifying this file, please validate using
+# curl -X POST --data-binary @.codecov.yml https://codecov.io/validate
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -112,7 +112,20 @@ jobs:
           source loki-activate
           pytest --cov=lint_rules/lint_rules --cov-report=xml lint_rules/tests
 
-      - name: Upload coverage reports to Codecov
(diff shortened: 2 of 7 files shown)


reuterbal:
I think this is ready for a look but please hold off merging until ecmwf-ifs/dwarf-p-cloudsc#24 is merged to develop. Then, we can update the reference to CLOUDSC in the workflow configuration.

Overall it does what it says on the tin: run the pylint checks and pytest battery for Python 3.8, 3.9, 3.10 when opening a PR. Pylint violations should be displayed inline in the diff:
Screenshot 2022-11-28 at 22-26-15 Github workflow to run code checks and pytest · ecmwf-ifs_loki@06bad2b

Pylint works now also without the pytest integration, which was often throwing warnings and therefore I have removed that dependency.
An odd failure of F_Front/OMNI occurs with the `test_sourcefile_lazy_construction` test with the frontend segfaulting. I can't reproduce this locally, not even in a similar docker container, which makes it hard to debug. I have allowed the test to gently xfail if the frontend exits with a signal 11 on that particular line. 

Lastly, the base directory path to Loki's source directory was wrong in the CLOUDSC test. This was a leftover from moving the subpackages.

mlange05 (approved):
Looks fantastic, and again many thanks for setting up all the neat tooling! Very much appreciated! GTG once CI comes back green.

PR #2, opened 2022-11-29
Cuda Fortran (CUF) backend implementation including tests

MichaelSt98:
### CUDA Fortran Kernels

* the thread index for each dimension starts at one
* kernel invocation: `kernel<<<grid size, thread block size>>>`
	* `call kernel<<<grid,block[,bytes][,streamid]>>>(arg1,arg2,...)`
	* e.g. using the `dim3` derived type: `type(dim3) =:: blocks, threads`
		* `blocks = dim3(n/256, n/16, 1)`, `threads = dim3(16, 16, 1)`  
* device subprograms are not allowed to have the `SAVE` attribute or with data initialization

| subroutine/function qualifier | description |
| ----------------------------- | ----------- |
| `attributes(host)` | to be executed on the host |
| `attributes(global)` | to be executed on the device, called using a kernel call | 
| `attributes(device)` | to be executed on the device, called from a subprogram with `global` or `device` attribute | 
| `attributes(host,device)` | either `host` or `device` | 
| `attributes(grid_global)` | like `global`, but threads within the grid group are guaranteed to be co-resident on the devic, allowing a grid synchronization (on cc70 hardware and greater) | 

* predefind variables (`cudafor`) in device subprograms

| variable | description |
| -------- | ----------- |
| `threadidx`, `threadidx%x` and/or `threadidx%z` | thread index within its thread block (for one or two dimensional thread blocks) |
| `blockdim` | dimension of the thread block, same value for all thread blocks in the same grid |
| `blockidx`, `blockidx%x` and/or `blockidx%z`| block index within the grid |
| `griddim` | dimension of the grid |
| `warpssize` | constant integer, threads are executed in groups of `warpsize` (currently 32) |

```fortran
! Kernel definition
attributes(global) subroutine ksaxpy( n, a, x, y )
    real, dimension(*) :: x,y
    real, value :: a
    integer, value :: n, i
    i = (blockidx%x-1) * blockdim%x + threadidx%x
    if( i <= n ) y(i) = a * x(i) + y(i)
end subroutine

! Host subroutine
subroutine solve( n, a, x, y )
    real, device, dimension(*) :: x, y
    real :: a
    integer :: n
    ! call the kernel
    call ksaxpy<<<n/64, 64>>>( n, a, x, y )
end subroutine
```

#### Kernel Loop directive

* automatic kernel generation and invocation from a region of host code containing one or more tightly nested loops
	* `!$cuf kernel do[(n)] <<< grid, block  [optional stream] >>>`
	* or `!$cuf kernel do[(n)] <<< grid, block, 0, streamid >>>`
	* or `!$cuf kernel do[(n)] <<< grid, block, stream=streamid >>>` 
* scalar reduction operation, such as summing the values in a vector or matrix, are handled by the compiler, e.g. generation of the final reduction kernel, inserting synchronization into the kernel as appropriate

e.g. (grid shape is computed by the compiler)

```fortran
!$cuf kernel do(2) <<< (*,*), (32,4) >>>
do j = 1, m
  do i = 1, n
    a(i,j) = b(i,j) + c(i,j)
  end do
end do
```

e.g. (automatic) reduction

```fortran
rsum = 0.0
!$cuf kernel do <<< *, * >>>
do i = 1, n
    rsum = rsum + x(i)* y(i)
end do
```

e.g. explicit reduction

* allowed reductions
	* Fortran integer type: `+`, `*`, `max`, `min`, `iand`, `ior`, and `ieor` 
	* Fortran real type: `+`, `*`, `max`, `min`
	* Fortran complex type: `+` 
	* Fortran logical type: `.and.`, `.or.`

```fortran
value = 0.0
!$cuf kernel do <<< *, * >>> reduce(+:value)
do i = 1, n
    a(i) = real(int(a(i) * 100.0 - 50.0),kind=4)
    if (a(i) .ge. 0.0) then
        value = value + a(i)
    else
        value = value + a(i) + 50.0
    end if
end do
```

* the directive specifying `n` must be followed by at least that many tightly-nested `DO` loops
* tighly-nested `DO` loops must have invariant loop limits
* the invariant loop limits cannot be a value from an array expression, unless those arrays have the managed attribute
* no `GOTO` or `EXIT` statements within or between any loops that have been mapped onto the grid and block configuration values
	*  body of the loops may contain assignment statements, IF statements, loops, and GOTO statements
* only CUDA Fortran data types allowed 
* device-specific intrinsics such as the syncthreads and other warp or block-level cooperating, syncing, or barrier functions should be avoided except in very limited situations
* subroutine and function calls to attributes(device) subprograms are allowed if they are in the same module as the code containing the directive
* implicit loops and F90 array syntax are not allowed within the directive loops

### Variable Qualifiers

* by default all variables are alllocated in the host main memory

| qualifier | description |
| --------- | ----------- |
| `attributes(device)` | allocated in the device global memory | 
| `attributes(managed)` | may be used in both host and device code |
| `attributes(constant)` | device constant variable, allocated in the device constant memory (modifiable by the host but not by the device) |
| `attributes(shared)` | may not be data initialized, allocated in the device shared memory for a thread block, a write in one thread only guaranteed to be visible by another thread in the thread block after `SYNCTHREADS()` |
| `attributes(pinned)` | a pinned variable must be an allocatable array, when allocated it will be allocated in host pagelocked memory allowing for faster copies |
| `attributes(texture)` | a texture variable must be an F90 pointer (real or integer), read-only |

* members of derived type may not have the device attribute unless they are allocatable
* device variables
	* device variables and arrays may be passed as actual argumnts to host and device subprograms
		* subprogram interface must be explicit and the matching dummy argument must also have the device attribute
	* device variables and arrays declareed in a host subprogram cannot have the `save` attribute unless they are allocatable 
* shared variables
	* shared data may not have the `pointer`, `target` or `allocatable` attributes
	* members of derived type may not have the shared attribute
	* `real, shared :: x(*)` is allowed but memory for this assumed-siye shared memory needs to be defined by the kernel invocation
* dummy arguments (like in Fortran) are passed by default by reference, therefor the actual argument must be stored in device global memory
	* scalar arguments can be passed by value adding the `value` argument 
	
### F90 pointers

* (recently added support for) F90 pointers that point to device data
	* limited to pointer declared at module scope
	* accessed through module assocation, or can be passed in to global subroutines
	* `associated()` function is also supported in device code  

```fortran
module devptr
! currently, pointer declarations must be in a module
  real, device, pointer, dimension(:) :: mod_dev_ptr
  real, device, pointer, dimension(:) :: arg_dev_ptr
  real, device, target,  dimension(4) :: mod_dev_arr
  real, device, dimension(4) :: mod_res_arr
contains
  attributes(global) subroutine test(arg_ptr)
    real, device, pointer, dimension(:) :: arg_ptr
    ! copy 4 elements from one of two spots
    if (associated(arg_ptr)) then
      mod_res_arr = arg_ptr
    else
      mod_res_arr = mod_dev_ptr
    end if
  end subroutine test
end module devptr
```

## Compilation

* CUDA Fortran is enabled in compilation specifying `-cuda` or renaming the file to have a `.cuf` or `.CUF` extension
* target architecture e.g. `-Mcuda=cc35`
* specifying emulation mode: `-Mcuda=emu`
* target CUDA toolkit version e.g. `-Mcuda=cuda5.0`
* `-Mcuda=fastmath` enables faster intrinsics
* `-Mcuda-nofma` disables fused multiply-add
* `-Mcuda=maxregcount:<n>` limits registr usage per thread
* `-Mcuda=ptxinfo` prints memory usage per kernel
* `-Mcuda=rdc` (relocatable device code) allows to call device functions between modules

### Conditional compilation

* pre-processing is enabled by either the `-Mpreprocess` option or by using capital letters in the filename extension
	* then `_CUDA` macro is defined

e.g. 

```fortran
program p
!@cuf use cudafor
real a(1000)
!@cuf attributes(managed) :: a
real b(1000)
!@cuf real, device :: b_dev(1000)
b = 2.0
!@cuf b_dev = b
!@cuf associate(b=>b_dev)
!$cuf kernel do(1) <<<*,*>>>
do i = 1, 1000
    a(i) = real(i) * b(i)
end do
!@cuf end associate
#ifdef _CUDA
print *,"GPU sum passed? ",sum(a).eq.1000*1001
#else
print *,"CPU sum passed? ",sum(a).eq.1000*1001
#endif
end program
``` 

## Device memory

* dynamically allocated in host buprograms using the `allocate` statement, if it does not have the `save` attribute, it will be automatically deallocated when the subprogram returns
	* also applicable to scalars

```fortran
real, allocatable, device :: b(:)
allocate(b(5024),stat=istat)
...
if(allocated(b)) deallocate(b)
```

* automatic and local arrays, having the lifetime of the subprogram (as usual)

```fortran
subroutine vfunc(a,c,n)
    real, device :: adev(n)
    real, device :: atmp(4)
    ...
end subroutine vfunc   ! adev and atmp are deallocated
```

* allocating device memory using runtime routines (C-like behaviour)

```fortran
real, allocatable, device :: v(:)
istat = cudaMalloc(v, 100)
...
istat = cudaFree(v)
```

* using `NV_CUDAFOR_DEVICE_IS_MANAGED=1` all device data allocation will be managed meory instead of device memory (potentially without coding changes)

### Memory transfer 

* copy H2D (and vice-versa) is possible through simple assigment statements in the host program (using CUDA stream zero)
	* unless `cudaforSetDefaultStream` was used to associate one or more device and managed variables to a particular stream 
* an assignment with device variables on both sides copies the ata between two device variables or arrays
* elemental transfers are supported (but perform poorly)
* array slices are also supported, whereas the performance is dependent on the size of the slice, the amount of contiguous data in the slices and the implementation

further following expressions are legal:

```fortran
a = adev
adev = a
b = a + adev
c = x * adev + b
```

## Synchronization

* `syncthreads` - barrier synchronization for all threads in a single thread block
* `syncthreads_count` - additionally evaluates the integer argument `int_value` for all threads of the block and returns the number of threads for which `int_value` evaluates to non-zero
* `syncthreads_and(int_value)` - additionally evaluates the integer argument `int_value` for all threads of the block and returns non-zero if and only if `int_value` evaluates to non-zero for all of them
* `synthreads_or(int_value)` - additionally evaluates the integer argument `int_value` for all threads of the block and returns non-zero if and only if `int_value` evaluates to non-zero for any of them
* `syncwarp` - cause all executing threads within a warp, and specified in the mask argument, to reach a barrier
* `threadfence` - all global and ahared memory accesses by the calling thread prior to `threadfence()` are visible to:
	* all the threads in the thread block for shared memory accesses
	* all the threads in the device for global memory accesses 
* `threadfence_block` - visible to all the threads in the thread block 
* `threadfence_system` 

* warp-vote operations:
	* `allthreads`: `if( allthreads(a(i)<0.0) ) allneg = .true.`
	* `anythread`: `if( anythread(a(i)<0.0) ) allneg = .true.`
	* `ballot(int_value)`: evaluates the argument `int_value` for all threads of the warp and returns an integer whose Nth bit is set if and only if `int_value` evaluates to non-zero for the Nth thread of the warp
	* `activemask` - returns a 32-bit integer mask of all the curently active threads
	* ...

## Device intrinsics

* Fortran intrinsics allowed in device subprograms
	* `abs`, `aimag`, `aint`, `anint`, `ceiling`, `cmplx`, `conjg`, `dim`, `floor`, `int`, `logical`, `max`, `min`, `mod`, `modulo`, `nint`, `real`, `sign`	* ...
* CUF additions by NVIDIA:
	* ...

## Runtime APIs

* device management:
	* `cudaChooseDevice`
	* `cudaDeviceGetAttribute`
	* `cudaDeviceGetCacheConfig`
	* `cudaDeviceGetLimit`
	* `cudaSetDevice`
	* ...

### DIFF
--- a/loki/backend/cufgen.py
+++ b/loki/backend/cufgen.py
@@ -1,3 +1,10 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
 from loki.backend.fgen import FortranCodegen
 
 __all__ = ['cufgen', 'CudaFortranCodegen']
@@ -5,11 +12,8 @@ __all__ = ['cufgen', 'CudaFortranCodegen']
 
 class CudaFortranCodegen(FortranCodegen):
     """
-    Tree visitor that extends `FortranCodegen` with Cuda Fortran (CUF) language variations.
+    Tree visitor that extends :any:`FortranCodegen` with Cuda Fortran (CUF) language variations.
     """
-
-    def __init__(self, depth=0, indent='  ', linewidth=90, conservative=True):
-        super().__init__(depth=depth, indent=indent, linewidth=linewidth, conservative=conservative)
 
     def visit_CallStatement(self, o, **kwargs):
         """
@@ -26,6 +30,8 @@ class CudaFortranCodegen(FortranCodegen):
             chevron = f"<<<{','.join([str(elem) for elem in o.chevron])}>>>"
         else:
             chevron = ""
+        if o.kwarguments:
+            args += tuple(f'{self.visit(arg[0], **kwargs)}={self.visit(arg[1], **kwargs)}' for arg in o.kwarguments)
         call = self.format_line('CALL ', name, chevron, '(', self.join_items(args), ')')
         return self.join_lines(pragma, call)
 
@@ -55,6 +61,21 @@ class CudaFortranCodegen(FortranCodegen):
 
 def cufgen(ir, depth=0, conservative=False, linewidth=132):
     """
-    Generate standardized Fortran code from one or many IR objects/trees.
+    Generate CUDA Fortran code from one or many IR objects/trees.
+
+    Implemented by extending the :class:`FortranCodegen` to support
+    CUDA Fortran specific syntax. Refer to the CUDA_FORTRAN_PROGRAMMING_GUIDE_ for more information.
+
+    Supported subset of the CUDA Fortran specifications:
+
+    * variable qualifiers e.g. ``attributes(device)``
+    * chevron syntax for to launch kernels e.g. ``call kernel<<<grid,block[,bytes][,streamid]>>>(arg1,arg2,...)``
+
+    Natively supported (via :class:`FortranCodegen`):
+
+    * subroutine/function qualifiers e.g. ``attributes(global)`` via :py:attr:`loki.Subroutine.prefix`
+    * kernel loop directives via :class:`loki.ir.Pragma`
+
+    .. _CUDA_FORTRAN_PROGRAMMING_GUIDE: https://docs.nvidia.com/hpc-sdk/compilers/cuda-fortran-prog-guide/index.html
     """
     return CudaFortranCodegen(depth=depth, linewidth=linewidth, conservative=conservative).visit(ir)
--- a/loki/backend/cufgen.py
+++ b/loki/backend/cufgen.py
@@ -1,0 +1,60 @@
+from loki.backend.fgen import FortranCodegen
+
+__all__ = ['cufgen', 'CudaFortranCodegen']
+
+
+class CudaFortranCodegen(FortranCodegen):
+    """
+    Tree visitor that extends `FortranCodegen` with Cuda Fortran (CUF) language variations.
+    """
+
+    def __init__(self, depth=0, indent='  ', linewidth=90, conservative=True):
+        super().__init__(depth=depth, indent=indent, linewidth=linewidth, conservative=conservative)
+
+    def visit_CallStatement(self, o, **kwargs):
+        """
+        Format call statement as
+          CALL(<chevron>) <name>(<args>)
+          with the chevron as launch configuration for device offloading,
+          resulting in something like
+          call kernel<<<grid,block[,bytes][,streamid]>>>(arg1,arg2,...)
+        """
+        pragma = self.visit(o.pragma, **kwargs)
+        name = self.visit(o.name, **kwargs)
+        args = self.visit_all(o.arguments, **kwargs)
+        if o.chevron is not None:
+            chevron = f"<<<{','.join([str(elem) for elem in o.chevron])}>>>"
+        else:
+            chevron = ""
+        call = self.format_line('CALL ', name, chevron, '(', self.join_items(args), ')')
+        return self.join_lines(pragma, call)
+
+    def visit_SymbolAttributes(self, o, **kwargs):
+        """
+        Format declaration attributes as
+          <typename>[(<spec>)] [, <attributes>]
+        """
+        attr_str = super().visit_SymbolAttributes(o, **kwargs)
+        attributes = []
+
+        attr_dic = {
+            "device": "DEVICE",
+            "managed": "MANAGED",
+            "constant": "CONSTANT",
+            "shared": "SHARED",
+            "pinned": "PINNED",
+            "texture": "TEXTURE"
+                    }
+
+        for key, value in attr_dic.items():
+            if getattr(o, key):
+                attributes += [value]
+
+        return self.join_items([attr_str] + attributes)
+
+
+def cufgen(ir, depth=0, conservative=False, linewidth=132):
+    """
+    Generate standardized Fortran code from one or many IR objects/trees.
+    """
+    return CudaFortranCodegen(depth=depth, linewidth=linewidth, conservative=conservative).visit(ir)
--- a/loki/expression/symbols.py
+++ b/loki/expression/symbols.py
@@ -573,6 +573,13 @@ class MetaSymbol(StrCompareMixin, pmbl.AlgebraicLeaf):
         """
         return self.symbol.type
 
+    @type.setter
+    def type(self, _type):
+        """
+        Update the :any:`SymbolAttributes` declared for this symbol
+        """
+        self.symbol.type = _type
+
     @property
     def variables(self):
         """
--- a/tests/test_cufgen.py
+++ b/tests/test_cufgen.py
@@ -1,3 +1,10 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
 import pytest
 
 from conftest import available_frontends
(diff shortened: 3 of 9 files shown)


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf-ifs/loki?pullRequest=2) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf-ifs/loki?pullRequest=2) before we can accept your contribution.<br/><hr/>**Michael Staneker** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf-ifs/loki?pullRequest=2) it.</sub>

reuterbal:
Many thanks. Two quick remarks before I take a proper look:
- You may need to add your ECMWF email address (assuming it is the one your commits are referencing) to your Github profile so the CLA check succeeds. Github allows to add additional email adresses, so no need to replace your current one.
- Can you please rebase the branch on top of `main` to include the test configuration? This should then trigger the test workflow. (Same for the other PR, please)

loki/backend/cufgen.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Format call statement as
+          CALL(<chevron>) <name>(<args>)
+          with the chevron as launch configuration for device offloading,
+          resulting in something like
+          call kernel<<<grid,block[,bytes][,streamid]>>>(arg1,arg2,...)
+        """
+        pragma = self.visit(o.pragma, **kwargs)
+        name = self.visit(o.name, **kwargs)
+        args = self.visit_all(o.arguments, **kwargs)
```
reuterbal, line 24: Does CUF not allow keyword arguments?

If it does allow them, these should be printed, too. Otherwise, I suggest to add a `assert not o.kwarguments`.

tests/test_cufgen.py
```diff
@@ -0,0 +1,117 @@
...
+        elif "kernel" in routine.name:
+            kernels.append(routine)
+        elif "device" in routine.name:
+            device_subroutines.append(routine)
+
+    assert driver
+    assert module.to_fortran(cuf=True) == module.to_fortran()
+
+    decl_map = {}
+    for decl in FindNodes(ir.VariableDeclaration).visit(driver.spec):
```
reuterbal, line 67: To make you aware of a simpler way of doing this: You can simply edit the variable types directly:
```
for var in driver.variables:
    if "device" in var.name:
        var.type = var.type.clone(device=True)
    if "managed" in var.name:
        ...
```

tests/test_cufgen.py
```diff
@@ -0,0 +1,117 @@
...
+        if "pinned" in decl.symbols[0].name:
+            decl_map[decl] = decl.clone(symbols=(decl.symbols[0].clone(type=decl.symbols[0].type.clone(pinned=True)),))
+        if "texture" in decl.symbols[0].name:
+            decl_map[decl] = decl.clone(symbols=(decl.symbols[0].clone(type=decl.symbols[0].type.clone(texture=True)),))
+    driver.spec = Transformer(decl_map).visit(driver.spec)
+
+    call_map = {}
+    for call in FindNodes(ir.CallStatement).visit(driver.body):
+        if "kernel" in str(call.name):
+            with pytest.raises(Exception):
```
reuterbal, line 86: Can we use a narrower specification than `Exception`?

tests/test_cufgen.py
```diff
@@ -0,0 +1,117 @@
...
+
+end module transformation_module_cufgen
+"""
+
+    module = Module.from_source(fcode, frontend=frontend)
+
+    driver = None
+    kernels = []
+    device_subroutines = []
+    for routine in module.routines:
```
reuterbal, line 55: You can extract routines simply by name from modules:
```
driver = module['driver']
kernel = module['kernel']
```

tests/test_cufgen.py
```diff
@@ -0,0 +1,117 @@
...
+
+    driver.body = Transformer(call_map).visit(driver.body)
+
+    for kernel in kernels:
+        kernel.prefix = ("ATTRIBUTES(GLOBAL)",)
+
+    for device_subroutine in device_subroutines:
+        device_subroutine.prefix = ("ATTRIBUTES(DEVICE)",)
+
+    cuf_str = module.to_fortran(cuf=True)
```
reuterbal, line 105: A suggestion to make the string checks a bit more targeted and to increase test coverage.

- Generate per-subroutine strings to check for the declaration attributes, or use `fgen` directly on declaration nodes
- Read the `fcode` as a `Sourcefile` and write out a temporary CUF file in the end and compare strings to that file content

But that's optional, really

loki/backend/cufgen.py
```diff
@@ -0,0 +1,60 @@
...
+        for key, value in attr_dic.items():
+            if getattr(o, key):
+                attributes += [value]
+
+        return self.join_items([attr_str] + attributes)
+
+
+def cufgen(ir, depth=0, conservative=False, linewidth=132):
+    """
+    Generate standardized Fortran code from one or many IR objects/trees.
```
reuterbal, line 58: Please update the dosctring, it's not standard Fortran ;-)

Also, since we're only implementing a subset of the CUF specification, it would be great if you provided a little more detail which features are supported and what properties are used for that (I know that the other backends aren't prime examples there).

reuterbal (changes requested):
This looks fantastic, many thanks! Implementation is neat and clean, not much to add there. My comments are either just formal stuff or mostly relate to some aspects of the tests simply to make you aware of a few simpler ways to do certain things.

Down the line I would hope to get a bit more testing into place for the CUF backend but there is not much you can do without running code. We'll have to do this via regression tests on a dwarf later on.

Our backends aren't particularly well explained in the documentation, but I think we should improve on that. For that reason, please provide a comprehensive docstring on `cufgen`. Also mention the method in the documentation and add it to the list: https://github.com/ecmwf-ifs/loki/blame/main/docs/source/backends.rst#L41

tests/test_cufgen.py
MichaelSt98: Yes I tried this ... 
However, I don't know why, but `var.type = var.type.clone(device=True)` does **not** work *AttributeError: can't set attribute 'type'*.

```py
for var in driver.variables:
    if "device" in var.name:
        var.clone(type=var.type.clone(device=True))
```

works, but probably shouldn't work.

reuterbal: Then there's probably a `type` setter missing on one of the variable classes. What type does `var` have in that context.
The second works but is unnecessary - it uses the same mechanism of updating the symbol table under the hood.

MichaelSt98: Yes, the `MetaSymbol` class is lacking this/those setter(s) (in general).

Adding

```py
    @type.setter
    def type(self, _type):
        if self.symbol.scope is None:
            # Store locally if not attached to a scope
            self.symbol.type = _type
        elif _type is None:
            # Store deferred type if unknown
            self.symbol.scope.symbol_attrs[self.symbol.name] = SymbolAttributes(BasicType.DEFERRED)
        elif _type is not self.symbol.scope.symbol_attrs.lookup(self.symbol.name):
            # Update type if it differs from stored type
            self.symbol.scope.symbol_attrs[self.symbol.name] = _type
```

to the `MetaSymbol` class solves the issue. However, all other properties in `MetaSymbol` do not have setters as well.

reuterbal: Yes, IR and expression nodes have intentionally been kept immutable. We may consider relaxing this gradually but it easily risks ending up in an inconsistent state.
However, `type` property is somewhat separate to that because it modifies the symbol table of the enclosing scope rather than the node itself, which is why this has been added recently. `MetaSymbol` seems to have been forgotten there.
However, there's no need to re-implement the logic, simply use the setter on `self.symbol`:
```python
@type.setter
def type(self, _type):
    self.symbol.type = _type
```

reuterbal (changes requested):
Thank you very much for making the changes, this looks really fantastic now!
A small nitpick about the type setter to avoid code duplication. Everything to do with the type system is a bit brittle, so I'm slightly pedantic about that one - sorry ;-)

loki/expression/symbols.py
```diff
@@ -573,6 +573,24 @@ def type(self):
         """
         return self.symbol.type
 
+    @type.setter
+    def type(self, _type):
+        """
+        Update the :any:`SymbolAttributes` declared for this symbol
```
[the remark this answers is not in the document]
MichaelSt98: Yes, sorry, thought it was within the last commit ...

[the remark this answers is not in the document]
MichaelSt98: Yes, sorry, thought it was within the last commit ...

reuterbal (approved):
Looks great now, many thanks again. Fantastic work.

mlange05 (approved):
A very neat and clean implementation, and a fantastic first contribution. Since @reuterbal picked out the last minor details, I have nothing further to add. Very nice work, and many thanks to both!

PR #3, opened 2022-11-29
Generic hoist functionality via two transformation

MichaelSt98:
Implemented:

* `HoistAnalysis` class 
    * in a derived class: overload `find_variables(routine)` to influence which variables to be hoisted
* `HoistSynthesis` class
    * in a derived class: overload `driver_variable_declaration(routine, var)` to influence the actual variable declaration (and potentially allocation, ...)

so that the actual hoisting can be accomplished via

```python
# Analysis: find variables to be hoisted
scheduler.process(transformation=HoistAnalysis(), reverse=True)
# Synthesis: hoist found variables
scheduler.process(transformation=HoistSynthesis())
```


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf-ifs/loki?pullRequest=3) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf-ifs/loki?pullRequest=3) before we can accept your contribution.<br/><hr/>**Michael Staneker** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf-ifs/loki?pullRequest=3) it.</sub>

loki/transform/hoist.py
```diff
@@ -0,0 +1,79 @@
...
+        role = kwargs.get('role', None)
+        item = kwargs.get('item', None)
+        successors = kwargs.get('successors', None)
+        successor_map = {successor.routine.name: successor for successor in successors}
+
+        if item and not item.local_name == routine.name.lower():
+            return
+
+        if role == 'driver':
+            for var in item.user_data["to_hoist"]:
```
reuterbal, line 63: Two things here:
- I think it would be sufficient to simply append the variables to the driver's variable list. Declarations will be created automatically
- I would explicitly rescope the variables here to make sure you don't run into issues

This could look like that (untested!):
```
routine.variables += tuple(v.rescope(routine) for v in item.user_data['to_hoist'])
```

loki/transform/hoist.py
```diff
@@ -0,0 +1,79 @@
...
+                self.driver_variable_declaration(routine, var)
+        else:
+            routine.arguments += ir.as_tuple([var.clone(type=var.type.clone(intent='inout', allocatable=None))
```
reuterbal, line 66: `ir` shouldn't export the `as_tuple` symbol. Please import it from the utils module. Surprised this works!

Also, the square brackets are redundant if you directly construct a tuple as `tuple(var.clone(...) for ... )`
Almost sure that line would fail, which suggests we might lack test coverage here ;-)

Otherwise same comment as for the `if` branch about rescoping.

loki/transform/hoist.py
```diff
@@ -0,0 +1,79 @@
...
+                                           for var in item.user_data["to_hoist"]])
+
+        call_map = {}
+        for call in FindNodes(ir.CallStatement).visit(routine.body):
+            new_args = [arg.clone(dimensions=None) for arg in successor_map[call.name].user_data["hoist_variables"]]
```
reuterbal, line 71: Rescoping might be automatic here but I would still do it explicitly by adding `scope=routine` to the clone

tests/sources/projHoist/module/driver_mod.f90
```diff
@@ -0,0 +1,68 @@
...
+    integer, parameter :: len = 3
+contains
+
+    subroutine driver(a, b, c)
+        integer, intent(in) :: a
+        integer, intent(inout) :: b(a)
+        integer, intent(inout) :: c(a, len)
+        real :: x, y
+        call kernel1(a, b, c)
+        !    kernel1(a, b, c, kernel1_x, kernel1_y, kernel1_k1_tmp)
```
reuterbal, line 12: Is this the expected call signature after the transformation?
I would prefer not having this as a comment in the Fortran but rather as a string-comparison in the test to validate also the expected order of the additional arguments, e.g.:
```
assert call.arguments == ('a', 'b', 'c', 'kernel1_x', 'kernel1_y', 'kernel1_k1_tmp')
```

You can rely here on the automatic comparison against strings, which allows tricks like that:
```
>>> [Variable(name='a'), Variable(name='b')] == ['a', 'b']
True
```

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
+"""
```
reuterbal, line 1: Coyright header please

tests/sources/projHoist/module/driver_mod.f90
```diff
@@ -0,0 +1,68 @@
...
+    subroutine kernel1(a, b, c)
+        integer, intent(in) :: a
+        integer, intent(inout) :: b(a)
+        integer, intent(inout) :: c(a, len)
+        real :: x(a)
+        integer :: y(a, len)
+        real :: k1_tmp(a, len)
+        y = 11
+        c = y
+        ! call device2(x)
```
reuterbal, line 29: Should this be active?

tests/sources/projHoist/module/driver_mod.f90
```diff
@@ -0,0 +1,68 @@
...
+        !    device2(a, b, x, device2_z, device2_d2_tmp)
+    end subroutine device1
+
+    !          device2 (a, b, x, z, d2_tmp)
+    subroutine device2(a, b, x)
+        integer, intent(in) :: a
+        integer, intent(inout) :: b(a)
+        real, intent(inout) :: x(a)
+        integer z(a)
+        integer :: d2_tmp(len)
```
reuterbal, line 63: For my own understanding: Are we intentionally hoisting an unused temporary here?

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+def test_hoist_find_arrays(here, frontend, config):
+    """
+    Testing hoist functionality with modified *Analysis* part using a `find_variables` method only searching for arrays.
+    """
+
+    class HoistAnalysisFindArrays(HoistAnalysis):
+
+        @staticmethod
+        def find_variables(routine):
+            return [var for var in routine.variables if var not in routine.arguments and isinstance(var, sym.Array)]
```
reuterbal, line 112: This should probably also exclude parameters

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+        elif "device1" in item.name:
+            assert len(item.routine.arguments) == 8
+        elif "device2" in item.name:
+            assert len(item.routine.arguments) == 5
+
+    compile_and_test(scheduler=scheduler, here=here, a=(5, 10, 100))
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_hoist_find_arrays(here, frontend, config):
```
reuterbal, line 103: Genuine question what the right pattern is that we would want for that:
Should this specialization method be done via derived classes that overwrite the `find_variables` method, or should the criteria be formulated as a callback, similar to the [`ExpressionFinder`](https://github.com/ecmwf-ifs/loki/blob/main/loki/expression/expr_visitors.py#L61) visitor?

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+            assert len(item.routine.arguments) == 5
+
+    compile_and_test(scheduler=scheduler, here=here, a=(5, 10, 100))
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_hoist_find_variables(here, frontend, config):
+    """
+    Testing hoist functionality with modified *Analysis* part using a `find_variables` method only searching for arrays
+    with variable `a` within the shape/dimensions.
```
reuterbal, line 156: FYI: the RST markup that Sphinx uses requires double ticks for fixed width formatting, e.g. ` ``a`` `, ` ``find_variables`` `. It's irrelevant for tests but something to be aware of in general.

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+        {
+            'name': 'driver',
+            'role': 'driver',
+            'expand': True,
+        },
+    ]
+
+    scheduler = Scheduler(paths=[proj], config=config, seed_routines=['driver'], frontend=frontend)
+
+    # compile_and_test(scheduler=scheduler, here=here, a=(5, 10, 100))
```
reuterbal, line 126: Enable or remove, please

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
```
reuterbal, line 178: Enable or remove, please

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
```
reuterbal, line 241: Enable or remove, please

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+        @staticmethod
+        def find_variables(routine):
+            return [var for var in routine.variables if var not in routine.arguments and isinstance(var, sym.Array)
+                    and "a" in FindVariables().visit(var.dimensions)]
+
+    class HoistSynthesisAllocatable(HoistSynthesis):
+
+        @staticmethod
+        def driver_variable_declaration(routine, var):
+            dimensions = [sym.IntrinsicLiteral(":")] * (len(var.dimensions))
```
reuterbal, line 223: `:` is `sym.RangeIndex((None, None))`

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+            routine.spec.append(ir.VariableDeclaration((var.clone(dimensions=as_tuple(dimensions),
+                                                                  type=var.type.clone(allocatable=True)),)))
+            routine.body.prepend(ir.Allocation((var.clone(),)))
```
reuterbal, line 226: That's where it gets a bit hacky... Wouldn't it make more sense to provide these variations as "modes" of the hoist transformation?

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+import numpy as np
+
+from conftest import available_frontends, jit_compile, clean_test
+from loki import FindVariables
+from loki.expression import symbols as sym
+from loki import ir, as_tuple, is_iterable
+from loki import Scheduler
+from loki.transform import HoistAnalysis, HoistSynthesis
+
+write_transformation_output = False
```
reuterbal, line 15: What's the purpose of this? If it's not going to be used, I would suggest removing it rather than leaving dead code in the tests.

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+    scheduler = Scheduler(paths=[proj], config=config, seed_routines=['driver'], frontend=frontend)
+
+    # compile_and_test(scheduler=scheduler, here=here, a=(5, 10, 100))
+
+    # Analysis
+    scheduler.process(transformation=HoistAnalysisFindVariables(), reverse=True)
+    # Synthesis
+    scheduler.process(transformation=HoistSynthesisAllocatable())
+
+    for item in scheduler.items:
```
reuterbal, line 248: Can you add here some tests that verify the attributes of the hoisted variables and the other code changes? For example, that the allocation nodes exist?

Otherwise it might just have been working by accident and not really testing the actual intrinsics of that transformation mode.

tests/test_hoist.py
```diff
@@ -0,0 +1,264 @@
...
+            'mode': 'idem',
+            'role': 'kernel',
+            'expand': True,
+            'strict': True,
+        },
+        'routines': []
+    }
+
+
+def compile_and_test(scheduler, here, a=(5,)):
```
reuterbal, line 39: I would suggesting here an additional argument with something like `test_name` that is injected into the file names. That avoids the risk that leftover files from a different test are used.

reuterbal (changes requested):
This is very cool, many thanks! I was wondering about a mechanism to forward information from analysis to transformation and that's a nice solution.

I'm not 100% happy with the naming of this metadata data structure, but that's something @mlange05 might also have a suggestion for.

The generic hoisting mechanism is very neat and a really great showcase of that scheduler mechanism. 

What I'm not so happy about is the mode-switching for the analysis and transformation pass. I think these different variations should be available to choose from in the transformation modules. Right now it feels like implementations are done and tested inside the testbase, which isn't really adding value in my eyes.

MichaelSt98:
@reuterbal 
Referring to

> What I'm not so happy about is the mode-switching for the analysis and transformation pass. I think these different variations should be available to choose from in the transformation modules. Right now it feels like implementations are done and tested inside the testbase, which isn't really adding value in my eyes.

and 

> Should this not be limited to arrays? Or are we hoisting also scalar variables?

and 

> That's where it gets a bit hacky... Wouldn't it make more sense to provide these variations as "modes" of the hoist transformation?

I see your point. The primary application of this hoisting mechanism is intended for, but not limited to, the CUDA Fortran transformation. I have a few others planned or can imagine many more. For example, using the hoist mechanism

* for the (OpenACC) *scc hoist* transformation
* for module imports to avoid rewriting modules for e.g. CUDA Fortran
* for `parameter` variables, e.g. to make those variables *texture* (in CUDA Fortran)
* ...

Moreover, regarding the CUDA Fortran transformation not all local arrays are supposed to be hoisted, but those local arrays including the vertical dimension. Therefore, it would be necessary to pass those information as argument(s) limiting the functionality. 
But as I said, I agree with you, however I don't think it makes sense to work with callbacks, but maybe a solution would be to add more classes with more specialised application?

loki/bulk/scheduler.py
```diff
@@ -407,6 +407,7 @@ def process(self, transformation, reverse=False):
         always processed before their target :any:`Subroutine`.
         """
 
+        root = list(nx.topological_sort(self.item_graph))[0]
```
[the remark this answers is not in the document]
MichaelSt98: Yes. For the current implementation I don't need the depth. However, for the CUF transformation I will need the depth to know whether a function is a kernel (`__global__` subroutine) or a device function (a function called within a kernel function, thus a `__device__` subroutine). 

Having one than more root a possible implementation would be/include (e.g. within the `Scheduler` class `__init__` method):

```py
topological_generations = list(nx.topological_generations(self.item_graph))
# all of the generations/depths, this gives the maximum depth per item if the subroutine has multiple depths
for i_gen, gen in enumerate(topological_generations):
    print("{}: {}".format(i_gen, gen))
# those are all roots
roots = topological_generations[0]
# in order to get all the depths per item (and per root)
depths = {}
for item in self.item_graph:
    depths[item] = {}
    depths[item]["all"] = []
    for root in topological_generations[0]:
        simple_paths = list(nx.all_simple_paths(self.item_graph, source=root, target=item))
        depths[item]["all"].extend([len(_path) for _path in simple_paths])
        depths[item]["root"] = {}
        depths[item]["root"][root] = [len(_path) for _path in simple_paths]
```

For my purposes it will be sufficient to just know whether a subroutine is called by a driver (e.g. `level = 1`) or called by a kernel (e.g. `level >= 2`). 
**However, introducing this depth information for any transformation we may consider to pass/collect more information than that?** I would like to discuss which information and how to store/pass those information.
Nonetheless, I would suppose to exclude this from this pull request and make another pull request for this or include it in the CUF transformation pull request.

loki/transform/hoist.py
```diff
@@ -0,0 +1,79 @@
...
+        role = kwargs.get('role', None)
+        item = kwargs.get('item', None)
+        successors = kwargs.get('successors', None)
+
+        if item and not item.local_name == routine.name.lower():
+            return
+
+        variables = self.find_variables(routine)
+        if role != 'driver':
+            item.user_data["to_hoist"] = variables
```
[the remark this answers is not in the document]
MichaelSt98: I agree. 
As this transformation or mechanism includes two transformation, I would propose following implementation:
Any thoughts about that?

```py
class HoistTemporaryArraysAnlaysis(Transformation):
    
    def __init__(self):
        self._key = self.__class__.__name__

    def tranform_subroutine(self, routine, **kwargs):
        ...
        item.user_data[self._key]["to_hoist"] = ...
        ...

class HoistTemporaryArraysTransformation(Transformation):
    
    def __init__(self):
        self._key = HoistTemporaryArraysAnlaysis().__class__.__name__

    def tranform_subroutine(self, routine, **kwargs):
        ...
        ... = item.user_data[self._key]["to_hoist"]
        ...
```

loki/bulk/scheduler.py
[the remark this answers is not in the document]
reuterbal: Agreed, the question of what information to collect and pass mandates a more in-depth discussion.

I'm fairly far out of my area of expertise with graph algorithms, but I wonder if it would make sense to maintain also an adjoint view of the item_graph? That way, `all_simple_paths` might be able to produce the information for all roots in a single call (just viewed in the reverse direction).

loki/transform/hoist.py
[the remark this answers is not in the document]
reuterbal: Exactly. You could make `_key` even a class property:
```python
class HoistTemporaryArraysAnalysis(Transformation):
    _key = HoistTemporaryArraysAnalysis.__class__.__name__

class HoistTemporaryArraysTransformation(Transformation):
    _key = HoistTemporaryArraysAnalysis.__class__.__name__
```

Nice additional feature of this is that we can have a simple check in transformations that depend on an analysis:
```python
    if self._key not in item.user_data:
        # bail out with error
```

loki/transform/transform_hoist_variables.py
```diff
@@ -0,0 +1,333 @@
...
+        To be applied **reversed**, in order to recursively find all variables to be hoisted.
+
+    Parameters
+    ----------
+    key : str
+        Access identifier/key for the ``item.user_data`` dictionary. Only necessary to provide if several of
+        these transformations are carried out in succession.
+    """
+
+    _key = 'HoistVariablesTransformation'
```
MichaelSt98, line 110: Problem with:

```py
class HoistTemporaryArraysAnalysis(Transformation):
    _key = HoistTemporaryArraysAnalysis.__class__.__name__
    
```

as `NameError: name 'HoistTemporaryArraysAnalysis' is not defined`.

Further problems with derived classes using the `__init__` method and different behaviour for interactive sessions and non-interactive sessions. Therefore proposed solution as implemented.

loki/transform/transform_hoist_variables.py
```diff
@@ -0,0 +1,335 @@
...
+        if self._key not in item.user_data:
+            raise RuntimeError(f'{self.__class__.__name__} requires key "{self._key}" in item.user_data!\n'
+                               f'Make sure to call HoistVariablesAnalysis (or any derived class) before and to provide '
+                               f'the correct key.')
+
+        if role == 'driver':
+            for var in item.user_data[self._key]["to_hoist"]:
+                self.driver_variable_declaration(routine, var)
+        else:
+            routine.arguments += as_tuple([var.clone(type=var.type.clone(intent='inout', allocatable=None),
```
mlange05, line 230: Do we need this to be an `allocatable` here? Variables with an `intent` attribute can be declared without, I think.

tests/sources/projHoist/module/driver_mod.f90
```diff
@@ -0,0 +1,66 @@
...
+        z = 42
+        b = z
+    end subroutine device2
+
+end module transformation_module_hoist
```
mlange05, line 66: Linting: Missing newline at end of file

tests/test_transform_hoist_variables.py
```diff
@@ -0,0 +1,276 @@
...
+                assert (b == 42).all()
+                assert (c == 11).all()
+            clean_test(filepath=here.parent / item.source.path.with_suffix(suffix).name)
+
+
+def check_arguments(scheduler, subroutine_arguments, call_arguments):
+    """
+    Check the subroutine and call arguments of each subroutine.
+    """
+    for item in scheduler.items:
```
mlange05, line 79: Instead of loop+if/elif you could just access the items directly. Something like `scheduler.item_map['#driver'` or scheduler.item_map['#device2'].

loki/bulk/item.py
```diff
@@ -92,6 +92,7 @@ def __init__(self, name, source, config=None):
         self.name = name
         self.source = source
         self.config = config or {}
+        self.user_data = {}
```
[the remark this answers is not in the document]
mlange05: Ok, I originally proposed this name, but you are right, @reuterbal , it's a bit generic. Maybe something like `trafo_data`? For now, I'm happy to keep it as is though, as we'll probably internalise this in the rewrite of the scheduler.

mlange05 (approved):
Ok, this is looking very good, and quite a powerful addition to our supported transformations. There's a few API corner cases that we might need to rethink in the grand scheduler re-write, which have been exposed here, but otherwise, this looks very good.

I've left a few small style suggestions, but no showstoppers. So I'll approve this now, and let you decide whether to address them, @MichaelSt98 . Please signal here once you're happy for this to be merged.

tests/test_transform_hoist_variables.py
```diff
@@ -0,0 +1,315 @@
...
+                parameter_length = 3
+                b = np.zeros((_a,), dtype=np.int32, order='F')
+                c = np.zeros((_a, parameter_length), dtype=np.int32, order='F')
+                module.Transformation_Module_Hoist.driver(_a, b, c)
+                assert (b == 42).all()
+                assert (c == 11).all()
+            clean_test(filepath=here.parent / item.source.path.with_suffix(suffix).name)
+
+
+# def check_arguments(scheduler, subroutine_arguments, call_arguments):
```
reuterbal, line 75: Please remove the unused implementation.

reuterbal (approved):
Thank you very much for the changes. This looks really fantastic now and good to go from my side. I've left a few remarks but mostly just general comments.

loki/transform/transform_hoist_variables.py
```diff
@@ -0,0 +1,335 @@
...
+from loki.visitors import Transformer, FindNodes
+from loki.transform.transformation import Transformation
+import loki.expression.symbols as sym
+
+
+__all__ = ['HoistVariablesAnalysis', 'HoistVariablesTransformation',
+           'HoistTemporaryArraysAnalysis', 'HoistTemporaryArraysTransformationAllocatable']
+
+
+class HoistVariablesAnalysis(Transformation):
```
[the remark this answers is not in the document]
MichaelSt98: Yes, you're right. However, as the base class can be used for an actual transformation, I/we decided to not make it an abstract class, rather informing potential users to be aware that the default usage should include deriving a modified class ...

loki/transform/transform_hoist_variables.py
```diff
@@ -0,0 +1,335 @@
...
+        **Override**: Find/Select all the variables to be hoisted.
+
+        Selects all local variables that are not ``parameter`` to be hoisted.
+
+        Parameters
+        ----------
+        routine : :any:`Subroutine`
+            The subroutine find the variables.
+        """
+        return [var for var in routine.variables if var not in routine.arguments if not var.type.parameter]
```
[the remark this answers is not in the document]
mlange05: We'll leave that for now. I'll likely do this once I start adding to those implementations.

mlange05:
Ok, I think this is good to go. Everything essential addressed, and the abstract-baseclass change we'll leave until I start adding more implementations to this (soon :wink: ). Merging now.

PR #4, opened 2022-11-30
Github actions workflow for documentation

reuterbal:
Remaining CI item was auto-generating and uploading documentation. Not entirely sure this workflow does what it is supposed to do, but it generates documentation for pull request branches and should hopefully do the same for main after merging - and ideally clean-up the pr documentation files.

As a courtesy, the action even prints the link to the generated documentation in the PR ;-)


mlange05 (approved):
Looks great, much obliged! :pray:

PR #5, opened 2022-12-02
Implied-do support and error for `PROGRAM`

reuterbal:
This expands the support for array constructors, in particular `implied-do` format and with type-spec. They had not been handled gracefully before. Implemented for OMNI and fparser, OFP seemed too silly.

Piggy-backed is more sane handling of `PROGRAM` blocks - i.e., printing an error and bailing out. Behaviour was different between frontends before, some treating them as subroutines and others skipping over them. It shouldn't be terribly hard to add an implementation of `ProgramUnit` for that but until we do, we should gracefully fail or ignore that node rather than have some scrambled-together half-representation (since this has confused users before).


mlange05 (approved):
Excellent. Neat and clean as always. Very nice, can go in as is.

PR #11, opened 2023-01-05
Numpy<1.24 due to f90wrap incompatibility

reuterbal:
See https://github.com/jameskermode/f90wrap/issues/175

NumPy 1.24 ships with "new f2py features and fixes": https://numpy.org/doc/stable/release/1.24.0-notes.html

This requirement can be relaxed once f90wrap has caught up.


mlange05 (approved):
Yes, I can confirm this works and fixes the test base.

PR #10, opened 2022-12-13
Transformation: parametrise

MichaelSt98:
**Probably best explained using an example** (same code used for the unit tests):

<details>
<summary>Source code/Starting point</summary>

```fortran
module parametrise
    implicit none
contains

    subroutine stop_execution(msg)
        character(200), INTENT(IN) :: msg
        PRINT *, msg
        stop 1
    end subroutine stop_execution

    subroutine driver(a, b, c, d)
        integer, intent(inout) :: a, b, d(b)
        integer, intent(inout) :: c(a, b)
        real :: x, y
        call kernel1(a, b, c)
        call kernel2(a, b, d)
        call kernel1(a, b, c)
    end subroutine driver

    subroutine another_driver(a, b, c)
        integer, intent(in) :: a
        integer, intent(in) :: b
        integer, intent(inout) :: c(a, b)
        real :: x(a)
        call kernel1(a, b, c)
    end subroutine another_driver

    subroutine kernel1(a, b, c)
        integer, intent(in) :: a
        integer, intent(in) :: b
        integer, intent(inout) :: c(a, b)
        integer :: local_a
        real :: x(a)
        integer :: y(a, b)
        real :: k1_tmp(a, b)
        y = 11
        c = y
    end subroutine kernel1

    subroutine kernel2(a_new, b, d)
        integer, intent(in) :: a_new
        integer, intent(in) :: b
        integer, intent(inout) :: d(b)
        real :: x(a_new)
        real :: y, z
        real :: k2_tmp(a_new, a_new)
        call device1(a_new, b, d, x, k2_tmp)
    end subroutine kernel2

    subroutine device1(a, b, d, x, y)
        integer, intent(in) :: a
        integer, intent(in) :: b
        integer, intent(inout) :: d(b)
        real, intent(inout) :: x(a)
        real, intent(inout) :: y(a, a)
        real :: z
        integer :: d1_tmp
        call device2(a, b, d, x)
        call device2(a, b, d, x)
    end subroutine device1

    subroutine device2(a, b, d, x)
        integer, intent(in) :: a
        integer, intent(in) :: b
        integer, intent(inout) :: d(b)
        real, intent(inout) :: x(a)
        integer z(b)
        integer :: d2_tmp(b)
        z = 42
        d = z
    end subroutine device2

end module parametrise
```

</details>

________

can be transformed via 

```py
dic2p = {'a': 12, 'b': 11}
[...]

PR #12, opened 2023-01-06
Transform utilities: single variable declaration

MichaelSt98:
* **single variable declaration** for specific variables via: `single_variable_declaration(routine, variables=())`

```fortran
subroutine foo(a, x, y)
   integer, intent(in) :: a
   real, intent(inout):: x(a), y(a, a)
   integer :: i1, i2, i3, i4
   real :: r1, r2, r3, r4
   x = a
   y = a
end subroutine foo
```

via `single_variable_declaration(routine=routine, variables=('y', 'i1', 'i3', 'r1', 'r2', 'r3', 'r4'))` to

```fortran
SUBROUTINE foo (a, x, y)
  INTEGER, INTENT(IN) :: a
  REAL, INTENT(INOUT) :: x(a)
  REAL, INTENT(INOUT) :: y(a, a)
  INTEGER :: i2, i4
  INTEGER :: i1
  INTEGER :: i3
  REAL :: r1
  REAL :: r2
  REAL :: r3
  REAL :: r4
  x = a
  y = a
END SUBROUTINE foo
```

________

* **single variable declarations** for subroutines: `single_variable_declarations(routine, strict=True)`

```fortran
subroutine foo(a, x, y)
   integer, intent(in) :: a
   real, intent(inout):: x(a), y(a, a)
   integer :: i1, i2, i3, i4
   real :: r1, r2, r3, r4
   real :: x1, x2(a), x3(a), x4(a, a)
   x = a
   y = a
end subroutine foo
```

via `single_variable_declarations(routine=routine, strict=True)` to

```fortran
SUBROUTINE foo (a, x, y)
  INTEGER, INTENT(IN) :: a
  REAL, INTENT(INOUT) :: x(a)
  REAL, INTENT(INOUT) :: y(a, a)
  INTEGER :: i1
  INTEGER :: i2
  INTEGER :: i3
  INTEGER :: i4
  REAL :: r1
  REAL :: r2
  REAL :: r3
  REAL :: r4
  REAL :: x1
  REAL :: x2(a)
  REAL :: x3(a)
  REAL :: x4(a, a)
  x = a
  y = a
END SUBROUTINE foo
```

and via `single_variable_declarations(routine=routine, strict=False)` to

```fortran
SUBROUTINE foo (a, x, y)
  INTEGER, INTENT(IN) :: a
  REAL, INTENT(INOUT) :: x(a)
  REAL, INTENT(INOUT) :: y(a, a)
  INTEGER :: i1, i2, i3, i4
  REAL :: r1, r2, r3, r4
  REAL :: x2(a), x3(a)
  REAL :: x4(a, a)
  REAL :: x1
  x = a
  y = a
END SUBROUTINE foo
```

### DIFF
--- a/loki/transform/transform_utilities.py
+++ b/loki/transform/transform_utilities.py
@@ -11,7 +11,7 @@ Collection of utility routines to deal with general language conversion.
 
 """
 import platform
-
+from collections import defaultdict
 from loki.expression import (
     symbols as sym, FindVariables, FindInlineCalls, FindLiterals,
     SubstituteExpressions, SubstituteExpressionsMapper, ExpressionFinder,
@@ -27,13 +27,19 @@ from loki.visitors import Transformer, FindNodes
 
 __all__ = [
     'convert_to_lower_case', 'replace_intrinsics', 'sanitise_imports',
-    'replace_selected_kind', 'single_variable_declarations', 'single_variable_declaration'
+    'replace_selected_kind', 'single_variable_declaration'
 ]
 
 
-def single_variable_declaration(routine, variables=()):
+def single_variable_declaration(routine, variables=None, group_by_shape=False):
     """
-    Modify/extend variable declarations to declare variables specified in ``variables`` in single declarations.
+    Modify/extend variable declarations to
+
+    * default: only declare one variable each time while preserving the order if ``variables=None`` and
+     ``group_by_shape=False``
+    * declare variables specified in ``variables``in single/unique declarations if ``variables`` is a tuple
+     of variables
+    * variable declarations to be grouped according to their shapes if ``group_by_shape=True``
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,96 @@
 
 __all__ = [
     'convert_to_lower_case', 'replace_intrinsics', 'sanitise_imports',
-    'replace_selected_kind'
+    'replace_selected_kind', 'single_variable_declarations', 'single_variable_declaration'
 ]
 
 
+def single_variable_declaration(routine, variables=()):
```
reuterbal, line 34: Similar remark about default arguments as in #10.

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,96 @@
...
+    routine: :any:`Subroutine`
+        The subroutine in which to modify the variable declarations
+    strict: bool
+        Whether to strictly make unique variable declarations or to only disassemble non-arrays and arrays and among
+        arrays, arrays with differing shapes.
+    """
+    decl_map = {}
+    for decl in FindNodes(VariableDeclaration).visit(routine.spec):
+        if len(decl.symbols) > 1:
+            if strict:
```
reuterbal, line 85: `strict` mode is essentially `single_variable_declaration` for all variables, right? Wouldn't it make sense to simply combine the two utilities into one?

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,96 @@
...
+                counter = 1
+                for i_sdecl, sdecl in enumerate(decl.symbols):
```
reuterbal, line 87: I think this could be a single list comprehension, similar to above:
```python
decl_map[decl] = tuple(decl.clone(symbols=(s,)) for s in decl.symbols)
```

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,96 @@
...
+                    if i_sdecl == 0:
+                        decl_map[decl] = decl.clone(symbols=(sdecl,))
+                    else:
+                        routine.spec.insert(routine.spec.body.index(decl) + counter, (decl.clone(symbols=(sdecl,)),))
+                        counter += 1
+            else:
+                types = [type(smbl) for smbl in decl.symbols]
+                smbls_by_type = {smbl_type: [] for smbl_type in set(types)}
```
reuterbal, line 95: This pattern is a typical use-case for `defaultdict`:
```python
smbls_by_type = defaultdict(list)
for smbl in decl.symbols:
    smbls_by_type[type(smbl)] += [smbl]
```

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,96 @@
...
+                for smbl in decl.symbols:
+                    smbls_by_type[type(smbl)].append(smbl)
+                declarations = []
+                for key in smbls_by_type:
```
reuterbal, line 99: If I'm reading this correctly, I think you're essentially trying to sort the symbols into different groups depending on their shape. May I suggest again a slightly simpler sorting procedure to make this a bit more straightforward? It's untested, so may not work correctly:
```python
symbols_by_shape = defaultdict(list)
for smbl in decl.symbols:
    symbols_by_shape[getattr(smbl, 'shape', None)] += [smbl]
decl_map[decl] = tuple(decl.clone(symbols=as_tuple(smbls)) for smbls in symbols_by_shape.values())
```

reuterbal (changes requested):
This is a very useful utility and I know that some users had asked for something like this before. So, much appreciated!
Great test coverage as well. Could you please add them also to the [docs](https://github.com/ecmwf-ifs/loki/blame/main/docs/source/transform.rst#L110)?

I'm not sure if you're aware of the 1-to-many mapping capability of `Transformer`, which allows to make the implementation much shorter. I've proposed a few variants, please see if you agree but check them carefully if you decide to use any of them.

Finally, I'm a bit unsure about the very similar utility names and am wondering if it couldn't become a single utility?
For example, with a signature like this?
```
def single_variable_declarations(routine, variables=None, group_by_shape=False)
```
With default behaviour equivalent to `single_variable_declarations(routine, strict=True)`, specifying `group_by_shape=True` would be equivalent to `single_variable_declarations(routine, strict=False)`, and specifying a variable list would only create the single declarations for the specified variables, i.e. equivalent to `single_variable_declaration(routine, variables=(...))`. Specifying `variables` and `group_by_shape=True` could result in an error.

mlange05 (commented):
Again, this looks very good, and I'm happy with it in principle. I agree with the small style suggestions in @reuterbal s review. 

One additional observation I have is, whether the two functions aren't just a general and a special-case of the same thing? As in `single_variable_declarations(routine, variables=None)` could simply do `variables = <all variables in multi-var declarations in routine>`, and the actual replacement logic would then not need to be replicated. Or am I missing a specific (set of) corner case(s) here? If so, happy to go ahead as is.

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,96 @@
...
+    ----------
+    routine: :any:`Subroutine`
+        The subroutine in which to modify the variable declarations
+    variables: tuple
+        Variables to grant unique/single declaration for
+    """
+    if variables:
+        decl_map = {}
+        for decl in FindNodes(VariableDeclaration).visit(routine.spec):
+            if len(decl.symbols) > 1:
```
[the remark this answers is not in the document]
MichaelSt98: That's true, however I am checking for `len(decl.symbols) > 1` and not `len(decl.symbols) > 0`.

However, `if decl.symbols` (which corresponds to `if len(decl.symbols) > 0`) would also work but would make a unnecessary mapping or one to one mapping.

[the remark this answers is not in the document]
reuterbal: Oh, you're right, scratch my remark. I hadn't read carefully enough.

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,53 @@
...
+                    else:
+                        decl_map[decl] = new_decls
+        else:
+            smbls_by_shape = defaultdict(list)
+            for smbl in decl.symbols:
+                smbls_by_shape[getattr(smbl, 'shape', None)] += [smbl]
+            decl_map[decl] = tuple(decl.clone(symbols=as_tuple(smbls)) for smbls in smbls_by_shape.values())
+    routine.spec = Transformer(decl_map).visit(routine.spec)
+    # if variables defined and group_by_shape, first call ignores the variables, thus second call
+    if variables and group_by_shape:
```
reuterbal, line 73: That's a really clever idea!

loki/transform/transform_utilities.py
```diff
@@ -27,10 +27,53 @@
...
+    variables: tuple
+        Variables to grant unique/single declaration for
+    group_by_shape: bool
+        Whether to strictly make unique variable declarations or to only disassemble non-arrays and arrays and among
+        arrays, arrays with differing shapes.
+    """
+    decl_map = {}
+    for decl in FindNodes(VariableDeclaration).visit(routine.spec):
+        if not group_by_shape:
+            if len(decl.symbols) > 1:
```
reuterbal, line 57: Shouldn't this be the outer conditional? As in, we don't have to do anything for `group_by_shape==True` either if there's only one symbol declared.

reuterbal (approved):
Very nicely done, thanks for the changes. Looks good to me!

loki/transform/transform_utilities.py
MichaelSt98: True, makes it less checks `if not group_by_shape` but also two (independent) loops (which is probably cleaner code). Do you want me to change it?

```py
decl_map = {}
if not group_by_shape:
    for decl in FindNodes(VariableDeclaration).visit(routine.spec):
        if len(decl.symbols) > 1:
            unique_symbols = [s for s in decl.symbols if variables is None or s.name in variables]
            if unique_symbols:
                new_decls = tuple(decl.clone(symbols=(s,)) for s in unique_symbols)
                retain_symbols = tuple(s for s in decl.symbols if variables is not None and s.name not in variables)
                if retain_symbols:
                    decl_map[decl] = (decl.clone(symbols=retain_symbols),) + new_decls
                else:
                    decl_map[decl] = new_decls
else:
    for decl in FindNodes(VariableDeclaration).visit(routine.spec):
        smbls_by_shape = defaultdict(list)
        for smbl in decl.symbols:
            smbls_by_shape[getattr(smbl, 'shape', None)] += [smbl]
        decl_map[decl] = tuple(decl.clone(symbols=as_tuple(smbls)) for smbls in smbls_by_shape.values())
routine.spec = Transformer(decl_map).visit(routine.spec)
# if variables defined and group_by_shape, first call ignores the variables, thus second call
if variables and group_by_shape:
    single_variable_declaration(routine=routine, variables=variables, group_by_shape=False)
```

reuterbal: Ah, sorry, I had the following order in mind:
```python
for decl in FindNodes(VariableDeclaration).visit(routine.spec):
    if len(decl.symbols) > 1:
        if not group_by_shape:
```
because you end up with a 1-to-1 mapping for the else case otherwise. It's a detail and all three variants should work, so feel free to leave it as is.

MichaelSt98: no, you're right 👍

mlange05:
Indeed, very neat and well tested! GTG, going in...

PR #16, opened 2023-01-11
Different way of creating documentation comment

reuterbal:
This changes the way the documentation comment is provided, to avoid these duplicate messages.

Also another stab at the cleanup after merging, but that one we'll know only after merging.


mlange05 (approved):
Can't claim I understand this all, but seems functional and clean. So GTG from me.

PR #17, opened 2023-01-12
Parsing items in reverse topological order instead of topological order

MichaelSt98:
**Parsing items in reverse topological order instead of topological order**

* should make explicit enrichment in principle completely unnecessary

____

PR as discussion/test about whether

* changing order generally to be in reversed topological order
* or implementing parsing in reversed topological order as additional possibility 

e.g. 

`scheduler.py`:

```py
def _parse_items(self, reversed=False):
    if reversed:
        for item in reversed(list(nx.topological_sort(self.item_graph))):
             item.source.make_complete(**self.build_args)
    else:
         for item in nx.topological_sort(self.item_graph):
             item.source.make_complete(**self.build_args)
```

### DIFF
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -372,7 +372,7 @@ class Scheduler:
         the execution plan and enriching subroutine calls.
         """
         # Force the parsing of the routines
-        for item in nx.topological_sort(self.item_graph):
+        for item in reversed(list(nx.topological_sort(self.item_graph))):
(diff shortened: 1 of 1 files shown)


reuterbal (approved):
Thanks a lot! As discussed offline, I'm voting to make this the default and I don't think there is a need to do this in a different order.

mlange05 (approved):
Agreed, sounds good to me.

PR #18, opened 2023-01-16
Tour de bug

reuterbal:
A bunch of bug fixes I collected along the way, filing them as a collective PR. Namely the following things are fixed:

- Formatting of `DATA` declaration nodes in fgen backend
- A fix for formatting of long literal lists (or more generally, long lines with nested `JoinableStringList`s): These should now wrap gracefully and produce well-formated line breaks, whereas they would before wrap lines and then indent them, which would make it necessary to break the lines again, leading to all kinds of havoc.
- Newly added support for subroutine suffixes (fixing `BIND` attribute and adding support for `RESULT`)
- There were two bugs in our handling of inline calls with keyword-arguments to type-bound procedures and calls to type-bound procedures declared on array members of a derived type
- A new `StringSubscript` node has been added to handle subscript operations on strings - the only case where dimensions expressions can appear in series in Fortran (e.g. `substr = my_type_var%list_of_names(idx)(start:end)`).
- Our loki-transform script has a `--debug` option to automatically attach pdb(++) when uncaught exceptions occur. In a CMake integration context it's a bit cumbersome to use this option and therefore I've added a config option that allows to enable this generall via `LOKI_DEBUG=1` environment variable.

### DIFF
--- a/tests/test_fgen.py
+++ b/tests/test_fgen.py
@@ -1,0 +1,37 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import pytest
+
+from conftest import available_frontends
+
+from loki import Subroutine, fgen
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_fgen_literal_list_linebreak(frontend):
+    """
+    Test correct handling of linebreaks for LiteralList expression nodes
+    """
+    fcode = """
+subroutine literal_list_linebreak
+    implicit none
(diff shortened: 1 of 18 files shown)


tests/test_fgen.py
```diff
@@ -0,0 +1,119 @@
...
+    if frontend == OFP:
+        assert isinstance(routine.spec.body[-1], Intrinsic)
+    else:
+        assert isinstance(routine.spec.body[-1], DataDeclaration)
+    spec_code = fgen(routine.spec)
+    assert spec_code.lower().count('data ') == 2
+    assert spec_code.count('/') == 4
+    if frontend != OMNI:
+        # OMNI seems to evaluate constant expressions, replacing 31*0 by 0,
+        # although it's not a product here but a repeat specifier (great job, Fortran!)
```
mlange05, line 118: :disappointed:

mlange05 (approved):
Nice! Very extensive list of bugs, and I cannot decide which one of these looks like the worst to find and fix. Regardless, uncontroversial, and very neatly done; many thanks! GTG from me.

PR #22, opened 2023-01-18
Enrichment during parsing

reuterbal:
Note: This is based on #18, so review only after that one is merged to reduce diff volume.

With #17 we parse files now in the correct order to theoretically resolve all dependencies right away, thus making manual enrichment essentially superfluous. However, for this to take effect, we have to provide the already parsed program units as `definitions` to `make_complete`, which is added in this PR.

Although it should allow to entirely retire the `_enrich` step in the Scheduler, I have kept it for the moment to avoid fallout in edge cases where this doesn't work for some reason. It works on ecrad, though.

Accompanied with this are tests for nested derived types, where intermediate type definitions can only be resolved via this enrichment method.

With this in place, I have hit a situation in the frontends, where imports include symbols that we have for some reason not in the symbol table of the module they are imported from, although we have the corresponding module parsed. This happens, when these symbols themselves stem from an unqualified import from another module. Such situations are now handled gracefully in the same way as when the corresponding module was not available. We should be able to resolve this in the future cleanly by chasing also imports as dependencies in the Scheduler.

Lastly, piggy-backed is a workaround for wrongful classification of inline calls as array subscripts in the fparser frontend. We repair this already later on with type information from the symbol table, however, in the process we would only convert the `Array` node to a `ProcedureSymbol`, thus loosing the information about the inline call and its arguments. This is now being taken care of in the `LokiIdentityMapper` by detecting such conversions and creating a new `InlineCall` node with the previous `dimensions` attributes as `parameters`.

### DIFF
--- a/loki/__init__.py
+++ b/loki/__init__.py
@@ -46,6 +46,9 @@ config.register('print-config', False, env_variable='LOKI_PRINT_CONFIG',
 config.register('log-level', 'INFO', env_variable='LOKI_LOGGING',
                 callback=set_log_level, preprocess=lambda i: log_levels[i])
 
+config.register('debug', None, env_variable='LOKI_DEBUG',
+                callback=set_excepthook, preprocess=lambda i: auto_post_mortem_debugger if i else None)
+
 # Define Loki's temporary directory for generating intermediate files
 config.register('tmp-dir', None, env_variable='LOKI_TMP_DIR')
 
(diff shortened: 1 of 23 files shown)


mlange05 (approved):
Ok, changes are clean and well tested. I confirmed full-physics behaves as expected on Cy47R1 dev branches. GTG from me!

Issue #19, opened 2023-01-18
Demoted arrays in SCC transform do not get their declaration updated when DIMENSION keyword is used.


JoeffreyLegaux:
Small reproducer available here : https://mercure.cerfacs.fr/cfxfile/get.php?bc22b33fc3a0  (link available up to the 02 feb 2023).

When arrays are demoted to scalar variables during the SCC transformation, if their original declaration was done using the 'DIMENSION' keyword, their declaration do not get updated accordingly.

PR #24, opened 2023-01-23
Internalize array demotion transform

mlange05:
This PR creates an internal array transformation to demote individual array dimensions from a list of array variables within a routine. This was previously done in SCC via a custom utility, which we have now generalized and internalised to the core package, including the respective tests. 

As an added bonus, the new routine is now also capable of honouring `DIMENSION` keywords, and will adjust the respective `VariableDeclaration` node accordingly. This should address issue #19 .

### DIFF
--- a/loki/transform/transform_array_indexing.py
+++ b/loki/transform/transform_array_indexing.py
@@ -471,4 +471,23 @@ def demote_variables(routine, variable_names, dimensions):
     routine.body = SubstituteExpressions(vmap).visit(routine.body)
     routine.spec = SubstituteExpressions(vmap).visit(routine.spec)
 
+    # Ensure all declarations with `DIMENSION` keywords are modified too!
+    decls = tuple(
+        d for d in FindNodes(VariableDeclaration).visit(routine.spec)
+        if d.dimensions and any(s.name.lower() in vnames for s in d.symbols)
+    )
+    decl_map = {}
+    for decl in decls:
+        # If all symbols have the same shape (after demotion)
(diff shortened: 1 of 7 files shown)


loki/transform/transform_array_indexing.py
```diff
@@ -428,3 +428,66 @@ def promote_nonmatching_variables(routine, promotion_vars_dims, promotion_vars_i
...
+    dimensions = as_tuple(dimensions)
+
+    # Compare lower-case only, since we're not comparing symbols
+    vnames = tuple(name.lower() for name in variable_names)
+
+    variables = FindVariables(unique=False).visit(routine.ir)
+    variables = tuple(v for v in variables if v.name.lower() in vnames)
+
+    # Record original array shapes
+    shape_map = CaseInsensitiveDict({v.name: v.shape for v in variables})
```
reuterbal, line 458: This will probably fail if `variables` includes non-Array variables. I think the utility should handle this gracefully and discard any scalar symbols in `variables`. Would be good to verify this also in a test, e.g. by adding just a scalar to the list of names passed to the utility and checking that nothing has happened to it.

reuterbal (changes requested):
Thank you very much. Useful and could already be used in #14. I have left two questions, one of them I think needs adressing. I'll let the user confirm it fixes the issue for him, too.

reuterbal (approved):
As discussed offline, I've added resiliency against calling with non-Array variables.

mlange05:
Looking good now, going in.

PR #27, opened 2023-01-30
Fixing OMNI get_source()

MichaelSt98:
in order to have unique nodes e.g. call statements to the same subroutine ...

which solves some problems e.g. regarding transformer with mapper/dictionary entries which would otherwise be overwritten as the nodes e.g. subroutines are not distinguishable.


reuterbal (approved):
Excellent find, many thanks! Uncontroversial and GTG from me.

mlange05 (approved):
Fully agree, very nice find, and uncontroversial fix. GTG, going in.

PR #30, opened 2023-01-31
Fix Regex pattern for subroutines to account for interfaces (fix #26)

### LINKED ISSUE #26: Incorrect identification of interface blocks in regex frontend
In the presence of interface blocks, the regex frontend incorrectly identifies the `end subroutine` statement inside the interface block as the end of the parent subroutine. This happens even in the presence of the subroutine name at the end of the interface block i.e. `end subroutine name`. Strangely, this only seems to happen once the raw source has been sanitized by `FortranReader`; the regex frontend correctly parses the unsanitized string correctly and the sanitized string incorrectly (the sanitization itself appears to be correct though). I've attached an example.
[regex_frontend_bug.txt](https://github.com/ecmwf-ifs/loki/files/10517618/regex_frontend_bug.txt)

### LINKED ISSUE #31: Scheduler: Find dependencies of calls declared as generic interfaces across multiple modules
It's a fascinating Fortran pattern:

```f90

module a_mod
type a
...
end type
 
interface load
  module procedure load_a
end interface
 
contains
subroutine load_a
...
end subroutine
end module
 
! ----
 
module b_mod
type b
...
end type
 
interface load
  module procedure load_b
end interface
 
contains
subroutine load_b
...
end subroutine
end module
 
! ----
 
module c_mod
...
end module
 
! ----
 
subroutine driver
  use a_mod
  use b_mod
  use c_mod
 
  type(a):: avar
  type(b):: bvar
  type(c):: cvar
 
  call load(avar)
  call load(bvar)
  call load(cvar)
end subroutine

```

In short: Scheduler's `obj_map` would have to also list declared interfaces as names defined in the corresponding source file, and the lookup needs to take into account, that we can in fact have more than one place where a specific symbol is defined (yay!). The mapping interface -> module procedure introduces another level of indirection similar to type bindings and needs to be dealt with. Here, we have the added complication that we don't have a `%` in the name which implicitly makes clear that there is a procedure binding involved.

#30 already adds the mechanics on the REGEX frontend side for parsing interfaces.

reuterbal:
The purpose of this is to fix the regex frontend bug reported in #26. Solution is to explicitly match interface blocks in the subroutine pattern, thus shortcutting any contained subroutine declarations.

 In the process I had already developed a more sophisticated pattern for matching interface blocks. We may need this eventually to resolve #31.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -398,7 +398,8 @@ class Item:
                 warning(f'{candidates} not found in available_names')
                 return candidates
             if len(matched_names) != 1:
-                warning(f'Duplicate symbol: {name[name.index("#")+1:]}, can be one of {matched_names}')
+                name = matched_names[0]
+                warning(f'Duplicate symbol: {name[name.find("#")+1:]}, can be one of {matched_names}')
                 return tuple(name + member_name for name in matched_names)
             return matched_names.pop() + member_name
 
--- a/loki/frontend/ofp.py
+++ b/loki/frontend/ofp.py
@@ -1061,7 +1061,10 @@ class OFP2IR(GenericVisitor):
 
     def visit_interface(self, o, **kwargs):
         scope = kwargs['scope']
-        abstract = o.get('type') == 'abstract'
+        if o.get('type') is not None:
+            abstract = o.get('type').lower() == 'abstract'
+        else:
+            abstract = False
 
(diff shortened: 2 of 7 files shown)


mlange05 (approved):
Changes look good, and well tested. Happy to proceed, but needs a rebase now. I'll approve it now though, assuming rebase happens before merge.

PR #32, opened 2023-02-06
Minor scheduler and dependency injection bugfixes

reuterbal:
Two little bugfixes/enhancements for the Scheduler:

1. Resolve a case-sensitivity issue in `Item.qualify_names`
2. Remove duplicates from `Item.children`

Two little fixes for the dependency injection transformation:

1. Canonical basename needs to look for the suffix via `rindex` to avoid aliasing problems (e.g., when looking for `_mod` in `some_model_thing_mod`).
2. Calls to contained member routines can be ignored.


mlange05 (approved):
Yes, looks uncontroversial. Will merge once PR #14 has landed.

PR #34, opened 2023-02-08
Fix infinite recursion in SubstituteExpression

reuterbal:
As discussed offline, this fixes the infinite recursion issue when we have expression maps like `{var: var + 1}`. Solution is to remove the recursion on the replaced node and rather have a dedicated utility that allows applying that recursion to the expression map before carrying out the substitution.

@MichaelSt98 could you please let me know if this fixes your issue?

### DIFF
--- a/loki/expression/symbols.py
+++ b/loki/expression/symbols.py
@@ -67,7 +67,7 @@ class ExprMetadataMixin:
 
     @property
     def init_arg_names(self):
-        return super().init_arg_names + ('_source', )
+        return super().init_arg_names + ('source', )
 
     def __init__(self, *args, **kwargs):
         self._source = kwargs.pop('source', None)
@@ -80,6 +80,10 @@ class ExprMetadataMixin:
     def source(self):
         """The :any:`Source` object for this expression node."""
         return self._source
+
+    @source.setter
+    def source(self, source):
+        self._source = source
 
     make_stringifier = loki_make_stringifier
 
@@ -301,7 +305,7 @@ class TypedSymbol:
             List of member variables in a derived type
         """
         _type = self.type
-        if isinstance(_type.dtype, DerivedType):
+        if _type and isinstance(_type.dtype, DerivedType):
             if _type.dtype.typedef is BasicType.DEFERRED:
                 return ()
             return tuple(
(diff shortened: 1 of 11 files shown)


MichaelSt98:
> As discussed offline, this fixes the infinite recursion issue when we have expression maps like `{var: var + 1}`. Solution is to remove the recursion on the replaced node and rather have a dedicated utility that allows applying that recursion to the expression map before carrying out the substitution.
> 
> @MichaelSt98 could you please let me know if this fixes your issue?

Yes! Thank you @reuterbal 👍

loki/expression/symbols.py
```diff
@@ -81,6 +81,10 @@ def source(self):
         """The :any:`Source` object for this expression node."""
         return self._source
 
+    @source.setter
+    def source(self, source):
+        self._source = source
```
mlange05, line 86: Eventually, this needs to simply be an attribute, not a property. But this will come with the mode to dataclasses.

mlange05 (commented):
This is fantastic. Very neat and cleanly tested. Also, many thanks for the additional testing added as part of this!

The only thing I'm wondering is if this routine could be moved to the `SubstitutionMapper` definition itself, and triggered via a constructor flag, rather than having it in `transform_utilities`? Maybe something like this would be nice (default would be off, I guess): 
```
routine.body = SubstituteExpressions(vmap, recursive=True).visit(routine.body)
```

If not, I'd be happy to merge this as is, so please just reply here and I'll approve and merge once #34 is in.

reuterbal:
Thanks for the review and remarks. Moving this to the constructor could be a good idea, however, a common pattern is something like this:
```
vmap = recursive_expression_map_update(vmap)
routine.spec = SubstituteExpressions(vmap).visit(routine.spec)
routine.body = SubstituteExpressions(vmap).visit(routine.body)
```
So, having this in the constructor would
1. raise the question whether it should change the existing `vmap` or create a clone
2. duplicate the substution effort or deviate from existing behaviour, that the `vmap` itself is not modified by the Transformer

Just to clarify: I'm not opposed, only concerned it may result in unexpected behaviour. Perhaps a big fat note on the SubstituteExpressions docstring for this sanitization requirement might be safer?

If moved to the constructor it would definitely need to be off by default, though, to make the original issue of mapping `var -> var+1` work correctly (and not generate `var -> var + 1 + 1 + 1 + 1 + 1 ...`.

mlange05:
Ok, great, I'm happy with that explanation, and agree that there could be some side effects. So for now, simply adding this to solve the problem sounds like the right approach. Many thanks, GTG.

mlange05 (approved):

PR #38, opened 2023-02-22
Fix #36: regex frontend hangs for function declaration followed by subroutine

### LINKED ISSUE #36: REGEX frontend hangs for function declaration, followed by subroutine
An odd corner case, where a function declaration that uses inline return type declaration, and happens to be followed by another subroutine, can send the REGEX frontend into an infinite while loop and cause it to hang (no error returned!).

The following MFE triggers this behaviour, only when both the inline return type declaration and(!) the routine after the function are present. Removing either seem to side-step the sissue.

```
In [23]: fcode = """
REAL(KIND=JPRB)  FUNCTION  DOT_PRODUCT_ECV() 

END FUNCTION DOT_PRODUCT_ECV

SUBROUTINE DOT_PROD_SP_2D()

END SUBROUTINE DOT_PROD_SP_2D
"""

In [24]: Sourcefile.from_source(fcode, frontend=REGEX)
```

reuterbal:


mlange05 (approved):
Neat. GTG from me, going in.

PR #39, opened 2023-02-22
Fix #37: regex frontend hangs if end is missing subroutine

### LINKED ISSUE #37: REGEX frontend hangs if `END` is missing `SUBROUTINE`
While the following coding style should rightly be discouraged(!), it is legal Fortran, and causes the REGEX frontend to start spinning indefinitely and thus hang. The following MFE demonstrates this:
 
```
In [25]: fcode = """
subroutine some_routine(a)
  real, intent(in) :: a
end 
"""
Sourcefile.from_source(fcode, frontend=REGEX)
```
Of course, turning the final line into `end subroutine`, even without the name side-steps this, but .... Fortran...

reuterbal:
Add a timeout utility with the option to specify (or disable) a default timeout for the REGEX frontend to gracefully fail on catastrophic backtracking.

### DIFF
--- a/loki/__init__.py
+++ b/loki/__init__.py
@@ -73,6 +73,9 @@ config.register('disk-cache', False, env_variable='LOKI_DISK_CACHE',
 config.register('case-sensitive', False, env_variable='LOKI_CASE_SENSITIVE',
                 preprocess=lambda i: bool(i) if isinstance(i, int) else i)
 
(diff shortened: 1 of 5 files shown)


mlange05 (approved):
Neat and tidy, as always. Many, many thanks, this will be very useful. GTG from me.

PR #43, opened 2023-02-24
Various Pygen-backend fixes

mlange05:
A small collection of fixes for the Python backend that brings the full CLOUDSC back to validating. I've added a few internal test extensions, but have probably missed a few, so please feel free to request more. CLOUDSC @ Python is not yet part of the CI cycle, so internal testing is paramount.

A few nuggets:
* Better way of ensuring dynamic kernel loading in the tests
* No longer replicate `Subroutine` object as part of F2Py transformation
* Fixed downcasing and added support for statement functions (yes!) and inline calls
* Extended line length limit, since automatic line-breaks work different for Python
* Extended intrinsic handling to include exp/sqrt/sign (the latter works differently, so somewhat cumbersome!)
* Allow explicitly triggering / overriding the index-inversion, as this is dependent on outer calling code

### DIFF
--- a/loki/backend/pygen.py
+++ b/loki/backend/pygen.py
@@ -27,7 +27,7 @@ def numpy_type(_type):
             return 'np.float32'
         return 'np.float64'
     if isinstance(_type.dtype, DerivedType):
-        return _type.dtype.name.lower()
+        return _type.dtype.name
     raise ValueError(str(_type))
 
 
@@ -139,11 +139,11 @@ class PyCodegen(Stringifier):
         arg_str = []
         for arg in arguments:
             if isinstance(arg.type.dtype, DerivedType):
-                arg_str += [f'{arg.name.lower()}']
+                arg_str += [f'{arg.name}']
             else:
                 dtype = self.visit(arg.type, **kwargs)
-                arg_str += [f'{arg.name.lower()}: {dtype}']
-        header += [self.format_line('def ', o.name.lower(), '(', self.join_items(arg_str), '):')]
+                arg_str += [f'{arg.name}: {dtype}']
+        header += [self.format_line('def ', o.name, '(', self.join_items(arg_str), '):')]
 
         # ...and generate the spec without imports and only declarations for variables that
         # either are local arrays or are assigned an initial value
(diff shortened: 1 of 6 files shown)


loki/backend/pygen.py
```diff
@@ -75,6 +77,21 @@ def map_array_subscript(self, expr, enclosing_prec, *args, **kwargs):
...
+        arguments = ', '.join(self.rec(p, PREC_NONE, *args, **kwargs) for p in expr.parameters)
+
+        if expr.kw_parameters:
+            arguments += ', ' + ', '.join(
+                f'{self.rec(k, PREC_NONE, *args, **kwargs)}={self.rec(v, PREC_NONE, *args, **kwargs)}'
+                for k, v in expr.kw_parameters.items()
+            )
+
+        f = self.rec(expr.function, PREC_NONE, *args, **kwargs)
+        return self.format(f'{str(f).lower()}({arguments})')
```
reuterbal, line 90: Should the `.lower()` not be something that is done in the Transformation rather than the backend? This effectively prohibits the use of function calls to routines with upper case function names.

loki/backend/pygen.py
```diff
@@ -75,6 +77,21 @@ def map_array_subscript(self, expr, enclosing_prec, *args, **kwargs):
...
+
+    def map_deferred_type_symbol(self, expr, *args, **kwargs):
+        return str(expr.name).lower().replace('%', '.')
```
reuterbal, line 93: Same comment about `.lower()`.

reuterbal (commented):
Great stuff, many thanks!
I haven't tested it on CLOUDSC, yet, so leaving this as status "Comment" for now.

I don't think the backend should change names to lower case, otherwise nothing that caught my eye

loki/backend/pygen.py
mlange05: Yes, indeed! Good catch, these were actually masking some severe issues in our downcasing utility transform. I've just pushed two commits that remove these two short-cuts, and instead fix the issue in the F2P transformation, as suggested. :grin:

mlange05:
Ok, just pushed another rebase to bring in line with `main`.

loki/backend/pygen.py
```diff
@@ -115,13 +132,18 @@ def <name>(<args>):
...
-                     for arg in o.arguments if arg not in out_args]
-        header += [self.format_line('def ', o.name.lower(), '(', self.join_items(arguments), '):')]
+        scalar_args = [a for a in o.arguments if isinstance(a, sym.Scalar)]
+        inout_args = [a for a in scalar_args if a.type.intent and a.type.intent.lower() == 'inout']
+        out_args = [a for a in scalar_args if a.type.intent and a.type.intent.lower() == 'out']
+        arguments = [arg for arg in o.arguments if arg not in out_args]
+        arg_str = []
+        for arg in arguments:
+            if isinstance(arg.type.dtype, DerivedType):
+                arg_str += [f'{arg.name.lower()}']
```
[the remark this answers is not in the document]
mlange05: Yes, I managed to remove all in `pygen.py`, except the comparison of the "intent" values (which are not symbols, and therefore need explicit down-case comparison). Pushed a single commit with those, please have a look.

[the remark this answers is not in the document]
reuterbal: Sorry for the delay. Looks great now, many thanks. Intent is absolutely fine and we usually take care to downcase them in the frontends but doesn't hurt to be careful there.

PR #44, opened 2023-03-08
Linter: Internalize functionality and scheduler integration stub

reuterbal:
Note: filing this early for test coverage and while testing offline.

This PR moves some of the functionality from `loki-lint.py` into the `loki.lint` module to facilitate testing. In that process, some of the functionality could be replaced by internal utility functions. With this, we now have a new entry point for linting: `lint_files`

This PR also bundles a first stub of a scheduler integration for the linter. It implements linting as a transformation that is applied across the scheduler's graph.
However, since linting is file based, this required equipping the Scheduler with a file-centric view of dependencies. For that purpose, the scheduler has been equipped with the ability to generate a file graph for the scheduler's item_graph, and optionally use that in `process`. This may be handy for the FileWriteTransformation, too.

The CLI script for checking in loki-lint.py is now only responsible for reading the YAML config and filling the config dict.


reuterbal:
I believe this is ready for a look.

mlange05 (approved):
Looks great, neat and tidy as always. Especially the `Scheduler` changes to traversal and the file graph are much appreciated. CI and regression tests are looking good, so GTG from me.

PR #45, opened 2023-03-09
Transformation to remove utility routine calls

mlange05:
This PR aims to rename the `transformations.dr_hook` submodule to `transformations.utility_routines` and adds a second transformation to just remove calls to utility routines generically. This new `RemoveCallsTransformation` also handles inclusion in inline-conditionals and optionally can also handle calls masked as `Intrinsic` Statements, eg. `WRITE(NULOUT,*)` - a very common pattern in IFS. Tests have been extended / added as appropriate, and it was integrated into the `loki-transform.py ecphys` mode. 

Final note: For the GPU-oriented transformations, we also remove DrHook calls, but on CPU we adjust it's label. For this reason, I've left the very specific `DrHookTransformation` as part of this renamed subpackage, but use the new one for SCC modes.


reuterbal (approved):
Very useful, neatly done and great test coverage. Looks good to me, many thanks!

PR #48, opened 2023-03-21
Dataflow analysis bug fixes

awnawab:
Bug fixes to the dataflow analysis module including:
- Correct handling of interface blocks
- Removal of loop variable symbol from `live` set in while loops
- Grouping of bodies in multi-conditional nodes preserved, meaning IR can now be correctly reconstructed after dataflow analysis pass

### DIFF
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -14,7 +14,7 @@ from loki.expression import FindVariables, Array, FindInlineCalls, ProcedureSymb
 from loki.tools import as_tuple, flatten
 from loki.types import BasicType
 from loki.visitors import Visitor, Transformer
-from loki import Subroutine
+from loki.subroutine import Subroutine
 
 __all__ = [
     'dataflow_analysis_attached', 'read_after_write_vars',
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -10,7 +10,7 @@ Collection of dataflow analysis schema routines.
 """
 
 from contextlib import contextmanager
-from loki.expression import FindVariables, Array, FindInlineCalls
+from loki.expression import FindVariables, Array, FindInlineCalls, ProcedureSymbol
 from loki.tools import as_tuple, flatten
 from loki.types import BasicType
 from loki.visitors import Visitor, Transformer
@@ -98,10 +98,11 @@ class DataflowAnalysisAttacher(Transformer):
     # Internal nodes
 
     def visit_Interface(self, o, **kwargs):
+        # Subroutines/functions calls defined in an explicit interface
         defines = set()
         for b in o.body:
             if isinstance(b, Subroutine):
-                defines = defines | self._symbols_from_expr(b.name)
+                defines = defines | set(as_tuple(ProcedureSymbol(name=b.name)))
         return self.visit_Node(o, defines_symbols=defines, **kwargs)
 
     def visit_InternalNode(self, o, **kwargs):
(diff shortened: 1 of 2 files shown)


loki/analyse/analyse_dataflow.py
```diff
@@ -96,6 +97,14 @@ def visit_Node(self, o, **kwargs):
 
     # Internal nodes
 
+    def visit_Interface(self, o, **kwargs):
+        # Subroutines/functions calls defined in an explicit interface
+        defines = set()
+        for b in o.body:
+            if isinstance(b, Subroutine):
+                defines = defines | set(as_tuple(ProcedureSymbol(name=b.name)))
```
reuterbal, line 105: please use `b.procedure_symbol` instead of creating the `ProcedureSymbol`

loki/analyse/analyse_dataflow.py
```diff
@@ -119,7 +128,7 @@ def visit_WhileLoop(self, o, **kwargs):
         # A while loop uses variables in its condition
         live = kwargs.pop('live_symbols', set())
         uses = self._symbols_from_expr(o.condition)
-        body, defines, uses = self._visit_body(o.body, live=live|{o.variable.clone()}, uses=uses, **kwargs)
+        body, defines, uses = self._visit_body(o.body, live=live, uses=uses, **kwargs)
```
reuterbal, line 131: If possible, could you add a minimal test for dataflow analysis on while loops? This is completely untested right now, which is probably how this bug found it's way in in the first place... (sorry!)

loki/analyse/analyse_dataflow.py
```diff
@@ -150,9 +159,12 @@ def visit_MultiConditional(self, o, **kwargs):
         vset = set(v for v in FindVariables().visit(o.values) if not v in query_args)
 
         uses = self._symbols_from_expr(as_tuple(eset)) | self._symbols_from_expr(as_tuple(vset))
-        body, defines, uses = self._visit_body(o.bodies, live=live, uses=uses, **kwargs)
+        body = ()
+        defines = set()
+        for b in o.bodies:
+            _b, defines, uses = self._visit_body(b, live=live, uses=uses, defines=defines, **kwargs)
```
reuterbal, line 165: Would it be possible to check the impact of this in [test_analyse_multiconditional](https://github.com/ecmwf-ifs/loki/blob/7fd6b8607967ee1d30281e36c63a86c54f1d054a/tests/test_analyse_dataflow.py#L413)? Not sure I'm understanding the original problem: did it re-order the cases?

reuterbal (changes requested):
Important: this is built on top of #47 and should be merged after this!
While it is relatively obvious from the diff, it would be good to point this out as a disclaimer at the top of the PR as it makes it easier to keep track of the order in which things need merging.

Overall very nicely done, many thanks! As this module is currently undertested (my fault!) but we might use this more and more in the future, I have asked for some additional testing if possible.

I'm not sure I understand the problem with MultiConditionals?

loki/analyse/analyse_dataflow.py
awnawab: The bodies of the multiconditional should be arranged in a nested tuple, with the outer level corresponding to the number of branches for the multiconditional. When we run a dataflow analysis pass, it flattens the bodies of the multiconditional. Whilst this isn't a problem for the dataflow analysis, it's a problem if we then want to regenerate Fortran code from the IR (the node is updated).

awnawab: And yes, we can test the impact of this by querying the size of the inner and outer layer of `o.bodies`. I will amend the test accordingly.

mlange05 (approved):
Looks very good, and increased test coverage is much appreciated! GTG from me.

PR #49, opened 2023-03-22
Add inline function calls as scheduler dependencies

awnawab:
One of the recommended ways of declaring inline function calls is via an explicit interface. This PR adds inline function calls declared via an explicit interface as scheduler dependencies. The more modern way of declaring inline function calls is to wrap them in a module and use module imports. Adding this functionality requires a larger update to the scheduler, and this PR is being prepared.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -93,8 +93,6 @@ class Item:
         self.source = source
         self.config = config or {}
         self.trafo_data = {}
-        self._children = ()
-        self._targets = ()
 
     def __repr__(self):
         return f'loki.bulk.Item<{self.name}>'
@@ -301,7 +299,7 @@ class Item:
         disabled = as_tuple(str(b).lower() for b in self.disable)
 
         # Base definition of child is a procedure call (for now)
-        children = self.calls + self._children
+        children = self.calls
 
         # Filter out local members and disabled sub-branches
         children = [c for c in children if c not in self.members]
@@ -514,7 +512,7 @@ class SubroutineItem(Item):
         return tuple(
             self._variable_to_type_name(call.name).lower()
             for call in FindNodes(CallStatement).visit(self.routine.ir)
-        )
+        ) + self.function_interfaces
 
     def _variable_to_type_name(self, var):
         """
@@ -548,24 +546,6 @@ class SubroutineItem(Item):
         )
 
         return names
-
-    @property
-    def children(self):
-        """
-        Extend the base class' definition of children for items of type :class:`SubroutineItem`.
-        """
-
-        self._children = self.function_interfaces
-        return super().children
-
-    @property
-    def targets(self):
-        """
-        Extend the base class' definition of targets for items of type :class:`SubroutineItem`.
-        """
-
-        self._targets = self.function_interfaces
-        return super().targets
 
 class ProcedureBindingItem(Item):
     """
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -20,8 +20,6 @@ from loki.tools import as_tuple
 from loki.tools.util import CaseInsensitiveDict
 from loki.visitors import FindNodes
 from loki.ir import CallStatement
-from loki.subroutine import Subroutine
-from loki.module import Module
 
 
 __all__ = ['Item', 'SubroutineItem', 'ProcedureBindingItem']
@@ -208,7 +206,7 @@ class Item:
 
     @property
     @abstractmethod
-    def inline_function_interfaces(self):
+    def function_interfaces(self):
         """
         All inline functions defined in this :class:`Item` via an explicit interface
         """
@@ -532,7 +530,7 @@ class SubroutineItem(Item):
         return type_name + var.name[pos:]
 
     @property
-    def inline_function_interfaces(self):
+    def function_interfaces(self):
         """
         Inline functions declared in the corresponding :any:`Subroutine`,
         or its parent :any:`Module`, via an explicit interface.
@@ -541,14 +539,13 @@ class SubroutineItem(Item):
         names = ()
         interfaces = self.routine.interfaces
 
-        if isinstance(self.scope, Module):
-            interfaces += self.scope.interfaces
+        if (scope := self.scope) is not None:
+            interfaces += scope.interfaces
 
-        for i in interfaces:
-            for b in i.body:
-                if isinstance(b, Subroutine):
-                    if b.is_function:
-                        names += as_tuple(b.name.lower())
+        names = tuple(
+            s.name.lower() for intf in interfaces for s in intf.symbols
+            if s.type.dtype.is_function
+        )
 
         return names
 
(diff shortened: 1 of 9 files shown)


loki/transform/dependency_transform.py
```diff
@@ -194,6 +219,39 @@ def rename_imports(self, source, imports, **kwargs):
...
+
+        # Transformer map to remove any outdated interfaces
+        removal_map = {}
+
+        for i in intfs:
+            for b in i.body:
+                if isinstance(b, Subroutine):
+                    target_symbol = b.name.lower()
+
+                    if targets is not None and target_symbol.upper() in targets:
```
reuterbal, line 242: Better to use the implicit case-insensitive string comparison: `b.name in targets`

reuterbal (changes requested):
_Note: This is based on #47 and #48 and should be merged after those have landed._

Very nicely done and a tremendously useful addition! Many thanks.
I have left a few comments how I think this could be improved. Most of them have to do with the fact that we are not cleanly encoding the concepts in the Scheduler at the moment and therefore there's not a good way for you to hook these changes in.
However, I would like to change this later on in a proper Scheduler overhaul and have therefore left a few proposals how to do this more aligned with the current way.

loki/frontend/regex.py
```diff
@@ -443,7 +443,8 @@ class SubroutineFunctionPattern(Pattern):
 
     def __init__(self):
         super().__init__(
-            r'^[ \t\w()=]*?(?P<keyword>subroutine|function)[ \t]+(?P<name>\w+)\b.*?$'
+            r'^(?P<return_type>real(\(kind=.+\))?|integer(\(kind=.+\))?|logical(\(kind=.+\))?|type\(.+\))?'
```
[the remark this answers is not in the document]
awnawab: Actually adding a named group for return_type was overkill, I never actually used the return_type but was just trying to capture it for completeness. It turns out that the only functionality I needed was checking for the `=` sign, and you added that in this commit: https://github.com/ecmwf-ifs/loki/commit/35605560c031bbff4d88b6866798851b5bf51abc. I prefer your implementation of a catch-all `prefix` named group, which can then be split as and when needed, so I'll return that line to how it is in main. I'll also update the `test_regex_function_inline_return_type` test. Thanks!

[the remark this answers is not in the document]
awnawab: Actually adding a named group for return_type was overkill, I never actually used the return_type but was just trying to capture it for completeness. It turns out that the only functionality I needed was checking for the `=` sign, and you added that in this commit: https://github.com/ecmwf-ifs/loki/commit/35605560c031bbff4d88b6866798851b5bf51abc. I prefer your implementation of a catch-all `prefix` named group, which can then be split as and when needed, so I'll return that line to how it is in main. I'll also update the `test_regex_function_inline_return_type` test. Thanks!

loki/bulk/item.py
```diff
@@ -520,6 +531,44 @@ def _variable_to_type_name(self, var):
...
+    def inline_function_interfaces(self):
+        """
+        Inline functions declared in the corresponding :any:`Subroutine`,
+        or its parent :any:`Module`, via an explicit interface.
+        """
+
+        names = ()
+        interfaces = self.routine.interfaces
+
+        if isinstance(self.scope, Module):
```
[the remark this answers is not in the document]
awnawab: Thanks for the tip! And for the opportunity to learn a new Python operator ;-)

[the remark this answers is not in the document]
reuterbal: That one is a controversial one but I consider it to be actually useful ;-)

loki/transform/dependency_transform.py
reuterbal: Yes, but `b.name` is a symbol that has the mixin, and it works in this direction as well:
```python
>>> l = ['a', 'b', 'c']
>>> from loki import Variable
>>> Variable(name='a') in l
True
>>> Variable(name='A') in l
True
>>> Variable(name='aa') in l
False
```

awnawab: Ah yes, my bad, sorry, was suffering from a case of "ramadan brain" there 😆

loki/frontend/regex.py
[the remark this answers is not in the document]
reuterbal: Great, makes sense

loki/transform/dependency_transform.py
```diff
@@ -94,10 +96,29 @@ def transform_subroutine(self, routine, **kwargs):
...
+        self.rename_interfaces(routine, intfs=intfs, **kwargs)
+
         if role == 'kernel' and self.mode == 'strict':
             # Re-generate C-style interface header
             self.generate_interfaces(routine)
 
+    def update_result_var(self, routine):
+        """
+        Update name of result variable for function calls.
+        """
```
reuterbal, line 110: Spotted this only now. Does this work as expected with functions that are declared with a different result name?

```f90
function my_func result(ret)
    integer ret
    ret = 5
end function my_func
```

reuterbal (commented):
Many thanks for addressing the comments. Looks great now. One more question about the dependency injection transformation but otherwise GTG!

loki/transform/dependency_transform.py
awnawab: Ah, great catch! You are right, it won't work. It needs a test to check if `result_name` matches `Subroutine.name`. I'll update and add a test 👍

reuterbal (approved):
Looks great, thanks for confirming and fixing!

tests/test_scheduler.py
```diff
@@ -1558,3 +1558,31 @@ def test_scheduler_qualify_names():
...
+    my_config = config.copy()
+    my_config['routine'] = [
+        {
+            'name': 'driver',
+            'role': 'driver'
+        }
+    ]
+
+    scheduler = Scheduler(paths=here/'sources/projInlineCalls', config=my_config, frontend=frontend)
+    print(scheduler.items)
```
mlange05, line 1578: Debug leftover?

mlange05 (commented):
Looks very good, and again, many thanks for increased test coverage. There's still one corner case flagged by @reuterbal that needs addressing (unless I missed something?), and I'highlighted a (presumably) debug print that we should probably drop. Otherwise GTG after these are addressd.

tests/test_scheduler.py
awnawab: Ah yes that is a debug print, I'll remove that and push an update. And are you referring to B's spot that `update_result_var` won't work for a function with an explicitly declared result name? If so, I've fixed that one.

loki/transform/dependency_transform.py
mlange05: Sorry, did not understand the fix, but cleared up offline. All good now

tests/test_scheduler.py
mlange05: Yes, sorry had not understood the fix, but happy now!

mlange05 (approved):
OK, all addressed, GTG from me.

PR #50, opened 2023-03-23
Scheduler enhancements for dealing with derived types

reuterbal:
A set of Scheduler changes aimed at enhancing the capabilities when working with derived types and dependencies from typebound procedure calls.

* When handing successors for an `Item` to a transformation, it follows the indirection of typebound procedures and yields the actual subroutine items in addition to the type binding
* In addition to enriching calls, a new `enrich_types` method provides type annotation for derived types after the initial parse
* A new `add_dependencies` method allows to introduce additional dependencies to an existing Scheduler graph. This may become useful in multiple situations in the future. I am using this to post-hoc introduce dependencies due to inline calls via a bespoke transformation that I will file as a separate PR. This includes a new mechanism to invalidate cached properties.
* Not related to derived types, the scheduler is now able to break circular dependencies that are introduced by recursive Fortran procedures.

This will likely have conflicts with #49 and I'm happy to rebase on top of these changes once they are merged to main.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -113,6 +113,13 @@ class Item:
     def __hash__(self):
         return hash(self.name)
 
+    def clear_cached_property(self, property_name):
+        """
+        Clear the cached value for a cached property
+        """
+        if property_name in self.__dict__:
+            del self.__dict__[property_name]
+
     @property
     def scope_name(self):
         """
@@ -169,7 +176,7 @@ class Item:
         list of :any:`Import`
         """
 
-    @cached_property
+    @property
     def qualified_imports(self):
         """
         Return the mapping of named imports (i.e. explicitly qualified imports via a use-list
@@ -181,7 +188,7 @@ class Item:
             for symbol in import_.symbols + tuple(s for _, s in as_tuple(import_.rename_list))
         )
 
-    @cached_property
+    @property
     def unqualified_imports(self):
         """
         Return names of imported modules without explicit ``ONLY`` list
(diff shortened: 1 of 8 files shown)


loki/bulk/item.py
```diff
@@ -113,6 +113,13 @@ def __eq__(self, other):
     def __hash__(self):
         return hash(self.name)
 
+    def clear_cached_property(self, property_name):
```
mlange05, line 116: The fact that we need this is worrying. If a property needs to be wiped, it might not need to be cached in the first place, no? From what I can see this is only the `Imports` in a module/routine spec, right? Could we simply live with re-deriving these on the fly, as they are accessed quite rarely, I would think?

mlange05 (commented):
Very nice, and very much appreciated, as we're gearing up to really overhaul the `Scheduler`. I only left one comment regarding the cache invalidation of properties, and I think this now needs a rebase over main to make merging clean. I'd also love to run through the full ec-phys regression suite, but otherwise, GTG.

So, GTG in principle, but marking as comment until discussion/rebase and hopefully a test run through the big one. :wink:

loki/bulk/item.py
reuterbal: I suspect we could - the reason I didn't simply remove the caching is that we do in fact depend on the caching behaviour, at least after applying dependency injection (and the associated renaming). So, a callback mechanism that updates items if they are renamed might be something for the Scheduler wishlist. See LOKI-42 where this was flagged a while ago. Might be worth checking the runtime impact as well.

mlange05 (approved):
Great, question answered and regression tests are fine too. GTG from my side.

loki/bulk/item.py
mlange05: Ok, that sounds sensible. We'll address this in a follow-up.

PR #52, opened 2023-03-23
loki-transform: Small CMake refactoring, C preprocessing, absolute paths

reuterbal:
A small utility PR for the loki-transform script and CMake integration:
* CMake has been slightly refactored to allow specifying the command to use when transforming a target (defaults to ecphys to not break existing behaviour)
* preprocessing can now be enabled optionally for planning and transformation via a CMake option
* A new feature allows to copy unmodified files from the target into the build directory (useful when replicating a target without using dependency injection mechanisms)
* Support for using absolute paths instead of relative paths (required when doing some trickery with relative paths in the CMake setup to add CMake build configuration from external to an existing repository)


cmake/loki_transform.cmake
```diff
@@ -573,30 +582,84 @@ endfunction()
...
+    _loki_transform_parse_target_args( loki_transform_command )
     _loki_transform_env_setup()
 
-    ecbuild_debug( "COMMAND ${_LOKI_TRANSFORM} ecphys ${_ARGS}" )
+    ecbuild_debug( "COMMAND ${_LOKI_TRANSFORM} ${_PAR_COMMAND} ${_ARGS}" )
 
     add_custom_command(
         OUTPUT ${_PAR_OUTPUT}
-        COMMAND ${_LOKI_TRANSFORM} ecphys ${_ARGS}
+        COMMAND ${_LOKI_TRANSFORM} ${_PAR_COMMAND} ${_ARGS}
```
mlange05, line 606: Does this make an implicit assumption about all sub-modes of `loki-transform.py` having the same set of arguments. Is this actually true? Could I trigger C-transpilation via this from CLOUDSC? Not a showstopper, just curious, as the safety checks have now moved to the cmake layer....

mlange05 (commented):
Looks very nice and GTG in principle. I'd like to test this with the full ec-phys regression test, and I left a small question about argument checking. But otherwise, GTG once I can confirm that full ec-phys setup works as expected.

cmake/loki_transform.cmake
reuterbal: It doesn't imply a requirement on the sub-modes, only provides the opportunity to enable additional modes of `loki-transform.py` via the same CMake layer. I consider the ecphys interface to be the de-facto definition, but if we can homogenize this across more modes eventually it's certainly not a drawback. But it would definitely not work right now to call the transpilation that way.
Motivation is mostly a way for me to inject an `ecrad` mode without having to duplicate all of the CMake layer.

reuterbal: With regards to arguments checking, I would prefer leaving that to the Python layer as it would otherwise require replicating that logic. So if a mode is called via the wrong CMake function, this will fail in the call to loki-transform.

mlange05: Yes, absolutely agree about the Python layer being the one to check that arguments are sane; and happy to go ahead with this PR as is. I was merely trying to point out that we need to be careful which args the cmake layer passes down, if different sub-modes accept different arguments. But more of an academic consideration at this point, so not a showstopper.

PR #54, opened 2023-03-24
Transformation to resolve calls to typebound procedures

reuterbal:
~~_Note: This is based on #51._~~
_Update: This is now separated from #51._

A new transformation that allows resolving calls to typebound procedures by replacing them with direct calls to the corresponding routine. For example:
Given a derived type
```f90
TYPE SOME_TYPE
CONTAINS
    PROCEDURE :: SOME_ROUTINE => SOME_TYPE_SOME_ROUTINE
END TYPE SOME_TYPE
```
then a call statement
```f90
CALL SOME_TYPE_VAR%SOME_ROUTINE(ARG1, ARG2, KW=KWARG)
```
is replaced by 
```f90
CALL SOME_TYPE_SOME_ROUTINE(SOME_TYPE_VAR, ARG1, ARG2, KW=KWARG)
```
This takes also care of adding the relevant import to the caller. The same is applied also to InlineCalls in expressions. As a side-effect, these calls are identified as additional dependencies and made available after the Transformer pass. These can then be provided to the scheduler, as introduced in #50.

### DIFF
--- a/transformations/tests/test_transform_derived_types.py
+++ b/transformations/tests/test_transform_derived_types.py
@@ -5,11 +5,37 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+from shutil import rmtree
 import pytest
 
-from loki import OMNI, Module, FindNodes, CallStatement
+from loki import OMNI, Module, Scheduler, FindNodes, FindInlineCalls, CallStatement, gettempdir
 from conftest import available_frontends
-from transformations import DerivedTypeArgumentsTransformation
+from transformations import (
+    DerivedTypeArgumentsTransformation,
+    TypeboundProcedureCallTransformation
+)
+
+
+@pytest.fixture(name='config')
+def fixture_config():
+    """
+    Default configuration dict with basic options.
+    """
+    return {
+        'default': {
+            'mode': 'idem',
+            'role': 'kernel',
+            'expand': True,
+            'strict': True,
+        },
+        'routines': [
+            {
+                'name': 'driver',
+                'role': 'driver',
+                'expand': True,
+            },
+        ]
+    }
 
 
 @pytest.mark.parametrize('frontend', available_frontends())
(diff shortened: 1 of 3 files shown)


transformations/transformations/derived_types.py
```diff
@@ -168,3 +169,271 @@ def flatten_derived_args_routine(self, routine):
...
+        return proc_symbol.type.bind_names[0]
+
+    parent = proc_symbol.parents[0]
+    if parent.type.dtype.typedef is not BasicType.DEFERRED:
+        # Fiddle our way through derived type nesting until we obtain the symbol corresponding
+        # to the procedure-binding in the TypeDef
+        local_parent = None
+        local_var = parent
+        try:
+            for local_name in proc_symbol.name_parts[1:]:
```
mlange05, line 209: [No action needed] I've seen this pattern before. Shouldn't we, in theory, be able to use the symbol-table / scoping mechanism to do this implicitly? Or do we need to first make the scoping stuff cleaner for that to work?

mlange05 (approved):
This looks very powerful, and as always very cleanly implemented. I've left two comments that don't really need any addressing in code. So this should be GTG pretty much as is.

One side-note I have is that some of the back-bending in these utilities highlights that we need to reconsider a few things about scopes and the way type-bindings/parentage are handled currently. But that is for another day, I think.

transformations/tests/test_transform_derived_types.py
```diff
@@ -124,3 +153,208 @@ def test_transform_derived_type_arguments(frontend):
...
+    call data%stuff(1)%arr(1)%add(1)
+    mysum = data%stuff(1)%total_sum() + data%stuff(2)%total_sum()
+    associate (some => data%some)
+        mysum = mysum + some%some_func()
+    end associate
+    call data%print
+end subroutine driver
+    """.strip()
+
+    workdir = gettempdir()/'test_transform_typebound_procedure_calls'
```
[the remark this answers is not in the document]
reuterbal: Yes, I've had the same thought. And we do also have in one of the test modules the mechanics for automatic set-up before and clean-up after (even failed) tests, which would be a proper way of handling even larger test source trees.

transformations/transformations/derived_types.py
reuterbal: The handling of procedure bindings does indeed need an overhaul. Much of it is down to Fortran and the fact that this allows for multiple flavours, all of which come with varying degrees of indirection - which we currently replicate in the IR/symbol table. 
The particular situation here is unlikely to completely disappear because we don't require type information to be present, which would be required to create a tighter link between declared symbols and their bindings.

PR #56, opened 2023-03-29
Make rescoping in post_apply optional

reuterbal:
Due to some unfortunate nesting, we are currently calling excessively often `rescope_symbols` after a transformation.
This is because `post_apply` was called for every subroutine and every module to which a transformation was applied, even if the actual transformation only implements a handler for one of them.

Out of caution, we had put a `rescope_symbols()` call into this `post_apply` method to catch some of the more subtle scope abuse. However, it seems that most transformations are actually fairly good in avoiding scoping errors, as tests pass even without that.

Now, either already at that time or slightly after, the `rescope_symbols()` call has learned to be properly recursive, i.e. also fix scopes for its member routines if called on a subroutine or module. This meant an abundantly often call to `rescope_symbols`, which can cause a significant overhead, as it isn't cheap either.

On practical test cases this hasn't showed up so far, as typical IFS files only contain a single subroutine, so it would only be called once. However, in code written in an object-oriented fashion, everything is encapsulated in modules, which contain multiple subroutines/functions, and there it makes a huge difference.

My proposal (as implemented in this PR) would be to make the rescoping entirely optional but allow activating this for quick fixing of "legacy" transformations where there is a scoping issue.

This PR contains three contributions:
1. Rework the `apply` and `post_apply` logic in `Transformation` to avoid excessive calls to `post_apply`
2. Make rescoping entirely optional and off by default
3. Fix a few type errors in IR node constructor uses.

### DIFF
--- a/loki/transform/transform_parametrise.py
+++ b/loki/transform/transform_parametrise.py
@@ -275,7 +275,7 @@ class ParametriseTransformation(Transformation):
                         name = str(call.name)
                         successor_map[name].trafo_data[self._key][str(arg_map_reversed[call.arguments[index]])] = \
                             dic2p[call.arguments[index].name]
-                    arguments = [arg for i, arg in enumerate(call.arguments) if arg not in vars2p]
+                    arguments = tuple(arg for arg in call.arguments if arg not in vars2p)
                     call_map[call] = call.clone(arguments=arguments)
             routine.body = Transformer(call_map).visit(routine.body)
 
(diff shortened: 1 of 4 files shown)


mlange05 (approved):
Looks great, and very neatly done! GT from me.

mlange05:
Confirmed clean with regression tests, and thus GTG. Next on in....

PR #61, opened 2023-04-11
Remove unused import and variable from hook

rolfhm:
The dangling variables and imports make the compiler vomit out a ton of warnings. This will remove those as well as the calls to DR_HOOK

### DIFF
--- a/transformations/transformations/utility_routines.py
+++ b/transformations/transformations/utility_routines.py
@@ -64,6 +64,17 @@ class DrHookTransformation(Transformation):
 
         routine.body = Transformer(mapper).visit(routine.body)
 
+        #Get rid of unused import and variable
+        if self.remove:
+            for imp in FindNodes(Import).visit(routine.spec):
+                if 'LHOOK' in [s.name for s in imp.symbols]:
+                    mapper[imp] = None
+
+            routine.spec = Transformer(mapper).visit(routine.spec)
+
+            #Remove unused zhook_handle
+            routine.variables = as_tuple(v for v in routine.variables if v.name != 'ZHOOK_HANDLE')
+
 
 class RemoveCallsTransformation(Transformation):
     """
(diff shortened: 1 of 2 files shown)


transformations/transformations/utility_routines.py
```diff
@@ -64,6 +64,17 @@ def transform_subroutine(self, routine, **kwargs):
 
         routine.body = Transformer(mapper).visit(routine.body)
 
+        #Get rid of unused import and variable
+        if self.remove:
+            for imp in FindNodes(Import).visit(routine.spec):
+                if 'LHOOK' in [s.name for s in imp.symbols]:
```
reuterbal, line 70: It should be sufficient to test `if 'LHOOK' in imp.symbols`, since this will also take care of making the comparison case-insensitive.

reuterbal (changes requested):
Many thanks for this, a very useful addition to the transformation! Can you please amend the tests in https://github.com/ecmwf-ifs/loki/blob/main/transformations/tests/test_utility_routines.py with checks that verify that
a) Import and variable exist before applying the transformation
b) Import and variable are removed after applying the transformation

rolfhm:
I added the tests and cleaned up a bit.

What is the mode thing used for? 

It looks like it only tests for one line conditionals, so if there's a `if (lhook) then`, it won't remove it and compilation might fail. In a version I wrote for myself, I just removed the whole block if I found a conditional, with `lhook`. That won't be possible with the mode thing though.

transformations/transformations/utility_routines.py
```diff
@@ -64,6 +64,17 @@ def transform_subroutine(self, routine, **kwargs):
 
         routine.body = Transformer(mapper).visit(routine.body)
 
+        #Get rid of unused import and variable
+        if self.remove:
+            for imp in FindNodes(Import).visit(routine.spec):
+                if imp.module == 'yomhook':
```
reuterbal, line 70: I would use `imp.module.lower() == 'yomhook'` just to be sure that this is case-insensitive.

rolfhm: Added `.lower()`

mlange05:
Looks very nice, and confirmed good with ec-physics regression test. GTG, going in.

PR #62, opened 2023-04-11
Add module imports as scheduler dependencies

awnawab:
Applying Loki transformations to more complex control flows requires generating GPU offload instructions for global (module) variable imports in an automated way. To write these transformations, we first need to extend the definition of scheduler dependencies to also include global variable imports.

This PR addresses the latter, and tries to provide a clean way to add global variable imports as scheduler dependencies. Importantly, the default behaviour remains unchanged; module imports are not added as dependencies unless explicitly stated in the scheduler config. Therefore existing scheduler workflows should not be affected.

Finally, this will likely have some conflicts with #50, and so a rebase is needed once that is merged.

NB: This now relies on a bug-fix in #68.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -698,6 +698,17 @@ class GenericImportItem(Item):
         """
         return ()
 
+    @property
+    def function_interfaces(self):
+        """
+        Empty tuple as generic import items cannot include interface blocks
+
+        Returns
+        -------
+        tuple
+        """
+        return ()
+
     @cached_property
     def imports(self):
         """
@@ -725,8 +736,8 @@ class GenericImportItem(Item):
                 if i.spec.name == self.local_name:
                     for b in i.body:
                         if isinstance(b, ProcedureDeclaration):
-                           for s in b.symbols:
-                               _children += as_tuple(s.name)
+                            for s in b.symbols:
+                                _children += as_tuple(s.name)
                         elif isinstance(b, Subroutine):
                             if b.is_function:
                                 _children += as_tuple(b.name)
@@ -757,6 +768,17 @@ class GlobalVarImportItem(Item):
     def members(self):
         """
         Empty tuple as variable imports have no member routines
+
+        Returns
+        -------
+        tuple
+        """
+        return ()
+
+    @property
+    def function_interfaces(self):
+        """
+        Empty tuple as global variable imports cannot include interface blocks
 
         Returns
         -------
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -676,7 +676,7 @@ class GenericImportItem(Item):
 
     def __init__(self, name, source, config=None):
         name_parts = name.split('#')
-        assert len(name_parts) > 1 #only accept fully-qualified module imports
+        assert len(name_parts) > 1 and all(name_parts) #only accept fully-qualified module imports
         super().__init__(name, source, config)
         assert self.scope
 
@@ -753,7 +753,7 @@ class GlobalVarImportItem(Item):
 
     def __init__(self, name, source, config=None):
         name_parts = name.split('#')
-        assert len(name_parts) > 1 #only accept fully-qualified module imports
+        assert len(name_parts) > 1 and all(name_parts) #only accept fully-qualified module imports
         super().__init__(name, source, config)
         assert self.scope
 
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -35,11 +35,12 @@ class Item:
     sub-trees.
 
     Depending of the nature of the work item, the implementation of :class:`Item` is
-    done in subclasses :class:`SubroutineItem` and :class:`ProcedureBindingItem`.
+    done in subclasses :class:`SubroutineItem`, :class:`ProcedureBindingItem`,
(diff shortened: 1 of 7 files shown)


loki/bulk/item.py
```diff
@@ -628,3 +653,141 @@ def calls(self):
...
+                            if b.is_function:
+                                _children += as_tuple(b.name)
+
+        return _children
+
+
+class GlobalVarImportItem(Item):
+    """
+    Implementation of :class:`Item` to represent a global variable import.
+    These only appear at the end of dependency trees and do not have children.
```
reuterbal, line 742: Considering the ball of wool that our dependency trees (dependency shrubbery?) tend to resemble, I think "end of dependency tree" may not be well defined. The intention is that it doesn't have any dependencies itself, right?

What about (not uncommon) situations like this:

```f90
module some_mod
  use parkind1, only: jprb

  real(kind=jprb) :: var
end module some_mod
```

with

```f90
subroutine some_routine
  use some_mod, only: var
  ...
end subroutine some_routine
```

Should this generate another import dependency? I would (at least in principal) expect something like
```
SubroutineItem(#some_routine) --> GlobalVarImportItem(some_mod#var) --> GlobalVarImportItem(parkind1#jprb)
```
although that is probably tricky to derive in practice (and likely not necessary for our purposes).

 Maybe worth adding a test for that scenario to clarify the behaviour?

loki/bulk/scheduler.py
```diff
@@ -222,7 +228,26 @@ def _discover(self):
...
-            for r in module.subroutines + tuple(module.typedefs.values())
+            for r in module.subroutines + tuple(module.typedefs.values()) + module.variables
+        )
+        self.obj_map.update(
+            (f'{module.name}#{r.spec.name}', obj)
+            for obj in obj_list for module in obj.modules
+            for r in module.interfaces if r.spec
+        )
+
+        # Create maps of routines and module variables to help with fast sorting in create_item later
```
reuterbal, line 239: Not convinced that performance justifies the bookkeeping overhead for additional maps here: Finding this out after a lookup in obj_map should be fairly quick, especially compared to the cost of transformations? Happy to be convinced otherwise, though!

reuterbal (commented):
Filing these as comments only for now, as we will have to estimate the fallout from merging #50. Please have a look at the comments to see which of those make sense in the meantime.

In general looks good and is an elegant solution given the structure into which it has to be embedded!

loki/bulk/item.py
awnawab: Firstly, dependency shrubbery is spot on 😆 what I wanted to say here is that a `GlobalVarImportItem` does not have any children (I intentionally left out the mechanism to add children). So in your example, `some_mod#var` does indeed depend on `parkind#jprb`. But I intended `GlobalVarImportItem` to just be dedicated to variables that are used to store data, and not a catchall for any import.

According to my (probably quite wooly to continue the analogy) nomenclature, `parkind1#jprb` would be a `GenericImportItem`. But I didn't add a mechanism to add `GenericImportItem`s as children of `GlobalVarImportItem` because I didn't think there would be any practical need for it.

I can certainly add a test to clarify this behaviour. But do let me know please if you would like me to update the naming convention to something clearer, or even change the behaviour of `GlobalVarImportItem` to have `GenericImportItem` children, even if we don't do anything with them.

loki/bulk/scheduler.py
```diff
@@ -222,7 +228,26 @@ def _discover(self):
...
+        self.routine_map = CaseInsensitiveDict(
+            (f'#{r.name}', obj) for obj in obj_list for r in obj.subroutines
+        )
+        self.routine_map.update(
+            (f'{module.name}#{r.name}', obj)
+            for obj in obj_list for module in obj.modules
+            for r in module.subroutines
+        )
+        self.modvars_map = CaseInsensitiveDict(
```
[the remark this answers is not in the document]
awnawab: Agreed, but I've removed these extra maps now as per your suggestion below.

loki/bulk/scheduler.py
awnawab: Yes agreed, those extra maps felt a bit clunky looking back at it and I've gotten rid of them. Thanks for the tip!

loki/bulk/scheduler.py
```diff
@@ -345,7 +370,11 @@ def create_item(self, name):
         debug(f'[Loki] Scheduler creating Item: {name} => {sourcefile.path}')
         if '%' in name:
             return ProcedureBindingItem(name=name, source=sourcefile, config=item_conf)
-        return SubroutineItem(name=name, source=sourcefile, config=item_conf)
+        if self.routine_map.get(name):
```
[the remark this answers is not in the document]
awnawab: Taken care of indirectly by removing the extra maps.

mlange05 (approved):
Yes, this looks good to me, and presents an excellent new data point for the eventual re-write of the scheduler. 

I've not yet run this through regression with EC-physics, but will do so next. I'll mark this as approved, but it still needs update with develop and then final check with full-scale regression. After that, GTG, I think.

mlange05:
Ohhh, ignore me. Seems all conflicts have been resolved. Will run regression next then.

mlange05:
Confirmed with regression suite, all good, going in.

PR #63, opened 2023-04-12
Insert contained subroutine bodies

rolfhm:
Here's the routine I use to insert contained subroutine bodies.

Should I clean up the documentation, add a test and make a pull request?

BTW, is there any way to squash commits on github?


reuterbal:
Thanks! I believe @mlange05 has looked into this some more and may have come up with a similar utility himself - so I suggest to wait for his comments before spending time on clean-up and tests.

reuterbal:
PS: I don't know if squashing is available via a Github functionality but you can do this in the regular Git way, e.g. by doing a `git rebase -i origin/main` where you can select the commits to squash, and then force-push the branch.

PR #65, opened 2023-04-14
Limited recursion and made boz code gfortran compliant

rolfhm:
Modified `test_derived_type_linked_list` to iterate to `min(1000, getrecursionlimit()-100)`.
On my system, `getrecursionlimit()` returned 1000. I set 100 as a buffer somewhat arbitrarily. 50 was not enough.

Changed the boz test code to be compliant with the Fortran standard. gfortran 12 refused to compile and looks like `n1 = B'00000'` has not been acceptable since gfortran 10.

### DIFF
--- a/tests/test_derived_types.py
+++ b/tests/test_derived_types.py
@@ -5,11 +5,13 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+from sys import getrecursionlimit
+from inspect import stack
+
 from pathlib import Path
 import re
 import pytest
 import numpy as np
-from sys import getrecursionlimit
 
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Thanks, this is great and I could indeed reproduce the issues.

I have left a remark about import order (I suspect pylint will fail on this anyway) and a suggestion if you wanted to engineer this further (but that's entirely optional, of course). Otherwise good to go!

reuterbal (changes requested):
As discussed offline, leave the recursion limit as is. Could you please fix the import order in the test?
> Error: tests/test_derived_types.py:12:0: C0411: standard import "from sys import getrecursionlimit" should be placed before "import pytest" (wrong-import-order)

PR #66, opened 2023-04-16
Global variable offload transformation

awnawab:
This is a transformation that tries to automate the offload of module variables imported below the driver layer. Although the transformation only supports OpenACC derivatives for the moment, it has a simple logic that can be easily be extended to other offload paradigms:
- `process_kernel` collects the list of variables to offload
- `transform_module` adds device-side variable declarations
- `process_driver` adds the relevant offload/copy-back directives.

I have also added limited support for module imports of derived-types. I imagined this might be a reasonably common pattern in the IFS. If we don't want to support derived-types, then the transformation can be made even simpler.

NB: This is stacked on top of #74.


transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+                                           content='exit data copyout(' + ','.join(self._exit_data) + ')'),)
+        if self._acc_copyout:
+            update_host += as_tuple(Pragma(keyword='acc', content='update self(' + ','.join(self._acc_copyout) + ')'),)
+
+        # replace Loki pragmas with acc data/update pragmas
+        pragma_map = {}
+        for pragma in [pragma for pragma in FindNodes(Pragma).visit(routine.body) if pragma.keyword == 'loki']:
+            if 'update_device' in pragma.content:
+                if update_device:
+                    pragma_map.update({pragma: update_device})
```
mlange05, line 115: Why not just `pragma_map[pragma] = update_device`? (Also below...)

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        Collect the set of module variables to be offloaded.
+        """
+
+        # filter out procedure imports
+        imports = FindNodes(Import).visit(routine.spec)
+        calls = [c.name for c in FindNodes(CallStatement).visit(routine.body)]
+
+        imports = [i for i in imports if i.symbols not in calls]
+        imports = [i for i in imports if i.symbols not in FindInlineCalls().visit(routine.body)]
```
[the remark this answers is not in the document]
mlange05: Or, even simply do `calls += FindInlineCalls().visit(routine.body)` above, and only build `imports` once.

mlange05 (changes requested):
OK, first of all, this looks really great and adds a very valuable feature. I've added a few minor comments on coding style, but most of these are cosmetic. This is, however, a very impressive piece of work. So many, many thanks, @awnawab 

I've also highlighted a few points that might require further attention when we integrate this with the full SCC transform (eg. for physics), but these need not be addressed in this first round. 

I am wondering a little about the use of the `GlobalVarImportItem` and the changes to the dependency transform. In particular, I'm not sure I understand why the latter are necessary, as I'm not sure we should regard imported symbols as "targets". This one might require a quick off-line discussion, but should not become a showstopper.

transformations/transformations/global_var_offload.py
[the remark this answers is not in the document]
awnawab: Agreed! I'm still finding my way with python/Loki performance considerations so tips like these are very useful, thanks! 🙏

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+        for k, v in new_import_map.items():
+            new_imports += as_tuple(Import(k, symbols=tuple(Variable(name=s, scope=routine) for s in v)))
+
+        # add new imports to driver subroutine sepcification
+        import_pos = 0
+        if (old_imports := FindNodes(Import).visit(routine.spec)):
+            import_pos = routine.spec.body.index(old_imports[-1]) + 1
+        if new_imports:
+            routine.spec.insert(import_pos, Comment(text=
+                                '!.....Adding global variables to driver symbol table for offload instructions'))
```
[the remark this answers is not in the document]
awnawab: Agreed, done.

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+                                             content='enter data create(' + ','.join(self._enter_data_create) + ')'),)
+        if self._exit_data:
+            update_host += as_tuple(Pragma(keyword='acc',
+                                           content='exit data copyout(' + ','.join(self._exit_data) + ')'),)
+        if self._acc_copyout:
+            update_host += as_tuple(Pragma(keyword='acc', content='update self(' + ','.join(self._acc_copyout) + ')'),)
+
+        # replace Loki pragmas with acc data/update pragmas
+        pragma_map = {}
+        for pragma in [pragma for pragma in FindNodes(Pragma).visit(routine.body) if pragma.keyword == 'loki']:
```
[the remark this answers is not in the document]
awnawab: Agreed, done.

transformations/transformations/global_var_offload.py
awnawab: Agreed, done.

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+
+        item = kwargs['item']
+
+        # bail if not global variable import
+        if not isinstance(item, GlobalVarImportItem):
+            return
+
+        # confirm that var to be offloaded is declared in module
+        symbol = item.name.split('#')[-1].lower()
```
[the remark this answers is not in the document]
awnawab: A `GlobalVarImportItem` is generated per symbol.

tests/test_transformation.py
```diff
@@ -260,6 +260,114 @@ def test_dependency_transformation_module_imports(frontend):
...
+    USE kernel_mod, only: some_const
+    INTEGER, INTENT(INOUT) :: a, b, c
+
+    CALL kernel(a, b ,c)
+END SUBROUTINE driver
+""", frontend=frontend)
+
+    transformation = DependencyTransformation(suffix='_test', module_suffix='_mod')
+    kernel.apply(transformation, role='kernel')
+    driver.apply(transformation, role='driver', targets=('kernel', 'some_const'))
```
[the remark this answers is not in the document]
reuterbal: Just to chime in here (I still haven't looked through all the changes, sorry!): In the Scheduler rewrite we can (and want) to consider also other dependency types than control flow, which therefore become "live edges". And considering the scope of the transformation (traverse the dependency tree to global variable leaf nodes), I think it is actually in line with the concept of "targets" - filtering these symbols from the `targets` list when not required by a tree traversal should become a functionality of the SFilter class.

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+        assert symbol in [s.name.lower() for s in module.variables]
+
+        # do nothing if var is a parameter
+        if module.symbol_map[symbol].type.parameter:
+            return
+
+        # check if var is already declared
+        pragmas = [p for p in FindNodes(Pragma).visit(module.spec) if p.keyword.lower() == 'acc']
+        for p in pragmas:
+            if re.search(fr'\b{symbol}\b', p.content.lower()):
```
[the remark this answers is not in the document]
awnawab: The native string comparison was returning false positives when one variable's name was a subset of another's, .e.g. `DFIM` and `DFIMOFR`.

transformations/tests/test_global_var_offload.py
```diff
@@ -0,0 +1,181 @@
...
+    assert imports[0].module.lower() == 'module_derived_type'
+    assert set(s.name for s in imports[0].symbols) == {'p', 'p0'}
+
+    # check that existing acc pragmas have not been stripped and update device/update self added correctly
+    pragmas = FindNodes(Pragma).visit(driver.body)
+    assert len(pragmas) == 5
+    assert all(p.keyword.lower() == 'acc' for p in pragmas)
+
+    assert 'enter data copyin' in pragmas[0].content
+    assert re.search(r'\bp0%x\b', pragmas[0].content)
```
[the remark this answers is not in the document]
awnawab: Agreed, the regex search is overkill here. I've replaced with native string search.

tests/test_transformation.py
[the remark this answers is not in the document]
awnawab: Yeah I add `GlobalVarImportItem`s as targets so that within the transformation I can filter out the imports I am interested in based on the `disable\ignore\block` lists in the scheduler config.

loki/transform/dependency_transform.py
```diff
@@ -177,6 +184,20 @@ def rename_imports(self, source, imports, **kwargs):
...
+            calls = ()
+            for routine in source.subroutines:
+                calls += as_tuple(str(c.name).upper() for c in FindNodes(CallStatement).visit(routine.body))
+                calls += as_tuple(str(c).upper() for c in FindInlineCalls().visit(routine.body))
+        else:
+            calls = as_tuple(str(c.name).upper() for c in FindNodes(CallStatement).visit(source.body))
+            calls += as_tuple(str(c).upper() for c in FindInlineCalls().visit(source.body))
+
+        # Import statements still point to unmodified call names
+        calls = [call.replace(f'{self.suffix.upper()}', '') for call in calls]
```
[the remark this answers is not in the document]
awnawab: We don't want to rename the global variables themselves or the modules they are declared in because they are probably used throughout non-device code, and we want to keep those variables coherent with the ones used on device. But yes, let's please discuss offline.

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,208 @@
...
+    :any:`Transformation` class that facilitates insertion of offload directives
+    for module variable imports. The following offload paradigms are currently supported:
+
+    * OpenACC
+
+    It comprises of three main components. ``process_kernel`` which collects a set of
+    imported variables to offload, ``transform_module`` which adds device-side declarations
+    for the imported variables to the relevant modules, and ``process_driver`` which adds
+    offload instructions at the driver-layer. ``!$loki update_device`` and ``!$loki update_host``
+    pragmas are needed in the ``driver`` source to insert offload and/or copy-back directives.
```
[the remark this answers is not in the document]
awnawab: Sure, thanks for spotting this, I'll clarify the description with an example.

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,275 @@
...
+                       if isinstance(s.type.dtype, BasicType) and s.name.lower() in self._var_set]
+        deriv_types = [s for i in imports for s in i.symbols if s in kwargs['targets']
+                       if isinstance(s.type.dtype, DerivedType) and s.name.lower() in self._var_set]
+
+        with dataflow_analysis_attached(routine):
+
+            # collect symbols to add to acc update pragmas in driver layer
+            for item in basic_types:
+                if item in routine.body.uses_symbols:
+                    self._acc_copyin.add(item)
```
reuterbal, line 245: Same comment here: I think this should use `item.trafo_data` to store this analysis, and then merge this in every level of the scheduler traversal from successors. See e.g. the pool allocator transformation for how this works in principle (where it is used to determine the combined size of temporaries under a call hierarchy).

The current implementation would not be optimal with multiple driver entry points within a scheduler traversal as it will always add all variables to the offload directive, even if they aren't used in the relevant call tree.

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,275 @@
...
+        # separate out derived and basic types
+        basic_types = [s.name.lower() for i in imports for s in i.symbols if s in kwargs['targets']
+                       if isinstance(s.type.dtype, BasicType) and s.name.lower() in self._var_set]
+        deriv_types = [s for i in imports for s in i.symbols if s in kwargs['targets']
+                       if isinstance(s.type.dtype, DerivedType) and s.name.lower() in self._var_set]
+
+        with dataflow_analysis_attached(routine):
+
+            # collect symbols to add to acc update pragmas in driver layer
+            for item in basic_types:
```
reuterbal, line 243: Just a matter of taste: I would not use `item` here, as it is a somewhat reserved name in transformations to refer to the `Item` object

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,275 @@
...
+            else:
+                new_import_map.update({self._modules[s]: as_tuple(s)})
+
+        new_imports = ()
+        for k, v in new_import_map.items():
+            new_imports += as_tuple(Import(k, symbols=tuple(Variable(name=s, scope=routine) for s in v)))
+
+        # add new imports to driver subroutine sepcification
+        import_pos = 0
+        if (old_imports := FindNodes(Import).visit(routine.spec)):
```
reuterbal, line 211: [No action required] is there a technical reason that we have to add these imports after existing imports instead of just prepending them to the `spec`?

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,275 @@
...
+            import_pos = routine.spec.body.index(old_imports[-1]) + 1
+        if new_imports:
+            routine.spec.insert(import_pos, Comment(text=
+               '![Loki::GlobalVarOffload].....Adding global variables to driver symbol table for offload instructions'))
+        import_pos += 1
+        for index, i in enumerate(new_imports):
+            routine.spec.insert(import_pos + index, i)
```
reuterbal, line 218: `Section.insert` accepts also a tuple of nodes, so could just be `routine.spec.insert(import_pos, new_imports)`

mlange05 (approved):
Ok, all requested changes and comments addressed. Again, very valuable work and looking forward to taking this for a spin on large code paths. GT from me.

reuterbal (changes requested):
This is fantastic and I'm keen on trying this myself. Many, many thanks!

Overall very happy also with the implementation and approach, which is neatly done and well tested. Appreciate also the detailed documentation.

My main remark is related to the way the "analysis data" is carried forward through the scheduler traversal. Our established mechanism is to use `Item.trafo_data` for this, which allows "local" (as in: call-tree specific) transformation data, whereas putting this on the Transformation class makes it global in the Scheduler scope. I would therefore prefer to use that approach here as well.

transformations/transformations/global_var_offload.py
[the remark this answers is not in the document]
awnawab: Yes that makes a lot of sense, will do, thanks!

transformations/transformations/global_var_offload.py
```diff
@@ -0,0 +1,275 @@
...
+      use moduleC, only: var4,var5
+      implicit none
+      !$acc routine seq
+
+      var4 = var2
+      var5 = var3
+
+      end subroutine kernel
+
+    is transformed to:
```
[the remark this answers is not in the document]
awnawab: Thanks for spotting this, fixed!

transformations/transformations/global_var_offload.py
awnawab: No technical reason. In that case I would also like to add a comment after the Loki generated imports, just to make a clear demarcation. Let me know please which implementation you would prefer.

transformations/transformations/global_var_offload.py
awnawab: Agreed, names now changed.

transformations/transformations/global_var_offload.py
awnawab: Ah yes, that's much cleaner, thanks for the suggestion!

transformations/transformations/global_var_offload.py
reuterbal: Thanks for explaining. Either is fine with me - @mlange, any preferences?
My question was motivated by the fact that I would have been lazy and therefore would have just prepended the imports, so was wondering if there was a specific reason.

transformations/transformations/global_var_offload.py
awnawab: Yes `item.trafo_data` is definitely a better way of doing it, I'll copy the pool allocator pattern of gathering this data from the call tree. I'm implementing this now. Thanks a lot!

Also for some reason I can't reply to your comment about `get_pragma_parameters` so I'll write here: I didn't know that utility existed. Always better to reuse existing code rather than duplicate functionality. I've changed the implementation to use it now, thanks!

transformations/transformations/global_var_offload.py
mlange05: No preference, was just curious.

awnawab:
Thanks for the comments @mlange05  and @reuterbal. I've tried to implement all of your suggestions. The analysis information now travels up the call-tree using `item.trafo_data` as B suggested. This uncovered a strange bug, where in certain cases the `GlobalVarImportItem` returned by the graph `topological_sort` and `successors` methods would be copies of the items rather than the originally spawned items (the ID would not match the contents of `Scheduler.item_map`). I've put in a fix and a test for this, please let me know what you think. Perhaps this could be another test for the scheduler re-write.

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+                if 'update_host' in pragma.content:
+                    if update_host:
+                        pragma_map[pragma] = update_host
+                    else:
+                        pragma_map[pragma] = None
+
+        routine.body = Transformer(pragma_map).visit(routine.body)
+
+        # build set of symbols to be offloaded
+        _var_set = reduce(operator.or_, [s.trafo_data[self._key]['var_set']
```
reuterbal, line 399: This could be simplified:
```python
_var_set = set.union(*[s.trafo_data[self._key]['var_set'] for s in successors])
```

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+    def process_driver(self, routine, successors):
+        """
+        Add offload and/or copy-back directives for the imported variables.
+        """
+
+        update_device = ()
+        update_host = ()
+
+        # build offload pragmas
+        _acc_copyin = reduce(operator.or_, [s.trafo_data[self._key]['acc_copyin']
```
reuterbal, line 355: `reduce` + `operator.or_` is actually unnecessary, you could just use `set.union` and tuple unpacking - e.g.:
```python
_acc_copyin = set.union(*[s.trafo_data.get(self._key, {}).get('acc_copyin', set()) for s in successors])
```

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+            routine.spec.insert(import_pos, new_imports)
+
+    def process_kernel(self, routine, successors, item, targets):
+        """
+        Collect the set of module variables to be offloaded.
+        """
+
+        # filter out procedure imports
+        calls = [c.name for c in FindNodes(CallStatement).visit(routine.body)]
+        calls += FindInlineCalls().visit(routine.body)
```
reuterbal, line 439: Does this not need to be something like the following?
```python
calls += [c.function for c in FindInlineCalls().visit(routine.body)]
```

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+
+        imports = [i for i in FindNodes(Import).visit(routine.spec) if i.symbols not in calls]
```
reuterbal, line 441: `Import.symbols` should be a tuple, so I don't think `i.symbols` does here what you think it does (it will check for that exact tuple of symbols to be present in `calls`, which it never is.

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+
+        # build map of modules corresponding to imports
+        import_mod = {s.name.lower(): i.module for i in imports for s in i.symbols}
```
reuterbal, line 444: Considering the above comments, I think this could be built more easily like this:

```python
import_mod = CaseInsensitiveDict(
    (s.name, imprt) for imprt in routine.imports for s in imprt.symbols
    if s not in calls
)
```

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+
+        #build set of offloaded symbols
+        item.trafo_data[self._key]['var_set'] = reduce(operator.or_,
```
reuterbal, line 447: Would recommend to use `set.union` here

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+                                                       [s.trafo_data[self._key]['var_set'] for s in successors], set())
+
+        #build map of module imports corresponding to offloaded symbols
+        item.trafo_data[self._key]['modules'].update({k: v
+                                                     for s in successors if isinstance(s, SubroutineItem)
+                                                     for k, v in s.trafo_data[self._key]['modules'].items()})
+
+        # separate out derived and basic types
+        basic_types = [s.name.lower() for i in imports for s in i.symbols if s in targets
```
reuterbal, line 456: Since `imports` likely isn't quite what you want it to be, here another proposal:
```python
imported_vars = [var for var in routine.imported_symbols if var in item.trafo_data[self._key]['var_set']]
basic_types = [var.name.lower() for var in imported_vars if isinstance(var.type.dtype, BasicType)]
deriv_types = [var.name.lower() for var in imported_vars if isinstance(var.type.dtype, DerivedType)]
```

transformations/transformations/data_offload.py
```diff
@@ -176,3 +181,335 @@ def remove_openmp_pragmas(self, routine, targets):
...
+                       if isinstance(s.type.dtype, BasicType)
+                       and s.name.lower() in item.trafo_data[self._key]['var_set']]
+        deriv_types = [s for i in imports for s in i.symbols if s in targets
+                       if isinstance(s.type.dtype, DerivedType)
+                       and s.name.lower() in item.trafo_data[self._key]['var_set']]
+
+        # accumulate contents of acc directives
+        item.trafo_data[self._key]['enter_data_copyin'] = reduce(operator.or_,
```
reuterbal, line 464: Again, would personally use `set.union` here

reuterbal (changes requested):
Thanks a lot for taking care of all the comments!

Overall I'm very happy with the changes. I've left a few remarks how things could be implemented in a, at least in my view, slightly more elegant fashion. But the more important remark is about `process_kernel`, where I think the logic is currently subtly flawed. There I have left a few remarks how I think this could be resolved, but none of this is actually tested. Have a look and let me know if you agree.

Thanks also for pointing out the Scheduler bug (it really is one), which I took a little time to investigate further: The underpinning problem is, that we store `Item` objects in the `item_map`, and as nodes in the `DiGraph`, which does indeed create duplicates. So, ideally we would only want to store them once, e.g. either always looking up the `Item` in the graph (where it should really be in the node's _data_, not the node itself), or always taking the item from the item_map and keeping only the name/key in graph.

Your fix resolves the issue by treating the item_map as the authority and always looking up the item there, but it is only removing the symptom. Fixing this properly is outside of the scope of this PR, though, and considering that this is in the process of being completely rewritten, I'm happy to keep this as is.

transformations/transformations/data_offload.py
awnawab: Yes I think you are right. This sneaked through because it isn't tested. In fact I don't need to build a list of calls to filter; when I sort the `imported_vars` into basic and derived types, the calls are filtered out automatically.

transformations/transformations/data_offload.py
awnawab: I prefer the use of `CaseInsensitiveDict`, thanks for the tip!

transformations/transformations/data_offload.py
awnawab: Tips like these are highly appreciated; there are still many holes in my python knowledge. Thanks a lot 😄

awnawab:
Thanks a lot for your feedback @reuterbal. I've implemented all of your suggestions, making the code a lot more elegant.

reuterbal (approved):
Thanks a lot, nothing further to add. Fantastic work!

PR #67, opened 2023-04-18
Frontend: Add sanitization of fypp annotations to REGEX frontend

mlange05:
_Note: Creating this as draft, as I'm not quite sure how to properly test this._

Essentially, the heavy use of Fypp in FIELD API can trigger the REGEX-frontend timeout when populating the scheduler. To avoid this, we need to run the sanitization step that removes the heavy Fypp annotations. 

I'm wondering if this could be handled more generically across frontends though, @reuterbal ?


reuterbal:
What would the expected behaviour be for Loki? FYPP annotations are not valid Fortran, so I would expect that other frontends (at least FP and OMNI) trip over this and fail as well. Would we want to strip these annotations completely, or retain them to be able to re-insert them via the fgen backend? I suspect the annotations are in source files that are not actually transformed, or even part of the set of files that are parsed in full?

Either way, I think your approach is pointing in the right direction, since encoding source code changes before feeding into frontends is precisely what Loki's "preprocessing" feature is there for. My proposal would be to amend this to prepend all such annotations (anything beginning with `#:`, `@:` or `$:`) with a `!fypp` to convert them to comments. In the backend, we can then (optionally?) check comments for that leading string and remove it again.

Testing should then be relatively straightforward?

mlange05:
Ok, great, thanks, and yes these are not further processed, but merely avoid hard crashes in the `Scheduler`. The same rule has existed in FP sanitisation forever (out of necessity), but now gets short-circuited as the scheduler touches everything with REGEX first and then breaks (lack of regression testing catching up with me :wink: ). I like the idea of turning them into comments though, instead of just stripping them, so I'll try that next.

mlange05:
Ok, testing this is actually trickier than I first though, since the usage of the REGEX frontend is different to the usage of FP. In a full setup, the full-package scan (regex) needs to sort through the source tree, where fypp files live non-expanded, while the final FP frontend passes are used when extracting type definitions from expanded files, that only have CPP-style line number annotations. The former is the target here, and I've added a test to see if this is sufficient, please, @reuterbal

reuterbal (approved):
Thanks for clarifying the scope and the test highlights well why this is much harder than I first thought. 
Looks good to me and happy to merge as is. Many thanks!

PR #68, opened 2023-04-20
Procedure interface scheduler bug fix

awnawab:
One of my recent PRs added the ability to include inline function calls as scheduler dependencies if they are defined in an explicit interface. A pretty nasty bug went unnoticed; if the calling subroutine is part of a module, it will add any procedure interfaces defined in that module as a child of the subroutine. This is a very small PR to fix that bug and test for it.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -537,8 +537,10 @@ class SubroutineItem(Item):
         names = ()
         interfaces = self.routine.interfaces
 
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Many thanks, looks good to me!

mlange05 (approved):
Agreed, looks clean and neat. GTG from me.

PR #70, opened 2023-04-21
Dataflow analysis multi-conditional `live_symbols` bug

awnawab:
Consider the following example:
```
subroutine multicond(ic)
integer, intent(in) :: ic
integer :: a,b,c

select case(ic)
case(0)
a = 0
case(1)
b=0
case(2)
c=0
end select

end subroutine multicond
```
The variables defined exclusively in one of the branches of the multi-conditional should not be in the `live_symbols` set for subsequent branches. In the current implementation of the dataflow analysis they were however live.

This PR fixes the above bug and updates the relevant test to check against this behaviour.


reuterbal (approved):
Nice find! Looks good to me.

PR #71, opened 2023-04-27
Fix multi-call hoisting in scc-hoist and extend test

mlange05:
Small bug fix in scc-hoist that ensure that a driver loop with more than one call to kernel routines hoists argument for all of them. The bug was simply cause by lazily using the loop to find and adjust the outer block loop with the loop over all candidate calls. The respective test has been extended in place to capture the corner case that tripped this in CLOUDSC2.

As a small bonus, I've also added a log line that prints out the variables hoisted for each driver call, which should not be too polluting in practical cases.

### DIFF
--- a/transformations/transformations/single_column_coalesced.py
+++ b/transformations/transformations/single_column_coalesced.py
@@ -15,7 +15,7 @@ from loki import (
     FindExpressions, Transformer, NestedTransformer,
     SubstituteExpressions, SymbolAttributes, BasicType, DerivedType,
     pragmas_attached, CaseInsensitiveDict, as_tuple, flatten,
-    demote_variables
+    demote_variables, info
 )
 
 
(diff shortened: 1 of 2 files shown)


mlange05:
@piotrows to confirm that this fixes the problem in CLOUDSC2-NL for loki-scc transformations.

piotrows:
> @piotrows to confirm that this fixes the problem in CLOUDSC2-NL for loki-scc transformations.

I can confirm that this works for CLOUDSC2-NL loki-scc-hoist transformation.

transformations/transformations/single_column_coalesced.py
```diff
@@ -640,7 +640,12 @@ def process_driver(self, routine, targets=None):
                             loop._update(pragma_post=(ir.Pragma(keyword='acc', content='end parallel loop'),
                                                       loop.pragma_post[0]))
 
-                # Apply hoisting of temporary "column arrays"
+        # Apply hoisting of temporary "column arrays"
+        with pragmas_attached(routine, ir.Loop, attach_pragma_post=True):
```
reuterbal, line 644: What's the reason for attaching pragmas here? Can't seem to find where this is used.

reuterbal (approved):
Thanks, useful addition for an important use case. A question about whether the pragma attachment is actually required, otherwise good to go if CI is happy

transformations/transformations/single_column_coalesced.py
mlange05: Good point, the pragmas handled in the call to `hoist_temporary_column_arrays` are actually not loop-attached ones, so this is probably redundant. Will remove, thanks.

reuterbal: Thanks, looks good now!

PR #72, opened 2023-04-28
Bug fix in dependency transform for routines declared using explicit interface

awnawab:
The mechanism in the dependency transform to replace a subroutine interface with a module import relied on the subroutine name in the interface block being lower-case. This is now fixed, along with an update to the relevant test.


reuterbal (approved):
Nice find and uncontroversial. Many thanks!

mlange05 (approved):
Agreed, neat and clean. GTG

Issue #69, opened 2023-04-21
Some metaclass conflict using loki.types.DataType and IntEnum with Python 3.11


meteospike:
While loading basic loki types, file `enum.py` raise `too many data types for 'BasicType': {<class 'int'>, <class 'loki.types.DataType'>}`

This occurs in the calling tree of the metaclass `EnumType` (more than one datatype found).

reuterbal:
Thanks. I have observed the same, which is a change in behaviour with Python 3.12. I will take care of fixing this soon.
As we don't support or test Python 3.12, yet, please switch to an older Python version in the meantime.

meteospike:
Thanks, I'm indeed currently using Python 3.10. Testing the install with Python 3.11 was just a matter of prospect, mainly because of the claim of the impressive efficiency of this new release... Anyway, it may be hard to get spectacular improvement on such a code (parsing, AST), with low amount of calculations.

PR #73, opened 2023-05-04
Fix documentation CI for PRs from forks and separate pytest and regression tests in Github actions

reuterbal:
What the title says.

Underlying issue for documentation is that workflows triggered from forks don't have permissions to acces pull request secrets, which are required to create/edit comments and upload the documentation.

The separate pytest and regression tests take a similar amount of time, overall reducing the turn-around time for CI by almost 50%.

Also, I marked Loki as incompatible with Python 3.11 as pointed out in #69 . Fix seems less straightforward than what I thought, so requires a dedicated PR.


mlange05 (approved):
Looking great! Many thanks, GTG! (with priority, if you want @reuterbal )

PR #74, opened 2023-05-05
Fix order of `USE` statements and `IMPLICIT NONE`

awnawab:
The Fortran standard mandates that `USE` statements should appear before `IMPLICIT NONE` but header includes and interface blocks should appear after it. So when the dependency transform finds header includes after `IMPLICIT NONE` and replaces them in-place with a `USE SOME_MOD` without reordering, we can run into problems.

### DIFF
--- a/loki/transform/dependency_transform.py
+++ b/loki/transform/dependency_transform.py
@@ -11,7 +11,7 @@ from loki.transform.transformation import Transformation
 from loki.subroutine import Subroutine
 from loki.module import Module
 from loki.visitors import FindNodes, Transformer
-from loki.ir import CallStatement, Import, Section, Interface
+from loki.ir import CallStatement, Import, Section, Interface, Intrinsic
 from loki.expression import Variable, FindInlineCalls, SubstituteExpressions
 from loki.backend import fgen
 from loki.tools import as_tuple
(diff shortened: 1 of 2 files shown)


loki/transform/dependency_transform.py
```diff
@@ -105,6 +105,14 @@ def transform_subroutine(self, routine, **kwargs):
             # Re-generate C-style interface header
             self.generate_interfaces(routine)
 
+        # Ensure Fortran module imports appear before 'IMPLICIT NONE'
+        intrs = [i for i in FindNodes(Intrinsic).visit(routine.spec) if i.text.lower() == 'implicit none']
+        imports = [i for i in FindNodes(Import).visit(routine.spec) if not i.c_import]
+        if len(intrs) == 1 and imports:
```
reuterbal, line 111: I think this implementation will fail in cases, where interfaces appear after a variable declaration (which is allowed by the standard). There, the interface will be replaced by the import, which is still after the variable declaration. Subsequently, the implicit none will end up behind a variable declaration, which is not standard-compliant.

reuterbal (changes requested):
Still puzzled that this hasn't caused problems before, but I suppose in most places interfaces are defined via C-style includes rather than in the code directly.

Unfortunately, I'm not too happy with the implementation. I think it will fail in at least one common case and the "monkey-patching" of the intrinsic node feels unnecessarily brittle. 
Having looked at the transformation, it seems that C includes of interfaces are handled correctly, with the newly inserted imports prepended to the spec. Doing the same when replacing interfaces seems therefore like the better solution.

In short: in `rename_interfaces`, do not insert `new_imports` via the `removal_map` but instead use that map only to remove the interfaces, and then insert the new imports via `source.spec.prepend(new_imports)`.

loki/transform/dependency_transform.py
awnawab: Yes absolutely, I should have studied the c-style import headers a bit more carefully, this is the way to do it. Thanks a lot for pointing this out, I'll update this PR.

mlange05 (approved):
Nice and clean, GTG from me.

PR #75, opened 2023-05-05
Install ant via APT for Github actions

reuterbal:
Possibly a quick fix for the install issues in CI we have seen


mlange05 (approved):
Great! Many thanks, fast-tracking this to make CI happy.

PR #76, opened 2023-05-09
Changes needed to run pool allocator transformation on ecWam physics

awnawab:
A few small fixes required to run the pool allocator transformation on ecWam physics. Performance is great out of the box! (performance discussed in more detail in the JIRA story DE-757).

### DIFF
--- a/transformations/transformations/pool_allocator.py
+++ b/transformations/transformations/pool_allocator.py
@@ -124,6 +124,7 @@ class TemporariesPoolAllocatorTransformation(Transformation):
         self.local_ptr_var_name_pattern = local_ptr_var_name_pattern
         self.directive = directive
         self.check_bounds = check_bounds
+
         if key:
             self._key = key
 
@@ -133,8 +134,12 @@ class TemporariesPoolAllocatorTransformation(Transformation):
         item = kwargs.get('item', None)
         targets = kwargs.get('targets', None)
 
-        if item and item.local_name != routine.name.lower():
-            return
+        self.stack_type_kind = 'JPRB'
+        if item:
+            if item.local_name != routine.name.lower():
+                return
+            if (real_kind := item.config.get('real_kind', None)):
+                self.stack_type_kind = real_kind
 
         successors = kwargs.get('successors', ())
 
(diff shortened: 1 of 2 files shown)


transformations/transformations/pool_allocator.py
```diff
@@ -530,11 +544,29 @@ def inject_pool_allocator_into_calls(self, routine, targets):
         """
         call_map = {}
         stack_var = self._get_local_stack_var(routine)
+        intfs = FindNodes(Interface).visit(routine.spec)
         for call in FindNodes(CallStatement).visit(routine.body):
             if call.name in targets:
+               # If call is declared via an explicit interface, the ProcedureSymbol corresponding to the call is the
+               # interface block rather than the Subroutine itself. This means we have to update the interface block
+               # accordingly
+                if call.name in [s for i in intfs for s in i.symbols]:
```
reuterbal, line 553: This list comprehension is potentially performed relatively often. Since `intfs` itself isn't required, this comprehension could be done once, e.g.:
```
intfs_symbols = [s for i in FindNodes(Interface).visit(routine.spec) for s in i.symbols]
```

awnawab: Ah yes, the performance benefits of list comprehensions. Sure, will update, thanks!

transformations/transformations/pool_allocator.py
```diff
@@ -530,11 +544,29 @@ def inject_pool_allocator_into_calls(self, routine, targets):
         """
         call_map = {}
         stack_var = self._get_local_stack_var(routine)
+        intfs = FindNodes(Interface).visit(routine.spec)
         for call in FindNodes(CallStatement).visit(routine.body):
             if call.name in targets:
+               # If call is declared via an explicit interface, the ProcedureSymbol corresponding to the call is the
```
reuterbal, line 550: Interesting, I wasn't aware of this. We should keep an eye out for this. I would almost consider this a bug, since the interface is the declaration of the routine and not the actual routine, so getting this after a call to `call.routine` seems counter-intuitive to me. (We should in any case update the interface, of course).

reuterbal (approved):
Thanks a lot, very happy with this!

I'd like to test this on CLOUDSC and wait for CI to come back green, but otherwise GTG!

transformations/transformations/pool_allocator.py
awnawab: I think this behaviour arises from the fact that if we have an interface, then the symbol already exists in the calling subroutine's symbol-table and so `Subroutine.enrich_calls` doesn't overwrite it, even though `update_symbol` is still `True` and we do actually execute `call._update(...)` in `loki/subroutine.py' L455.

I am not familiar at all with the symbol table, but in my mind a clean "fix" could perhaps be to allow entries in the symbol table to be overwritten?

reuterbal:
Thanks. Would you mind adding a test (or extending the existing ones) that 
- sets the `stack_type_kind`
- has a callee with optional argument
- has an inline call
Codecov pointed me to the fact that test coverage is incomplete right now.

PR #78, opened 2023-05-10
Introduce RegexParserClass for interfaces and add test for interfaces in modules

reuterbal:


mlange05 (approved):
Yes, looks great. GTG

PR #81, opened 2023-05-15
SCC modularization

awnawab:
I've split the SCC transform into six transforms:

- `SCCBaseTransformation`
- `SCCDevectorTransformation`
- `SCCDemoteTransformation`
- `SCCRevectorTransformation`
- `SCCHoistTransformation`
- `SCCAnnotateTransformation`

The original `SingleColumnCoalescedTransformation` also still exists, but it's basically now a wrapper script that calls the methods contained in the above transformations.

I've updated `loki-transform.py` to run each one of the six above one-by-one rather than running the unified SCC transform. This is regression tested on the ecWam physics against the original SCC transform and produces string-identical results.

I still have to refactor the tests to better test the new modular transforms individually, but before I do that I would really appreciate your input on whether I am the right path.

NB: this is stacked on top of a small bug fix in #89.


awnawab:
Regression tested against cloudsc and ecWam physics. I've left the old `SingleColumnCoalescedTransformation` (including some tests that call it) in. It's now basically a wrapper script that calls the methods in the modular transformations. In principle this can, and perhaps should, be removed.

transformations/transformations/scc_annotate.py
```diff
@@ -0,0 +1,323 @@
...
+        ----------
+        routine : :any:`Subroutine`
+            Subroutine to apply this transformation to.
+        role : string
+            Role of the subroutine in the call tree; should be ``"kernel"``
+        """
+
+        # TODO: we only need this here until the scheduler can combine multiple transformations into single pass
+        # Bail if routine has already been processed
+        if self._processed.get(routine, None):
```
mlange05, line 166: [No action] Indeed, this mechanism isn't great, but will likely only become redundant once we can preempt repeated processing, which should come with the scheduler rewrites. Ok for now, I htink.

mlange05 (commented):
Ok, first of all, this is looking really great, and in addition is exposing all the right dependencies to make start working on Transformation composition. The test coverage is also very, very nice, although we might be able to reduce the replication at a later stage once composition can be done neatly. All in all, I'm super happy with this, so, many thanks!!!

I'm marking this as "Comments" for now, until I've run the full ec-phys regression, as I've left a few minor comments about the location of some utilities. None of these are showstoppers though, and I'll approve once regression test has run (ongoing).

transformations/transformations/scc_annotate.py
reuterbal: The scheduler already ensures that each item is visited only once, so shouldn't this instead be replaced by the typical pattern?
```python
if (item := kwargs.get('item')) and item.name != routine.name.lower():
    return
```
I think this should become part of the dispatch in the `Transformation.apply` routine eventually.

mlange05 (approved):
Ok, EC-physics regression test is clean and comments are not considered showstoppers, so I'll green-light this now. I'll wait for secondary review and/or responses, but ready from my side.

transformations/transformations/scc_demote.py
```diff
@@ -0,0 +1,124 @@
...
+        ----------
+        routine : :any:`Subroutine`
+            Subroutine to apply this transformation to.
+        role : string
+            Role of the subroutine in the call tree; should be ``"kernel"``
+        """
+
+        # TODO: we only need this here until the scheduler can combine multiple transformations into single pass
+        # Bail if routine has already been processed
+        if self._processed.get(routine, None):
```
reuterbal, line 87: Same here

transformations/transformations/scc_base.py
```diff
@@ -0,0 +1,264 @@
...
+        ----------
+        routine : :any:`Subroutine`
+            Subroutine to apply this transformation to.
+        role : string
+            Role of the subroutine in the call tree; should be ``"kernel"``
+        """
+
+        # TODO: we only need this here until the scheduler can combine multiple transformations into single pass
+        # Bail if routine has already been processed
+        if self._processed.get(routine, None):
```
reuterbal, line 206: Same here

transformations/transformations/scc_devector.py
```diff
@@ -0,0 +1,161 @@
...
+        ----------
+        routine : :any:`Subroutine`
+            Subroutine to apply this transformation to.
+        role : string
+            Role of the subroutine in the call tree; should be ``"kernel"``
+        """
+
+        # TODO: we only need this here until the scheduler can combine multiple transformations into single pass
+        # Bail if routine has already been processed
+        if self._processed.get(routine, None):
```
reuterbal, line 132: Same here

transformations/transformations/scc_hoist.py
```diff
@@ -0,0 +1,250 @@
...
+        ----------
+        routine : :any:`Subroutine`
+            Subroutine to apply this transformation to.
+        role : string
+            Role of the subroutine in the call tree; should be ``"kernel"``
+        """
+
+        # TODO: we only need this here until the scheduler can combine multiple transformations into single pass
+        # Bail if routine has already been processed
+        if self._processed.get(routine, None):
```
reuterbal, line 184: Same here

transformations/transformations/scc_revector.py
```diff
@@ -0,0 +1,98 @@
...
+        ----------
+        routine : :any:`Subroutine`
+            Subroutine to apply this transformation to.
+        role : string
+            Role of the subroutine in the call tree; should be ``"kernel"``
+        """
+
+        # TODO: we only need this here until the scheduler can combine multiple transformations into single pass
+        # Bail if routine has already been processed
+        if self._processed.get(routine, None):
```
reuterbal, line 72: Same here

reuterbal (approved):
This is great and well-tested! I'm a bit concerned about the large number of modules, which isn't necessarily the Pythonic way (where I see module files more like directories than individual source files in other languages) but that's a matter of taste.

My only real request would be to remove the `self._processed` checks and instead use the established pattern of checking against the item name, which we use elswhere and which achieves the same.

For things flagged in the comments and that are not being addressed now, it might make sense to create issues to not forget them down the line.

transformations/transformations/scc_annotate.py
mlange05: Yes, I agree. I was planning on moving these bail-out mechanism when I switch to composed SCC for full ec-physics, but we can bring them in line here already

reuterbal: I think it's not necessary to move this already to the base class as part of this PR, but would rather not introduce a second pattern.

transformations/transformations/scc_annotate.py
```diff
@@ -0,0 +1,279 @@
...
+            Subroutine to apply this transformation to.
+        targets : list or string
+            List of subroutines that are to be considered as part of
+            the transformation call tree.
+        """
+
+        # Resolve associates, since the PGI compiler cannot deal with
+        # implicit derived type component offload by calling device
+        # routines.
+        resolve_associates(routine)
```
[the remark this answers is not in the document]
awnawab: Yes agreed. I'm not sure how this ended up in the diff, but I think a subsequent commit already moved this to the base.

loki/transform/transform_array_indexing.py
```diff
@@ -28,9 +28,24 @@
     'shift_to_zero_indexing', 'invert_array_indices',
     'resolve_vector_notation', 'normalize_range_indexing',
     'promote_variables', 'promote_nonmatching_variables',
-    'promotion_dimensions_from_loop_nest', 'demote_variables'
+    'promotion_dimensions_from_loop_nest', 'demote_variables',
+    'is_dimension_constant'
 ]
 
+def is_dimension_constant(d):
```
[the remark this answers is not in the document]
awnawab: Yes agreed, will move it there and add tests. Thanks!

transformations/transformations/scc_annotate.py
awnawab: Yes, fully agreed. I should have realised that this is just duplicating an existing mechanism. I'll replace all instances of the `self._processed` checks with B's suggestion above.

transformations/transformations/scc_revector.py
awnawab: Yup, will replace all of these with the existing mechanism. Thanks again!

transformations/tests/test_single_column_coalesced.py
```diff
@@ -10,32 +10,31 @@
...
+from transformations import (
+     DataOffloadTransformation, SCCBaseTransformation, SCCDevectorTransformation, SCCDemoteTransformation,
+     SCCRevectorTransformation, SCCHoistTransformation, SCCAnnotateTransformation, SingleColumnCoalescedTransformation
+)
 
 @pytest.fixture(scope='module', name='horizontal')
 def fixture_horizontal():
     return Dimension(name='horizontal', size='nlon', index='jl', bounds=('start', 'end'))
 
-
```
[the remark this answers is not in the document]
awnawab: Done, thanks!

scripts/loki_transform.py
```diff
@@ -38,6 +38,12 @@
 from transformations.single_column_claw import ExtractSCATransformation, CLAWTransformation
 from transformations.single_column_coalesced import SingleColumnCoalescedTransformation
 from transformations.scc_cuf import SccCufTransformation, HoistTemporaryArraysDeviceAllocatableTransformation
+from transformations.scc_base import SCCBaseTransformation
```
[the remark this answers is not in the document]
awnawab: I've put `SingleColumnCoalescedTransformation` into a separate module `single_column_coalesced_wrapper.py` to prevent a circular import. Since this transformation is earmarked for deletion in the near future anyway, also makes sense to put it in a separate module.

awnawab:
Thanks a lot for your comments! I've implemented all of them, please let me know what you think.

reuterbal:
Thanks so much, looks great! Going in...

PR #82, opened 2023-05-15
Loki-transform: Expose `directive` argument to loki_transform_convert

mlange05:
Follow-up to PR #79, again exposing the `directive` argument through the CMake layer, but this time for `loki_transform_convert`. The use case for this is the new CLOUDSC2-TL-AD repo, where cleanly suppressing OpenACC insertion aids PR and feature staging. Apologies for not including this before.


reuterbal (approved):
Uncontroversial and GTG. Thanks!

PR #83, opened 2023-05-17
Support for multiple kernel calls in DataOffloadTransformation

mlange05:
This PR adds support for multiple calls to kernel subroutines in the `DataOffloadTransformation`. This is required for CLOUDSC2, where we need this transformation to correctly place timers to separate compute and data movement timings.

The implementation itself adds a new test, updates the use of tuples and f-strings for code generation, and adds some logic to ensure repeated array arguments used with different intents on the kernel side are classed correctly and only put into one category


reuterbal (approved):
Thanks! Neat implementation and additional testing is much appreciated. Unless CI trips this is GTG.

PR #85, opened 2023-05-19
SCC-hoist: Add test and fix corner case, where no arrays are hoisted

mlange05:
When no arrays are hoisted, we would still insert empty `!$acc create/delete` pragmas, which is incorrect and invalid.


reuterbal (approved):
Small, uncontroversial change but well tested. Many thanks!

PR #86, opened 2023-05-22
IR: Make `ScopedNode` classes frozen datasets

mlange05:
The primary concern of this PR is to align the two `ScopedNode` IR classes (`Associate` and `TypeDef`) with the remaining IR node classes to fully use the immutable "frozen dataset" principle to provide native equivalence, pickling and hashing behaviour. 

The tricky part is here that, in order to allow frozen dataclasses to inherit from `Scope`, we need to make `Scope` a frozen dataclass too. However, since "parent" linkage of scopes and symbol table of `Scope` objects are not actually immutable, we provide specific and controlled backdoors to bypass that where necessary, and allow the `.parent` property to be reset. For this we're introducing a custom `WeakrefProperty` descriptor to formally type-define attributes that are mere weakref links. As a result, a symbol table object (`Scope.symbol_attrs`) is now hard-wired to a `Scope` object and may not be exchanged or re-linked, but needs to be copied over if the `Scope` node is rebuilt. We also need to control the constructor usage a little differently now.

On the flip side, we gain a lot more cleanliness in the IR tree itself. All the custom declarations of `_argnames` can now be removed, as equivalence, hashing and pickling can now use the native Python mechanisms provided by dataclasses. 

Detailed changes: 
* Made `Scope` a frozen dataclass and introduced `WeakrefProperty` type descriptor to allow `.parent` linkage
* Made `Associate` and `TypeDef` native and type-checked frozen dataclasses using private attribute descriptor classes 
* Changed `Subroutine`/`Module` constructor usage, as we no longer can re-use `__init__` in frozen `Scope` objects. Instead we set all the delayed attributes in a custom `__initialize__` routine via our usual `_update` mechanisms. 
* Derive `Node.args` via `self.__dataclass_fields__.keys()` and remove all other custom hash/equivalence behaviour and `_canonical` definition.
* Made `Scope.parent` a pseudo-frozen attribute, rather than a property, and add `Scope._reset_parent()` for those special cases where we really need to adopted a `Scope` object. 
* A small bug fix to `InlineCall` symbols by defining the right `init_arg_names` and `__getinitargs__` for rebuilding.

As a final note, I've run full EC-physics regression with this, and it checks out. Performance is equivalent with no noticeable degradation, and a slight (30%) improvement in regex-frontend scanning large directory paths.


reuterbal (approved):
Very neat! I don't think I am able to comprehend every detail of the dataclass side of things, but the solution is very nice.

I also quite like that the "re-initialization" pattern is now a little more formalized, this was a sore point. Nothing else to add, really, so GTG!

PR #87, opened 2023-05-22
Homogenize access to typedefs and imports in program units

reuterbal:
A small utility PR:

On `ProgramUnit` objects, we have various convenience properties: imports, variables, declarations, ...
The naming "convention" so far was, e.g.:
- `.variables`: list of `Variable` symbols declared in the module or subroutine
- `.variable_map`: `CaseInsensitivDict` to map variable names to variable objects

Only exception was `.typedefs`, which returned a mapping rather than the list of typedefs. This PR homogenizes this by introducing `.typedef_map` and making `.typedefs` a list.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -387,7 +387,7 @@ class Item:
                     continue
 
                 # Search for definition of the type in parent scope
-                if scope is not None and type_name in scope.typedefs:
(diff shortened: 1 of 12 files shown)


mlange05 (approved):
Uncontroversial and neat. GTG from me.

PR #89, opened 2023-05-25
Pragma regions detach bug fix

awnawab:
`detach_pragma_regions` replaces any `PragmaRegion` node with the following tuple `(region.pragma, region.body, region.pragma_post)`. This nested tuple is unnecessary, and it causes problems with some of the pydantic checks if the original `PragmaRegoin` was nested with a `Conditional` or other `InternalNode`. Here is a small fix with updated tests.

### DIFF
--- a/loki/pragma_utils.py
+++ b/loki/pragma_utils.py
@@ -477,7 +477,7 @@ def detach_pragma_regions(ir):
     All replacements are performed in-place, without rebuilding any IR
     nodes.
     """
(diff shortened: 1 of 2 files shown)


reuterbal:
Nice find, and good testing of a 1-line change. Many thanks!

reuterbal (approved):

mlange05 (approved):
Agreed. Nice find and fix, great testing. GTG

PR #91, opened 2023-06-02
Dataflow associate block fix

awnawab:
The dataflow analysis was missing a `visit_Associate` method. @MichaelSt98 and I both ran into this issue. This is an attempted fix with matching test, please let me know what you think.


reuterbal (approved):
Many thanks, intruiging that this missing visitor method hadn't been noticed before.
Very neatly done and well-tested!

mlange05 (approved):
Looks good. GTG from me

PR #93, opened 2023-06-06
Various bugfixes

reuterbal:
A collection of commits with various minor bugfixes/improvements:

- Add a different colour for import items in the Scheduler's callgraph
- Equip `InlineCall` with the same set of inter-procedural properties as `CallStatement`, with associated testing
- Fix a bug with nested tuples in `Associates` for fparser
- Fix a bug with nested tuples in the `resolve_associates` transformation
- Fix REGEX frontend when square brackets are used in an initialiser list
- Apply `RemoveCallsTransformation` only on Scheduler graph items
- Fix `GlobalVarOffload` to not fail when `trafo_data` is empty for a successor

### DIFF
--- a/loki/transform/transform_associates.py
+++ b/loki/transform/transform_associates.py
@@ -8,7 +8,7 @@
 from loki.expression import FindVariables, SubstituteExpressions
 from loki.tools import CaseInsensitiveDict
 from loki.transform.transform_utilities import recursive_expression_map_update
-from loki.visitors import Transformer
+from loki.visitors import NestedTransformer
 
 
 __all__ = ['resolve_associates', 'ResolveAssociatesTransformer']
@@ -31,7 +31,7 @@ def resolve_associates(routine):
     routine.rescope_symbols()
 
 
-class ResolveAssociatesTransformer(Transformer):
+class ResolveAssociatesTransformer(NestedTransformer):
     """
     :any:`Transformer` class to resolve :any:`Associate` nodes in IR trees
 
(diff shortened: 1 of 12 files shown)


loki/transform/transform_associates.py
```diff
@@ -58,5 +58,9 @@ def visit_Associate(self, o, **kwargs):  # pylint: disable=unused-argument
         # Apply the expression substitution map to itself to handle nested expressions
         vmap = recursive_expression_map_update(vmap)
 
-        # Return the body of the associate block with all expressions replaced
-        return SubstituteExpressions(vmap).visit(body)
+        # Mark the associate block for replacement with its body, with all expressions replaced
+        self.mapper[o] = SubstituteExpressions(vmap).visit(body)
+
+        # Return the original object unchanged and let the tuple injection mechanism take care
```
mlange05, line 64: Clever! :smile:

transformations/transformations/utility_routines.py
```diff
@@ -100,6 +100,9 @@ def transform_subroutine(self, routine, **kwargs):
         """
         Apply transformation to subroutine object
         """
+        if 'item' in kwargs and kwargs['item'].local_name != routine.name.lower():
```
mlange05, line 103: [no action] This is no longer needed once PR #100 lands. No need to take out here, just "note to self" to mop this up in the rebase!

mlange05 (approved):
Very neat and useful. I particularly love the `NestedTransformer` trick for associates (I tried before, but failed... :( ). GTG once rebased! :shipit:

PR #95, opened 2023-06-12
SCC updates needed for ecWam

awnawab:
For both correctness and performance, the SCC transform needed a few updates before it could be applied to the ecWam physics. With that aim, this branch adds the following features:
- `trim_vector_sections` option to remove scalar assignments from vector loops
- Removal of calls explicitly marked with `!$loki inline` pragmas from separator nodes
- Loki-pragma driven insertion of reduction loops

The companion piece to this PR is the following ecWam branch where similar to CLOUDSC, we use Loki's cmake tools to build a range of transformations: https://github.com/ecmwf-ifs/ecwam/tree/naan-loki-trafo

NB: This is stacked on top of #92.

### DIFF
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -189,6 +189,15 @@ def convert(out_path, path, header, cpp, directive, include, define, omni_includ
     # First, remove all derived-type arguments; caller first!
     scheduler.process(transformation=DerivedTypeArgumentsTransformation())
 
+    # Remove DR_HOOK and other utility calls first, so they don't interfere with SCC loop hoisting
+    if 'scc' in mode:
+        scheduler.process(transformation=RemoveCallsTransformation(
+            routines=config.default['utility_routines'] or ['DR_HOOK', 'ABOR1', 'WRITE(NULOUT'],
+            include_intrinsics=True
+        ))
+    else:
+        scheduler.process(transformation=DrHookTransformation(mode=mode, remove=False))
+
     # Insert data offload regions for GPUs and remove OpenMP threading directives
     use_claw_offload = True
     if data_offload:
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -18,7 +18,8 @@ import click
 
 from loki import (
     Sourcefile, Transformation, Scheduler, SchedulerConfig, SubroutineItem,
-    Frontend, as_tuple, set_excepthook, auto_post_mortem_debugger, flatten, info
+    Frontend, as_tuple, set_excepthook, auto_post_mortem_debugger, flatten, info,
+    GlobalVarImportItem
 )
 
 # Get generalized transformations provided by Loki
@@ -31,7 +32,7 @@ from loki.transform import (
 from transformations.argument_shape import (
     ArgumentArrayShapeAnalysis, ExplicitArgumentArrayShapeTransformation
 )
-from transformations.data_offload import DataOffloadTransformation
+from transformations.data_offload import DataOffloadTransformation, GlobalVarOffloadTransformation
 from transformations.derived_types import DerivedTypeArgumentsTransformation
 from transformations.utility_routines import DrHookTransformation, RemoveCallsTransformation
 from transformations.pool_allocator import TemporariesPoolAllocatorTransformation
@@ -146,8 +147,10 @@ def cli(debug):
               help='Path to custom scheduler configuration file')
 @click.option('--trim-vector-sections', is_flag=True, default=False,
               help='Trim vector loops in SCC transform to exclude scalar assignments.')
+@click.option('--global-var-offload', is_flag=True, default=False,
+              help="Generate offload instructions for global vars imported via 'USE' statements.")
 def convert(out_path, path, header, cpp, directive, include, define, omni_include, xmod,
-            data_offload, remove_openmp, mode, frontend, config, trim_vector_sections):
+            data_offload, remove_openmp, mode, frontend, config, trim_vector_sections, global_var_offload):
     """
     Single Column Abstraction (SCA): Convert kernel into single-column
     format and adjust driver to apply it over in a horizontal loop.
@@ -254,6 +257,10 @@ def convert(out_path, path, header, cpp, directive, include, define, omni_includ
     else:
         raise RuntimeError('[Loki] Convert could not find specified Transformation!')
 
+    if global_var_offload:
+        scheduler.process(transformation=GlobalVarOffloadTransformation(),
+                          item_filter=(SubroutineItem, GlobalVarImportItem), reverse=True)
+
     if mode in ['idem-stack', 'scc-stack']:
         if frontend == Frontend.OMNI:
             # To make the pool allocator size derivation work correctly, we need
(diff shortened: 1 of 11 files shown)


transformations/transformations/pool_allocator.py
```diff
@@ -579,7 +580,7 @@ def create_pool_allocator(self, routine, stack_size):
                 if pragma.content.lower().startswith('parallel') and 'gang' in pragma.content.lower():
                     parameters = get_pragma_parameters(pragma, starts_with='parallel', only_loki_pragmas=False)
                     if 'private' in [p.lower() for p in parameters]:
-                        content = pragma.content.lower().replace('private(', f'private({stack_var.name}, ')
+                        content = re.sub(r'\bprivate\(', f'private({stack_var.name}, ', pragma.content.lower())
```
mlange05, line 583: [no action] Is this just to catch optional whitespace? Just for my own understanding, why is this an issue?

mlange05 (changes requested):
Ok, overall this looks really great, and contains several very useful things that I will need to try out quickly on my code sections! Some of these could've probably been standalone PRs, but that's not super-important here, as the PR is still neatly sized.

I have one comment though, that I think needs addressing before we can land this. The use of `!$loki inline` suggests way more than we do atm, and while we might want to fully support explicit full inlining macros later, I'd prefer if we either rename it, or look at the style of routine to determine the override that this effectively brings.

All other comments are optional or just that, "comments", so GTG once the pragma is addressed. I'll run EC-phys regression next before merging and post back here, but should be fine.

transformations/tests/test_single_column_coalesced.py
```diff
@@ -1519,3 +1520,320 @@ def test_scc_base_routine_seq_pragma(frontend, horizontal):
...
+    subroutine some_kernel(start, end, nlon, work, cond)
+       logical, intent(in) :: cond
+       integer, intent(in) :: nlon, start, end
+       real, dimension(nlon), intent(inout) :: work
+
+       integer :: jl
+
+       do jl=start,end
+          if(cond)then
+             !$loki inline
```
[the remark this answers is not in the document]
awnawab: Yes I think it should be possible to do this by examining whether we have any horizontal loops inside the attached routine. I'll have a go at that, thanks for the suggestion!

transformations/tests/test_single_column_coalesced.py
```diff
@@ -1519,3 +1520,320 @@ def test_scc_base_routine_seq_pragma(frontend, horizontal):
...
+
+    # check we found the right assignment
+    assert assign.lhs.name.lower() == 'flag0'
+
+    # check we found the right comment
+    assert comment.text == '! random comment'
+
+    if trim_vector_sections:
+        assert assign not in loop.body
+        assert assign in routine.body.body
```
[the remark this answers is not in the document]
awnawab: No it doesn't work, I get an `argument of type 'Section' is not iterable` error.

transformations/transformations/pool_allocator.py
awnawab: If the pragma had both a `private` and a `firstprivate` clause, then the previous implementation would put `YLSTACK` in both.

awnawab:
Thanks a lot for the review @mlange05! Whilst I address the `loki inline` pragma comment, I also wanted to just highlight that with these changes the modular scc transform now diverges from the old scc, which now lives in `single_column_coalesced_wrapper.py`. Please let me know if you would like me to play these changes through the scc wrapper too.

mlange05:
> Thanks a lot for the review @mlange05! Whilst I address the `loki inline` pragma comment, I also wanted to just highlight that with these changes the modular scc transform now diverges from the old scc, which now lives in `single_column_coalesced_wrapper.py`. Please let me know if you would like me to play these changes through the scc wrapper too.

No, that's fine. In fact it's a good motivator for me to flip the `ecphys` mode over to use the same a convert. No need to hold up this PR for this.

transformations/tests/test_single_column_coalesced.py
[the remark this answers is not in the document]
mlange05: Oh, ok. We'll need to fix this at some point (famous last words :grin: ). Was just out of interest, nothing needed for this PR.

transformations/transformations/pool_allocator.py
mlange05: Ahhhhh! Makes sense, thanks.

transformations/transformations/single_column_coalesced_vector.py
```diff
@@ -76,12 +76,11 @@ def extract_vector_sections(cls, section, horizontal):
...
-            if is_loki_pragma(call.pragma, starts_with='inline'):
-                # Because extract_vector_sections was called within pragmas_attached, the pragma node
-                # will not be a part of the returned subsections. Therefore we remove it here to prevent
-                # problems later on
-                call._update(pragma=None)
-                continue
+
+            # check if calls have been enriched
+            if not call.routine is BasicType.DEFERRED:
+                if not horizontal.size in call.routine.symbol_map:
```
mlange05, line 82: Uhhh, that's nifty, but could also be quite error-prone. If I understand correctly, if the callee routine deviates from the coding norms, it will now be treated as effectively sequential/non-vector? That could be dangerous as that actually happens quite often in unsanitised code. 

I had originally thought, looking explicitly for the `!$loki routine seq` pragma on the callee (which we already use to bypass SCC-vectorisation), would be more appropriate here (although not quite the one-liner that this is, admittedly)? Would be keen to hear counter-points if I'm not seeing something here? @awnawab

awnawab: Ah yes, fully agreed, if we already have one mechanism to mark routines as sequential, then we should just use that. Thanks for the suggestion, I'll implement that! And also add more tests to boost coverage 😆

mlange05: Also, paging Dr @reuterbal for input on this one?

reuterbal: Yeah, I might have to come back to that one as I do have quite a few routines that are unsanitised in that sense. I agree that checking only for horizontal size may be too error-prone, though, so sticking to the existing annotation mechanism would probably be better.

transformations/tests/test_utility_routines.py
```diff
@@ -84,6 +87,13 @@ def fixture_source(srcdir):
 
     if (lhook) call dr_hook('never_gonna_give',1,zhook_handle)
 end subroutine
+subroutine i_hope_you_havent_let_me_down
+    real(kind=jprb) :: zhook_handle
+    if (lhook) call dr_hook('i_hope_you_havent_let_me_down',0,zhook_handle)
+
+    if (lhook) call dr_hook('i_hope_you_havent_let_me_down',1,zhook_handle)
+end subroutine i_hope_you_havent_let_me_down
+end module rick_rolled
```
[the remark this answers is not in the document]
reuterbal: No `use rick_astley`, though?

reuterbal (approved):
Thanks a lot, looks good to me!

scripts/loki_transform.py
```diff
@@ -283,7 +300,8 @@ def transform_subroutine(self, routine, **kwargs):
     scheduler.process(transformation=dependency)
 
     # Write out all modified source files into the build directory
-    scheduler.process(transformation=FileWriteTransformation(builddir=out_path, mode=mode, cuf='cuf' in mode))
+    scheduler.process(transformation=FileWriteTransformation(builddir=out_path, mode=mode, cuf='cuf' in mode),
```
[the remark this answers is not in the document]
awnawab: Yes! file graph makes so much more sense, I no longer have to see unintelligibly long log entries from the `FileWriteTransformation`. Thanks! 😄

Issue #96, opened 2023-06-12
FGen: Erroneous line continuation in complicated inline-conditional-call-statement


mlange05:
The following code snippet (never mind the missing declarations), generates wrong line-continuation when run through a simple parse-unparse cycle, as shown below:

```
from loki import Subroutine, fgen, FP

fcode = """
subroutine test_fgen(DIMS, ZSURF_LOCAL)
contains
subroutine test_inline_multiline(KDIMS, LBUD23)

  DO JKGLO=1,NGPTOT,NPROMA
    ! Add saturation adjustment tendencies to cloud scheme (LBUD23)
    IF (LBUD23) CALL UPDATE_FIELDS(YDPHY2,1,DIMS%KIDIA,DIMS%KFDIA,DIMS%KLON,DIMS%KLEV,&
     & PTA1=TENDENCY_LOC%T, PO1=ZSURF_LOCAL%GSD_XA%PGROUP(:,:,19),&
     & PTA2=TENDENCY_LOC%Q, PO2=ZSURF_LOCAL%GSD_XA%PGROUP(:,:,20),&
     & LDV3=YGFL%YL%LT1, PTA3=TENDENCY_LOC%CLD(:,:,NCLDQL), PO3=ZSURF_LOCAL%GSD_XA%PGROUP(:,:,21),&
     & LDV4=YGFL%YI%LT1, PTA4=TENDENCY_LOC%CLD(:,:,NCLDQI), PO4=ZSURF_LOCAL%GSD_XA%PGROUP(:,:,22))
  ENDDO
end subroutine test_inline_multiline
end subroutine test_fgen
""".strip()

routine = Subroutine.from_source(fcode, frontend=FP)

out = fgen(routine, linewidth=132)
print(out)
for line in out.splitlines():
    assert line.count('&') <= 2
```

The resulting printout after fgen formatting is:
```
SUBROUTINE test_fgen (dims, zsurf_local)
  CONTAINS
  SUBROUTINE test_inline_multiline (kdims, lbud23)
    
    
    DO JKGLO=1,NGPTOT,NPROMA
      ! Add saturation adjustment tendencies to cloud scheme (LBUD23)
      IF (LBUD23)       CALL UPDATE_FIELDS(YDPHY2, 1, DIMS%KIDIA, DIMS%KFDIA, DIMS%KLON, DIMS%KLEV, PTA1=TENDENCY_LOC%T,  &
      &
[...]

PR #97, opened 2023-06-12
FGen: Fix bug for inline-conditionals with multi-line output

mlange05:
This was being formatted as single-line, which can create erroneous line breaks if the body is a multi-line statement, as exemplified in the (completely made-up) test. :)

This should address issue #96


reuterbal:
Thanks, looks uncontroversial and well-tested with a properly hard example!

reuterbal (approved):

PR #98, opened 2023-06-12
SCC: Insert vector_length directive into the gang loop

reuterbal:
Setting the vector_length directive to the block size has shown to significantly improve performance for NPROMA numbers that don't match the default value of the architecture. This aims to add this directive to the OpenACC directives inserted in the driver as part of the SCC transformations.

For that, it attempts to find any of the `size_expressions` of the horizontal dimension and use that. Note, that this only works with NVIDIA, the Cray compiler will ignore this value if it is not a compile-time constant.


mlange05 (approved):
Looks great! Well tested, and does exactly what it says on the tin. Significant performance improvements for SCC-hoist and SCC-stack when combined with corresponding CLOUDSC PR.

GTG from me.

PR #100, opened 2023-06-14
Scheduler: Only apply transformations to requested subroutines 

mlange05:
Mnay(!) of our transformations require the same entry-point check that the subroutine about to be processed has indeed been requested by the scheduler. This PR moves this check up to the `Transformation` base class and adds a test to check for the most common violation of this. The problem really arrives if we have modules with multiple subroutines, of which only a subset is meant to be transformed.

As this touches the SCC cases, it would be good to run full regression on this. I've verified that it works with the current EC-physics demonstrator case.

### DIFF
--- a/loki/transform/build_system_transform.py
+++ b/loki/transform/build_system_transform.py
@@ -77,10 +77,6 @@ class CMakePlanner(Transformation):
         item = kwargs['item']
         role = kwargs.get('role')
 
-        # Back out, if this Subroutine is not part of the plan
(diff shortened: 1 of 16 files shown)


reuterbal (approved):
Thanks, very straightforward but well-tested. Good to go!

awnawab (approved):
Not much to add. Thanks for this much needed mop-up job! Super thoroughly tested, GTG 👍

PR #107, opened 2023-06-30
Python 3.11 compatibility

reuterbal:
A small change to achieve compatibility with Python 3.11. Reason is that since 3.11, Enum values do get the type of all mixin base classes and because of the composing nature how this is done, the assertion we had there triggered.
See also https://discuss.python.org/t/inconsistent-behavior-with-python-3-11-enum-mixin-class-behavior/24613/4

I had to also use a plain `Enum` with `int` datatype instead of `IntEnum` to retain the original `str`-behaviour. (Changed with PEP 663)

Testing has been expanded to include Python 3.11

### DIFF
--- a/.github/workflows/code_checks.yml
+++ b/.github/workflows/code_checks.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false  # false: try to complete all jobs
       matrix:
-        python-version: ["3.10"]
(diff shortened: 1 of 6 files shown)


mlange05 (approved):
Looks great to me, many thanks! If someone could convince the over-zealous codecov checker, it'd be GTG

Issue #102, opened 2023-06-15
Install doc for tests recommandation 


meteospike:
At the end of the manual installation, the documentation recommands to run `py.test` from the root directory.
But then pytest will load both `tests/conftest.py` and `transformations/test/conftest.py`, without hierarchical, and I may be wrong but it seems that this leads to some incohenrency.
Either we could change the doc (move to each directory), either we create a coherent set of tests with hierarchical conftest.

reuterbal:
Thanks, the documentation is indeed out of date there and a plain call to py.test leads to conflicts.
While there is certainly scope for refactoring some of the conftest functionality, tests for transformations, lint_rules and the core loki tool should be called separately. This can simply be done as follows:

```
py.test tests
py.test transformations
py.test lint_rules
```

I will clarify the instruction.

reuterbal:
This is related to #33

Issue #33, opened 2023-02-06
Update documentation how to run pytest


reuterbal:
With transformations, lint_rules and loki being separate packages inside a common repository, pytest becomes confused about conftest.py, when running pytest without any arguments.
This can easily be mitigated by specifying the test suite to run:
```
pytest tests
pytest transformations/tests
pytest lint_rules/tests
```
However, documentation should reflect this.

PR #110, opened 2023-06-30
Migrate to pyproject.toml package installation

reuterbal:
This simplifies the package configuration, using only pyproject.toml (and retiring setup.py and setup.cfg), thus implementing PEP 660.

This was not possible until recently, when setuptools added support for editable installations using pyproject.toml alone. That works since 61.0. As a consequence, because most system pip installations are usually older, I have added a pip upgrade to the installation instructions and scripts.

pyproject.toml allows now also to define entry points for pytest, which avoids the problem that running pytests without arguments would fail due to incompatible conftest specifications, as reported in #33 and #102.


mlange05 (approved):
Very neat, and much anticipated. Looks great as always! GTG from me. :shipit:

PR #111, opened 2023-07-01
ECPhys: Add a "kernel-only" option to remove calls

mlange05:
This avoids the removal of DrHook calls in the driver layer, which we actually want for benchmarking.


reuterbal (approved):
Valuable addition and cleanly implemented. GTG

awnawab (approved):
Nothing to add, very useful addition, GTG!

PR #109, opened 2023-06-30
Depend on pydantic<2.0

reuterbal:
Hotfix to mitigate #108

PR #112, opened 2023-07-02
Loki: Add a 1% code coverage degradation threshold

mlange05:
Filing this as a separate PR to get input on the level of degradation we should allow. I've gone for 1% per PR, which would easily work for most of the maintenance PRs, where this is the biggest issue. Opinions and comments welcome.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%


reuterbal:
Looks good. I would suggest merging into #109 to fix checks there, so main is green afterwards

reuterbal (approved):

reuterbal:
1% overall degradation is maybe too generous for larger PRs but it should only be a review aid rather than an absolute metric anyway

PR #118, opened 2023-07-27
Remove `ExprMetadataMixin`

awnawab:
Expression nodes have a `source` attribute, and the smallest unit of source is a source line. Strictly speaking, the `source` property of an expression should be just the expression itself, i.e. smaller than a full-line.

In many instances, the control-flow node `source` can be invalidated but `Expression.source` will still be intact, e.g. after a `Transformer` pass. This can then cause problems when running `fgen` in `conservative=True` mode. When the `fgen` visitor arrives at the `Expression` node in question and finds the `source` property intact, it will simply print `Expression.source` rather than generating Fortran code. Because `Expression.source` points to a full source-code line, incorrect code is generated.

@reuterbal suggested that because we cannot correctly parse the source for `Expression` nodes, they should not have a `source` property at all. So one solution to the above problem is to remove the `source` property from `Expression` nodes. Since the only purpose of `ExprMetaDataMixin` was to add a `source` property to expression nodes, this PR tries to remove `ExprMetaDataMixin`.

As this was a change to the IR, it also affected many ostensibly unrelated parts of the code. Small fixes and updates to tests were thus also required.

NB: These changes have been regression tested against ecWam.

### DIFF
--- a/lint_rules/lint_rules/ifs_coding_standards_2011.py
+++ b/lint_rules/lint_rules/ifs_coding_standards_2011.py
@@ -472,6 +472,15 @@ class Fortran90OperatorsRule(GenericRule):  # Coding standards 4.15
         '<': re.compile(r'(?P<f77>\.lt\.)|(?P<f90><(?!=))', re.I),
     }
 
+    _op_map = {
+        '==': '.eq.',
+        '/=': '.ne.',
+        '>=': '.ge.',
+        '<=': '.le.',
+        '>': '.gt.',
+        '<': '.lt.'
+    }
+
     @classmethod
     def check_subroutine(cls, subroutine, rule_report, config, **kwargs):
         '''Check for the use of Fortran 90 comparison operators.'''
@@ -486,20 +495,31 @@ class Fortran90OperatorsRule(GenericRule):  # Coding standards 4.15
             root_expr_map = defaultdict(list)
             for expr in expr_list:
                 expr_root = FindExpressionRoot(expr).visit(node)[0]
-                if expr_root.source and expr_root.source.string:
+                if node.source and node.source.string:
                     # Include only if we have source string information for this node
                     root_expr_map[expr_root] += [expr]
 
             # Then we look at the comparison operators for each expression root and match
             # them directly in the source string
             for expr_root, exprs in root_expr_map.items():
+                # find source lines for expression root
+                lstart, lend = node.source.find(str(expr_root))
+                lines = node.source.clone_lines((lstart, lend))
+
                 # For each comparison operator, check if F90 or F77 operators are matched
                 for op in sorted({op.operator for op in exprs}):
-                    source_string = strip_inline_comments(expr_root.source.string)
+
+                    # find source line for operator
+                    op_str = op if op != '!=' else '/='
+                    line = [line for line in lines if op_str in strip_inline_comments(line.string)]
+                    if not line:
+                        line = [line for line in lines
+                                if op_str in strip_inline_comments(line.string.replace(cls._op_map[op_str], op_str))]
+
+                    source_string = strip_inline_comments(line[0].string)
                     matches = cls._op_patterns[op].findall(source_string)
                     for f77, _ in matches:
                         if f77:
-                            op_str = op if op != '!=' else '/='
                             msg = f'Use Fortran 90 comparison operator "{op_str}" instead of "{f77}"'
                             rule_report.add(msg, node)
 
--- a/tests/test_visitor.py
+++ b/tests/test_visitor.py
@@ -10,7 +10,7 @@ from pymbolic.primitives import Expression
 
 from conftest import available_frontends
 from loki import (
-    OMNI, FP,
+    OMNI,
     Module, Subroutine, Section, Loop, Assignment, Conditional, Sum, Associate,
     Array, ArraySubscript, LoopRange, IntLiteral, FloatLiteral, LogicLiteral, Comparison, Cast,
     FindNodes, FindExpressions, FindVariables, ExpressionFinder, FindExpressionRoot,
@@ -385,11 +385,6 @@ end subroutine routine_simple
     literal_root = FindExpressionRoot(literals[0][1].pop()).visit(literals[0][0])
     assert literal_root[0] is cast_root[0]
 
-    # Check source properties of expressions (for FP only)
-    if frontend == FP:
-        assert comp_root[0].source.lines == (12, 12)
-        assert cast_root[0].source.lines == (13, 13)
-
 
 @pytest.mark.parametrize('frontend', available_frontends())
 def test_find_expression_root_constructor_args(frontend):
@@ -735,7 +730,7 @@ end subroutine routine_simple
     # Find the conditional in new bodies and check that source is untouched
     assert cond.source is not None
     cond_without_src = get_conditional(body_without_source)
-    assert cond_without_src.source == cond.source
+    assert cond_without_src.source is None
     cond_with_src = get_conditional(body_with_source)
     assert cond_with_src.source == cond.source
 
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -33,7 +33,7 @@ class DataflowAnalysisAttacher(Transformer):
     _mem_property_queries = ('size', 'lbound', 'ubound', 'present')
 
     def __init__(self, **kwargs):
-        super().__init__(inplace=True, **kwargs)
+        super().__init__(inplace=True, invalidate_source=False, **kwargs)
 
     # Utility routines
 
@@ -277,7 +277,7 @@ class DataflowAnalysisDetacher(Transformer):
     """
 
     def __init__(self, **kwargs):
-        super().__init__(inplace=True, **kwargs)
+        super().__init__(inplace=True, invalidate_source=False, **kwargs)
 
     def visit_Node(self, o, **kwargs):
         o._update(_live_symbols=None, _defines_symbols=None, _uses_symbols=None)
--- a/loki/expression/operations.py
+++ b/loki/expression/operations.py
@@ -15,7 +15,7 @@ from sys import intern
 import pymbolic.primitives as pmbl
 
 from loki.expression.symbols import (
-    StringLiteral, ExprMetadataMixin, Sum, Product, Quotient, Power,
+    StringLiteral, Sum, Product, Quotient, Power,
     loki_make_stringifier
 )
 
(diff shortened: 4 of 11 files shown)


lint_rules/lint_rules/ifs_coding_standards_2011.py
```diff
@@ -486,20 +495,31 @@ def check_subroutine(cls, subroutine, rule_report, config, **kwargs):
             root_expr_map = defaultdict(list)
             for expr in expr_list:
                 expr_root = FindExpressionRoot(expr).visit(node)[0]
```
reuterbal, line 497: This could go inside the conditional

loki/transform/transform_loop.py
```diff
@@ -654,7 +654,7 @@ class FissionTransformer(NestedMaskedTransformer):
 
     def __init__(self, loop_pragmas, active=True, **kwargs):
         super().__init__(active=active, require_all_start=True, greedy_stop=True, **kwargs)
-        self.loop_pragmas = loop_pragmas
+        self.loop_pragmas = dict((k, v) for k, v in loop_pragmas.items())
```
reuterbal, line 657: Why is that required? And would `self.loop_pragmas = loop.pragmas.copy()` achieve the same?

reuterbal (approved):
Thanks so much, this looks great and given the potential impact of a change like this, the fallout outside immediately affected files is surprisingly small. Even more of a reason to remove this now before it becomes a dependency elsewhere.

loki/transform/transform_loop.py
awnawab: dict lookups kept failing, even though the keys were definitely in the dict. It was a strange bug; I could get it to pass by printing the keys before the lookup. This seems to have solved the problem.

awnawab: No, the bug reappears if I use `loop_pragmas.copy()`.

lint_rules/lint_rules/ifs_coding_standards_2011.py
awnawab: Done!

tests/test_scheduler.py
```diff
@@ -1572,7 +1572,7 @@ def test_scheduler_nested_type_enrichment(frontend, config):
     assert calls[1].name.parent.parent.parent.type.dtype.name == 'third_type'
     assert isinstance(calls[1].name.parent.parent.parent.type.dtype.typedef, TypeDef)
 
-    inline_calls = FindInlineCalls().visit(driver.body)
+    inline_calls = FindInlineCalls(unique=False).visit(driver.body)
```
mlange05, line 1575: This seems entirely unrelated to the PR. Is there a deeper reason for this?

tests/test_visitor.py
```diff
@@ -845,7 +840,7 @@ def test_transformer_multinode_keys(frontend):
                 body=tuple(a.clone() for a in assigns))
     # Need to use NestedTransformer here, since replacement contains
     # the original nodes.
-    transformed = NestedTransformer({assigns: loop}).visit(routine.body)
+    transformed = NestedTransformer({assigns[0]: loop, assigns[1]: None}).visit(routine.body)
```
mlange05, line 843: This really does change the nature of the test (sub-tuple detection). Is this `Transformer` feature truly dependent on the source property of sub-expressions? The behaviour of transformers should not really be influenced by `expr.source` property, so I'm not sure why this is necessary?

transformations/transformations/single_column_coalesced.py
```diff
@@ -728,9 +728,8 @@ def hoist_temporary_column_arrays(cls, routine, call, horizontal, vertical, bloc
         v_end = arg_map[kernel.variable_map[horizontal.bounds[1]]]
         bounds = sym.LoopRange((v_start, v_end))
         vector_loop = ir.Loop(variable=v_index, bounds=bounds, body=(new_call,))
-        call_map[call] = vector_loop
 
-        routine.body = Transformer(call_map).visit(routine.body)
```
mlange05, line 733: Why is this Transformer call being hoisted here? The new control flow is really counter-intuitive (`if call_map:` :raised_eyebrow: ), so I'm again not seeing how this is related to the removal of the `expr.source` property?

transformations/transformations/single_column_coalesced.py
```diff
@@ -797,10 +796,14 @@ def process_driver(self, routine, targets=None, item=None):
...
         for call in FindNodes(ir.CallStatement).visit(routine.body):
             if not call.name in targets:
                 continue
 
-            self.hoist_temporary_column_arrays(routine, call, self.horizontal, self.vertical,
-                                               self.block_dim, item=item)
+            call_map[call] = self.hoist_temporary_column_arrays(routine, call, self.horizontal, self.vertical,
+                                                                self.block_dim, item=item)
+
+        if call_map:
```
mlange05, line 808: I find this quite counter intuitive to collect replacements from fairly atomic-looking transformation steps. Is this just to save traversals, or is this actually related to the `expr.source` change?

transformations/transformations/single_column_coalesced_wrapper.py
```diff
@@ -218,20 +219,26 @@ def process_driver(self, routine, targets=None, item=None):
...
-                if item:
-                    column_locals = item.trafo_data['SCCHoistTransformation'].get('column_locals', None)
-                if self.directive == 'openacc':
-                    SCCAnnotateTransformation.device_alloc_column_locals(routine, column_locals)
+                call_map[call] = SCCHoistTransformation.hoist_temporary_column_arrays(routine, call, self.horizontal,
+                                                                                      self.vertical, self.block_dim,
+                                                                                      item=item)
+
+        if call_map:
+            routine.body = Transformer(call_map).visit(routine.body)
```
mlange05, line 234: Same as above....

lint_rules/lint_rules/debug_rules.py
```diff
@@ -284,13 +284,25 @@ def fix_subroutine(cls, subroutine, rule_report, config):
...
+        single_variable_declaration(subroutine)
+
+        #TODO: 'VariableDeclaration.symbols' should be of type 'Variable' rather than 'Expression'
         # to enable case-insensitive search here
         new_var_names = [v.name.lower() for v in new_vars]
-        subroutine.variables = [var for var in subroutine.variables if not var.name.lower() in new_var_names]
 
-        subroutine.body = Transformer(node_map).visit(subroutine.body)
-        subroutine.variables += new_vars
+        routine = subroutine.clone()
```
mlange05, line 294: This seems pretty invasive as a change, given none of the expression manipulation below seems to deal with source strings. Is this strictly needed, or should this be in a separate PR?

mlange05 (commented):
All in all, I think I agree with this change and the rationale behind it. However, most of the associated fixes and changes seem fairly unrelated to the removal of the `source` property from expression nodes. Since this is a fairly invasive change in nature, I'd like to try and keep this free of piggy-backing, so I've flagged a few places where I can't see the connection, and would like to know why this is needed. 

I'll have a quick play with a few of these changes as well to try and understand why these might be needed. In the meantime I'll class this as comment, for approval once we're sure what is actually needed for this. 

I've also run the full ec-physics regression suite with this, and it seems to work very nicely. While the runtime of the SCC transformations seems unaffected, I believe I've noticed a roughly 8-10% speed-up in raw file parsing on average, which would be a very welcome improvement. :grin:

loki/transform/transform_loop.py
mlange05: This really does sound odd. Any chance we can figure out why such a seemingly unrelated change "fixes" things?

lint_rules/lint_rules/debug_rules.py
awnawab: Yes in hindsight, piggy-backing `DebugRule` fixes onto this PR was not the wisest move. I'll remove those commits, and add them to some other `DebugRule` fixes that I plan on PR'ing soon.

awnawab:
Thanks a lot for the feedback @mlange05. I'll look in detail into the issues you've highlighted, and I'll un piggy-back the `DebugRule` fixes and put them in a separate PR.

tests/test_scheduler.py
awnawab: Now the tests pass even without the `unique=False` argument 🤔 Don't know why I needed this at the time, but I've gotten rid of it.

tests/test_visitor.py
awnawab: In the `Transformer._rebuild` method, before invalidating a node's `source` we check if its children have a `source` property. After removing `Expression.source` the children of an `Assignment` no longer have any source. So that test now passes and we invalidate `Assignment.source` in a transformer pass whereas previously we did not. The reason we don't hit this problem when the mapper key is a node rather than a tuple is that in the former case, we do the `is o in self.mapper` check before rebuilding `o`.
The cleanest way I can think to resolve this: in a `Transformer` pass, if `self.invalidate_source` is true we invalidate the source of mapper keys.

loki/transform/transform_loop.py
awnawab: The `DataflowAnalysisAttacher/Detacher` by default had `invalidate_source` set to true. And now that the expression nodes no longer have a source, calls to `dataflow_analysis_attached` were invalidating source. This was the source (pun-intended) of those funky dict lookup failures.

I think the `DataflowAnalysis` should not modify the nodes, since it is an analysis pass. So I'm setting `invalidate_source` to false in the constructor.

lint_rules/lint_rules/ifs_coding_standards_2011.py
reuterbal: Can't see the change, did this go under somehow? The entire loop could be wrapped in the conditional:
```
if node.source and node.source.string:
    for expr in expr_list:
        expr_root = FindExpressionRoot(expr).visit(node)[0]
        root_expr_map[expr_root] += [expr]
```

reuterbal (approved):

lint_rules/lint_rules/ifs_coding_standards_2011.py
awnawab: I'd piggybacked some other stuff onto this PR but separated that into #126. This change accidentally wound up in that PR. Are you happy to leave it as is or should I move that commit to this PR?

reuterbal: Either is fine - but I can't find the change in #126 either?

awnawab: I think I misunderstood your original comment. I thought you meant move `_op_map` inside the conditional. But you were referring to moving the `for expr_root, exprs in root_expr.items()` loop into the `if node.source and node.source.string` conditional right?

reuterbal: Right, sorry, yes - that's what I meant. It wasn't possible before because the `source` property was used from the expression but now it's loop-invariant.

awnawab: Right. Sure, I'll move that loop inside the conditional, and also rebase #126 👍

mlange05 (approved):
Ok, I had another look and all looks to be addressed and regression still looks sane too. Sorry for taking long to look through these. GTG from me now.

PR #119, opened 2023-07-28
Add ecWam regression tests

awnawab:
This PR tries to add ecWam regression tests to the Loki CI suite.

### DIFF
--- a/.github/workflows/regression_tests.yml
+++ b/.github/workflows/regression_tests.yml
@@ -34,15 +34,11 @@ jobs:
           ref: develop
 
       - name: Clone ECWAM
-        run: |
-          mkdir ${{ github.workspace }}/ecwam
-          mkdir -p ~/.ssh/ && touch ~/.ssh/known_hosts
-          cat << 'EOF' > ~/.ssh/known_hosts
-          ${{ secrets.BITBUCKET_HOSTKEY }}
-          EOF
-          eval `ssh-agent -s`
-          ssh-add - <<< '${{ secrets.ECWAM_REPOSITORY_SSH_KEY }}'
-          git clone --branch naan-phys-gpu ssh://git@git.ecmwf.int/~nawd/ecwam-bundle.git ${{ github.workspace }}/ecwam
+        uses: actions/checkout@v3
+        with:
+          repository: ecmwf-ifs/ecwam
+          path: ecwam
+          ref: naan-phys-gpu
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
--- a/.github/workflows/regression_tests.yml
+++ b/.github/workflows/regression_tests.yml
@@ -33,6 +33,17 @@ jobs:
           path: cloudsc
           ref: develop
 
+      - name: Clone ECWAM
+        run: |
+          mkdir ${{ github.workspace }}/ecwam
+          mkdir -p ~/.ssh/ && touch ~/.ssh/known_hosts
+          cat << 'EOF' > ~/.ssh/known_hosts
+          ${{ secrets.BITBUCKET_HOSTKEY }}
+          EOF
+          eval `ssh-agent -s`
+          ssh-add - <<< '${{ secrets.ECWAM_REPOSITORY_SSH_KEY }}'
+          git clone --branch naan-phys-gpu ssh://git@git.ecmwf.int/~nawd/ecwam-bundle.git ${{ github.workspace }}/ecwam
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
@@ -64,11 +75,25 @@ jobs:
           source loki-activate
           pip install cmake
 
-      - name: Run CLOUDSC regression tests
+      - name: Install ECWAM dependencies
+        run: |
+          sudo apt-get -o Acquire::Retries=3 install -y libopenmpi-dev
+          source loki-activate
+          pip install fypp
+
+      - name: Run CLOUDSC and ECWAM regression tests
         env:
           CLOUDSC_DIR: ${{ github.workspace }}/cloudsc
           CLOUDSC_ARCH: ${{ github.workspace }}/cloudsc/arch/github/ubuntu/gnu/9.4.0
+          ECWAM_DIR: ${{ github.workspace }}/ecwam
+          ECWAM_ARCH: ${{ github.workspace }}/ecwam/arch/github/ubuntu/gnu/9.4.0
           OMP_STACKSIZE: 4G
         run: |
+          mkdir -p ~/.ssh/ && touch ~/.ssh/known_hosts
+          cat << 'EOF' > ~/.ssh/known_hosts
+          ${{ secrets.BITBUCKET_HOSTKEY }}
+          EOF
+          eval `ssh-agent -s`
+          ssh-add - <<< '${{ secrets.ECWAM_REPOSITORY_SSH_KEY }}'
           source loki-activate
-          pytest --cov=transformations/transformations transformations/tests -k cloudsc
+          pytest --cov=transformations/transformations transformations/tests -k 'cloudsc or ecwam'
--- a/transformations/tests/conftest.py
+++ b/transformations/tests/conftest.py
@@ -5,10 +5,10 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+from pathlib import Path
 import shutil
 import pytest
 import yaml
-from pathlib import Path
 
 from loki import as_tuple, Frontend
 import loki.frontend
--- a/transformations/tests/conftest.py
+++ b/transformations/tests/conftest.py
@@ -13,7 +13,7 @@ import yaml
 from loki import as_tuple, Frontend
 import loki.frontend
 
-__all__ = ['available_frontends', '_write_script', '_local_loki_bundle']
+__all__ = ['available_frontends', 'local_loki_setup', 'local_loki_cleanup']
 
 def available_frontends(xfail=None, skip=None):
     """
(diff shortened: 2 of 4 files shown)


.github/workflows/regression_tests.yml
```diff
@@ -33,6 +33,17 @@ jobs:
...
           ref: develop
 
+      - name: Clone ECWAM
+        run: |
+          mkdir ${{ github.workspace }}/ecwam
+          mkdir -p ~/.ssh/ && touch ~/.ssh/known_hosts
+          cat << 'EOF' > ~/.ssh/known_hosts
+          ${{ secrets.BITBUCKET_HOSTKEY }}
+          EOF
+          eval `ssh-agent -s`
```
reuterbal, line 43: Is there a specific reason why this is wrapped in an `eval`?

.github/workflows/regression_tests.yml
```diff
@@ -64,11 +75,25 @@ jobs:
           source loki-activate
           pip install cmake
 
-      - name: Run CLOUDSC regression tests
+      - name: Install ECWAM dependencies
+        run: |
+          sudo apt-get -o Acquire::Retries=3 install -y libopenmpi-dev
+          source loki-activate
+          pip install fypp
```
reuterbal, line 82: [No action required] This really ought to be happening in the FIELD_API CMake (I'll get to that, promised!)

transformations/tests/test_cloudsc.py
```diff
@@ -27,54 +25,32 @@ def fixture_here():
...
+def fixture_write_script():
+    return _write_script
+
+
 @pytest.mark.usefixtures('bundle_create')
 @pytest.mark.parametrize('frontend', available_frontends(
     skip=[(OMNI, 'OMNI needs FParser for parsing dependencies')] if not HAVE_FP else None
 ))
-def test_cloudsc(here, frontend):
+def test_cloudsc(here, write_script, frontend):
```
reuterbal, line 53: `write_script` doesn't need to be a fixture, that can just be a regular function call.

transformations/tests/test_ecwam.py
```diff
@@ -0,0 +1,119 @@
...
+    bundle = str(next(local_loki_bundle(here)))
+    execute(
+        ['./ecwam-bundle', 'create', '--bundle', bundle],
+        cwd=here,
+        silent=False
+    )
+
+
+@pytest.fixture(scope='module', name='write_script')
+def fixture_write_script():
```
reuterbal, line 42: Same comments apply here as for CLOUDSC

reuterbal (changes requested):
Thanks so much, this is fantastic and well done for keeping patiently tweaking the run config until it eventually worked!

Very nice that you generalized the CLOUDSC setup instead of simply replicating the code. However, I have a few comments on the specific implementation which I think could be improved.

.github/workflows/regression_tests.yml
awnawab: The `eval` is to load the environment variables output by `ssh-agent`.

reuterbal: Ah, of course!

.github/workflows/regression_tests.yml
awnawab: So FIELD_API should install fypp if not found? I also use fypp in ecWam: I use a yaml config file to define the derived-types and use fypp to generate the FIELD_API glue code. So fypp and pyyaml need to be installed on the same interpreter. That's why I source loki and then install fypp, since the ecWam build is done with `loki-env` activated and it already has pyyaml.

transformations/tests/conftest.py
```diff
@@ -68,3 +72,50 @@ def my_test(frontend):
             params += [f]
 
     return params
+
+
+def _write_script(here, binary, args):
```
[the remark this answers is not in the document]
awnawab: Yes sure, good idea!

transformations/tests/conftest.py
```diff
@@ -68,3 +72,50 @@ def my_test(frontend):
...
+source env.sh >&2
+bin/{binary} {' '.join(args)}
+exit $?
+    """.strip())
+    script.chmod(0o750)
+
+    return script
+
+
+def _local_loki_bundle(here):
```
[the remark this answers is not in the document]
awnawab: Ah yes that would be much nicer! I've reached out to Willem asking for advice.

transformations/tests/conftest.py
```diff
@@ -68,3 +72,50 @@ def my_test(frontend):
...
+    # Change bundle to symlink for Loki
+    bundle = yaml.safe_load(bundlefile.read_text())
+    loki_index = [i for i, p in enumerate(bundle['projects']) if 'loki' in p]
+    assert len(loki_index) == 1
+    if 'git' in bundle['projects'][loki_index[0]]['loki']:
+        del bundle['projects'][loki_index[0]]['loki']['git']
+    bundle['projects'][loki_index[0]]['loki']['dir'] = str(lokidir.resolve())
+    local_loki_bundlefile.write_text(yaml.dump(bundle))
+
+    yield local_loki_bundlefile
```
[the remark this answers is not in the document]
awnawab: Thanks for clearing this up; as you could probably tell I wasn't quite clear on the distinction between fixtures and functions 😅

.github/workflows/regression_tests.yml
reuterbal: Yeah, the idea would essentially be to export `fypp` as a CMake target. If it doesn't exist, yet, try a `find_program` and export it, otherwise create a venv and install it in there. That pattern can be used in all projects that use fypp and the first in any bundle will automatically become the one to install it. But I'll do this soon ™️

transformations/tests/conftest.py
```diff
@@ -68,3 +72,45 @@ def my_test(frontend):
...
+    bundlefile = here/'bundle.yml'
+
+    # Do not overwrite any existing Loki copy
+    if target.exists():
+        if backup.exists():
+            shutil.rmtree(backup)
+        shutil.move(target, backup)
+
+    # Change bundle to symlink for Loki
+    bundle = yaml.safe_load(bundlefile.read_text())
```
reuterbal, line 105: Is this still needed now? I thought by calling the create step with a pointer to the loki dir we could cut out the need to modify bundle.yml entirely?

awnawab: I wanted to keep the assert where we check that loki is in the original `bundle.yml`. If we don't want to check that then yeah we can get rid of this completely.

transformations/tests/test_cloudsc.py
```diff
@@ -28,45 +28,21 @@ def fixture_here():
...
+    lokidir, target, backup = local_loki_setup(here)
+    yield lokidir
+    local_loki_cleanup(target, backup)
 
 
 @pytest.fixture(scope='module', name='bundle_create')
 def fixture_bundle_create(here, local_loki_bundle):
+    """Inject ourselves into the CLOUDSC bundle"""
+    os.environ['CLOUDSC_BUNDLE_LOKI_DIR'] = local_loki_bundle
+    env = os.environ.copy()
```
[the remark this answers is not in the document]
awnawab: Yes agreed, silly oversight, I'll fix that. Thanks!

reuterbal (approved):
Very neat, thanks so much!

.github/workflows/regression_tests.yml
```diff
@@ -33,6 +33,17 @@ jobs:
...
+      - name: Clone ECWAM
+        run: |
+          mkdir ${{ github.workspace }}/ecwam
+          mkdir -p ~/.ssh/ && touch ~/.ssh/known_hosts
+          cat << 'EOF' > ~/.ssh/known_hosts
+          ${{ secrets.BITBUCKET_HOSTKEY }}
+          EOF
+          eval `ssh-agent -s`
+          ssh-add - <<< '${{ secrets.ECWAM_REPOSITORY_SSH_KEY }}'
+          git clone --branch naan-phys-gpu ssh://git@git.ecmwf.int/~nawd/ecwam-bundle.git ${{ github.workspace }}/ecwam
```
mlange05, line 45: Uhhhhh, does ecWAM not self-bundle? Seems pretty back-bending to go back to BITBUCKET just for the bundle. Do we really need this, or could this be fixed upstream in github to avoid this?

awnawab: No ecWam does not self-bundle. I don't think Willem would be opposed to that, I can speak to him about including the bundle in the ecwam repo, or as a middle-ground to have the bundle on github so we don't need to go to bitbucket.

awnawab:
@mlange05 I've added ecwam-bundle to the ecwam repo, so we no longer need to access to bitbucket for that. We still need access to bitbucket for FIELD_API but once I've brought the new FIELD_API up-to-date with our stuff we can remove that too. I've also unstacked it from #118, so this one can go in independently of that.

.github/workflows/regression_tests.yml
```diff
@@ -64,11 +71,25 @@ jobs:
...
+          ECWAM_DIR: ${{ github.workspace }}/ecwam
+          ECWAM_ARCH: ${{ github.workspace }}/ecwam/arch/github/ubuntu/gnu/9.4.0
           OMP_STACKSIZE: 4G
         run: |
+          mkdir -p ~/.ssh/ && touch ~/.ssh/known_hosts
+          cat << 'EOF' > ~/.ssh/known_hosts
+          ${{ secrets.BITBUCKET_HOSTKEY }}
+          EOF
+          eval `ssh-agent -s`
+          ssh-add - <<< '${{ secrets.ECWAM_REPOSITORY_SSH_KEY }}'
```
mlange05, line 93: [no action] I take it this is needed for the FIELD API download the ecWAM @ Loki still needs?

mlange05 (approved):
Very nice and super useful! Very happy about this! GTG! :shipit:

.github/workflows/regression_tests.yml
mlange05: I believe this has now been rectified and updated on ecWAM side. So all good on this one!

.github/workflows/regression_tests.yml
awnawab: Yes, we can get rid of this once I've tested the new FIELD_API with CLOUDSC and ecWAM.

reuterbal:
Agreed, this is great and much cleaner with the bundle on ecwam's side. For the record, adding the corresponding PR (https://github.com/ecmwf-ifs/ecwam/pull/3) here to provide the link, but the integration in this PR works with a branch still and is therefore not depending on that one to be merged.

PR #121, opened 2023-08-02
Fix #120: Test and fix constructor arg conflict in FindExpressionRoot

### LINKED ISSUE #120: Prevent multiple values conflict in FindExpressionRoot init
In module `loki.expression.expr_visitors` the initialization of class `FindExpressionRoot` does not check for `unique` keyword argument presence, which could lead to a multiple values conflict since this argument is statically set in the `super.__init__`  call.

This could be easily fixed, adding the line:
`kwargs.pop('unique', None)`

reuterbal:


reuterbal:
Not sure why codecov thinks that coverage went down

mlange05 (approved):
Looks and nicely tested. Many thanks to @meteospike for reporting this. GTG from me.

PR #124, opened 2023-08-11
Remove SCC wrapper transformation

mlange05:
This PR does pretty much what it says on the tin: Remove the old SCC wrapper transformation and instead use the component-wise composition in the `ecphys` entry point. In addition, it includes some slight cosmetics improvements and removes the redundant duplicated tests in the `transformations` sub-package. 

I've tested this with the Cy48R1 ec-physics demonstrator and things are fine. Please note that I haven't enabled the vector-section trim or global var offloads yet, because I'm planning to consolidate this with the convert entry point soon, and possibly more these into the config file.


scripts/loki_transform.py
```diff
@@ -467,39 +474,63 @@ def ecphys(mode, config, header, source, build, cpp, directive, frontend):
...
-    # Now we instantiate our transformation pipeline and apply the main changes
+    # Instantiate transformation pipeline and apply the main changes
     transformation = None
     if mode == 'idem':
         transformation = IdemTransformation()
 
     if mode == 'sca':
         # Define the target dimension to strip from kernel and caller
-        horizontal = scheduler.config.dimensions['horizontal']
         transformation = ExtractSCATransformation(horizontal=horizontal)
```
awnawab, line 491: Do we not need to wrap this in a tuple? Same for the idem trafo above.

transformations/tests/test_single_column_coalesced.py
```diff
@@ -222,59 +222,6 @@ def test_scc_base_masked_statement(frontend, horizontal):
     assert fgen(kernel_conds[0].else_body) == 'q(jl, jk) = t(jl, jk)'
 
 
-@pytest.mark.parametrize('frontend', available_frontends())
```
reuterbal, line 225: Just to make sure we're not loosing test coverage: The deleted tests were essentially redundant in the sense that they tested functionality with the wrapper that's already been checked on the invidividual relevant transformations?

reuterbal (approved):
Looks good to me. I agree with @awnawab's comment regarding wrapping in tupels for other transformation modes, currently this looks like it might break.

transformations/tests/test_single_column_coalesced.py
mlange05: Yes, I believe so. @awnawab might be better placed to confirm this, but I've glanced over them and they looked fairly identical still.

scripts/loki_transform.py
mlange05: Good catch! I've added an `as_tuple` around the transformations in the application loop, which will have the same effect, but makes it a little safer.

transformations/tests/test_single_column_coalesced.py
awnawab: Yup, all the tests with 'wrapper' in the name were essentially redundant.

awnawab (approved):
Thanks for adding the `as_tuple`, looks good to go 👍

PR #125, opened 2023-08-11
SCC: Better logging and warning for "bad" code constructs 

mlange05:
_Note, this PR stack on top of #124 and should only be reviewed once this is merged._

A small housekeeping PR that homogenises the detailed logging info in Loki-SCC passes (after the removal of the wrappers), and adds some more warnings for "correct, but bad" code constructs that should be resolved in the original source. This includes private array declarations at the "vector" level and wrong ordering of parallel directives (`loop vector` inside `loop seq`).

Please note that the latter requires more advanced features (proper source-level subroutine inlining) that are currently still under development to truly mitigate. In the meantime, removing some of these by hand seems to have small positive effects in ec-physics regression.

### DIFF
--- a/loki/transform/transform_array_indexing.py
+++ b/loki/transform/transform_array_indexing.py
@@ -495,4 +495,4 @@ def demote_variables(routine, variable_names, dimensions):
             )
     routine.spec = Transformer(decl_map).visit(routine.spec)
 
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Uncontroversial and very useful. Many thanks!

PR #126, opened 2023-08-14
Linter debug rule fixes

awnawab:
A small PR to fix errors in the linter debug rules:
- `ArgSizeMismatchRule`: Loki now checks up to a maximum of two indirections for dimension names
- `DynamicUboundCheckRule`: Fixer method now returns `node_map` to trigger source invalidation
- `DynamicUboundCheckRule`: Fixer method now preserves subroutine signature when updating argument declarations
- `Fortran90OperatorsRule`: Class member was added unnecessarily during removal of `Expression.source`. This is now a local object.

NB: This PR is stacked on top of #118 because the `DynamicUboundCheckRule` fixer method relies on the removal of `Expression.source`.


lint_rules/lint_rules/debug_rules.py
```diff
@@ -183,9 +183,22 @@ def check_subroutine(cls, subroutine, rule_report, config, **kwargs):
                 dummy_size = Product(dummy_arg_size)
                 stat = cls.compare_sizes(arg_size, alt_arg_size, dummy_size)
 
-                # if necessary, update dimension names and check
-                if not stat:
-                    dummy_size = Product(SubstituteExpressions(assign_map).visit(dummy_arg_size))
+                # we check for a maximum of two indirections for the dummy and arg dimension names
+                for _ in range(2):
```
reuterbal, line 187: Can we make this indirection depth a config value? That would allow overwriting the value from outside if necessary. See for example https://github.com/ecmwf-ifs/loki/blob/5ed914c00d371fb356889aa457118fe1de0502f3/lint_rules/lint_rules/ifs_coding_standards_2011.py#L36-L38

lint_rules/lint_rules/ifs_coding_standards_2011.py
```diff
@@ -486,20 +486,39 @@ def check_subroutine(cls, subroutine, rule_report, config, **kwargs):
...
+
                 # For each comparison operator, check if F90 or F77 operators are matched
                 for op in sorted({op.operator for op in exprs}):
-                    source_string = strip_inline_comments(expr_root.source.string)
+
+                    # find source line for operator
+                    op_str = op if op != '!=' else '/='
+                    line = [line for line in lines if op_str in strip_inline_comments(line.string)]
+                    if not line:
+                        _op_map = {
```
reuterbal, line 507: Just curious: any particular reason for moving the map here?

reuterbal (approved):
Looks good to me, needs to land after #118, though.

lint_rules/lint_rules/ifs_coding_standards_2011.py
awnawab: As explained in #118, this ended up here because I misunderstood one of your earlier comments 😅

lint_rules/lint_rules/debug_rules.py
awnawab: Aah that is a great idea! Please wait to merge until I've added this.

lint_rules/lint_rules/debug_rules.py
```diff
@@ -284,13 +303,24 @@ def fix_subroutine(cls, subroutine, rule_report, config):
                 vtype = arg.type.clone(shape=new_shape, scope=subroutine)
                 new_vars += as_tuple(arg.clone(type=vtype, dimensions=new_shape, scope=subroutine))
 
-        #TODO: add 'VariableDeclaration.symbols' should be of type 'Variable' rather than 'Expression'
+        # simplify variable declarations
+        single_variable_declaration(subroutine)
```
mlange05, line 307: Ok, apologies if I'm misreading this, but from what I understand, fix_subroutine is the auto that is intended to "correct" small violations for code that gets put back into source, right? With that in mind, splitting variable declarations looks like a very invasive change for a linter! I know we like these for GPU-specific transformations, but forcing this into every file of a large code base for mere convenience seems like a pretty hard no, if we want to bring "fixed" code back into any repo. Is this really intended here, or am I misunderstanding this?

awnawab: Your understanding is correct....this was an easy yet robust way to deal with any potential corner case of array declarations. I thought it would be ok given how invasive this fixer method is already.

Thanks a lot for pointing this out, I'll stop being lazy and make the fixer method more sophisticated 😅

mlange05 (approved):
OK, thanks for addressing my previous comment. All looks ok form my end now. GTG :shipit:

PR #129, opened 2023-08-23
Skip tests if graphviz is not installed or functional

rolfhm:
An attempt to detect if graphviz is actually installed, and not just the python wrapper.

On my laptop, no tests are skipped, even if I have not installed the wrapper. The same is true for the main branch. Is there an issue with the skip mechanism?


tests/test_scheduler.py
```diff
@@ -102,6 +101,19 @@ def fixture_frontend():
...
 
+def graphviz_present():
+    """
+    Test if graphviz is present and works
+    """
+    try:
+        import graphviz as gviz
+        callgraph = gviz.Graph()
+        callgraph.pipe()
+    except ImportError:
```
reuterbal, line 112: Is `ImportError` really enough here? My understanding was that the import would succeed if the Python wrappers are installed but the actual graph generation was what failed. That throws a different exception, presumably?

tests/test_scheduler.py
```diff
@@ -102,6 +101,19 @@ def fixture_frontend():
     return FP if HAVE_FP else OFP
 
 
+def graphviz_present():
+    """
+    Test if graphviz is present and works
+    """
+    try:
+        import graphviz as gviz
```
reuterbal, line 109: This needs annotating with `# pylint: disable=import-outside-toplevel`

tests/test_scheduler.py
```diff
@@ -102,6 +101,19 @@ def fixture_frontend():
...
+        callgraph = gviz.Graph()
+        callgraph.pipe()
```
reuterbal, line 111: This could probably be shortened to gviz.Graph().pipe() since you don't care about the actual return value.

reuterbal (changes requested):
Thanks, this is great.
Regarding your question: note that Loki declares the Python wrappers as a dependency, i.e. even if you don't install them yourself they should be installed automatically with Loki - hence the import is likely to succeed but the actual use of Graphviz then throws an exception.

tests/test_scheduler.py
rolfhm: Thanks, done

rolfhm:
Now it works for the scheduler tests, but `test_cmake_plan` still fails because it looks for the graph pdf
https://github.com/ecmwf-ifs/loki/blob/0db337cdeed57d2a06d7694c05993ea4d719be46/tests/test_cmake.py#L207
Is it necessary to look for the pdf here? The test passes if I comment out the look for pdf part.

If it is necessary, is it possible to skip part of a test? Where should I put the `graphviz_present()` function go to be available to both tests.

What are all the regression tests that keep failing here on github?

tests/test_scheduler.py
```diff
@@ -102,6 +101,23 @@ def fixture_frontend():
     return FP if HAVE_FP else OFP
 
 
+def graphviz_present():
```
reuterbal, line 104: If this needs to be available for more than one test, we can put it in conftest.py

tests/test_cmake.py
```diff
@@ -204,7 +204,6 @@ def test_cmake_plan(srcdir, config, cmake_project, loki_install, ecbuild, silent
 
             # Make sure the plan files have been created
             assert (builddir/'loki_plan.cmake').exists()
-            assert (builddir/'loki_callgraph.pdf').exists()
```
reuterbal, line 207: It should be enough to guard this with an
```python
if graphviz_present():
    assert (builddir/'loki_callgraph.pdf').exists()
```

reuterbal (changes requested):
Thanks, I think we can put the `graphviz_present` utility in conftest to make it available to more than one test module. That would then allow to conditionally skip the test for the PDF.

Otherwise looks great!

Don't worry about the failing regression tests. These currently require access to a secret to use the field-api based regression tests, which isn't available in PRs from outside the organisation. That's a bit of an unfortunate temporary situation until field-api is available open-source.

tests/test_scheduler.py
rolfhm: Fixed and added a bit more explanation

reuterbal (approved):
Thanks, looks great now! I'll let @mlange05 flag any objections, otherwise will merge tomorrow.

PR #130, opened 2023-08-24
Member inlining utility transform

mlange05:
_Note: I'm filing this early to enable debugging with external use cases. It is entirely likely that I've missed edge cases and would like to weed these out early, if possible. Interested user should feel free to comment here with edge-cases that break this feature._

This PR brings in an initial implementation of an internally supported source-level member subroutine inlining utility that substitutes a call to a member routine (subroutine contained within another subroutine with a shared variable declaration space) with its body. This is then exposed to the SCC-style family of transformations via an option `inline_members`, both via the CMake interfaces and the general `loki_transform.py` head script ("convert" and "ecphys" mode).  

This will perform matching of sub-array argument references and map them to the corresponding partial local array indices; for example passing a sub-array to a one-dimensional array argument (`call routine(a=b(:, j)`) will map the local expression `a(i)` to `b(i, j)` in the substituted subroutine body.

The current implementation is limited to subroutines only (no functions, yet!) and only to members, although we are planning to extend this to more a generic source-inlining utility for generic subroutine calls soon.

### DIFF
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -233,7 +233,7 @@ def inline_member_routine(routine, member):
     parent_variables = routine.variable_map
     duplicate_locals = tuple(
         v for v in member.variables
-        if v.name in parent_variables and v.name not in member._dummies
+        if v.name in parent_variables and v.name.lower() not in member._dummies
     )
     shadow_mapper = SubstituteExpressions(
         {v: v.clone(name=f'{member.name}_{v.name}') for v in duplicate_locals}
@@ -243,7 +243,7 @@ def inline_member_routine(routine, member):
 
     # Get local variable declarations and hoist them
     decls = FindNodes(VariableDeclaration).visit(member.spec)
-    decls = tuple(d for d in decls if all(s.name not in routine._dummies for s in d.symbols))
+    decls = tuple(d for d in decls if all(s.name.lower() not in routine._dummies for s in d.symbols))
     decls = tuple(d for d in decls if all(s not in routine.variables for s in d.symbols))
     routine.spec.append(decls)
 
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -218,10 +218,19 @@ def inline_member_routine(routine, member):
         For example, mapping the passed array ``m(:,j)`` to the local
         expression ``a(i)`` yields ``m(i,j)``.
         """
-        val_free_dims = tuple(d for d in val.dimensions if isinstance(d, sym.Range))
-        var_bound_dims = tuple(d for d in var.dimensions if not isinstance(d, sym.Range))
-        mapper = SubstituteExpressionsMapper(dict(zip(val_free_dims, var_bound_dims)))
-        return mapper(val)
+        new_dimensions = list(val.dimensions)
+
+        indices = [index for index, dim in enumerate(val.dimensions) if isinstance(dim, sym.Range)]
+
+        for index, dim in enumerate(var.dimensions):
+            new_dimensions[indices[index]] = dim
+
+        original_symbol = sym.ArraySubscript(val.symbol, val.dimensions)
+        new_symbol = sym.ArraySubscript(val.symbol, tuple(new_dimensions) )
+
+        mapper = SubstituteExpressionsMapper({original_symbol:new_symbol})
+
+        return mapper(original_symbol)
 
     # Get local variable declarations and hoist them
     decls = FindNodes(VariableDeclaration).visit(member.spec)
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -233,6 +233,18 @@ def inline_member_routine(routine, member):
 
         return mapper(original_symbol)
 
+    # Prevent shadowing of member variables by renaming them a priori
+    parent_variables = routine.variable_map
+    duplicate_locals = tuple(
+        v for v in member.variables
+        if v.name in parent_variables and v.name not in member._dummies
+    )
+    shadow_mapper = SubstituteExpressions(
+        {v: v.clone(name=f'{member.name}_{v.name}') for v in duplicate_locals}
+    )
+    member.spec = shadow_mapper.visit(member.spec)
+    member.body = shadow_mapper.visit(member.body)
+
     # Get local variable declarations and hoist them
     decls = FindNodes(VariableDeclaration).visit(member.spec)
     decls = tuple(d for d in decls if all(s.name not in routine._dummies for s in d.symbols))
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -241,9 +241,13 @@ def inline_member_routine(routine, member):
                 if isinstance(arg, sym.Array):
                     # Resolve implicit dimension ranges of the passed value,
                     # eg. when passing a two-dimensional array `a` as `call(arg=a)`
-                    qualified_value = val if val.dimensions else val.clone(
-                        dimensions=tuple(sym.Range((None, None)) for _ in arg.shape)
-                    )
+                    # Check if val is a DeferredTypeSymbol, as it does not have a `dimensions` attribute
+                    if not isinstance(val, sym.DeferredTypeSymbol) and val.dimensions:
+                        qualified_value = val
+                    else:
+                        qualified_value = val.clone(
+                            dimensions=tuple(sym.Range((None, None)) for _ in arg.shape)
+                        )
                     arg_vars = tuple(v for v in member_vars if v.name == arg.name)
                     argmap.update((v, _map_unbound_dims(v, qualified_value)) for v in arg_vars)
                 else:
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -14,13 +14,18 @@ from loki.expression import (
     FindVariables, FindInlineCalls, FindLiterals,
     SubstituteExpressions, LokiIdentityMapper
 )
-from loki.ir import Import, Comment, Assignment
+from loki.ir import Import, Comment, Assignment, VariableDeclaration, CallStatement
 from loki.expression import symbols as sym
 from loki.types import BasicType
 from loki.visitors import Transformer, FindNodes
+from loki.tools import as_tuple
+from loki.logging import warning
 
 
-__all__ = ['inline_constant_parameters', 'inline_elemental_functions']
+__all__ = [
+    'inline_constant_parameters', 'inline_elemental_functions',
+    'inline_member_procedures'
+]
 
 
 class InlineSubstitutionMapper(LokiIdentityMapper):
@@ -183,3 +188,67 @@ def inline_elemental_functions(routine):
         if all(hasattr(s, 'type') and s.type.dtype in removed_functions for s in im.symbols):
             import_map[im] = None
     routine.spec = Transformer(import_map).visit(routine.spec)
+
+
+def inline_member_routine(routine, member):
+    """
+    Inline an individual member :any:`Subroutine` at source level.
+
+    This will replace all :any:`Call` objects to the specified
+    subroutine with an adjusted equivalent of the member routines'
+    body. For this, argument matching, including partial dimension
+    matching for array references is performed, and all
+    member-specific declarations are hoisted to the containing
+    :any:`Subroutine`.
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine in which to inline all calls to the member routine
+    member : :any:`Subroutine`
+        The contained member subroutine to be inlined in the parent
+    """
+
+    # Get local variable declarations and hoist them
+    decls = FindNodes(VariableDeclaration).visit(member.spec)
+    decls = tuple(d for d in decls if all(not s.type.intent for s in d.symbols))
+    decls = tuple(d for d in decls if all(s not in routine.variables for s in d.symbols))
+    routine.spec.append(decls)
+
+    call_map = {}
+    for call in FindNodes(CallStatement).visit(routine.body):
+        if call.routine == member:
+            # Substitute argument calls into a copy of the body
+            member_body = SubstituteExpressions(call.arg_map).visit(member.body.body)
+
+            # Inline substituted body within a pair of marker comments
+            comment = Comment(f'! [Loki] inlined member subroutine: {member.name}')
+            c_line = Comment('! =========================================')
+            call_map[call] = (comment, c_line) + as_tuple(member_body) + (c_line, )
+
+    # Replace calls to member with the member's body
+    routine.body = Transformer(call_map).visit(routine.body)
+    # Can't use transformer to replace subroutine, so strip it manually
+    contains_body = tuple(n for n in routine.contains.body if not n == member)
+    routine.contains._update(body=contains_body)
+
+
+def inline_member_procedures(routine):
+    """
+    Inline all member subroutines contained in an individual :any:`Subroutine`.
+
+    Please note that member functions are not yet supported!
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine in which to inline all member routines
+    """
+
+    # Run through all members and invoke individual inlining transforms
+    for member in routine.members:
+        if member.is_function:
+            # TODO: Implement for functions!!!
+            warning('[Loki::inline] Inlining member functions is not yet supported, only subroutines!')
+        else:
+            inline_member_routine(routine, member)
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -12,7 +12,7 @@ Collection of utility routines to perform code-level force-inlining.
 """
 from loki.expression import (
     FindVariables, FindInlineCalls, FindLiterals,
-    SubstituteExpressions, LokiIdentityMapper
+    SubstituteExpressions, SubstituteExpressionsMapper, LokiIdentityMapper
 )
 from loki.ir import Import, Comment, Assignment, VariableDeclaration, CallStatement
 from loki.expression import symbols as sym
@@ -209,6 +209,20 @@ def inline_member_routine(routine, member):
         The contained member subroutine to be inlined in the parent
     """
 
+    def _map_unbound_dims(var, val):
+        """
+        Maps all unbound dimension ranges in the passed array value
+        ``val`` with the indices from the local variable ``var``. It
+        returns the re-mapped symbol.
+
+        For example, mapping the passed array ``m(:,j)`` to the local
+        expression ``a(i)`` yields ``m(i,j)``.
+        """
+        val_free_dims = tuple(d for d in val.dimensions if isinstance(d, sym.Range))
+        var_bound_dims = tuple(d for d in var.dimensions if not isinstance(d, sym.Range))
+        mapper = SubstituteExpressionsMapper(dict(zip(val_free_dims, var_bound_dims)))
+        return mapper(val)
+
     # Get local variable declarations and hoist them
     decls = FindNodes(VariableDeclaration).visit(member.spec)
     decls = tuple(d for d in decls if all(not s.type.intent for s in d.symbols))
@@ -218,8 +232,25 @@ def inline_member_routine(routine, member):
     call_map = {}
     for call in FindNodes(CallStatement).visit(routine.body):
         if call.routine == member:
+            argmap = {}
+            member_vars = FindVariables().visit(member.body)
+
+            # Match dimension indexes between the argument and the given value
+            # for all occurences of the argument in the body
+            for arg, val in call.arg_map.items():
+                if isinstance(arg, sym.Array):
+                    # Resolve implicit dimension ranges of the passed value,
+                    # eg. when passing a two-dimensional array `a` as `call(arg=a)`
+                    qualified_value = val if val.dimensions else val.clone(
+                        dimensions=tuple(sym.Range((None, None)) for _ in arg.shape)
+                    )
+                    arg_vars = tuple(v for v in member_vars if v.name == arg.name)
+                    argmap.update((v, _map_unbound_dims(v, qualified_value)) for v in arg_vars)
+                else:
+                    argmap[arg] = val
+
             # Substitute argument calls into a copy of the body
-            member_body = SubstituteExpressions(call.arg_map).visit(member.body.body)
+            member_body = SubstituteExpressions(argmap).visit(member.body.body)
 
             # Inline substituted body within a pair of marker comments
             comment = Comment(f'! [Loki] inlined member subroutine: {member.name}')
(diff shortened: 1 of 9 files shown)


loki/transform/transform_inline.py
```diff
@@ -183,3 +188,92 @@ def inline_elemental_functions(routine):
...
+
+    call_map = {}
+    for call in FindNodes(CallStatement).visit(routine.body):
+        if call.routine == member:
+            argmap = {}
+            member_vars = FindVariables().visit(member.body)
+
+            # Match dimension indexes between the argument and the given value
+            # for all occurences of the argument in the body
+            for arg, val in call.argument_map.items():
```
reuterbal, line 237: Is there a benefit to using `call.argument_map.items()` here instead of `call.arg_iter`?

scripts/loki_transform.py
```diff
@@ -489,7 +496,8 @@ def ecphys(mode, config, header, source, build, cpp, directive, frontend):
         block_dim = scheduler.config.dimensions['block_dim']
         transformation = SingleColumnCoalescedTransformation(
```
reuterbal, line 497: Spotting the SCC wrapper here, I think the PR branch needs rebasing onto latest main

tests/test_transform_inline.py
```diff
@@ -281,3 +287,126 @@ def test_constant_replacement_internal(frontend):
...
+  real(kind=8), intent(inout) :: matrix(3, 3), tensor(3, 3, 4)
+  integer :: i
+  do i=1, 3
+    call add_one(3, matrix(:,i), tensor(1:3,i,:))
+  end do
+  contains
+    subroutine add_one(n, a, b)
+      integer, intent(in) :: n
+      real(kind=8), intent(inout) :: a(3), b(3,1:n)
+      integer :: j
```
reuterbal, line 379: Related to the earlier comment: What would happen if the loop variable was called `i` here as well?

tests/test_transform_inline.py
```diff
@@ -281,3 +287,126 @@ def test_constant_replacement_internal(frontend):
...
+    """
+    Test inlining of member subroutines when sub-arrays of rank less
+    than the formal argument are passed.
+    """
+    fcode = """
+subroutine member_routines_arg_dimensions(matrix, tensor)
+  real(kind=8), intent(inout) :: matrix(3, 3), tensor(3, 3, 4)
+  integer :: i
+  do i=1, 3
+    call add_one(3, matrix(:,i), tensor(1:3,i,:))
```
reuterbal, line 373: I'm quite impressed that the expansion `b(3,1:n)` -> `tensor(1:3,i,1:n)` works because I don't seem to understand it from the implementation...

reuterbal (commented):
Many thanks, this is great and neatly implemented!

I have left a few comments and questions but the important one is related to the  behaviour when the member routine shadows variables from the host routine.

JoeffreyLegaux:
Hi Michael, 
Great addition, I am currently working on routines that need this sort of inlining and I've played a bit with it and spotted a couple issues on my cases.

loki/transform/transform_inline.py
```diff
@@ -183,3 +188,92 @@ def inline_elemental_functions(routine):
...
+                if isinstance(arg, sym.Array):
+                    # Resolve implicit dimension ranges of the passed value,
+                    # eg. when passing a two-dimensional array `a` as `call(arg=a)`
+                    qualified_value = val if val.dimensions else val.clone(
```
JoeffreyLegaux, line 241: When val happens to come from a derived type argument, its class is DeferredTypeSymbol which does not have a dimensions attributes, thus crashing at this line.
This change seems to correctly fix the issue :
```
                    if not isinstance(val, sym.DeferredTypeSymbol) and val.dimensions:
                        qualified_value = val
                    else:
                        qualified_value = val.clone(
                            dimensions=tuple(sym.Range((None, None)) for _ in arg.shape)
                        )
                        ```

**Comment by skarppinen** [comment]
Here is an example where a shadowed variable is defined twice (with different types) in the inlined result.

```
import loki
from loki.transform import inline_member_procedures

fcode = """
subroutine outer()
     real :: x = 3 ! 'x' is real in outer.
     real :: tmp = 0
     call inner(tmp)
     x = x + tmp 

contains
    subroutine inner(y) 
        real, intent(inout) :: y
        real :: x(3) ! 'x' is array in inner.
        x = [1, 2, 3]
        y = sum(x)
    end subroutine inner
end subroutine outer
"""
src = loki.Sourcefile.from_source(fcode)
routine = src.routines[0]
```

Then, `print(routine.to_fortran())` gives (defining `x` twice):

```
SUBROUTINE outer ()
  REAL :: x = 3
  REAL :: tmp = 0
  REAL :: x(3)
  ! [Loki] inlined member subroutine: inner
  x = (/ 1, 2, 3 /)
  tmp = SUM(x)
  x = x + tmp

  CONTAINS
END SUBROUTINE outer
```

File: loki/transform/transform_inline.py (hunk 1 above)
Comment (mlange05) [reply loki/transform/transform_inline.py:237]: No, I only realised that we require a function call for a property that would most naturally be expressed as a dict. In the first implementation I was passing this, not iterating it, but alas, we've come full circle. :wink:

**Comment by mlange05** [comment]
Ok, many thanks for all the comments and very helpful suggestions of corner cases. I've rebased the PR branch over the recent developments, and will now try to incorporate some more of the corner case tests and fixes. If you've branched off this branch before, please carefully fetch and reset and cherry-pick your changes on top of this.

@reuterbal I've rebased the simple fixes inline, but I'll make another push soon with additional tests/fixes according to the various suggestions. Please hold off until I ping here. :wink:

**Comment by skarppinen** [comment]
Hi @mlange05, here is a reproducer on an issue with the current inliner:

```
import loki
from loki.transform import inline_member_procedures

fcode = """
subroutine outer()
    real :: arr(3, 3)
    real :: out
    arr = 10
    call inner(out, arr(1, 3))
    
    print *, out
    contains
    
    subroutine inner(out, x)
        real, intent(inout) :: out
        real, intent(in) :: x(3)

        out = x(3)
    end subroutine inner
end subroutine outer
"""
src = loki.Sourcefile.from_source(fcode)
routine = src.routines[0]
```

The above code throws an `IndexError`.

**Comment by JoeffreyLegaux** [comment]
Hello @mlange05, I just identified another troublesome corner case.

**Comment by JoeffreyLegaux** [comment]
Hello @mlange05 I just identified another corner case, here is a reduced test case :

```
SUBROUTINE CPCFU()

IMPLICIT NONE

IF (YDCFU%YFCL%LACTIVE) THEN
  CALL COPY1_CFU (YDCFU%YFCL, YDCFU%FCL, YDMF_PHYS_OUT%FCLL(:,1))
  CALL COPY1_CFU (YDCFU%YFCL, YDCFU%FCL, YDMF_PHYS_OUT%FCLN(:,1), LDNORESET=.TRUE.)
ENDIF

CONTAINS

SUBROUTINE COPY1_CFU (YDDESC, PXFU, PPHY, KTIME, LDNORESET)

TYPE(FLUXES_DESCRIPTOR) ,INTENT(IN)               :: YDDESC
REAL(KIND=JPRB)         ,INTENT(INOUT)            :: PXFU(:)
REAL(KIND=JPRB)         ,INTENT(IN)               :: PPHY(1:YDCPG_OPTS%KLON)
INTEGER(KIND=JPIM)      ,INTENT(IN)    ,OPTIONAL  :: KTIME
LOGICAL                 ,INTENT(IN)    ,OPTIONAL  :: LDNORESET

LLNORESET = .FALSE.
IF (PRESENT (LDNORESET)) LLNORESET = LDNORESET

IF (LLRESET .AND. (.NOT. LLNORESET)) THEN
  PXFU(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA)=0._JPRB
ENDIF

END SUBROUTINE 
END SUBROUTINE CPCFU
```
inline_member_routine crashes when treating the second call with the optional logical argument, when calling  recursive_expression_map_update : 
```
 File "/home/legaux/meteo/loki_github/loki/transform/transform_utilities.py", line 454, in recursive_expression_map_update
    prev_map, expr_map = expr_map, {
  File "/home/legaux/meteo/loki_github/loki/transform/transform_utilities.py", line 455, in <dictcomp>
    expr: type(replacement)(**{
  File "/home/legaux/meteo/loki_github/loki/expression/symbols.py", line 1072, in __init__
    self.value = value.lower() in ('true', '.true.')
AttributeError: 'bool' object has no attribute 'lower'
> /home/legaux/meteo/loki_github/loki/expression/symbols.py(1072)__init__()
-> self.value = value.lower() in ('true', '.true.')
(Pdb) p value
True
(Pdb) p type(value)
<class 'bool'>
```
Not really sure about what is happening here, this is too deep in the loki core for my understanding.
Apparently, the function is trying to build a LogicLiteral by passing an actual bool, but the __init__ method seems to be written to receive a string representation of a boolean.

**Comment by mlange05** [comment]
> Hi @mlange05, here is a reproducer on an issue with the current inliner:
> 
> ```
> import loki
> from loki.transform import inline_member_procedures
> 
> fcode = """
> subroutine outer()
>     real :: arr(3, 3)
>     real :: out
>     arr = 10
>     call inner(out, arr(1, 3))
>     
>     print *, out
>     contains
>     
>     subroutine inner(out, x)
>         real, intent(inout) :: out
>         real, intent(in) :: x(3)
> 
>         out = x(3)
>     end subroutine inner
> end subroutine outer
> """
> src = loki.Sourcefile.from_source(fcode)
> routine = src.routines[0]
> inline_member_procedures(routine)
> ```
> 
> The above code throws an `IndexError`.

Hi @skarppinen, I've run your example and the problem actually arises from the source code being wrong in this case. In `call inner(out, arr(1, 3))` your are mapping a scalar value (`arr(1, 3)`) onto a 3-element array. The reason we're getting an index error is that the transformation looks for ranges, so something like `arr(:,3)` or even `arr(1:3,3)` should work.

We could possibly throw a more informative error for such cases, if the dimension cannot be matched. :thinking:

**Comment by mlange05** [comment]
> Not really sure about what is happening here, this is too deep in the loki core for my understanding. Apparently, the function is trying to build a LogicLiteral by passing an actual bool, but the **init** method seems to be written to receive a string representation of a boolean.

Hi @JoeffreyLegaux, many thanks for this! Yes, you seem to have unearthed a true low-level bug in the expression constructor for logicals, where indeed we were not dealing with native Python boolean values correctly (:scream_cat:). I've added a small fix and test that should fix this. Many thanks again! :grin:

File: loki/transform/transform_inline.py (hunk 5 above)
Comment (mlange05) [reply loki/transform/transform_inline.py]: Addressed in external contribution! Many thanks @JoeffreyLegaux

Comment (mlange05) [reply loki/transform/transform_inline.py]: Addressed in external contribution! Many thanks @JoeffreyLegaux

File: loki/transform/transform_inline.py [hunk 6]
```diff
@@ -183,3 +188,92 @@ def inline_elemental_functions(routine):
...
+        """
+        val_free_dims = tuple(d for d in val.dimensions if isinstance(d, sym.Range))
+        var_bound_dims = tuple(d for d in var.dimensions if not isinstance(d, sym.Range))
+        mapper = SubstituteExpressionsMapper(dict(zip(val_free_dims, var_bound_dims)))
+        return mapper(val)
+
+    # Get local variable declarations and hoist them
+    decls = FindNodes(VariableDeclaration).visit(member.spec)
+    decls = tuple(d for d in decls if all(not s.type.intent for s in d.symbols))
+    decls = tuple(d for d in decls if all(s not in routine.variables for s in d.symbols))
```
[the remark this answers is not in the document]
mlange05: Yes, shadowing is now explicitly detected before the substitution is happening, where the child-routine's name is prepended to the variable declaration before it gets hoisted.

tests/test_transform_inline.py
mlange05: Shadowing is now explicitly dealt with and tested.

mlange05:
@reuterbal Ok, I think I have addressed all concerns raised and this is ready for another look. 

Many thanks also for the through testing and external input @JoeffreyLegaux @skarppinen . Please feel free to give this one more round, if you have the time.

skarppinen:
> > Hi @mlange05, here is a reproducer on an issue with the current inliner:
> > ```
> > import loki
> > from loki.transform import inline_member_procedures
> > 
> > fcode = """
> > subroutine outer()
> >     real :: arr(3, 3)
> >     real :: out
> >     arr = 10
> >     call inner(out, arr(1, 3))
> >     
> >     print *, out
> >     contains
> >     
> >     subroutine inner(out, x)
> >         real, intent(inout) :: out
> >         real, intent(in) :: x(3)
> > 
> >         out = x(3)
> >     end subroutine inner
> > end subroutine outer
> > """
> > src = loki.Sourcefile.from_source(fcode)
> > routine = src.routines[0]
> > inline_member_procedures(routine)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The above code throws an `IndexError`.
> 
> Hi @skarppinen, I've run your example and the problem actually arises from the source code being wrong in this case. In `call inner(out, arr(1, 3))` your are mapping a scalar value (`arr(1, 3)`) onto a 3-element array. The reason we're getting an index error is that the transformation looks for ranges, so something like `arr(:,3)` or even `arr(1:3,3)` should work.
> 
> We could possibly throw a more informative error for such cases, if the dimension cannot be matched. 🤔

Sorry for late reply, for some reason I did not get a notification this time.

I don't think that the source code is wrong in this case, as the example code compiles (I tested using gfortran). 
I purposefully pass a scalar to the array argument. This actually happens in ACRANEB2, which is where this test case basically comes from. Of course, this is arguably bad programming practice, but that code appears to be valid Fortran still. 

I think the inliner should be able to handle this case (perhaps by postprocessing the calling arguments similarly as you suggest).

JoeffreyLegaux:
> 
> Sorry for late reply @mlange05, for some reason I did not get a notification this time.
> 
> I don't think that the source code is wrong in this case, as the example code compiles (I tested using gfortran). I purposefully pass a scalar to the array argument. This actually happens in ACRANEB2, which is where this test case basically comes from. Of course, this is arguably bad programming practice, but that code appears to be valid Fortran still.
> 
> I think the inliner should be able to handle this case (perhaps by preprocessing the calling arguments similarly as you suggest).

I confirm that this is a relatively common pattern in Fortran, passing an array's element to a rotuine that expects an actual array turns into the passing of (sort of) a pointer to that element (this requires the caller to have an explicit vision on the callee's signature).
I am not sure if this is actually standard behaviour, but we have plenty of this kind of occurences in our codes at CERFACS and all compilers have always been supporting this behaviour.
This has some convenient use cases such as offsetting an array in a call, or passing arrays with various numbers of dimension to a common routine that treats them as a flat 1D array.
But I would also not recommend this practice, especially in OpenACC parallel regions.

mlange05:
> I confirm that this is a relatively common pattern in Fortran, 

Hi @JoeffreyLegaux @skarppinen. Fair enough, I agree that scalar-to-array passing (formal term "sequence association", I believe) is a common pattern in legacy Fortran codes, but it is also one of the true Fortran anti-patterns (from a compiler perspective) that prevents a lot of optimisations since it makes reasoning about array dimensions impossible. In fact this is exactly the case here, as we are doing partial dimension matching, and scalar-to-array mapping breaks some basic assumptions of argument resolution. As a corollary I would argue that it is not the role of a source-level inlining utility  to retro-actively fit necessary dimension information that the programmer purposefully omitted, although such a utility might be useful if this pattern is commonly (ab)used in your code. So to move forward, I'll probably put that restriction in the documentation, but defer implementing a scalar-to-array remapper to another feature PR? 

@reuterbal to confirm if merging this feature without extended scalar-to-array support is ok?

The for ACRANEB I would recommend to either:
* Fix the source code (my personal preference), since the necessity to use if this feature almost always indicates a  deeper design problem (from my experience the most common use case for this feature is to prevent runtime bounds checking, so ...).
* Add a separate utility to Loki (either core or your personal scripts) that adds the necessary dimension information; eg. maps `a(i)` to either `a(:)` or `a(i:i+n)` if the receiving argument declaration is `b(n)`. Creating such conceptually "correct" mapping might be hard to establish, but would also allow further downstream use for inter-procedural optimisations in general beyond inlining. 
* Write a custom inliner that covers all corner cases. :shrug: :wink: 

One final hint I can give is that we have a slightly similar case with unqualified dimension passing in some SCC use cases, where some call chains start using `a(:,:)` instead of `a(n, m)` notation for declaring argument shapes. Since some of the SCC recipes require explicit dimension names to apply loop-flipping we wrote a utility to derived the symbolic dimension names from the caller via inter-procedural passes, to then allow SCC to use this information (see `transformations/argument_shape.py`). A similar "sanitisation pass" could possibly be useful for your code base if you keep hitting edge-case uses for Fortran standard that prevent other recipes to make certain necessary assumptions.

reuterbal:
Thanks for the updates and clarifying the sequence association questions. I agree that we shouldn't work around bad practices in Fortran code which should be fixed in the original source code. If possible, it would be useful to have a meaningful error message whenever such scalar-to-array argument association would cause the inliner to fail.

I will undertake a proper review once I have finished a few other high-priority items.

JoeffreyLegaux:
> Thanks for the updates and clarifying the sequence association questions. I agree that we shouldn't work around bad practices in Fortran code which should be fixed in the original source code. If possible, it would be useful to have a meaningful error message whenever such scalar-to-array argument association would cause the inliner to fail.
> 
> I will undertake a proper review once I have finished a few other high-priority items.

I agree that this is arguably a bad practice and would not expect it to be supported (I already forsee many sketchy corner cases to treat). My point was that this usage does exist, being able to spot it and output a relatable error message would be beneficial for loki's usage in the larger scope.

skarppinen:
> > I confirm that this is a relatively common pattern in Fortran,
> 
> Hi @JoeffreyLegaux @skarppinen. Fair enough, I agree that scalar-to-array passing (formal term "sequence association", I believe) is a common pattern in legacy Fortran codes, but it is also one of the true Fortran anti-patterns (from a compiler perspective) that prevents a lot of optimisations since it makes reasoning about array dimensions impossible. In fact this is exactly the case here, as we are doing partial dimension matching, and scalar-to-array mapping breaks some basic assumptions of argument resolution. As a corollary I would argue that it is not the role of a source-level inlining utility to retro-actively fit necessary dimension information that the programmer purposefully omitted, although such a utility might be useful if this pattern is commonly (ab)used in your code. So to move forward, I'll probably put that restriction in the documentation, but defer implementing a scalar-to-array remapper to another feature PR?
> 
> @reuterbal to confirm if merging this feature without extended scalar-to-array support is ok?
> 
> The for ACRANEB I would recommend to either:
> 
> * Fix the source code (my personal preference), since the necessity to use if this feature almost always indicates a  deeper design problem (from my experience the most common use case for this feature is to prevent runtime bounds checking, so ...).
> * Add a separate utility to Loki (either core or your personal scripts) that adds the necessary dimension information; eg. maps `a(i)` to either `a(:)` or `a(i:i+n)` if the receiving argument declaration is `b(n)`. Creating such conceptually "correct" mapping might be hard to establish, but would also allow further downstream use for inter-procedural optimisations in general beyond inlining.
> * Write a custom inliner that covers all corner cases. 🤷 😉
> 
> One final hint I can give is that we have a slightly similar case with unqualified dimension passing in some SCC use cases, where some call chains start using `a(:,:)` instead of `a(n, m)` notation for declaring argument shapes. Since some of the SCC recipes require explicit dimension names to apply loop-flipping we wrote a utility to derived the symbolic dimension names from the caller via inter-procedural passes, to then allow SCC to use this information (see `transformations/argument_shape.py`). A similar "sanitisation pass" could possibly be useful for your code base if you keep hitting edge-case uses for Fortran standard that prevent other recipes to make certain necessary assumptions.

Hello @mlange05. I think it is completely fine to not implement this corner case to the inliner, as long as a meaningful error message is printed for the user in this case. Perhaps the documentation of the inliner could also point out the existence of the scalar-to-array remapper (if it exists?), so it is clear what can be done if someone faces this error.

For ACRANEB2 I am currently using an inliner developed by Daan Degrauwe and Denis Haumont with a preprocessing step of Denis Haumont that avoids this issue.

loki/transform/transform_inline.py
```diff
@@ -183,3 +188,123 @@ def inline_elemental_functions(routine):
...
+        for index, dim in enumerate(var.dimensions):
+            new_dimensions[indices[index]] = dim
+
+        return val.clone(dimensions=tuple(new_dimensions))
+
+    # Prevent shadowing of member variables by renaming them a priori
+    parent_variables = routine.variable_map
+    duplicate_locals = tuple(
+        v for v in member.variables
+        if v.name in parent_variables and v.name not in member._dummies
```
reuterbal, line 236: Do we risk problems with case sensitive comparisons for `v.name not in member._dummies`? I think the latter uses all lower-case variables, so might be worth amending the shadowing tests with an upper-case declaration of a shadowing variable?

loki/transform/transform_inline.py
```diff
@@ -183,3 +188,123 @@ def inline_elemental_functions(routine):
...
+            member_vars = FindVariables().visit(member.body)
+
+            # Match dimension indexes between the argument and the given value
+            # for all occurences of the argument in the body
+            for arg, val in call.arg_map.items():
+                if isinstance(arg, sym.Array):
+                    # Resolve implicit dimension ranges of the passed value,
+                    # eg. when passing a two-dimensional array `a` as `call(arg=a)`
+                    # Check if val is a DeferredTypeSymbol, as it does not have a `dimensions` attribute
+                    if not isinstance(val, sym.DeferredTypeSymbol) and val.dimensions:
```
reuterbal, line 263: I suspect this would be the place to bail out with an error message if a scalar is passed as actual argument to an array dummy argument

reuterbal (changes requested):
Very neat and much appreciate the detailed testing by @JoeffreyLegaux and @skarppinen. I think this has much improved the code. As mentioned in a comment earlier, I think it would be useful and not terribly hard to detect situations where scalars are passed as array arguments, and to throw an appropriate error. 

I have also flagged a suspicious line which looks like it might cause issues with case sensitivity, and which the test likely doesn't flag.

Also, we don't have any testing for the embedding of member inlining in the actual SCC transformation - I don't know if we could retro-fit this easily to one of the SCC tests, or may be able to get at least appropriate regression testing in place at some point?

loki/transform/transform_inline.py
mlange05: Wow, good catch, this does indeed break a special corner case that we didn't test yet. I've pushed an extension to the shadowing test and a fix for this.

loki/transform/transform_inline.py
mlange05: Yes, fully agree. I've added a detect-and-error clause and a test to gracefully fail when we encounter this.

mlange05:
@reuterbal Ok, I've fixed the last two outstanding issues (good spot!), and hit the rebase button to make everything merge-compatible. I'm not sure additional SCC-level testing is needed right now, but it will be added to the ec-phys regression due time, and I might add more in a later PR.

reuterbal (approved):
Many thanks, looks all good now. Going in!

Issue #104, opened 2023-06-19
Start of computation misinterpreted when mixing multiple: variable and statement function declarations.


piotrows:
In some cases, two or more header files with statement function declarations need to be included in the kernel. With preprocessing enabled, the include files are injected into kernel code. At the moment Loki may misinterpret such code containing pattern:
1) variable declaration
2) statement function declaration
3) variable declaration
4) statement function declaration
so the vector_section is inserted before 4), resulting in the Fortran code with statement function declarations inside the loop.
```# (C) Copyright 2018- ECMWF.
# This software is licensed under the terms of the Apache Licence Version 2.0
# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
# In applying this licence, ECMWF does not waive the privileges and immunities
# granted to it by virtue of its status as an intergovernmental organisation
# nor does it submit to any jurisdiction.

import pytest

from loki import (
    Subroutine, Dimension, fgen, Sourcefile, SubroutineItem
)
from transformations import (
    SCCDevectorTransformation, SCCHoistTransformation
)
from conftest import available_frontends
#pylint: disable=too-many-lines

@pytest.fixture(scope='module', name='horizontal')
def fixture_horizontal():
    return Dimension(name='horizontal', size='nlon', index='jl', bounds=('start', 'end'))


@pytest.fixture(scope='module', name='vertical')
def fixture_vertical():
    return Dimension(name='vertical', size='nz', index='jk')


@pytest.fixture(scope='module', name='blocking')
def fixture_blocking():
    return Dimension(name='blocking', size='nb', index='b')



@pytest.mark.parametrize('frontend', available_frontends())
def test_scc_hoist_multiple_kernels(frontend, horizontal, vertical, blocking):
    """
    Test hoisting of column temporaries to "driver" level.
    """

    fcode_driver = """
  SUBROUTINE column_driver(nlon, nz, q, nb)
    INTEGER, INTENT(IN)   :: nlon, nz, nb  ! Size of the horizontal and vertical
    REAL, INTENT(INOUT)   :: q(nlon,nz,nb)
    REAL :: T(NLON)
    INTEGER :: b, start, end

    start = 1
    end = nlon
    do b=1, nb
      call compute_column2(start, end, nlon, nz, q(:,:,b))
    end do
  END SUBROUTINE column_driver
"""

    fcode_kernel2 = """
  SUBROUTINE compute_column2(start, end, nlon, nz, q)
[...]

PR #132, opened 2023-08-28
Statement function parsing on-the-fly

mlange05:
Previously, statement functions (on of the true beauties of the Fortran standard!) have been retro-fitted to the IR after an initial parse, since the FParser frontend cannot distinguish them from regular array accesses without type information. As pointer out in issue #104, the Loki FP frontend can also fall into the trap of getting this wrong by preprending the last stmt function to the subroutine body instead of appending it to the spec.

This PR now fixes the statement function detection by using our internal symbol table to detect them on-the-fly and thus construct them as part of the original parse. This avoids some very(!) costly IR transformations (rebuilds) that we otherwise perform routinely. Since the StmtFunc objects are already in the partial IRs for spec and body, re-shuffling the mismatch between body and spec is not much easier and can be done before creating the final `Subroutine` object.

This PR now does:
* Detect StmtFuncs in FP frontend by looking for explicitly defined scalars that are used with notation on the LHS of an assignment
* Before finalising the `Subroutine`, look for stray StmtFuncs at the beginning of the body
* Update the original Scalar declaration symbol after the subroutine is finalized; this is needed as the `NodeLazyLookup` requires all declarations to be attached to the spec
* Adjust the existing StmtFunction test to capture the corner-case in issue #104.
* Remove the now obsolete retro-fitting utility

The PR seems to drop the full-parsing time of CLOUDSC (dwarf version) by about 25%, which is a nice side-effect, and has been verified with EC-physics regression.

One final note (@reuterbal), it might make sense to create a dedicated `Procedure` => `Subroutine` | `Function` | StmtFunction` set of classes, with Procedure an ACB, to fully capture the specific procedure types. This would then allow the explicit "nesting" of StmtFuncs in the IR without the use of the `LazyNodeLookup`, as it would create a StmtFunc as a scope in its own right. But that's just meant as food for thought. :wink:

### DIFF
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -1753,6 +1753,15 @@ class FParser2IR(GenericVisitor):
             body = self.visit(body_ast, **kwargs)
             body = sanitize_ir(body, FP, pp_registry=sanitize_registry[FP], pp_info=self.pp_info)
 
+        # Workaround for lost StatementFunctions:
+        # Since FParser has no means to identify StmtFuncs, the last set of them
+        # can get lumped in with the body, and we simply need to shift them over.
+        stmt_funcs = tuple(n for n in body.body if isinstance(n, ir.StatementFunction))
+        if stmt_funcs:
+            idx = body.body.index(stmt_funcs[-1]) + 1
+            spec._update(body=spec.body + body.body[:idx])
+            body._update(body=body.body[idx:])
+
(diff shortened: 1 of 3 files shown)


reuterbal (approved):
Nicely done and a much better solution than the post-hoc fiddling we did previously. That this has to be done only for fparser is due to the fact that OMNI inlines them anyway and OFP doesn't support them in the first place?

Issue #29, opened 2023-01-30
Generic key-value options parameter in loki-transform.py


MichaelSt98:
As suggested by @reuterbal 

Generic key-value options parameter, essentially a dict of named options that can be handed as kwargs to the transformation. This can be achieved by using [tuples as multi-value options](https://click.palletsprojects.com/en/8.1.x/options/#tuples-as-multi-value-options).

```py
@click.option('--option', '-o', type=(str, str),
              type=click.Tuple([str, str]), multiple=True,
              help='Specify transformation options as key-value pairs')
```

reuterbal:
I think the config file capabilities to parameterise transformations that we have now could be considered as fulfilling this requirement? (Cc @mlange05 )

mlange05:
I'd think so. Happy to close this, if you agree.

Issue #28, opened 2023-01-30
Generic approach for transformation specific settings in SchedulerConfig


MichaelSt98:
**Generic approach for transformation specific settings to avoid adding new keys flat into the scheduler config**

by introducing a `transformation_configs` key or similar as proposed by @reuterbal.

Consequently, transformation-specific sections (analogue to the routine-specific configs) could be declared, e.g.

```
[transformation_options]

[transformation_options.ParametriseTransformation]
dic2p = { NLEV = 137 }

[transformation_options.SccCufTransformation]
derived_types = [ 'TECLDP' ]
```

This could then be picked out in the scheduler's process routine and handed down to the transformation invocation:

```py
trafo_config = self.config.get(trafo_name, {})
...
transformation.apply(..., **trafo_config)
```

mlange05:
Draft PR #191 should address this, so comments welcome.

PR #135, opened 2023-08-30
SCC-STACK for ecphys

MichaelSt98:
> Note: this PR stack on top of #125 and should only be reviewed once this is merged.

Implements/Complements SCC Stack transformation for the ecphys. 

* allows for stack size definition in the config file (per routine), which is currently the solution for multi-library targets
* allows for pragma `!$loki stack-insert` to define stack size assignment location in the code, necessary to grant variables used in the stack size expression to be initialised


loki/bulk/scheduler.py
```diff
@@ -55,6 +55,13 @@ def __init__(self, default, routines, disable=None, dimensions=None, dic2p=None,
             self.routines = CaseInsensitiveDict(routines)
         else:
             self.routines = CaseInsensitiveDict((r.name, r) for r in as_tuple(routines))
+        for routine in self.routines:
```
reuterbal, line 58: Is this a subtle way of providing transformation config via the config file by storing it as values to a routine with name `trafo`? I would argue that we should implement #29 and thus have a proper way of providing such configs rather than using a hacky kind of implementation like this.

transformations/tests/test_pool_allocator.py
```diff
@@ -757,22 +798,65 @@ def test_pool_allocator_temporaries_kernel_nested(frontend, block_dim, directive
...
-            'role': 'driver',
-            'real_kind': 'jwrb',
-        }]
-    }
-
-    scheduler = Scheduler(paths=[basedir], config=SchedulerConfig.from_dict(config), frontend=frontend,
-                          definitions=definitions)
+    stack_size = "400*nz*nlon"
+
+    if stack_size_config == "dict":
```
reuterbal, line 803: I think for readability it would be better to separate this into different tests.

transformations/transformations/pool_allocator.py
```diff
@@ -121,6 +122,7 @@ def __init__(self, block_dim,
         self.local_ptr_var_name_pattern = local_ptr_var_name_pattern
         self.directive = directive
         self.check_bounds = check_bounds
+        self.trafo_dict = {}
```
reuterbal, line 125: Does this really need to be a property on the Transformation class object? It gets overwritten for every item/subroutine that is being processed, so shouldn't really be stored as a state on the object.

transformations/transformations/pool_allocator.py
```diff
@@ -292,15 +298,22 @@ def _get_stack_storage_and_size_var(self, routine, stack_size):
...
             body_prepend += [stack_size_assign]
 
             # Stack-size no longer guaranteed to be a multiple of 8-bytes, so we have to check here
             padding = Assignment(lhs=stack_size_var, rhs=Sum((stack_size_var, Literal(1))))
             stack_size_check = Conditional(
                                  condition=LogicalNot(Comparison(InlineCall(Variable(name='MOD'),
-                                 parameters=(stack_size, stack_type_bytes)),
-                                 '==', Literal(0))), inline=True, body=(padding,),
+                                                                            parameters=(stack_size, stack_type_bytes)),
+                                                                 '==', Literal(0))), inline=True, body=(padding,),
```
reuterbal, line 316: Something odd is going on with formatting here

transformations/transformations/pool_allocator.py
```diff
@@ -292,15 +298,22 @@ def _get_stack_storage_and_size_var(self, routine, stack_size):
             stack_type_bytes = Cast(name='REAL', expression=Literal(1), kind=_kind)
             stack_type_bytes = InlineCall(Variable(name='C_SIZEOF'),
                                           parameters=as_tuple(stack_type_bytes))
+            stack_type_bytes = InlineCall(function=Variable(name='MAX'),
+                                          parameters=(stack_type_bytes, Literal(8)), kw_parameters=())
+            if "stack-size" in self.trafo_dict:
+                scope = Scope()
+                #TODO: parse_fparser_expression only for fparser ...
```
reuterbal, line 305: I'd argue it's acceptable to have a hard dependency on this here

transformations/transformations/pool_allocator.py
```diff
@@ -345,12 +358,24 @@ def _get_stack_storage_and_size_var(self, routine, stack_size):
         if variables_append:
             routine.variables += as_tuple(variables_append)
         if body_prepend:
-            routine.body.prepend(body_prepend)
+            if not self._insert_stack_at_loki_pragma(routine, body_prepend):
```
reuterbal, line 361: Why is there a requirement for this pragma? Can the injection point lookup be improved instead?

transformations/transformations/pool_allocator.py
```diff
@@ -13,7 +13,8 @@
     SymbolAttributes, BasicType, DerivedType, Quotient, IntLiteral, LogicLiteral,
     Variable, Array, Sum, Literal, Product, InlineCall, Comparison, RangeIndex, Cast,
     Intrinsic, Assignment, Conditional, CallStatement, Import, Allocation, Deallocation, is_dimension_constant,
-    Loop, Pragma, SubroutineItem, FindInlineCalls, Interface, ProcedureSymbol, LogicalNot, dataflow_analysis_attached
+    Loop, Pragma, SubroutineItem, FindInlineCalls, Interface, ProcedureSymbol, LogicalNot, dataflow_analysis_attached,
+    expression, parse_fparser_expression, Scope
```
reuterbal, line 17: I'd prefer to either import the relevant expressions directly (e.g. `RangeIndex`) or to use the elsewhere commonly used pattern `import loki.expression.symbols as sym`

transformations/transformations/pool_allocator.py
```diff
@@ -659,25 +695,49 @@ def create_pool_allocator(self, routine, stack_size):
...
-    def inject_pool_allocator_into_calls(self, routine, targets):
+    def inject_pool_allocator_into_calls(self, routine, targets, ignore):
         """
         Add the pool allocator argument into subroutine calls
         """
         call_map = {}
         stack_var = self._get_local_stack_var(routine)
         for call in FindNodes(CallStatement).visit(routine.body):
-            if call.name in targets:
+            if call.name in targets or call.routine.name.lower() in ignore: #['surfpp', 'surfexcdriver', 'surfseb']:
```
reuterbal, line 705: debug leftover?

transformations/transformations/pool_allocator.py
```diff
@@ -129,7 +131,11 @@ def transform_subroutine(self, routine, **kwargs):
 
         role = kwargs['role']
         item = kwargs.get('item', None)
+        ignore = item.ignore if item else ()
```
reuterbal, line 134: Shouldn't ignored routines be excluded from targets?

transformations/transformations/pool_allocator.py
```diff
@@ -659,25 +695,49 @@ def create_pool_allocator(self, routine, stack_size):
...
+                            else:
+                                arg_idx = arg_pos[0] - len(arguments)
+                                kwarguments = kwarguments[:arg_idx] + ((self.stack_argument_name, stack_var),) + \
+                                              kwarguments[arg_idx:]
+                        else:
+                            if kwarguments:
+                                kwarguments += ((self.stack_argument_name, stack_var),)
+                            else:
+                                arguments += (stack_var,)
+                        call_map[call] = call.clone(arguments=as_tuple(arguments), kwarguments=as_tuple(kwarguments))
```
reuterbal, line 740: What's going on here? I suspect this has to do with picking the right injection point when there are keyword arguments present?
That index-fuffing is a bit hard to follow, could this be made more elegant or maybe even outlined into a utility routine (because I think we discussed this interface update problem in a wider context before)?

reuterbal (changes requested):
Many thanks, this is really a huge step forward and it's fantastic that this works.

However, I think before we bring this in the actual implementation could be improved in a number of ways. I've flagged them in individual comments but will repeat here the most important points in my eyes:

1. This has a hard requirement on providing per-transformation config values in the toml configuration file. I think now is the right moment to take care of #29 instead of using the implemented workaround.
2. Some of the implementation details are very opaque. It is entirely unclear to me why there is a need for the loki pragmas to specify certain injection locations. The entire transformation has become a little hard to read due to various additions, which may indicate that some refactoring is overdue.
3. The injection location for the stack argument seems to require some intricate index fuffing. I can't remember the exact details of the offline discussion but I think the problem is generic enough to justify making this available as a utility transformation.
4. The testing of the pool allocator transformation has become quite messy. I think it would be good to not convolute too many special cases and variations into the same test and instead replicate the tests and check the various cases separately. We can use fixtures to share commonalitites between the tests to avoid redundant code.

MichaelSt98:
> Many thanks, this is really a huge step forward and it's fantastic that this works.
> 
> However, I think before we bring this in the actual implementation could be improved in a number of ways. I've flagged them in individual comments but will repeat here the most important points in my eyes:
> 

Thanks @reuterbal for having a look and thanks for the comments.

> 1. This has a hard requirement on providing per-transformation config values in the toml configuration file. I think now is the right moment to take care of [Generic approach for transformation specific settings in SchedulerConfig #28](https://github.com/ecmwf-ifs/loki/issues/28) instead of using the implemented workaround.

I agree. However, for this transformation we need transformation **and** routine specific config entries. Therefore it's (at least in my opinion) up for discussion what the best way is to do this or rather whether it is like described in [Generic approach for transformation specific settings in SchedulerConfig #28](https://github.com/ecmwf-ifs/loki/issues/28). 

> 2. Some of the implementation details are very opaque. It is entirely unclear to me why there is a need for the loki pragmas to specify certain injection locations. The entire transformation has become a little hard to read due to various additions, which may indicate that some refactoring is overdue.

The Loki pragmas to specify certain injection locations is due to the fact that variables that compose the stack size calculation may not be initialised when the stack size is calculated. For example, the original implementation could (and indeed produces something like this for ecphys), causing either compile time or runtime errors ...

```fortran
subroutine driver(...)
  type(dimension) :: dims 
  ISTZ = ... klon ...
  ALLOCATE(ZSTACK(ISTSZ)) 
  call kernel(dims%klon, ...)
end subroutine driver

subroutine kernel(klon, ...)
  integer, intent(in) ::  klon
  real temp(klon)
end subroutine kernel()
```

The simplest solution is to just define `klon` as variable in the driver and assign the value `dims%klon` to it in the body of the driver routine. However, this wouldn't solve the problem after all, as the calculation of the stack size would be inserted above `klon = dims%klon`, which can be solved by having the Loki pragma `!$loki stack-insert`. 

Of course, this can be further automated, however multiple cases can occur, amongst others: 

* variable is a module variable, thus the import would need to be recreated at the driver level
* variable is a derived type entry or rather more general, variable is renamed in the kernel level's and this renaming has to be resolved somehow 
* variable is a parameter definition at the kernel level and would need to be recreated at the driver level 
* ... 

> 3. The injection location for the stack argument seems to require some intricate index fuffing. I can't remember the exact details of the offline discussion but I think the problem is generic enough to justify making this available as a utility transformation.

True. We would/should have some `add_args()`, `remove_args()`, `check_args()` for call statements. However, in our previous offline discussion we assumed to have the relevant information already in the actual routine information, which is not necessarily the case for multi-library Loki transformation. This makes the actual implementation more difficult or at least something that needs to be discussed. 

> 4. The testing of the pool allocator transformation has become quite messy. I think it would be good to not convolute too many special cases and variations into the same test and instead replicate the tests and check the various cases separately. We can use fixtures to share commonalitites between the tests to avoid redundant code.

Yes 👍

transformations/transformations/pool_allocator.py
```diff
@@ -292,15 +298,22 @@ def _get_stack_storage_and_size_var(self, routine, stack_size):
             stack_type_bytes = Cast(name='REAL', expression=Literal(1), kind=_kind)
             stack_type_bytes = InlineCall(Variable(name='C_SIZEOF'),
                                           parameters=as_tuple(stack_type_bytes))
+            stack_type_bytes = InlineCall(function=Variable(name='MAX'),
```
awnawab, line 301: I believe this was added to ensure that each entry in the stack is aligned to a 32-byte boundary. I think this only works if the array has at least has 4 entries. This is a much safer assumption than relying on the size being a multiple of NPROMA, but nevertheless something to be aware of.

Also, this should be left on by default but configurable so that for kernels where we know that we don't need to add a floor to the stack entries we can disable it and have a smaller stack size.

PR #136, opened 2023-09-01
Disable default recursion for Transformation.apply

### LINKED ISSUE #131: Calling a Transformation on a subroutine automatically applies itself recursively on contained subroutines.
Whenever a Transformation is applied to a routine object, the transformation is also implicitely applied to all its contained subroutines.
This behaviour is not controllable, and happens whether the contained subroutines are effectively called or not.

Applying this script :
[contained.py.txt](https://github.com/ecmwf-ifs/loki/files/12440361/contained.py.txt)

```
from loki import Frontend, Sourcefile, Transformation

class TestTransfo(Transformation):
    def transform_subroutine(self, routine, **kwargs):
    	print(f'Routine treated by transformation : {routine}')

source = Sourcefile.from_file('./routine.f90', frontend=Frontend.FP)

routines = source.subroutines

for routine in routines:
	print(f'Treating routine {routine}')
	routine.apply(TestTransfo())
```

To this routine :
[routine.f90.txt](https://github.com/ecmwf-ifs/loki/files/12440359/routine.f90.txt)

```
SUBROUTINE MY_ROUTINE()

WRITE(*,*) "Main subroutine"

! CALL MY_CONTAINED_ROUTINE()

CONTAINS 

SUBROUTINE MY_CONTAINED_ROUTINE()

  WRITE(*,*) "Contained subroutine"

END SUBROUTINE MY_CONTAINED_ROUTINE

END SUBROUTINE MY_ROUTINE
```

Produces in the console output :
```
Treating routine Subroutine:: MY_ROUTINE
Routine treated by transformation : Subroutine:: MY_ROUTINE
Routine treated by transformation : Subroutine:: MY_CONTAINED_ROUTINE
```

reuterbal:
This is a proposal to fix #131. 

By default, the `Transformation` class recurses to all contained nodes, i.e., calling `apply` on a `Sourcefile` recurses into all `Module` and `Subroutine` nodes in that file, and so on. Arguably, this was a questionable design choice, however, the `Scheduler.process` method relies on this behaviour to apply transformations. There, we invoke the transformation always on the source file that contains the IR node corresponding to an `Item`. (And then do some silly filtering afterwards, which possibly even prevents us from applying transformations to contained member subroutines...? 🤔)

This PR switches off the recursion by default but allows to retain current behaviour via the `recurse_to_contained_nodes` option. 

I'll file a separate PR on top of the current one that changes also the Scheduler behaviour by directly applying transformations only to the IR node corresponding to a scheduler graph item. This requires some more careful regression testing for ecphys, and therefore this might only be sensible to do with the new Scheduler (hence the separate PR).

Important to note is that the change does have API implications when not using the Scheduler, because it requires to be more careful to which IR node a Transformation is applied - by either explicitly applying it to the target file, module, or subroutine directly, or explicitly enabling the recursion.


reuterbal:
Merged #137 and rebased over main

scripts/loki_transform.py
```diff
@@ -29,9 +28,6 @@
 )
 
 # pylint: disable=wrong-import-order
-from transformations.argument_shape import (
-    ArgumentArrayShapeAnalysis, ExplicitArgumentArrayShapeTransformation
```
mlange05, line 33: These should not have been removed. This is an error in #142 and is fixed in #149. Good catch, and thanks for "flagging" this though. :wink:

loki/bulk/scheduler.py
```diff
@@ -638,9 +650,16 @@ def process(self, transformation, reverse=False, item_filter=SubroutineItem, use
                     # Use entry from item_map to ensure the original item is used in transformation
                     _item = self.item_map[item.name]
 
+                    # Pick out the IR node to which to apply the transformation
+                    # TODO: should this become an Item property?
```
mlange05, line 654: [no action] Yes, it should, but it requires the SGraph development to make this behaviour neat. So good for now.

loki/transform/dependency_transform.py
```diff
@@ -126,27 +129,50 @@ def transform_module(self, module, **kwargs):
...
             self.module_wrap(sourcefile, **kwargs)
 
+        for module in sourcefile.modules:
+            # Recursion into contained modules using the sourcefile's "role"
+            self.transform_module(module, role=role, targets=targets, **kwargs)
+
+        if items:
+            # Recursion into all subroutine items in the current file
+            for item in items:
+                self.transform_subroutine(item.routine, item=item, role=item.role, targets=item.targets, **kwargs)
```
mlange05, line 161: [no action] This interwoven subtlety highlights that the `DependencyTransform` is actually deeply flawed and needs to be re-designed. As already discussed offline, like as a combination of several individual transformations, so thanks for keeping this functional! :pray:

tests/test_transform_dependency.py
```diff
@@ -0,0 +1,629 @@
...
+
+        kernel = scheduler['kernel_mod#kernel'].source
+        driver = scheduler['driver_mod#driver'].source
+
+    else:
+        kernel = Sourcefile.from_source(kernel_fcode, frontend=frontend)
+        driver = Sourcefile.from_source(driver_fcode, frontend=frontend)
+
+        # Because the renaming is intended to be applied to the routines as well as the enclosing module,
+        # we need to invoke the transformation on the full source file and activate recursion to contained nodes
```
mlange05, line 164: With the removal of the `recurse_to_contained_nodes` option, this comment seems outdated. It's also repeated a few times throughout the test file.

mlange05 (approved):
Ok, first of all many thanks for sticking with it through this change. This one was quite tricky, because the underlying assumptions of the `DependencyTransformation` are basically broken and ill designed (for context, this transformation also predates most of the scheduler "smarts"). 

Nevertheless, full removing recursion and letting transformations implement the complex file=>module=>routine=>member recursion manually is a good move, for now, until we have more flexibility to configure transformations generically (the "manifest" idea). 

But, there is a lot(!) of great housekeeping, testing and general quality of life stuff in here, so GT to merge this, and I'll attempt any simplification and clean-up on top of this.

scripts/loki_transform.py
reuterbal: Apologies for blindly trusting the linter instead of using my brain...

tests/test_transform_dependency.py
reuterbal: Yes! I removed the branching in the test related to this but overlooked the comments. Will fix

loki/transform/dependency_transform.py
```diff
@@ -79,11 +79,35 @@ def transform_subroutine(self, routine, **kwargs):
...
         role = kwargs.get('role')
 
-        if role == 'kernel':
+        # If applied recursively over all routines in a sourcefile, we may
+        # visit contained member subroutines. Since these are not outward facing
+        # it is not necessary to update their name.
+        is_member = isinstance(routine.parent, Subroutine)
+
+        # Pick out correct item for current subroutine if processed via recursion
+        if 'items' in kwargs:
```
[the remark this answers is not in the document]
reuterbal: That's gone after yesterday's push, the file entry point is the place now where we do the item-picking and dispatch to program unit-specific entry points.

Issue #128, opened 2023-08-22
Crash when derived_types.py passes arguments with None as intent to arguement setter in idem transformation on acraneb2.


rolfhm:
Loki crashes when trying to set arguements in `derived_types.py` that appear to haev None as intent.
In the argument setter on line 386
https://github.com/ecmwf-ifs/loki/blob/0db337cdeed57d2a06d7694c05993ea4d719be46/loki/subroutine.py#L386
called by derived_types.py on line 322
https://github.com/ecmwf-ifs/loki/blob/0db337cdeed57d2a06d7694c05993ea4d719be46/transformations/transformations/derived_types.py#L322

  File "/home/rolfhm/install/loki/loki_env/bin/loki-transform.py", line 11, in <module>
    sys.exit(cli())
  File "/home/rolfhm/install/loki/loki_env/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/rolfhm/install/loki/loki_env/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/rolfhm/install/loki/loki_env/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/rolfhm/install/loki/loki_env/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/rolfhm/install/loki/loki_env/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/rolfhm/progs/loki/scripts/loki_transform.py", line 195, in convert
    scheduler.process(transformation=DerivedTypeArgumentsTransformation())
  File "/home/rolfhm/progs/loki/loki/bulk/scheduler.py", line 642, in process
    transformation.apply(
  File "/home/rolfhm/progs/loki/loki/transform/transformation.py", line 108, in apply
    self.apply_file(source, **kwargs)
  File "/home/rolfhm/progs/loki/loki/transform/transformation.py", line 145, in apply_file
    self.apply_subroutine(routine, **kwargs)
  File "/home/rolfhm/progs/loki/loki/transform/transformation.py", line 172, in apply_subroutine
    self.transform_subroutine(subroutine, **kwargs)
[...]

PR #138, opened 2023-09-01
Fix type update behaviour in expression clones

mlange05:
_Note the problem this PR intends to fix is the underlying problem to the symptoms reported in #128. Many thanks to @rolfhm for providing the MFE that helped triage this._

This PR adds a test and fixes certain corner cases for type updates of symbols via `symbol.clone(type-<...>)`. The key problem is that there were edge-cases where we would preserve existing type info stored in the associated scope of the symbol, if the new type was deemed inadequate. This PR now simplifies this logic and only ignores the incoming type if it is `None`, as we do not allow `None` type info in the symbol table (`BaseType.DEFERRED` signifies unknown types). 

In addition, a test was added that pins down the intended behaviour for plain scalar symbols. In the future this could be extended to procedures etc., but this felt overkill for the current problem. 

I've also needed to change the scope handling logic in the fparser-based standalone expression parsing utility, where we can never gain type info (no declarations in expressions), but would wipe existing type info by setting it to DEFERRED. We now instead use a dummy to parse expressions and suppress that info in the subsequent rescoping (see inline comments).

### DIFF
--- a/loki/expression/symbols.py
+++ b/loki/expression/symbols.py
@@ -815,16 +815,10 @@ class Variable:
         scope = kwargs.get('scope')
         _type = kwargs.get('type')
 
-        if scope is not None and (_type is None or _type.dtype is BasicType.DEFERRED):
+        if scope is not None and _type is None:
             # Try to determine stored type information if we have no or only deferred type
-            stored_type = cls._get_type_from_scope(name, scope, kwargs.get('parent'))
-            if _type is None:
-                _type = stored_type
-            elif stored_type is not None:
-                if stored_type.dtype is not BasicType.DEFERRED or not _type.attributes:
-                    # If provided and stored are deferred but attributes given, we use provided
-                    _type = stored_type
-            kwargs['type'] = _type
+            _type = cls._get_type_from_scope(name, scope, kwargs.get('parent'))
+        kwargs['type'] = _type
 
         if _type and isinstance(_type.dtype, ProcedureType):
             # This is the name in a function/subroutine call
--- a/loki/expression/symbols.py
+++ b/loki/expression/symbols.py
@@ -286,7 +286,7 @@ class TypedSymbol:
             if _type.dtype.typedef is BasicType.DEFERRED:
                 return ()
             return tuple(
-                v.clone(name=f'{self.name}%{v.name}', scope=self.scope, parent=self)
+                v.clone(name=f'{self.name}%{v.name}', scope=self.scope, type=v.type, parent=self)
                 for v in _type.dtype.typedef.variables
             )
         return None
@@ -328,8 +328,12 @@ class TypedSymbol:
             kwargs['name'] = self.name
         if 'scope' not in kwargs and self.scope:
             kwargs['scope'] = self.scope
-        if 'type' not in kwargs and self.type:
-            kwargs['type'] = self.type
+        if 'type' not in kwargs:
+            # If no type is given, check new scope
+            if 'scope' in kwargs and kwargs['scope'] and kwargs['name'] in kwargs['scope'].symbol_attrs:
+                kwargs['type'] = kwargs['scope'].symbol_attrs[kwargs['name']]
+            else:
+                kwargs['type'] = self.type
         if 'parent' not in kwargs and self.parent:
             kwargs['parent'] = self.parent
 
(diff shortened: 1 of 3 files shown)


loki/frontend/fparser.py
```diff
@@ -138,9 +139,14 @@ def parse_fparser_expression(source, scope):
...
+
+    # We parse the standalone expression with a dummy scope, to
+    # avoid overriding existing type info from the given scope.
+    _ir = parse_fparser_ast(ast, source, scope=Scope())
+
     # TODO: use rescope visitor for this
-    rescope_map = {v: v.clone(scope=scope) for v in FindTypedSymbols().visit(_ir)}
+    # In the rescoping we force using the existing type info via `type=None`
+    rescope_map = {v: v.clone(scope=scope, type=None) for v in FindTypedSymbols().visit(_ir)}
     _ir = SubstituteExpressions(rescope_map).visit(_ir)
```
reuterbal, line 150: Might make sense to do this by using corresponding functionality that's already available (and thus take care of the TODO in the process), e.g.:
```
{v: v.rescope(scope=scope)}
```
or, even better because the parse expression may actually be using symbols from a parent scope:
```
_ir = AttachScopes().visit(_ir, scope=scope)
```

tests/test_expression.py
```diff
@@ -1157,6 +1157,38 @@ def test_variable_clone(initype, inireftype, newtype, newreftype):
...
+     SymbolAttributes(BasicType.DEFERRED)),
+])
+def test_variable_clone_type(initype, newtype, reftype):
+    """
+    Test type updates are handled as expected and types are never ``None``.
+    """
+    scope = Scope()
+    var = symbols.Variable(name='var', scope=scope, type=initype)
+    assert 'var' in scope.symbol_attrs
+    new = var.clone(type=newtype)
```
reuterbal, line 1189: This just needs a `# pylint: disable=no-member` annotation.
(We really ought to convert the `Variable` factory class to a plain function)

reuterbal (changes requested):
This is very nice, many thanks! The fact that the logic in the symbol constructor could be that simple makes the current implementation almost embarrassing. 

There are still a few failing tests and I've left a comment on the expression parsing change.

loki/frontend/fparser.py
mlange05: Agreed, the second suggestion looks much neater and works very well. Updated.

mlange05:
@reuterbal As discussed offline, this branch now includes a small, but subtle behaviour change to get the expression tests back up and running. For this we now treat a `Variable.clone(scope=scope)` call, where the variable previously had no scope attached, as a re-scope, and prefer the type info in the provided scope over the previously stored DEFERRED type. This requires a small change in the `TypeDef.variables` property, but once fixed I think this is actually good. 

So, please have another look and see if you agree. All other suggestions have been incorporated, let's see if CI can be convinced too.

mlange05:
@reuterbal As discussed offline, this branch now includes a small, but subtle behaviour change to get the expression tests back up and running. For this we now treat a `Variable.clone(scope=scope)` call, where the variable previously had no scope attached, as a re-scope, and prefer the type info in the provided scope over the previously stored DEFERRED type. This requires a small change in the `TypeDef.variables` property, but once fixed I think this is actually good. 

So, please have another look and see if you agree. All other suggestions have been incorporated, let's see if CI can be convinced too.

loki/expression/symbols.py
```diff
@@ -815,16 +819,10 @@ def __new__(cls, **kwargs):
         scope = kwargs.get('scope')
         _type = kwargs.get('type')
 
-        if scope is not None and (_type is None or _type.dtype is BasicType.DEFERRED):
+        if scope is not None and _type is None:
             # Try to determine stored type information if we have no or only deferred type
```
reuterbal, line 823: The comment is out of date. Given the complexity, it might be worth expanding this here a little bit in general?

reuterbal (approved):
Looks great now, many thanks for figuring this out! The comment in the `clone` method is a little out of date now (and factually wrong), so might be worth updating this.

loki/expression/symbols.py
mlange05: Good spot. Should be fixed now. Thanks!

PR #139, opened 2023-09-04
Make derived argument flattening optional in loki_transform convert

mlange05:
Unlike the `ecphys` entry point, `loki_transform.py convert` always applies the `DerivedTypeArgumentTransformation`. This PR adds a switch to allow overriding this behaviour (default `True`, as before), and extends the CMake API with a similar override option.


reuterbal (approved):
Change looks good, thanks for taking care of this. Current behaviour should be retained by default.

Issue #133, opened 2023-08-29
Hacky or... bad design for ExpressionFinder ?


meteospike:
The class `ExpressionFinder` inherits from `Visitor` in module `loki.expression.expr_visitors`.

The initialization method is doing something a bit weird, qualified in a comment as "pretty hacky" : setting a static method (`retrieval_function`) provided by a dummy argument. Moreover, this argument is in fact built on the fly in the initialization methods of the derived classes such as `FindTypedSymbols`, `FindVariables`, etc., then calling `super.__init__` with this argument, which will be attached as a static method to the parent class ! 

One may consider that everything is fine as long as we proceed like this : 
```
FindTypedSymbols().visit(item.ir)
FindVariables(unique=False).visit(item.ir)

```
But as soon as you want to "freeze" your finder (for example to delegate work to an other entity or reuse it several times), things are going wrong:
```
find_ts = FindTypedSymbols()
find_vars = FindVariables(unique=False)
find_ts.visit(item.ir)

```
The `find_ts` will in fact use the retrieval function defined by the initialization method of `FindVariables`. 
At least, even if we want to keep this kind of design for the initialization of derived classes, the retrieval function should be attached to the instance and then called by the `retrieve` method. Do I miss something ?

reuterbal:
Thanks for reporting this. I agree that they way the retrieval function is stored seems odd and certainly mandates an overhaul. However, I haven't been able to reproduce the behaviour you described, the following test works correctly for me:
```python


@pytest.mark.parametrize('frontend', available_frontends())
def test_expression_finder_retrieval_function(frontend):
    """
    Verify that expression finder visitors work as intended and remain
    functional if re-used
    """
    fcode = """
module some_mod
    implicit none
contains
    function some_func() result(ret)
        integer, intent(out) :: ret
        ret = 1
    end function some_func

    subroutine other_routine
        integer :: var, tmp
        var = 5 + some_func()
    end subroutine other_routine
end module some_mod
    """.strip()

    source = Sourcefile.from_source(fcode, frontend=frontend)

    expected_ts = {'var', 'some_func'}
    expected_vars = ('var',)

    # Instantiate the first expression finder and make sure it works as expected
    find_ts = FindTypedSymbols()
    assert find_ts.visit(source['other_routine'].body) == expected_ts

    # Verify that it works also on a repeated invocation
    assert find_ts.visit(source['other_routine'].body) == expected_ts

    # Instantiate the second expression finder and make sure it works as expected
[...]

PR #140, opened 2023-09-05
Expression mappers: Remove recurse_to_parent option and recurse by default

reuterbal:
This fixes the symptom described in #133.

The retrieval function for IR visitors that recurse into expression trees is defined as a static method on the class of each such visitor. While the current way how this is being installed is debatable, it is conceptually correct. However, a later addition allowed to parameterize this retrieval function by providing the option to not recurse to "parents" of expression nodes (the `recurse_to_parent` option). This causes problems, because multiple instances of the same class may then use different variations and consequently change the behaviour of existing objects.

This option relates to the used of derived type member variables, where e.g., a lookup of symbols would then yield, say `a%b%c`, `a%b` and `a` - but return only `a%b%c` if that recursion is switched off.

The option has only been switched off in very few places and there wasn't a hard dependency on it's functionality (only true use case was in the derived type flattening where the filtering can be done post-visitor as well). Therefore, I removed the option and made recursion the default behaviour.

### DIFF
--- a/lint_rules/lint_rules/ifs_coding_standards_2011.py
+++ b/lint_rules/lint_rules/ifs_coding_standards_2011.py
@@ -11,11 +11,12 @@ Implementation of rules in the IFS coding standards document (2011) for loki-lin
 """
 
 from pathlib import Path
-from collections import defaultdict
 import re
 
+from pymbolic.primitives import Expression
+
 from loki import (
-    Visitor, FindNodes, ExpressionFinder, FindExpressionRoot, ExpressionRetriever,
+    Visitor, FindNodes, ExpressionFinder, ExpressionRetriever,
     flatten, as_tuple, strip_inline_comments, Module, Subroutine, BasicType, ir
 )
 from loki.lint import GenericRule, RuleType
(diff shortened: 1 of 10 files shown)


loki/expression/expr_visitors.py
```diff
@@ -149,17 +149,10 @@ class FindExpressions(ExpressionFinder):
...
-    recurse_to_parent : bool, optional
-        For symbols that belong to a derived type, recurse also to the
-        ``parent`` of that symbol (default: `True`)
     """
 
-    def __init__(self, recurse_to_parent=True, **kwargs):
-        self._retriever = ExpressionRetriever(lambda e: isinstance(e, Expression),
-                                              recurse_to_parent=recurse_to_parent)
+    def __init__(self, **kwargs):
+        self._retriever = ExpressionRetriever(lambda e: isinstance(e, Expression))
```
mlange05, line 155: This still makes the behaviour object-bound, instead of class-bound and the fact that this is static is applied post-hoc. I wonder it would be easier instead to just set a static `ExpressionRetriever` (expression visitor) via a class method and change the `ExpressionFinder` (IR visitor) to always call `self.retriever.retrieve()`. So for the sub-classes we would then merely do
```
class FindTypedSymbols(ExpressionFinder):

    retriever = ExpressionRetriever(lambda x: ...)
```
without ever touching the `__init__` of the derived instance?

reuterbal:
Thanks for the comment, I had overlooked that. I have changed the implementation as suggested, making the retriever a class property.

I took the opportunity to nuke the `FindExpressionRoot` visitor, which is incompatible with that and essentially required double-traversing the tree and it's use can easily be worked around in a much smarter and more efficient way. I had to do this for one of the linter rules, which as a consequence became much neater and more readable (in my eyes, anyway).

I think this is ready for another look.

lint_rules/lint_rules/ifs_coding_standards_2011.py
```diff
@@ -372,15 +373,20 @@ def check_kind_literals(subroutine, types, allowed_type_kinds, rule_report):
...
-        finder = ExpressionFinder(unique=False, retrieve=retriever.retrieve, with_ir_node=True)
-
-        for node, exprs in finder.visit(subroutine.ir):
+
+        class FindLiteralsWithKind(ExpressionFinder):
+            """
+            Custom expression finder that that yields all literals of the types
+            specified in the config and stops recursion on loop ranges and array subscripts
+            (to avoid warnings about integer constants in these cases)
+            """
```
mlange05, line 382: [no action] Agree on this looking much nicer! :+1:

lint_rules/lint_rules/ifs_coding_standards_2011.py
```diff
@@ -472,55 +478,78 @@ class Fortran90OperatorsRule(GenericRule):  # Coding standards 4.15
...
+        """
+
+        retriever = ExpressionRetriever(lambda e: isinstance(e, sym.Comparison))
+
+        def visit_Node(self, o, **kwargs):
+            """
+            Generic visitor method that will call the :any:`ExpressionRetriever`
+            only on :class:`pymbolic.primitives.Expression` children, collecting
+            ``(node, expression root, comparison)`` tuples for all matches.
+            """
```
mlange05, line 503: [no action] Uhhh, nice, although quite advanced interface usage! Still, the level of documentation makes it at least accessible.

mlange05 (approved):
Very nice, and the consistent use of Retriever objects to handle our expression trees in a flexible manner looks very neat to me. GTG from me :shipit:

Issue #134, opened 2023-08-30
Enriched call `.name` reverts to procedure symbol defined in explicit interface


awnawab:
If an explicit interface for a subroutine/function is present, the relevant call's `.name` points to the defined `ProcedureSymbol` once the subroutine is parsed. If we then call `subroutine.enrich_calls`, we update `call.name` to point to the actual subroutine rather than the symbol defined in the interface.

However this update doesn't seem to be permanent. If we ever later access the symbols inside the `Interface`, then `call.routine` again reverts to pointing to that symbol rather than the actual subroutine. (demonstrator attached)

I think the fundamental problem here is that we have now introduced two symbols with the same name to a single scope. Given it's such a fundamental issue, I'm at a loss for a proper solution. A quick and dirty fix would be to set `scope=None` if the symbol already exists when updating `call.name` in `subroutine.enrich_calls`.

[enrich_calls_bug.txt](https://github.com/ecmwf-ifs/loki/files/12475942/enrich_calls_bug.txt)

awnawab:
That quick fix falls apart as soon as we hit the first `rescope_symbols`. Hmm....

PR #141, opened 2023-09-05
Fix for enrich calls bug

awnawab:
This PR fixes the bug highlighted in #134 (in the presence of an explicit interface, `call.name` still points to symbol defined by interface even after enriching calls).

### DIFF
--- a/loki/subroutine.py
+++ b/loki/subroutine.py
@@ -467,6 +467,12 @@ class Subroutine(ProgramUnit):
                         name_type.dtype.procedure is not routine  # ProcedureType not linked to routine
                     )
                     if update_symbol:
+                        # Remove existing symbol from symbol table if defined in interface block
+                        for node in [node for intf in self.interfaces for node in intf.body]:
+                            if getattr(node, 'name', None) == call.name:
+                                if node.parent == self:
+                                    node.parent = None
+
                         # Need to update the call's symbol to establish link to routine
                         name_type = name_type.clone(dtype=routine.procedure_type)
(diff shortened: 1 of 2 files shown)


loki/subroutine.py
```diff
@@ -467,6 +467,12 @@ def enrich_calls(self, routines):
                         name_type.dtype.procedure is not routine  # ProcedureType not linked to routine
                     )
                     if update_symbol:
+                        # Remove existing symbol from symbol table if defined in interface block
+                        for node in [node for intf in self.interfaces for node in intf.body]:
+                            if getattr(node, 'name', None) == call.name:
+                                if node.parent == self:
+                                    node.parent = None
```
reuterbal, line 474: This breaks with the nested scoping behaviour that we're using normally. However, the root problem is that interface blocks shouldn't really have a parent scope in the first place because they can't inherit symbols,  so I'm ok with this temporary solution.

reuterbal (approved):
Many thanks, well-tested fix for a subtle bug.

As discussed offline, the proper way of fixing this in the future would be to improve the representation of interfaces. While the interface node needs to define a scope, it shouldn't really have a parent because it can't inherit symbols.

However, at the same time we likely want to retain the behaviour of registering the symbols declared by the interface in the parent scope's symbol table.

This is a more invasive change, though, and for a future PR.

mlange05 (approved):
Agree with all that was said. GTG from me. :shipit:

loki/subroutine.py
reuterbal: Filed a follow-up ticket in #147

PR #142, opened 2023-09-06
Remove "ecphys" entry point from Loki_transform.py

mlange05:
**Note: This changes external-facing features and default behaviour!**

Following recent additions, the preferred "convert" mode (entry point) to `loki_transform.py` is now sufficient to apply SCC-vector to the full EC-physics demonstrator, provided certain arguments are renamed to mimic the previous API. This PR attempts exactly this, while keeping the old argument names as aliases for backward-compatibility. The PR then switches the command to the Python-layer entry points in the CMake layer, which is sufficient to toggle the behaviour.

In some more detail:
* Adjust option names in `loki_transform.py convert` to mimic previous "ecphys" API (it's mostly names for path options)
* Update the "convert" docstring, which had grown outdated by quite a bit
* Switch default behaviour of the `--remove-derived--args` option in `loki_transform.py convert` to be off by default
* Remove "ecphys" entry point entirely
* Remove redundant `cloudsc_config` default dict and always expect a path to a config file

**Important: Due to the external behaviour changes, please thoroughly check all downstream regression tests!**


scripts/loki_transform.py
```diff
@@ -136,36 +96,33 @@ def cli(debug):
...
         global_var_offload, remove_derived_args
 ):
     """
-    Single Column Abstraction (SCA): Convert kernel into single-column
-    format and adjust driver to apply it over in a horizontal loop.
-
-    Optionally, this can also insert CLAW directives that may be use
-    for further downstream transformations.
+    Batch-processing mode for Fortran-to-Fortran transformations that
+    employs a :class:`Scheduler` to process large numbers of source
```
reuterbal, line 114: The docstring of click commands also serves as the text served on command line as help. Therefore, I would consider this to be the only place where we should not use the Sphinx/RST link syntax with `:class:` & co.

reuterbal (approved):
This is great and I'm very happy that we seem to be converging to a cleaner outward facing interface.

Regression testing seems to be successful for CLOUDSC and ecWAM and I trust you have tested this successfully on ecphys.

awnawab (approved):
Really great that we now only have a single entry point for Loki transformations, thanks a lot for consolidating them!

PR #145, opened 2023-09-07
Add `DEVICEPTR` annotations to data region in driver loop

awnawab:
A small update to the `DataOffload` transformation to mark kernel array arguments as true device-pointers and skip address translation via the OpenACC map. This is needed when using the CUDA backend in the new FIELD_API. The new functionality has been tested here: https://github.com/ecmwf-ifs/ecwam/tree/naan-deviceptr.

NB: This is stacked on top of #142.

### DIFF
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -92,6 +92,8 @@ def cli(debug):
               help='Run transformation to insert custom data offload regions.')
 @click.option('--remove-openmp', is_flag=True, default=False,
               help='Removes existing OpenMP pragmas in "!$loki data" regions.')
+@click.option('--deviceptr', is_flag=True, default=False,
+              help='Mark the relevant arguments as true device-pointers in "!$loki data" regions.')
 @click.option('--frontend', default='fp', type=click.Choice(['fp', 'ofp', 'omni']),
               help='Frontend parser to use (default FP)')
 @click.option('--trim-vector-sections', is_flag=True, default=False,
@@ -102,7 +104,7 @@ def cli(debug):
               help="Remove derived-type arguments and replace with canonical arguments")
 def convert(
         mode, config, build, source, header, cpp, directive, include, define, omni_include, xmod,
-        data_offload, remove_openmp, frontend, trim_vector_sections,
+        data_offload, remove_openmp, deviceptr, frontend, trim_vector_sections,
         global_var_offload, remove_derived_args
 ):
     """
@@ -165,7 +167,7 @@ def convert(
     # Insert data offload regions for GPUs and remove OpenMP threading directives
     use_claw_offload = True
     if data_offload:
-        offload_transform = DataOffloadTransformation(remove_openmp=remove_openmp)
+        offload_transform = DataOffloadTransformation(remove_openmp=remove_openmp, deviceptr=deviceptr)
         scheduler.process(transformation=offload_transform)
         use_claw_offload = not offload_transform.has_data_regions
 
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -92,7 +92,7 @@ def cli(debug):
               help='Run transformation to insert custom data offload regions.')
 @click.option('--remove-openmp', is_flag=True, default=False,
               help='Removes existing OpenMP pragmas in "!$loki data" regions.')
-@click.option('--deviceptr', is_flag=True, default=False,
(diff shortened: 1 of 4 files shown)


transformations/transformations/data_offload.py
```diff
@@ -27,13 +27,17 @@ class DataOffloadTransformation(Transformation):
     ----------
     remove_openmp : bool
         Remove any existing OpenMP pragmas inside the marked region.
+    deviceptr : bool
```
reuterbal, line 30: It's a little hard to judge from the brief option name what that implies. I would propose to follow the typical "start with a verb" naming scheme, e.g., by calling this `assume_deviceptr`? That implies imho also the requirement this imposes on the calling code to handle the offload.

transformations/transformations/data_offload.py
```diff
@@ -134,10 +138,16 @@ def insert_data_offload_pragmas(self, routine, targets):
...
                 # Now geenerate the pre- and post pragmas (OpenACC)
-                copyin = f'copyin({", ".join(inargs)})' if inargs else ''
-                copy = f'copy({", ".join(inoutargs)})' if inoutargs else ''
-                copyout = f'copyout({", ".join(outargs)})' if outargs else ''
-                pragma = Pragma(keyword='acc', content=f'data {copyin} {copy} {copyout}')
+                if self.deviceptr:
+                    deviceptr = ''
+                    if inargs+outargs+inoutargs:
+                        deviceptr = f'deviceptr({", ".join(inargs+outargs+inoutargs)})'
+                    pragma = Pragma(keyword='acc', content=f'data {deviceptr}')
```
reuterbal, line 145:
```suggestion
                if self.deviceptr:
                    offload_args = inargs + outargs + inoutargs
                    if offload_args:
                        deviceptr = f' deviceptr({", ".join(offload_args)})'
                    else:
                        deviceptr = ''
                    pragma = Pragma(keyword='acc', content=f'data{deviceptr}')
```

Purely a matter of taste, but I find this a little hard to read with the repeated contracted "sum".

reuterbal (changes requested):
Very nicely done and useful addition, many thanks!

Only real remark is related to naming: I'd prefer a marginally more descriptive naming of that option all the way through from CMake to transformation constructor. Suggestion would be something like `assume_deviceptr` but better variants do probably exist.

transformations/transformations/data_offload.py
mlange05: Seconded (but yes, matter of taste), spaces around operators used to be a pylint rule (for a reason).

mlange05: Seconded (but yes, matter of taste), spaces around operators used to be a pylint rule (for a reason).

transformations/transformations/data_offload.py
mlange05: Seconded, either `assume_deviceptr` or `use_deviceptr` or `mark_as_deviceptr` (or similar). And ideally propagate change up to cmake layer (explicit is better than implicit :wink: )

mlange05 (approved):
I Agree with all of the previous suggestions, but otherwise GTG. I've not tested this explicitly with any regression tests, but might try CLOUDSC later myself. This should not stop this being merged tho, so GTG from me. :shipit:

transformations/transformations/data_offload.py
```diff
@@ -27,13 +27,17 @@ class DataOffloadTransformation(Transformation):
     ----------
     remove_openmp : bool
         Remove any existing OpenMP pragmas inside the marked region.
+    assume_deviceptr : bool
```
reuterbal, line 30: Docstring and actual argument are out of sync here now. Can we call the argument `assume_deviceptr`?

scripts/loki_transform.py
```diff
@@ -92,6 +92,8 @@ def cli(debug):
               help='Run transformation to insert custom data offload regions.')
 @click.option('--remove-openmp', is_flag=True, default=False,
               help='Removes existing OpenMP pragmas in "!$loki data" regions.')
+@click.option('--assume_deviceptr', is_flag=True, default=False,
```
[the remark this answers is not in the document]
awnawab: Nice catch! Corrected 👍

reuterbal (approved):
Waiting for CI to confirm but looks good to me now. Many thanks for applying the renaming and apologies that this has turned into such a rat's tail of driving the renaming through the layers...

PR #146, opened 2023-09-07
Provide a graphical representation of the IR using graphviz

joscao:
IR is now viewable as a graph, allowing a general overview of the IR structure. Formatting of the node labels is done via pprint of the IR, keeping the overview as sane as possible.

Examples of small fortran files:

[module_with_subroutines.pdf](https://github.com/ecmwf-ifs/loki/files/12550588/module_with_subroutines.pdf)
[case_statement_subroutine.pdf](https://github.com/ecmwf-ifs/loki/files/12550623/case_statement_subroutine.pdf)
[if_else_statement_subroutine.pdf](https://github.com/ecmwf-ifs/loki/files/12550624/if_else_statement_subroutine.pdf)
[nested_if_else_statements_subroutine.pdf](https://github.com/ecmwf-ifs/loki/files/12550625/nested_if_else_statements_subroutine.pdf)

Also larger files are possible (like example/src/phys_mod.F90):
[example_src_phys_mod.F90.pdf](https://github.com/ecmwf-ifs/loki/files/12550646/example_src_phys_mod.F90.pdf)

Comments and recommendations on coding style / integration into the project are greatly appreciated.

If somebody knows of a nicer way to force graphviz to utilize quotes for the labels than using an invisible character I would love to change it.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/loki?pullRequest=146) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/loki?pullRequest=146) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/loki?pullRequest=146) it.</sub>

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""
+Pretty-Graph-Visualizer classes for IR
+"""
+from itertools import chain
+from codetiming import Timer
+from graphviz import Digraph
```
reuterbal, line 13: This import should be guarded against missing graphviz, similar to here:
https://github.com/ecmwf-ifs/loki/blob/75fc36f389b560cbf6f18ee22c80a80860f36ac8/loki/frontend/ofp.py#L15-L21

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def visit_all(self, item, *args, **kwargs):
+        """
+        Convenience function to call :meth:`visit` for all given arguments.
+
+        If only a single argument is given that is iterable,
+        :meth:`visit` is called on all of its elements instead.
+        """
+        if is_iterable(item) and not args:
+            return list(
```
reuterbal, line 138: What effect does the nesting of `list(chain.from_iterable(as_tuple(...)))` have? It feels like it's building multple times the sequence but I'm probably missing something here.

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+            A list of a tuple of a node and potentially a edge information
+        """
+        label = kwargs.get("label", "")
+        if label == "":
+            label = self.format_node(repr(node))
+
+        shape = kwargs.get("shape", "oval")
+        label = (
+            label + "‎"
+        )  # dirty hack to force graphviz to utilize parenthesis around the label
```
reuterbal, line 176: I don't understand this. Appending an empty string to `label` should not change the string itself - if that makes a difference, I suspect the root cause is something else and the add operator triggers a type conversion somehow.

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        Returns
+        -------
+        list[tuple[dict[str,str], dict[str,str]]]]
+            An extended list of tuples of a node and potentially a edge information
+        """
+        node_edge_info = self.__add_node(o, **kwargs)
+        kwargs["parent"] = o
+
+        node_edge_info.extend(self.visit(o.spec, **kwargs))
+        node_edge_info.extend(self.visit(o.subroutines, **kwargs))
```
reuterbal, line 212:
```suggestion
        node_edge_info.extend(self.visit(o.contains, **kwargs))
```
The `contains` part can contain also pragmas, comments etc. which are omitted if picking out the subroutines only

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        list[tuple[dict[str,str], dict[str,str]]]]
+            An extended list of tuples of a node and potentially a edge information
+        """
+        node_edge_info = self.__add_node(o, **kwargs)
+        kwargs["parent"] = o
+
+        node_edge_info.extend(self.visit(o.docstring, **kwargs))
+        node_edge_info.extend(self.visit(o.spec, **kwargs))
+        node_edge_info.extend(self.visit(o.body, **kwargs))
+        node_edge_info.extend(self.visit(o.members, **kwargs))
```
reuterbal, line 231:
```suggestion
        node_edge_info.extend(self.visit(o.contains, **kwargs))
```

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        """
+        return JoinableStringList(
+            items,
+            sep=sep,
+            width=self.linewidth,
+            cont="\n",
+            separable=separable,
+        )
+
+    def format_node(self, name, *items):
```
reuterbal, line 88: [no action required] The duplication here compared to Stringifier makes me think that we should introduce a common base class for the joint functionality.

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def format_line(self, *items, comment=None, no_wrap=False):
+        """
+        Format a line by concatenating all items and applying indentation while observing
+        the allowed line width limit.
+
+        Note that the provided comment will simply be extended to the line and no line
+        width limit will be enforced for that.
+
+        :param list items: the items to be put on that line.
```
reuterbal, line 111: This uses a different docstring flavour, presumably because the Stringifier method had that? Would you mind aligning this with the style in the other methods?

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        Returns
+        -------
+        list[tuple[dict[str,str], dict[str,str]]]]
+            An extended list of tuples of a node and potentially a edge information or list of nothing.
+        """
+        if kwargs.get("show_expressions"):
+            content = self.symgen(o)
+            parent = kwargs.get("parent")
+            return self.__add_node(o, label=content, parent=parent, shape="box")
+        return [None]
```
reuterbal, line 280: Does this need the `None` in the list? I'd imagine an empty list `[]` would have the same effect?

tests/test_pretty_visualize.py
```diff
@@ -0,0 +1,258 @@
...
+from loki.visitors.pretty_visualize import pretty_visualize, Visualizer
+
+
+@pytest.fixture(scope="module", name="here")
+def fixture_here():
+    return Path(__file__).parent
+
+
+test_files = [
+    "trivial_fortran_files/case_statement_subroutine.f90",
```
reuterbal, line 23: Can we move the entire `trivial_fortran_files` into `tests/sources`?

tests/test_pretty_visualize.py
```diff
@@ -0,0 +1,258 @@
...
+            yield (None, edge_info[name])
+            continue
+
+        if node_info and edge_info:
+            raise KeyError(f"Keyword {name} not found!")
+
+
+@pytest.mark.skipif(not graphviz_present(), reason="Graphviz is not installed")
+@pytest.mark.parametrize("test_file", test_files)
+def test_Visualizer(here, test_file):
```
reuterbal, line 168:
```suggestion
def test_visualizer(here, test_file):
```

naming convention for functions is snake_case

loki/ir.py
```diff
@@ -158,13 +158,17 @@ def args_frozen(self):
     def __repr__(self):
         raise NotImplementedError
 
-    def view(self):
+    def view(self, visualization = False, **kwargs_visualization):
```
reuterbal, line 161: I think I'd separate the functionality into a new method, e.g. `visualize`? Given that this is likely to be used mostly interactively, it's shorter to write and muscle memory is easier/auto completion more reliable with a separate method than switching via optional arguments.

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+    def visit_Comment(self, o, **kwargs):
+        """
+        Enables turning off comments.
+
+        Returns
+        -------
+        list[tuple[dict[str,str], dict[str,str]]]]
+            An extended list of tuples of a node and potentially a edge information, or list of nothing.
+        """
+        if kwargs.get("show_comments"):
```
reuterbal, line 245: This is behaviour parameterisation for the entire class - hence I would suggest to make this an object attribute and to add it to the constructor instead of passing it around via `kwargs`.

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        """
+        Dispatch routine to add nodes utilizing expression tree stringifier,
+        mark parent and stop.
+
+        Returns
+        -------
+        list[tuple[dict[str,str], dict[str,str]]]]
+            An extended list of tuples of a node and potentially a edge information or list of nothing.
+        """
+        if kwargs.get("show_expressions"):
```
reuterbal, line 276: Same as for `show_comments`: This is behaviour parameterisation for the entire class - hence I would suggest to make this an object attribute and to add it to the constructor instead of passing it around via kwargs.

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        node_edge_info = self.__add_node(o, label=label, parent=parent, shape="diamond")
+        kwargs["parent"] = o
+        node_edge_info.extend(self.visit_all(o.body, **kwargs))
+
+        if o.else_body:
+            node_edge_info.extend(self.visit_all(o.else_body, **kwargs))
+        return node_edge_info
+
+
+def pretty_visualize(ir, **kwargs):
```
reuterbal, line 310: Given that you only expect a number of specific keys (filename, show_comments, show_expressions), I would prefer having them explicitly as optional arguments in the function signature instead of a plain `kwargs`

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,349 @@
...
+        graph.attr(rankdir="LR")
+        for node_info, edge_info in node_edge_info:
+            if node_info:
+                graph.node(**node_info)
+            if edge_info:
+                graph.edge(**edge_info)
+        if filename:
+            graph.render(**kwargs)
+        else:
+            return graph
```
reuterbal, line 349: I think we should either always or never return the graph, conditionally switching off return values is considered bad practice (and will likely be flagged by the linter)

reuterbal (changes requested):
Thank you very much, this is an impressive contribution after only very short time, neatly implemented and well-tested.

I have left a number of comments to make suggestions how I think this could be further improved, but this is all minor details.

loki/visitors/pretty_visualize.py
joscao: Fix with a9cb30cb65c4348747ff73a0d286fa2b89259322

loki/visitors/pretty_visualize.py
joscao: Fix with 781ba3d2635147668e8770f119af5d772b95bf0d

loki/visitors/pretty_visualize.py
joscao: Was unnecessary 8b566dd2c72513270d3db8721b1e7129847b1d08

loki/visitors/pretty_visualize.py
joscao: Fixed 37ee8f07da7f7bb1e804f46039de50e17e6d2d23

loki/visitors/pretty_visualize.py
joscao: Resolved by 98d8d7842942cc505ac8f041344b852234d1952a

loki/visitors/pretty_visualize.py
joscao: Resolved by 98d8d7842942cc505ac8f041344b852234d1952a

tests/test_pretty_visualize.py
joscao: sure no problem a33c38f07e79acb4246f0e705fbd85e9628c31b6

tests/test_pretty_visualize.py
joscao: alright 65696625354a8904af110a250bccd5a3585eb9a7

loki/ir.py
joscao: has been separated into own function with explicit arguments, c72fe87359edd81a3f233e85478edf849f7b2c33

loki/visitors/pretty_visualize.py
joscao: this and show comments are now object attributes passed to the constructor c72fe87359edd81a3f233e85478edf849f7b2c33

loki/visitors/pretty_visualize.py
joscao: parameters are now explicitly passed c72fe87359edd81a3f233e85478edf849f7b2c33

loki/visitors/pretty_visualize.py
joscao: I agree, since graph is always returned rendering of graph is no longer reasonable, therefore filename is removed c72fe87359edd81a3f233e85478edf849f7b2c33

loki/visitors/pretty_visualize.py
reuterbal: Nice find!

reuterbal (approved):
Thanks again, very nicely done. All comments have been addressed, GTG from me.

mlange05 (changes requested):
First of all, a very useful and neat feature! Many thanks for taking this on. It's very impressive!

I've left two small nitpicky commits, one about Python errors and one about naming conventions. The latter leads to an important realisation though: This is visual tool, and no end-to-end example is provided. So I suggest we add a small demo of this to one of the tutorial notebooks, or even add a small dedicated notebook for highlighting this functionality (and maybe add an example of the textual pretty-printer too :wink: ). 

All in all, a really great first contribution though! Many, many thanks!

reuterbal:
Good point, adding this to the [first example notebook](https://github.com/ecmwf-ifs/loki/blob/main/example/01_reading_and_writing_files.ipynb) after the use of `phys_mod.spec.view()` as an alternative way to view the IR would be fantastic.

And while you're at it: Make sure to add your name to https://github.com/ecmwf-ifs/loki/blob/main/AUTHORS.md

loki/visitors/pretty_visualize.py
```diff
@@ -0,0 +1,357 @@
...
+        node_edge_info = [item for item in visualizer.visit(ir) if item is not None]
+
+        graph = Digraph()
+        graph.attr(rankdir="LR")
+        for node_info, edge_info in node_edge_info:
+            if node_info:
+                graph.node(**node_info)
+            if edge_info:
+                graph.edge(**edge_info)
+        return graph
```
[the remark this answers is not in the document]
joscao: Yes the DiGraph pretty-prints itself if used in a notebook context. I do though agree that the name is badly chosen , i have renamed it. Additionally I have like @reuterbal recommended added the use case of the graph visualization to the first example notebook.

mlange05 (approved):
@joscao Many thanks for the additional changes. This looks very cool now! :sunglasses: 

A fantastic first contribution! Well done! 

GTG from me :shipit: :grin:

PR #148, opened 2023-09-08
fix for argument shape regarding module imports

MichaelSt98:
Currently, there is a problem regarding the array argument shape transformation used within the SCC transformations. 

E.g.  regarding

```fortran
MODULE DRIVER_MOD
USE KERNEL_A_MOD, ONLY: KERNEL_A
IMPLICIT NONE
CONTAINS
  SUBROUTINE driver(nlon, nlev, a, b, c)
    INTEGER, INTENT(IN)   :: nlon, nlev  ! Dimension sizes
    INTEGER, PARAMETER    :: n = 5
    REAL, INTENT(INOUT)   :: a(nlon)
    REAL, INTENT(INOUT)   :: b(nlon,nlev)
    REAL, INTENT(INOUT)   :: c(nlon,n)

    call kernel_a(a, b, c)

  END SUBROUTINE driver

END MODULE DRIVER_MOD
```

```fortran
MODULE KERNEL_A_MOD
IMPLICIT NONE
CONTAINS
  SUBROUTINE kernel_a(a, b, c)
    USE VAR_MODULE_MOD, only: n
    REAL, INTENT(INOUT)   :: a(:)
    REAL, INTENT(INOUT)   :: b(:,:)
    REAL, INTENT(INOUT)   :: c(:,:)

  END SUBROUTINE kernel_a

END MODULE KERNEL_A_MOD
```

with 

```fortran
MODULE VAR_MODULE_MOD
INTEGER, PARAMETER    :: n = 5
END MODULE VAR_MODULE_MOD
```

the transformation without fix would create a definition `INTEGER, PARAMETER,INTENT(IN)   :: n = 5` in the routine `kernel_a` which causes a compilation error.

This is fixed by checking additionally for module imports (of both the routine and module).


mlange05 (approved):
Clean, neat and well tested. Very nice! GTG from me. :shipit:

reuterbal (approved):
Fully agreed, GTG!

PR #149, opened 2023-09-12
Loki_transform: Re-instate argument shape derivation in `convert`

mlange05:
Post-hoc fix to an oversight in PR #142. When removing the dedicated `ecphys` entry point, we dropped the argument array shape transformation that is used to replace `:` dimensions with explicitly named dimensions in the surface modules of the code. This PR adds an option (on by default) to the `convert` entry point that re-enables this behaviour.

### DIFF
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -107,10 +107,12 @@ def cli(debug):
               help="Remove derived-type arguments and replace with canonical arguments")
 @click.option('--inline-members/--no-inline-members', default=False,
               help='Inline member functions for SCC-class transformations.')
+@click.option('--argument-array-shape/--no-argument-array-shape', default=True,
+              help="Recursively derive explicit shape dimension for argument arrays")
 def convert(
         mode, config, build, source, header, cpp, directive, include, define, omni_include, xmod,
         data_offload, remove_openmp, assume_deviceptr, frontend, trim_vector_sections,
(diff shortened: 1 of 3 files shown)


reuterbal (approved):
Uncontroversial and looks good. Sorry for not spotting this on the original PR.

Only question would be whether to opt for a more descriptive option name, like `--derive-argument-array-shape`? Similarly to `--remove-openmp`, `--removed-derived-args`, `--trim-vector-sections` I find a verb helps to convey the implication.

scripts/loki_transform.py
```diff
@@ -104,12 +104,12 @@ def cli(debug):
               help="Generate offload instructions for global vars imported via 'USE' statements.")
 @click.option('--remove-derived-args/--no-remove-derived-args', default=False,
               help="Remove derived-type arguments and replace with canonical arguments")
-@click.option('--argument-array-shape/--no-argument-array-shape', default=False,
+@click.option('--derive-argument-array-shape/--no-argument-array-shape', default=False,
```
reuterbal, line 107: Shouldn't this be `--no-derive-argument-array-shape` now as well? (Sorry, bit of a mouthful)

reuterbal:
Rebased over latest main, resolving conflicts. Further CMake refactoring filed as separate PR (#166) against this branch.

mlange05:
Ok, I think this is now GTG for merging once the CI comes back green.

PR #152, opened 2023-09-13
Member inlining extended

JoeffreyLegaux:
This extends PR #130 to circumvent some limitations.

The first commit solves the issue when a derived type argument has to be matched to an array inside the inlined subroutine. This led to a crash since a DeferredTypeSymbol does not have a 'dimensions' attribute.

The second commit revamps the management of dimensions mapping. This was done originally by creating a mapper that would map the relevant argument dimensions to the variable dimensions, however when the argument has several assumed-size dimensions they would end up being mapped to the same single dimension. I solved this by making _map_unbound_dims replace the dimensions with awareness of their respective positions, and return the tuple of new dimensions, then create a clone of the argument with the new dimensions to replace the variable instance.

The third commit treats the cases where there are multiple arguments to variables conversions nested in the same expression.
I solved this through a recursive process that re-applies the expression substitution until all the variables names from the argument-to-variables list have been eliminated. This is a straightforward approach but I guess there are probably more elegant solutions to this.

### DIFF
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -14,13 +14,18 @@ from loki.expression import (
     FindVariables, FindInlineCalls, FindLiterals,
     SubstituteExpressions, LokiIdentityMapper
 )
-from loki.ir import Import, Comment, Assignment
+from loki.ir import Import, Comment, Assignment, VariableDeclaration, CallStatement
 from loki.expression import symbols as sym
 from loki.types import BasicType
 from loki.visitors import Transformer, FindNodes
+from loki.subroutine import Subroutine
+from loki.tools import as_tuple
 
 
-__all__ = ['inline_constant_parameters', 'inline_elemental_functions']
+__all__ = [
+    'inline_constant_parameters', 'inline_elemental_functions',
+    'inline_member_procedures'
+]
 
 
 class InlineSubstitutionMapper(LokiIdentityMapper):
@@ -183,3 +188,62 @@ def inline_elemental_functions(routine):
         if all(hasattr(s, 'type') and s.type.dtype in removed_functions for s in im.symbols):
             import_map[im] = None
     routine.spec = Transformer(import_map).visit(routine.spec)
+
+
+def inline_member_routine(routine, member):
+    """
+    Inline an individual member :any:`Subroutine` at source level.
+
+    This will replace all :any:`Call` objects to the specified
+    subroutine with an adjusted equivalent of the member routines'
+    body. For this, argument matching, including partial dimension
+    matching for array references is performed, and all
+    member-specific declarations are hoisted to the containing
+    :any:`Subroutine`.
+
+    routine : :any:`Subroutine`
+        The subroutine in which to inline all calls to the member routine
+    member : :any:`Subroutine`
+        The contained member subroutine to be inlined in the parent
+    """
+
+    # Get local variable declarations and hoist them
+    decls = FindNodes(VariableDeclaration).visit(member.spec)
+    decls = tuple(d for d in decls if all(not s.type.intent for s in d.symbols))
+    decls = tuple(d for d in decls if all(s not in routine.variables for s in d.symbols))
+    routine.spec.append(decls)
+
+    call_map = {}
+    for call in FindNodes(CallStatement).visit(routine.body):
+        if call.routine == member:
+            # Substitute argument calls into a copy of the body
+            member_body = SubstituteExpressions(call.argument_map).visit(member.body.body)
+
+            # Inline substituted body within a pair of marker comments
+            comment = Comment(f'! [Loki] inlined member subroutine: {member.name}')
+            c_line = Comment('! =========================================')
+            call_map[call] = (comment, c_line) + as_tuple(member_body) + (c_line, )
+
+    # Replace calls to member with the member's body
+    routine.body = Transformer(call_map).visit(routine.body)
+    # Can't use transformer to replace subroutine, so strip it manually
+    contains_body = tuple(n for n in routine.contains.body if not n == member)
+    routine.contains._update(body=contains_body)
+
+
+def inline_member_procedures(routine):
+    """
+    Inline all member subroutines contained in an individual :any:`Subroutine`.
+
+    Please note that member functions are not yet supported!
+
+    routine : :any:`Subroutine`
+        The subroutine in which to inline all member routines
+    """
+
+    # Run through all members and invoke individual inlining transforms
+    for member in routine.members:
+        if isinstance(member, Subroutine):
+            inline_member_routine(routine, member)
+
+    # TODO: Implement for functions!!!
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -12,7 +12,7 @@ Collection of utility routines to perform code-level force-inlining.
 """
 from loki.expression import (
     FindVariables, FindInlineCalls, FindLiterals,
-    SubstituteExpressions, LokiIdentityMapper
+    SubstituteExpressions, SubstituteExpressionsMapper, LokiIdentityMapper
 )
 from loki.ir import Import, Comment, Assignment, VariableDeclaration, CallStatement
 from loki.expression import symbols as sym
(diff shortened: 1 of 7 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/loki?pullRequest=152) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/loki?pullRequest=152) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/loki?pullRequest=152) it.</sub>

loki/transform/transform_inline.py
```diff
@@ -210,15 +213,22 @@ def inline_member_routine(routine, member):
...
         For example, mapping the passed array `m(:,j)` to the local
         expression `a(i)` yields `m(i,j)`.
         """
-        val_free_dims = tuple(d for d in val.dimensions if isinstance(d, sym.Range))
-        var_bound_dims = tuple(d for d in var.dimensions if not isinstance(d, sym.Range))
-        mapper = SubstituteExpressionsMapper(dict(zip(val_free_dims, var_bound_dims)))
-        return mapper(val)
+        new_dimensions = list(val.dimensions)
+        indices = []
+        for index, dim in enumerate(val.dimensions):
```
mlange05, line 223: This helper method is called in a pretty tightly iterated place, so list comprehension should be preferred over <list>.append.

loki/transform/transform_inline.py
```diff
@@ -238,17 +248,38 @@ def _map_unbound_dims(var, val):
                 if isinstance(arg, sym.Array):
                     # Resolve implicit dimension ranges of the passed value,
                     # eg. when passing a two-dimensional array `a` as `call(arg=a)`
-                    qualified_value = val if val.dimensions else val.clone(
-                        dimensions=tuple(sym.Range((None, None)) for _ in arg.shape)
-                    )
+                    # Check if val is a DeferredTypeSymbol, as it does not have a `dimensions` attribute
+                    if not isinstance(val, sym.DeferredTypeSymbol) and val.dimensions:
```
mlange05, line 252: Interesting corner case! Do you happen to have a small MFE code snippet to reproduce this? If not I'll try and add one to the main PR once I have rebased it.

loki/transform/transform_inline.py
```diff
@@ -238,17 +248,38 @@ def _map_unbound_dims(var, val):
...
+                        qualified_value = val
+                    else:
+                        qualified_value = val.clone(
+                            dimensions=tuple(sym.Range((None, None)) for _ in arg.shape)
+                        )
                     arg_vars = tuple(v for v in member_vars if v.name == arg.name)
-                    argmap.update((v, _map_unbound_dims(v, qualified_value)) for v in arg_vars)
+                    argmap.update((v, val.clone(dimensions = _map_unbound_dims(v, qualified_value))) for v in arg_vars)
```
mlange05, line 259: I was really trying to avoid using deep-clones in this tight level of loop (it's expensive at the wrong granularity, and we already have a few cases where we're being careless with this :sweat_smile:). Is there a particular reasons this is needed here?

loki/transform/transform_inline.py
```diff
@@ -238,17 +248,38 @@ def _map_unbound_dims(var, val):
...
+            arg_names = [a.name for a in argmap]
+            depth = 0
+            while test :
+                depth += 1
+                test = False
+                for var in FindVariables().visit(member_body):
+                    if var.name in arg_names:
+                        test=True
+                if test:
+                    member_body = SubstituteExpressions(argmap).visit(member_body)
```
mlange05, line 277: I believe the `SubstituteExpression` utility already has option to apply recursion to the map itself. A depth-tested while loop is for sure an anti-pattern that I would very much like to avoid here (sorry).

The utility is here:
https://github.com/ecmwf-ifs/loki/blob/2322c1bd863b5f3e43157dde66f8c3f938a98680/loki/transform/transform_utilities.py#L417

mlange05 (commented):
Hi @JoeffreyLegaux first of all, many thanks for this contribution. I think I can see the edge-cases you're referring to here. Unfortunately, there are a few patterns in there that I don't think should go inside the low-level helper routine that we apply over these expressions, which I've flagged in-line.

Moreover, for a fix contribution like this, it would be really good to expand the test base, either by adding new tests or extending the ones in the original PR, so that we can verify the fix, or reason about a better implementation. I'll try and do that in the main PR for the inliner now and incorporate your suggestions, as I understand them.

Unfortunately, as pointed out in the main PR, the target branch for your PR here requires rebasing, so it might become incompatible with this PR. I'll try and bring your changes on board though, as best I can. So, thanks again for pointing out these corner cases, and I'll ping once I've rebased.

loki/transform/transform_inline.py
JoeffreyLegaux: Ok nevermind, I managed to make that work through ArraySubscripts.

JoeffreyLegaux:
Hello @mlange05, thanks to your comments I managed to write a much cleaner implementation that I just commited.

JoeffreyLegaux:
The test case that I'm using for these changes, which is stripped from actual code :

```
SUBROUTINE CPCFU()

IMPLICIT NONE

REAL(KIND=JPRB) :: ZFPREC9(YDCPG_OPTS%KLON) ! past cumulated flux of total precipitations
REAL(KIND=JPRB) :: ZFPREC0(YDCPG_OPTS%KLON) ! instantaneous flux of total precipitations

CALL ACCUM3_CFU (YDCFU%YFPLCL, YDCFU%FPLCL, YDMF_PHYS_OUT%FPLCL)
CALL ACCUM3_CFU (YDCFU%YFPLCL, YDCFU%F_FPLCL(:,:,BLK), YDMF_PHYS_OUT%FPLCL)

CONTAINS

SUBROUTINE ACCUM3_CFU (YDDESC, PXFU, PPHY)

TYPE(FLUXES_DESCRIPTOR) ,INTENT(IN)               :: YDDESC
REAL(KIND=JPRB)         ,INTENT(IN)               :: PXFU(:,:)
REAL(KIND=JPRB)         ,INTENT(IN)               :: PPHY(1:YDCPG_OPTS%KLON,YDDESC%ILEVBOT:YDCPG_OPTS%KFLEVG)

IF (YDDESC%LACTIVE) THEN
  JLEV = YDDESC%ILEV
  
  IF (.NOT. LLRESET) THEN
    ZFPREC9(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA)=ZFPREC9(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA)+&
                   & PXFU(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA,JLEV)
  ENDIF
  
  ZFPREC0(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA)=ZFPREC0(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA)+&
                 & PPHY(YDCPG_BNDS%KIDIA:YDCPG_BNDS%KFDIA,YDDESC%INUMLEV(JLEV))
ENDIF

END SUBROUTINE 

END SUBROUTINE CPCFU
```

Without the check on DeferredTypeSymbol, it crashes.
Without the change on _map_unbound_dims, arguments of PXFU and PPHY do not get correctly mapped.
Without the recursive update on the expression map, the usage of YDDESC in PPHY indices does not get transformed.

mlange05:
Hi @JoeffreyLegaux, many, many thanks for those fixes and examples. They were very, very valuable! 

As you might have seen, I ended up integrating them directly into the main member-inliner PR (#130), as the recent rebase over master threw up a lot of conflicts (I also rebased out the post-fixes; I hope you don't mind). In any case, please have a look and see if your tests still work with this. I've create new smaller tests to trigger the behaviour you showed though, and I think I've captured it all. 

I also added you to the authors list, since you had already signed the CLA. Could you please indicate here, or on the original PR that you are happy with including your contributions this way? Once that is done and the main PR merged, I'll close this one here then.

And again, many thanks for those comments, pointers and fixes! Very cool stuff! :sunglasses:

JoeffreyLegaux:
Hi @mlange05, no problem in being added to the authors list.
I just pulled the latest naml-member-inlining branch, it still works perfect on the small test case however it crashes in a larger actual file when calling recursive_expression_map_update, I am investigating the issue.

Issue #123, opened 2023-08-11
Add unique ID to `Node`


awnawab:
Currently, the only mechanism for differentiating two string-identical but distinct nodes is the `.source` property. However this property is only instantiated for parsed nodes; if two identical but distinct nodes are generated in a transformation there is no method (apart from the object `id`) to tell them apart.

The above could be remedied by adding a `uid` attribute to the `Node` class. The `uid` property would be generated every time a node is created. This means that even copies of the same node would have different values of `uid`, unless overridden via a constructor argument.

reuterbal:
Importantly, this would give us the equivalent of a memory address-based hash that is controllable - thus allowing to restore original uid after pickling and an "in-place" rebuild via `_update`, while the default would be to generate a new uid when calling `clone`.

It might also provide opportunity to speed up certain Transformer passes, e.g. by matching nodes in maps via the uid rather than having to compute a hash over and over again.

joscao:
So I started playing around with that a little bit. Introduction of uid's seems easy, simply relying on the `uuid` package. The intended behavior described above, is extensively tested for all node types.

I will maybe have a look at the transformer passes but i currently doubt that my understand is sufficient to implement the utiliying the uid's.

mlange05:
As described in issue #520 I think the mechanism to differentiate between node equality (`==`) and identity (`is`) in the standard Python way are actually quite well established, so adding a UID would only work around the fact that the `.source` property should not be part of the equality definition in the first place. This design flaw implies that the equality behaviour between parsed and internally generated nodes differs, but to "fix this properly" ™ I would prefer we address in the original design issue outlined in #520. 

Hence closing this issue now, hoping to address this soon ™ .

PR #153, opened 2023-09-14
Add unique ID to `Node`

joscao:
Adds a uid to the `Node` baseclass via the `uuid` module. Refer Issue #123 

Current implementation utilizes `uuid4()` which creates a 

> universally unique identifier that is generated using random numbers, (https://www.uuidgenerator.net/version4)

e.g. `UUID('d5181261-c06e-4f68-82ed-8f8d4592511b')` 

Two implementation questions arise:
1.  Is this `UUID` type for loki's case a reasonable choice or would something simpler, i.e. a generator of incrementing integers, be enough?
2. Is there a reason that the `Comment` class was not calling the `Node` post init function? As far as git blame goes there once was a call to the base class _init__ but it was removed @mlange05 (d6fe4e1ecdefba47efe941dec4f82a81c4757670).

Current tests are:

-  iterating over all fortran files in project, parsing them to ir and then walking over all nodes and checking if they have a uuid and if it is unique (by collecting a list of used uuids)
- update/ clone tests: if update retains uid and clone creates new uid: for that all `Node` types are collected with dummy constructor arguments and checked if behavior is as described above

### DIFF
--- a/loki/ir.py
+++ b/loki/ir.py
@@ -1058,6 +1058,7 @@ class Comment(LeafNode, _CommentBase):
     """
 
     def __post_init__(self):
+        super().__post_init__()
         assert isinstance(self.text, str)
 
     def __repr__(self):
--- a/loki/ir.py
+++ b/loki/ir.py
@@ -16,6 +16,7 @@ from dataclasses import dataclass
 from functools import partial
 from itertools import chain
 from typing import Any, Tuple, Union
+from uuid import uuid4
 
 from pymbolic.primitives import Expression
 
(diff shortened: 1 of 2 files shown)


mlange05:
Ok, many apologies to @joscao for 
1/ leaving this to hang forever without any action
2/ closing it after all this time, due to no issue in the implementation itself.

To explain, I think this is very well implemented given the original problem discussion, but after playing with this recently, I actually found that the underlying problem goes much deeper and this would merely constitute a quick workaround for this deeper issue. The details are explained in detail in #520 and for the outlined reasons, I've now also closed the original issue #123. 

So, as discussed offline, for the same reasons as #123, I'm now closing this PR. However, the implementation is still great and many thanks again to @joscao for even picking this up in the first place! 🙏

Issue #103, opened 2023-06-16
Dangerous scenario allowed when a local driver variable matches the name of hoisted variable.


piotrows:
It may happen that the driver variable bears the same name as one of the hoisted kernel variables. If the dimensions match, the Fortran code will compile successfully. However, it will very probably result in the overwrite of the top level information, when hoisted variable is used  by the kernel. As this may lead to very subtle and hard to detect  bugs in larger programs, such situation needs to be detected at the transformation time. The following failing example illustrates scenario when a variable `t` is shared by the two kernels(which should be legal), but at the same time conflicts with the driver variable.

```# (C) Copyright 2018- ECMWF.
# This software is licensed under the terms of the Apache Licence Version 2.0
# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
# In applying this licence, ECMWF does not waive the privileges and immunities
# granted to it by virtue of its status as an intergovernmental organisation
# nor does it submit to any jurisdiction.

import pytest

from loki import (
    Subroutine, Dimension, fgen, Sourcefile, SubroutineItem
)
from transformations import (
    SCCDevectorTransformation, SCCHoistTransformation
)
from conftest import available_frontends
#pylint: disable=too-many-lines

@pytest.fixture(scope='module', name='horizontal')
def fixture_horizontal():
    return Dimension(name='horizontal', size='nlon', index='jl', bounds=('start', 'end'))


@pytest.fixture(scope='module', name='vertical')
def fixture_vertical():
    return Dimension(name='vertical', size='nz', index='jk')


@pytest.fixture(scope='module', name='blocking')
def fixture_blocking():
    return Dimension(name='blocking', size='nb', index='b')



@pytest.mark.parametrize('frontend', available_frontends())
def test_scc_hoist_multiple_kernels(frontend, horizontal, vertical, blocking):
    """
    Test hoisting of column temporaries to "driver" level.
    """

    fcode_driver = """
  SUBROUTINE column_driver(nlon, nz, q, nb)
    INTEGER, INTENT(IN)   :: nlon, nz, nb  ! Size of the horizontal and vertical
    REAL, INTENT(INOUT)   :: q(nlon,nz,nb)
    REAL :: T(NLON,NZ)
    INTEGER :: b, start, end

    start = 1
    end = nlon
    do b=1, nb
      call compute_column(start, end, nlon, nz, q(:,:,b))
      call compute_column2(start, end, nlon, nz, q(:,:,b))
    end do
[...]

Issue #101, opened 2023-06-14
Hoist transformation with multiple kernels allowing hoisted variable of a duplicated name but different size.


piotrows:
When two or more kernels are called from a driver, hoisting machinery works largely independently on each kernel. If the hoisted variable in  kernel 1 has the same name but different size than in kernel 2, this results in generation of invalid Fortran code (duplicated hoisted variable declaration). The following example detects such situation and raises an error in case there are two variables of the same name but of different size and/or dimensions. It is desired that either:
 the transformation stops, or
 the conflicting hoisted variables are renamed to e.g. indicate their kernel of origin.
```# (C) Copyright 2018- ECMWF.
# This software is licensed under the terms of the Apache Licence Version 2.0
# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
# In applying this licence, ECMWF does not waive the privileges and immunities
# granted to it by virtue of its status as an intergovernmental organisation
# nor does it submit to any jurisdiction.

import pytest

from loki import (
    Subroutine, Dimension, fgen, Sourcefile, SubroutineItem
)
from conftest import available_frontends
from transformations import (
    SCCDevectorTransformation, SCCHoistTransformation
)
#pylint: disable=too-many-lines

@pytest.fixture(scope='module', name='horizontal')
def fixture_horizontal():
    return Dimension(name='horizontal', size='nlon', index='jl', bounds=('start', 'end'))


@pytest.fixture(scope='module', name='vertical')
def fixture_vertical():
    return Dimension(name='vertical', size='nz', index='jk')


@pytest.fixture(scope='module', name='blocking')
def fixture_blocking():
    return Dimension(name='blocking', size='nb', index='b')



@pytest.mark.parametrize('frontend', available_frontends())
def test_scc_hoist_multiple_kernels(frontend, horizontal, vertical, blocking):
    """
    Test hoisting of column temporaries to "driver" level.
    """

    fcode_driver = """
  SUBROUTINE column_driver(nlon, nz, q, nb)
    INTEGER, INTENT(IN)   :: nlon, nz, nb  ! Size of the horizontal and vertical
    REAL, INTENT(INOUT)   :: q(nlon,nz,nb)
    INTEGER :: b, start, end

    start = 1
    end = nlon
    do b=1, nb
      call compute_column(start, end, nlon, nz, q(:,:,b))

      call compute_column2(start, end, nlon, nz, q(:,:,b))

[...]

PR #156, opened 2023-09-21
Loki-SCC: Adapt to use recursive hoisting transformation

mlange05:
The previous SCC-specific hoisting sub-transformation was conceptually very tedious and used many hacks/workarounds to facilitate interprocedural metadata passing without making use of the scheduler's metadata mechanism. This was error prone, had become hard to maintain and, most importantly, was not able to collect hoisted array variables across more than one level of kernel invocations. 

In this PR we therefore remove this sub-transformation entirely and create a specialised sub-class of the generic variable hoisting mechanism that was build by @MichaelSt98 as part of the SCC-CUF transformation chain. This is now able to recurse across multiple levels of kernel calls and can be applied after the bulk of the SCC base transformations (the previous one had a horrid co-dendency on certain sub-steps). This new hoisting transformation now also inserts the respective OpenACC annotations for device allocations of the hoisted arrays, and thus allows us to trim various pieces from the `SCCAnnotateTransformation`.

In some more detail:
* Changes the internalised `HoistVariableTransformation` to now defer argument variable remapping, as well as declaration handling to eventual sub-classes. It now also hands over tuple of hoisted variables, instead of calling the map per variable, to allow for bulk-insertion of variable declarations, etc.
* `HoistVariableTransformation` now also uses the `successor` items handed over from the scheduler to determine which calls to process, thus removign the need to pass the config option `disabled` to the transformation.
* Implements `SCCHoistTemporaryArraysTransformation` as an extension of the basic `HoistTemporaryArraysTransformation`, which adds the block dimension to the driver-side declarations, and remaps the call arguments between driver and kernel to use the block-index. As a corollary of this, we no longer hoist the vector-loop to the driver and can use the `SCCRevectorizeTransformation` with the recursive hoisting transformation.
* Removes the old `SCCHoistTransformation` and the relevant parts for annotating the driver-side hoisted array declarations in `SCCAnnotateTransformation`. This makes stuff much more focused and removes a lot of special casing (thanks @MichaelSt98 for making the base-hoisting transformation clever enough for this to "just work"!) 
* Adaptation of the SCC-hoisting tests to account for the propelry scheduler-dependent behaviour and lack of loop hoisting in the new transformation.
* Quite some refactoring in the `loki_transfrom.py` script to make the calling sequence of the sub-transformation bit more explicit.

Final note (@reuterbal ), the clear follow-on work from this is to extend the re-vctorizer to capture the driver-side loop-hoisting separately, which would allow us to pick-and-match the looping and temporary-handling styles. There are also plans to make the composition of the SCC components a bit more elegant, but this PR is already beefy enough, I fear. :wi


loki/transform/transform_hoist_variables.py
```diff
@@ -151,8 +146,7 @@ def transform_subroutine(self, routine, **kwargs):
             item.trafo_data[self._key]["to_hoist"] = []
             item.trafo_data[self._key]["hoist_variables"] = []
 
-        calls = [call for call in FindNodes(CallStatement).visit(routine.body) if call.name
-                 not in self.disable]
+        calls = [call for call in FindNodes(CallStatement).visit(routine.body) if call.name]
```
MichaelSt98, line 149: should be sufficient to just have `calls = FindNodes(CallStatement).visit(routine.body)`

tests/test_transform_hoist_variables.py
```diff
@@ -165,15 +166,15 @@ def test_hoist_disable(here, frontend, config):
...
+    disable = ("device1", "device2")
+    config['routine'].append({'name': 'kernel2', 'role': 'kernel', 'ignore': disable})
     proj = here/'sources/projHoist'
     scheduler = Scheduler(paths=[proj], config=config, seed_routines=['driver', 'another_driver'], frontend=frontend)
 
-    disable = ("device1", "device2")
-
     # Transformation: Analysis
-    scheduler.process(transformation=HoistVariablesAnalysis(disable=disable), reverse=True)
+    scheduler.process(transformation=HoistVariablesAnalysis(), reverse=True)
```
reuterbal, line 175: [No action] Coming to think of it, I believe we wouldn't actually need the separate passes for analysis and transformation but could directly apply the hoisting via a reverse traversal.
On the other hand, I wonder whether it wouldn't be beneficial to create a more generic analysis pass which collects just information about temporaries per routine and recursively for the call tree, and then use this for both hoisting and pool allocator. But that's a separate PR.

reuterbal (approved):
This is great and it's always fantastic to see when things converge like this. Nothing to flag on the implementation side beyond what @MichaelSt98 pointed out, looks all sane to me.

Two considerations that are tangentially relevant:
- I think for OpenMP offloading on Nvidia platforms we need the hoisting of the vector loop to the driver level, as there is (afaik still) the missing support for vector routines.
- We should consider generalising the analysis pass for temporaries into something that just yields the information about temporaries per routine and calls below, as well as their accumulated size. That would make it suitable for both, pool allocator and hoisting and any other future solutions for handling of temporaries.

reuterbal:
BTW: Have you checked that this resolves in particular the tests in #101 and #103?

mlange05:
> BTW: Have you checked that this resolves in particular the tests in #101 and #103?

Yes, I had a look and recreated one of them locally. Both issues should go away, as the internalised hoisting mechanism prefixes hoisted variable names with the kernel name to prevent shadowing. I've not added an explicit SCC test, as the internal tests should already cover this, but can be added easily if requested (please feel free to shout).

tests/test_transform_hoist_variables.py
mlange05: Agreed, a generic temporary analysis could be useful in more than one pass; good idea for the future, but beyond this PR.

loki/transform/transform_hoist_variables.py
mlange05: Good catch, fixed now (CI to confirm). Thanks!

reuterbal:
Thanks for confirming. #101 isn't strictly covered by the tests (as we don't have the precise situation) but given that we check the prefixing of hoisted variables it would be redundant indeed. GTG like this!

PR #157, opened 2023-09-22
Fix cudafor import in driver for SCC CUF

reuterbal:
Hotfix for failing transformation of CLOUDSC (see https://github.com/ecmwf-ifs/dwarf-p-cloudsc/actions/runs/6275881222/job/17044373035)

This is fallout from the recent removal of Transformation recursion, where previously the import was injected into the driver module. No tests had been verifying this. Now it is instead added to the driver subroutine and checked in the tests.


mlange05 (approved):
Good catch, nice fix! I can confirm this works, so GTG from me.

PR #158, opened 2023-09-25
Make `DataflowAnalysisAttacher/Detacher` true in-place transform

awnawab:
The `DataflowAnalysisAttacher/Detacher` transforms are in-place transforms that update the existing nodes without rebuilding them. However the `attach_dataflow_analysis` and `detach_dataflow_analysis` still assign a return value to `module_or_routine.spec/body`. This prevents the dataflow analysis from being applied to an `InternalNode`; the `.body` member of a frozen IR node cannot be reassigned to.

This PR attempts to remove the above limitation.

### DIFF
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -307,11 +307,11 @@ def attach_dataflow_analysis(module_or_routine):
         )
 
     if hasattr(module_or_routine, 'spec'):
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Many thanks, well spotted and neatly worked around. GTG!

mlange05 (approved):
Agree; nice catch, nice fix, nice tests! GTG from me. :shipit:

PR #159, opened 2023-09-29
Bugfix: Memory error in Statement functions test

reuterbal:
We had seen random failures of CI tests with a segfault, which typically went away when re-running the test. I have tracked this down to the statement functions test, where an uninitialised value was used for the index when writing to an array.

The test has been fixed and the problem hasn't resurfaced so far, hoping this was the root cause.


mlange05 (approved):
Looks good to me! Many thanks for tracking that down!!! GTG :pray:

PR #163, opened 2023-10-05
Call remove call and remove hook  transformations on contained subroutines

rolfhm:
Main will not call the remove_call and remove_hook on contained routines, resulting in code that does not compile. This PR adds calls on any member routines as well

### DIFF
--- a/transformations/tests/test_utility_routines.py
+++ b/transformations/tests/test_utility_routines.py
@@ -82,10 +82,29 @@ subroutine never_gonna_give
 
     if (dave) call abor1('[INLINE CONDITIONAL]')
 
+    call never_gonna_run_around()
+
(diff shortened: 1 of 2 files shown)


mlange05 (changes requested):
Hi @rolfhm, agreed that this makes sense, and is probably an oversight in the original implementation. 

However, would you mind adding a test to ensure that this does not break in the future?  I believe there's already a set of similar tests for these routines in `transformations/tests/test_utility_routines.py`.

rolfhm:
Added testing functionality now
I think there is the same issue for `associate` statements, but those require more than a line of code to deal with

mlange05 (approved):
Thanks for adding the test. This looks GTG now, once rebased. :shipit:

PR #166, opened 2023-10-06
CMake refactoring for loki_transform

reuterbal:
This refactors the CMake integration of Loki. As this is intended to incorporate the changes in #149 and resolve the outstanding issues in there, this is filed as a PR against that branch.

The following changes are applied:
- A single entry point `loki_transform` is now the default invocation of loki-transform.py, replacing `loki_transform_convert` and `loki_transform_command`. This is also used by `loki_transform_target` to apply the actual transformations.
- The various utility macros to translate arguments of the CMake functions to command-line arguments for loki-transform.py have been consolidated into two routines to parse options and arguments. This is possible because validating arguments is now left purely to the Python layer, i.e., the loki-transform.py script. However, the CMake functions calling these translation macros still only accept the explicitly provided list of options and arguments, thus we don't risk silently ignoring options, e.g. when they have a typo or use an outdated interface.
- `loki_transform_convert` is marked as deprecated but retained as an entry point for backwards-compatibility, since this is used by regression tests. Once the regression test repositories have been brought forward, this entry point can be completely dropped
- `loki_transform_plan` will remain as a separate entry point since this invokes loki at configuration (and not at build time).
- `loki_transform_transpile` remains as well, since the interface in the CLI script is very different from convert. However, I have applied the same renaming of options there to align this further, in preparation for future integration into the convert command.

Overall, the loki_transform.cmake script has shrunk by almost 25% (mostly by moving purely utility routines to a separate file), but can be further reduced when the deprecated `convert` and `transpile` entry points are removed.

Tested locally for CLOUDSC, waiting for CI now to confirm this works also on ecWAM. 
@mlange05: Please test this for ecphys!

### DIFF
--- a/cmake/loki_transform.cmake
+++ b/cmake/loki_transform.cmake
@@ -5,188 +5,32 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-macro( _loki_transform_parse_args _func_name )
-
-    if( _PAR_UNPARSED_ARGUMENTS )
-        ecbuild_critical( "Unknown keywords given to ${_func_name}(): \"${_PAR_UNPARSED_ARGUMENTS}\"" )
-    endif()
-
-    if( _PAR_DIRECTIVE )
-        list( APPEND _ARGS --directive ${_PAR_DIRECTIVE} )
-    endif()
-
-    if( NOT _PAR_FRONTEND )
-        ecbuild_critical( "No FRONTEND specified for ${_func_name}()" )
-    endif()
-    list( APPEND _ARGS --frontend ${_PAR_FRONTEND} )
-
-    if( _PAR_HEADER )
-        if ( _PAR_HEADERS )
-            ecbuild_critical( "Both HEADERS and HEADER given to ${_func_name}(): Please use HEADERS only" )
-        else()
-            ecbuild_warn( "${_func_name}(): HEADER is deprecated, please use HEADERS" )
-            set( _PAR_HEADERS ${_PAR_HEADER} )
-        endif()
-    endif()
-
-    if( _PAR_HEADERS )
-        foreach( _HEADER ${_PAR_HEADERS} )
-            list( APPEND _ARGS --header ${_HEADER} )
-        endforeach()
-    endif()
-
-    if( _PAR_INCLUDE )
-        if ( _PAR_INCLUDES )
-            ecbuild_critical( "Both INCLUDES and INCLUDE given to ${_func_name}(): Please use INCLUDES only" )
-        else()
-            ecbuild_warn( "${_func_name}(): INCLUDE is deprecated, please use INCLUDES" )
-            set( _PAR_INCLUDES ${_PAR_INCLUDE} )
-        endif()
-    endif()
-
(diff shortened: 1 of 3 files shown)


reuterbal:
PS: This also fixes an unrelated bug related to the member inliner, which wasn't applied even when `--inline-members` was specified. This was due to the fact that the option wasn't forwarded to the SCCBaseTransformation.

cmake/loki_transform.cmake
```diff
@@ -5,188 +5,32 @@
...
-#       [DEFINITIONS <define1> [<define2> ...]]
-#       [OMNI_INCLUDE <omni-inc1> [<omni-inc2> ...]]
-#       [XMOD <xmod-dir1> [<xmod-dir2> ...]]
-#       [REMOVE_OPENMP] [DATA_OFFLOAD] [GLOBAL_VAR_OFFLOAD]
-#       [TRIM_VECTOR_SECTIONS] [REMOVE_DERIVED_ARGS]
-#       [INLINE_MEMBERS] [DERIVE_ARGUMENT_ARRAY_SHAPE]
 #   )
 #
-# Call ``loki-transform.py convert ...`` with the provided arguments.
+# Call ``loki-transform.py <ecphys|...> ...`` with the provided arguments.
```
mlange05, line 33: I think ecphys is basically deprecated now too, no?

cmake/loki_transform.cmake
```diff
@@ -198,320 +42,59 @@ endmacro()
...
+        COMMAND ${_LOKI_TRANSFORM} ${_ARGS}
+        DEPENDS ${_PAR_DEPENDS} ${_LOKI_TRANSFORM_DEPENDENCY}
+        COMMENT "[Loki] Pre-processing: command=${_PAR_COMMAND} mode=${_PAR_MODE} directive=${_PAR_DIRECTIVE} frontend=${_PAR_FRONTEND}"
     )
 
 endfunction()
 
-##############################################################################
-
-macro( _loki_transform_parse_target_args _func_name )
```
mlange05, line 483: EC-phys regression tells me that there's still a dangling call to this, but it seems to be removed entirely. Was this intended to be moved to helpers, or is it entirely redundant now? In the latter case `loki_transform_target` needs a small update.

mlange05 (changes requested):
This looks fantastic - many, many thanks for taking this over! I've flagged one showstopper that breaks regression (yes, we need to start using the "target" entry point in CLOUDSC!); and flagged a minor outdated comment. Should be good once the target mode is fixes and regression runs!

cmake/loki_transform.cmake
reuterbal: Thanks, good spot. Yeah, ecphys is already gone. Will change to `convert`.

cmake/loki_transform.cmake
reuterbal: Thanks for testing and flagging! Yes, we need regression testing of this entry point asap. Please have another go, I have removed the dangling call which shouldn't have provided more than argument checking in this instance.

mlange05 (approved):
Ok, again, many, many thanks for this (beyond initial scope, but dearly needed!). ECphys regression ran fine now, so GTG from me. :shipit:

PR #169, opened 2023-10-12
Fix vector section trimming in driver loop

awnawab:
This PR is a one-line fix to the vector section trimming within the driver loop. The corresponding test is parametrised to show the effect of vector section trimming.


reuterbal (approved):
Thanks, this is great - fills in a test coverage gap and restores the functionality. GTG from my side but I let one of the other SCC experts agree.

mlange05 (approved):
I agree, neat and nicely tested. GTG :shipit:

PR #170, opened 2023-10-12
TransformInline: Fix rescoping in expression substitution

mlange05:
A small fix the the member inlining utility and accompanying test.

The problem was the wrong use of symbol substitution, which we now face to rescope and rebuild when replicating the body of the child routine to the parent. This would otherwise in-place update associate statements, as they are type definitions, which creates issues for replicated calls, as the same object is used for separate substitutions.

Many thanks to @rolfhm for finding, flagging and triaging the problem and providing the accompanying test case.


reuterbal (approved):
Thanks, nice find and fix. This is really subtle... GTG

PR #172, opened 2023-10-13
InlineMember: rename duplicate locals in the body

rolfhm:
The shadow mapper will only find variables in the declaration and leaves the body unchanged.

I do not understand the need for the `[dl.name for dl in duplicate_locals]` construct though. I would have thought that a `v in duplicate_locals` or `v.name in duplicate_locals` would be sufficient.

### DIFF
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -239,7 +239,12 @@ def inline_member_routine(routine, member):
         {v: v.clone(name=f'{member.name}_{v.name}') for v in duplicate_locals}
     )
     member.spec = shadow_mapper.visit(member.spec)
-    member.body = shadow_mapper.visit(member.body)
+
+    var_map = {}
+    for v in FindVariables(unique=False).visit(member.body):
+        if v.name in [dl.name for dl in duplicate_locals]:
+            var_map[v] = v.clone(name=f'{member.name}_{v.name}')
+    member.body = SubstituteExpressions(var_map).visit(member.body)
 
     # Get local variable declarations and hoist them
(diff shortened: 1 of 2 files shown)


loki/transform/transform_inline.py
```diff
@@ -239,7 +239,12 @@ def _map_unbound_dims(var, val):
         {v: v.clone(name=f'{member.name}_{v.name}') for v in duplicate_locals}
     )
     member.spec = shadow_mapper.visit(member.spec)
-    member.body = shadow_mapper.visit(member.body)
+
+    var_map = {}
+    for v in FindVariables(unique=False).visit(member.body):
+        if v.name in [dl.name for dl in duplicate_locals]:
+            var_map[v] = v.clone(name=f'{member.name}_{v.name}')
```
reuterbal, line 246: The reason that you need to do that list comprehension is that the (string) comparison takes into account dimension expressions of variables. If you want to make it slightly less error prone (because you would likely run into case-sensitivity issues with the current implementation): 

```suggestion
    duplicate_locals_names = {dl.name.lower() for dl in duplicate_locals}
    for v in FindVariables(unique=False).visit(member.body):
        if v.name.lower() in duplicate_locals_names:
            var_map[v] = v.clone(name=f'{member.name}_{v.name}')
```

reuterbal (changes requested):
Thanks for finding and fixing this. I have a small remark regarding the implementation, otherwise fine from my side.

reuterbal (approved):
Thanks for taking care of this. Looks good to go!

PR #175, opened 2023-10-17
Fix deep-cloning of subroutiens and modules (fix #174)

### LINKED ISSUE #174: Node transformation applied on a subroutine's clone also impact original routine (possibly caused by ASSOCIATE)
When applying this script :

```
from loki import Frontend, Sourcefile, FindNodes, Node, Subroutine, Transformer
from loki.ir import Section, Comment, VariableDeclaration, Pragma, Assignment, CallStatement
from loki.expression.symbols import DeferredTypeSymbol

for file in ['./routine.F90' ]:
    source = Sourcefile.from_file(file, frontend=Frontend.FP)

    for routine in source.subroutines:
        new_routine = routine.clone()

        map_nodes={}
        for assign in FindNodes(Assignment).visit(new_routine.body):
            map_nodes[assign] = CallStatement(name = DeferredTypeSymbol(name='TESTCALL'), arguments=(assign.lhs,), scope=new_routine)
        new_routine.body = Transformer(map_nodes).visit(new_routine.body)
            
        f = open('./new_routine.F90', 'w')
        f.write(new_routine.to_fortran())
        f.write('\n') #Add eol so vim doesn't complain
        f.close()

        f = open('./ref_routine.F90', 'w')
        f.write(routine.to_fortran())
        f.write('\n') #Add eol so vim doesn't complain
        f.close()
```

To the following routine :

```
SUBROUTINE MYROUTINE(YDCFU, YDCPG_OPTS)

USE PARKIND1                , ONLY : JPIM, JPRB
USE YOMCFU                  , ONLY : TCFU
USE CPG_OPTS_TYPE_MOD       , ONLY : CPG_BNDS_TYPE, CPG_OPTS_TYPE

IMPLICIT NONE

TYPE(CPG_OPTS_TYPE)     ,INTENT(IN)               :: YDCPG_OPTS

REAL(KIND=JPRB) :: ZFPREC9(YDCPG_OPTS%KLON) ! past cumulated flux of total precipitations
REAL(KIND=JPRB) :: ZFPREC0(YDCPG_OPTS%KLON) ! instantaneous flux of total precipitations
REAL(KIND=JPRB) :: ZSQPRE(YDCPG_OPTS%KLON)  ! instantaneous flux of total squared precipitations

ZFPREC9(:)=0.0_JPRB

ASSOCIATE(LREACFU=>YDCFU%LREACFU, NFRRC=>YDCFU%NFRRC)

IF (YDCFU%YFDUTP%LACTIVE) THEN
  ZFPREC0(:)=0.0_JPRB
ELSE
  ZSQPRE(:)=0.0_JPRB
END IF

END ASSOCIATE

END SUBROUTINE MYROUTINE
```

The outputted transformed cloned routine is correct (all Assignment nodes replaced by CallStatements).
However, when outputting back the original routine the two assignements in the conditional block are also transformed.

This could be tied to the ASSOCIATE block, since when it is commented the original routine remains correct in the output file.

mlange05:
As the title implies, we were indeed not rebuilding nested `ScopedNode` objects in the duplication `Transformer` used for deep-copying (cloning) of  `Subroutine`/`Module` objects. This PR fixes this and adds tests that include nested `ScopedNode` objects (`Associate` for subroutine and `TypeDef` for modules) in the duplicated IR. 

This should fix issue #174.

Many thanks to @JoeffreyLegaux for reporting and providing the reproducer for the subroutine case. I've slightly shortened the example code, but in essence it's the same problem. :wink: Please do confirm that this fixes things as you'd expect.

### DIFF
--- a/loki/program_unit.py
+++ b/loki/program_unit.py
@@ -290,12 +290,13 @@ class ProgramUnit(Scope):
         kwargs.setdefault('incomplete', self._incomplete)
 
         # Rebuild IRs
+        rebuild = Transformer({}, rebuild_scopes=True)
         if 'docstring' in kwargs:
-            kwargs['docstring'] = Transformer({}).visit(kwargs['docstring'])
+            kwargs['docstring'] = rebuild.visit(kwargs['docstring'])
         if 'spec' in kwargs:
-            kwargs['spec'] = Transformer({}).visit(kwargs['spec'])
+            kwargs['spec'] = rebuild.visit(kwargs['spec'])
         if 'contains' in kwargs:
-            kwargs['contains'] = Transformer({}).visit(kwargs['contains'])
+            kwargs['contains'] = rebuild.visit(kwargs['contains'])
 
         # Rescope symbols if not explicitly disabled
         kwargs.setdefault('rescope_symbols', True)
--- a/loki/subroutine.py
+++ b/loki/subroutine.py
@@ -288,7 +288,7 @@ class Subroutine(ProgramUnit):
 
         # Rebuild body (other IR components are taken care of in super class)
         if 'body' in kwargs:
-            kwargs['body'] = Transformer({}).visit(kwargs['body'])
+            kwargs['body'] = Transformer({}, rebuild_scopes=True).visit(kwargs['body'])
 
         # Escalate to parent class
         return super().clone(**kwargs)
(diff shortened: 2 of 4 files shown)


reuterbal (approved):
Thanks for fixing this and providing tests for it! That was a subtle oversight, many thanks to @JoeffreyLegaux for flagging this.

JoeffreyLegaux:
Hi @mlange05, I just tested this branch on a large actual case, and everything seems to work fine. Thanks for the quick fix !

PR #177, opened 2023-10-23
Sourcefile does not have a filepath

joscao:
- sourcefile has a source not a filepath

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -72,6 +72,8 @@ def jit_compile(source, filepath=None, objname=None):
     """
     if isinstance(source, Sourcefile):
         filepath = source.path if filepath is None else Path(filepath)
+        if filepath is None:
+            filepath = Path(gettempdir()/filehash(source, prefix='', suffix='.f90'))
         source.write(path=filepath)
     else:
         source = fgen(source)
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -71,7 +71,7 @@ def jit_compile(source, filepath=None, objname=None):
         Return a specific object (module or subroutine) in :attr:`source`
     """
     if isinstance(source, Sourcefile):
-        filepath = source.filepath if filepath is None else Path(filepath)
+        filepath = source.source if filepath is None else Path(filepath)
```
reuterbal, line 74: I disagree, `Sourcefile.source` is an object of type `Source` rather than a path. However, the current implementation is also wrong, it should be using `source.path` rather than `source.filepath`: https://github.com/ecmwf-ifs/loki/blob/af8ec9fc47bc529f914969d9e88ed898dc788ebf/loki/sourcefile.py#L60

reuterbal (changes requested):
Thanks, that's a nice find. Any chance we could test this behaviour in a test, since it's clearly not covered at the moment?

Also, I think the property to be used is `Sourcefile.path` rather than `source`.

tests/conftest.py
joscao: Used path instead of source. #[bc8f27b](https://github.com/ecmwf-ifs/loki/pull/177/commits/bc8f27be4a03c782de64a0c834e3cae530b55d16)

reuterbal (approved):
Thanks, also for the added handling of empty filepaths. Looks great now!

PR #178, opened 2023-10-23
Extract/Improve Polyhedron class

joscao:
- [x] extract polyhedron class from `loki/transform/transform_loop.py` into `loki/analyse/util_polyhedron.py`
- [x] extract testing from `tests/test_transform_loop.py ` into `tests/test_util_polyhedron.py`
- [x] add  `__repr__ ` to Polyhedron class
- [x] add utility function for construction `from_nested_loops`
- [x] add `is_empty` functionality


joscao:
Code-checks is complaining about a cyclic import, but in the `tests/test_build.py` which I cant seem to reproduce (there linter is running fine) on my local machine. Could somebody help me out please?


```
Error: tests/test_build/test_build.py:1:0: R0401: Cyclic import (loki -> loki.lint -> loki.lint.linter -> loki.transform -> loki.transform.transform_loop -> loki.analyse.util_polyhedron) (cyclic-import)
Error: tests/test_build/test_build.py:1:0: R0401: Cyclic import (loki -> loki.transform -> loki.transform.transform_loop -> loki.analyse.util_polyhedron) (cyclic-import)
```

loki/analyse/util_polyhedron.py
```diff
@@ -0,0 +1,328 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import numpy as np
+
+from loki import TypedSymbol, Loop
```
reuterbal, line 10: Thanks, I haven't had a chance, yet, to read and review this properly - will try to do this today. But just as a quick response to your question about the cyclic import: This is the place where this is introduced.

Within Loki, you should use fully-qualified imports  wherever possible to avoid these (i.e., here use `sym.TypedSymbol` because the module is already imported, and `from loki.ir import Loop` for the loop node). The reason is that the `__init__.py` imports the entire Loki namespace for ease of use from the outside.

joscao: Thank you. Fixed in 78422d3

loki/analyse/util_polyhedron.py
```diff
@@ -0,0 +1,328 @@
...
+    P = { x=[x1,...,xd]^T c R^d | Ax <= b }
+    ```
+    with n-by-d matrix `A` and d-dimensional right hand side `b`.
+
+    In loop transformations, polyhedrons are used to represent iteration spaces of
+    d-deep loop nests.
+
+    :param np.array A: the representation matrix A.
+    :param np.array b: the right hand-side vector b.
+    :param list variables: list of variables representing the dimensions in the polyhedron.
```
reuterbal, line 39: If you don't mind, change this to the up-to-date docstring format:

```suggestion
    Parameters
    ----------
    A : numpy.array
        The representation matrix A.
    b : numpy.array
        The right hand-side vector b.
    variables : list, optional
        List of variables representing the dimensions in the polyhedron.
```

loki/analyse/util_polyhedron.py
```diff
@@ -0,0 +1,328 @@
...
+            lhs, rhs = cls.generate_entries_for_lower_bound(
+                loop_range.stop, variable_names, j
+            )
+            A[2 * i + 1, :] = -lhs
+            b[2 * i + 1] = -rhs
+
+        return cls(A, b, variables)
+
+    @classmethod
+    def from_nested_loops(cls, nested_loops: list[Loop]):
```
reuterbal, line 322: Could we add a dedicated test for the `from_nested_loops` factory method, that verifies that the constructed polyhedron looks correct for a few simple test cases?

reuterbal, line 322: The type hint problem in Python 3.8 should go away by using
```python
from typing import List
```
and
```List[Loop]```

reuterbal (changes requested):
Thanks, this looks great. We need to take care of resolving the type hint problem with 3.8, and it would be wise to add another test for the construction from nested loops that ensures the resulting polyhedron looks fine.

loki/analyse/util_polyhedron.py
joscao: eb8bb06d5ac8af02e35d02a8b9386c1bb7a8eb9b

joscao: eb8bb06d5ac8af02e35d02a8b9386c1bb7a8eb9b

loki/analyse/util_polyhedron.py
joscao: eb8bb06d5ac8af02e35d02a8b9386c1bb7a8eb9b and eb8bb06d5ac8af02e35d02a8b9386c1bb7a8eb9b

reuterbal (approved):
Fantastic, many thanks. I think this is ready for merging.

PR #180, opened 2023-10-24
Array data dependence analysis

joscao:
Provides a array dependency analysis as* described by "Compilers: Principles, Techniques, and Tools".

Requires functionality of #179 and #178.

Implemented features:

-  `has_data_dependency`in `analyse_array_data_dependency_detection.py` allowing to determine if two array accesses in different loops exhibit a data dependency
- a notebook `06_array_data_dependency_detection.ipynb` showing a full integration (from source code, manual selecting of arrays, to determining if there exists a dependency)
- a test `test_analyse_array_data_dependency_detection.py` testing 8 different test cases mostly taken from the Dragon book
 
Current Problems/Things I need help with:

- [ ] How do I integrate the requirement of `ortools` for the solving of the integer linear problem into `loki`? 
- [ ] I have written an additional latex document, for exact formulation of linear algebra equations, since the Dragon is lacking them for this chapter, I will append it, how should it be preserved? [array_data_dependency_description.pdf](https://github.com/ecmwf-ifs/loki/files/13120680/array_data_dependency_description.pdf)

Asterisk:
The following parts of the Dragon book are implemented:
- "11.6.1 Denition of Data Dependence of Array Accesses", i.e. the full linear algebra description of the problem (iteration space, array access function)
- "11.6.3 The GCD Test", i.e. a Gaussian Elimination algorithm specialized for Diophantine equations (i.e. pure integer matrix vector equation) and therefore the substitution of the result into the inequalities as described by equation (5) of the array_data_dependency_description mentioned above.
- Independent-Variables Test from "11.6.4 Heuristics for Solving Integer Linear Program"
- The Acyclic Test, Loop-Residue-Test, Memoization and the full Branch and bound solution to integer linear programming problems were not implemented, to ensure full usability a general integer linear program solver provided by `ortools` is used, therfore full usability is ensured, to the best of my abilities.


reuterbal:
Again, just a quick response to your immediate questions:

To integrate ortools, I suggest to add them to `pyproject.toml` as an optional dependency for now, similar to the examples:
https://github.com/ecmwf-ifs/loki/blob/af8ec9fc47bc529f914969d9e88ed898dc788ebf/pyproject.toml#L62-L65

Call the option `dataflow_analysis` or something similarly meaningful. The import of ortools should then be wrapped with a safeguard similar to open-fortran-parser and clean exception if it isn't available.
https://github.com/ecmwf-ifs/loki/blob/af8ec9fc47bc529f914969d9e88ed898dc788ebf/loki/frontend/ofp.py#L15-L21

For the document, the best would probably be to convert this to a Sphinx document and integrate it into the documentation. But we can take care of that, so keeping the PDF like this is fine for now.

joscao:
Since #178 and #179 are now stable, this branch has been rebased and most issues fixed.

Not functional (require help):
- [ ] ortools installation (cant seem to figure it out how to add it to install)
- [ ] test of notebook "06_..." complains about ungrouped-imports, but it is a notebook? that shouldnt matter there, at least to my understand

Enhancements:
- [ ] convert function documentation to numpy style

joscao:
"Example 11.35 from Compilers: Principles, Techniques, and Tools" should return false due to indepdent variable test, that is not the case. Whats going on?

PR #181, opened 2023-10-25
Utility transformation for creating standalone subroutines from contained subroutines

skarppinen:
This PR adds a function named `lift_contained_subroutines` (for a lack of a better name) which processes the contained subroutines of a `loki.Subroutine` such that they are converted to "stand alone subroutines" that have no global dependencies. To do this:
1. all global bindings from the point of view of the contained subroutine(s) are introduced as imports or dummy arguments to the contained subroutine(s).
2. all calls to the contained subroutines in the parent are modified accordingly.

To understand the basic idea of what the function does, consider the next example. 
The following subroutine (with a contained subroutine "inner" in the CONTAINS block)
```
subroutine outer()
    integer :: y
    integer :: o
    o = 0
    y = 1
    call inner(o)
    contains
    subroutine inner(o)
       integer, intent(inout) :: o
       integer :: x
       x = 4
       o = x + y ! Note, 'y' is "global" here!
    end subroutine inner
end subroutine outer
```
is transformed into a `list` of `loki.Subroutine`s, where the transformed "parent" (i.e "outer") comes first:
```
subroutine outer()
    integer :: y
    integer :: o
    o = 0
    y = 1
    call inner(o, y) ! 'y' now passed as argument.
    contains
end subroutine outer
```
and the transformed children (i.e in this case only "inner") come next:
```
subroutine inner(o, y)
       integer, intent(inout) :: o
       integer, intent(inout) :: y
       integer :: x
       x = 4
       o = x + y ! Note, 'y' is no longer "global"
end subroutine inner
```
Naturally, multiple contained subroutines are supported as well.

Some remarks:
- The current implementation always sets the intent of "resolved variables" (i.e 'y' in the above example) as "inout", unless the variable to be resolved has an explicitly specified intent in the parent routine ('outer').
- If a variable is to be resolved that lacks a definition in the parent scope, an exception is thrown.
- If a variable to be resolved is a member of a derived type (say 'a%b'), the whole derived type (i.e 'a') is introduced as a dummy argument to the contained subroutine. 
- If a variable to be resolved is defined via an import, the import is added to the contained subroutine.
- If the definition of a variable to be resolved depends on imports or other variables, the necessary imports are added to the contained subroutine, and the variables are added as arguments (if need be). These cases occur for example, when the variable to be resolved (below 'x') is defined for example by:
```
USE parkind1, only: jprb, jpim
INTEGER(KIND=jpim) :: klon
REAL(KIND=jprb) :: x(klon) ! to resolve 'x', import 'jprb' and introduce 'klon' as argument.
```
- `resolve_associates` is called for each contained subroutine.

The tests cover the above cases (excluding resolving associates) and some other things (see docstring for each test). As additional "field testing", I have tested this (as a replacement for inlining) in the ACRANEB2 dwarf, where as part of an SCC pipeline this transform handles all Fortran constructs found there and produces code that compiles and provides correct results on NVIDIA GPUs.


mlange05 (commented):
Apologies for taking so long to get back to this. The utility certainly looks very useful, and a great contribution to have in our housekeeping arsenal. As @reuterbal already pointed out, we'll need appropriate copyright headers and and some of the outer interfacing could be slightly improved, so that we can use this in the future in a more composable way with other transformations. The test coverage, however, seems very thorough which is really great to see! :smile: 

I also noted that the utility routine itself does many little things in that main loop, which could indicate that breaking it down further might make the corner-cases more manageable. I'd like to have another quick look once most of the main issues have been addressed (and I promise to be quicker this time :wink: ), so leaving this in "comment" state for now. But overall this is looking very nice!

skarppinen:
Thanks @reuterbal and @mlange05 for your valuable comments. Based on a quick look I think I agree with all of them. (many of them I could anticipate, but decided to post this early without too much polishing)

I am writing this message just to acknowledge that I've seen your comments. However, I am quite busy at the moment, and it may well be until next month that I have time to address them.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
+from loki import (
```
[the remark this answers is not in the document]
skarppinen: Thanks. I did this.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+    Subroutine, FindNodes, FindVariables,
+    CallStatement, VariableDeclaration, Import,
+    FindInlineCalls, Transformer, CaseInsensitiveDict,
+    DerivedType
+)
+from functools import reduce
+from loki.transform import resolve_associates
+
+def lift_contained_subroutines(routine):
```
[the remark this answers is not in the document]
skarppinen: I introduced some minor modifications to also make it work for functions. The only difference is that for functions, need to transform `InlineCall`s in the parent instead of `CallStatement`s. I renamed everything to `procedures` instead of routines as well.I also added a two tests with functions. The test coverage for functions is not as good. However, it seems that functions are implemented using `Subroutine` (i.e they are very similar in Loki) so maybe this is not a problem? I can improve the tests if need be, the extreme of course being that each test is replicated such that the inner procedure is a function.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    This transform creates "standalone" `Subroutine`s from the contained subroutines of `routine`.
```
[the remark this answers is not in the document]
skarppinen: Thanks, fixed this.

loki/transform/transform_lift_contained_subroutines.py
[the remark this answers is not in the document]
skarppinen: Done as well.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+
+    def find_var_defining_import(varname: str, impos):
+        for impo in impos: 
+            varnames_in_import = [var.name.lower() for var in FindVariables().visit(impo)]
+            if varname.lower() in varnames_in_import:
+                return impo.clone()
+        raise Exception(f"variable '{varname}' was not found from the specified import objects.")
+
+    routine = routine.clone()
```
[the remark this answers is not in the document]
skarppinen: I really like the name "extract" a lot more than "lift". Thanks for this. Now this function also modifies its argument, as indeed seems to be the preferred approach in Loki. 

I think @mlange05's comment about `Sourcefile` / `Module` transforming is also very good, and I agree that this should be wrapped into a `Transformation`. However, I did not do it yet, but could do it together with your next comments about the implementation.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Get all inner routines and resolve all associate statements in them.
+    # This is strictly not needed, but is done to simplify the logic in later steps.
+    inner_routines = []
+    for r in routine.contains.body:
+        if isinstance(r, Subroutine):
+            newinner = r.clone()
+            resolve_associates(newinner)
+            inner_routines.append(newinner)
+    assert len(inner_routines) > 0
```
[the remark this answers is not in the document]
skarppinen: Thanks, indeed it can. Actually this bit got refactored away when I addressed your other comments.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+    # Get all calls.
+    all_calls = FindNodes(CallStatement).visit(routine.body)
+    call_transmap = {}
+
+    # Construct a list of routines returned from this function.
+    routines = []
+
+    outer_decls = FindNodes(VariableDeclaration).visit(routine.spec)
+
+    for inner in inner_routines:
```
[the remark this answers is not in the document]
skarppinen: Yes, very good suggestion. I did exactly this, now there is `extract_contained_procedure` and `extract_contained_procedures`, as you suggest.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+        # following loop may add new items that to `gvar_names` that need to be resolved. 
+        for gn in gvar_names:  
+            same_named_globals = [var.clone() for var in outer_spec_vars if var.name.lower() == gn]
+            if len(same_named_globals) != 1:
+                # If the code crashes to the next assert, there is a bug in the implementation of this function.
+                assert len(same_named_globals) == 0 
+                msg = f"Could not resolve the symbol '{gn}' "\
+                f"in the contained subroutine '{inner.name}'. "\
+                f"The symbol '{gn}' is undefined in the parent subroutine '{routine.name}'."
+                raise Exception(msg)
```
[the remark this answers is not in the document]
skarppinen: Indeed it is. I went with `RuntimeError`. However, maybe there could be something like a custom error `SymbolResolutionError` or something. I'd imagine this kind of an error pops up in other places in Loki as well, or? I did not, however, implement this, because it is not strictly speaking part of this PR.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+        gvar_names = set()
+        for var in all_vars:
+            # The next while loop is in case `var` is a member of a derived type. 
+            # For example for variable named `a%b%c`, 
+            # the global variable we want is `a`, not `a%b%c` itself.
+            varname = var.name.lower()
+            vartmp = var.clone()
+            while vartmp.parent is not None:
+                varname = vartmp.parent.name.lower()
+                vartmp = vartmp.parent
```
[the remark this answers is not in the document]
skarppinen: Perhaps you meant something different with this, because that change would break the code.
The idea here is to have `varname` point to the "root" of a derived type (in case `var` IS a derived type).
In other words, if `var` was something like `a%b%c`, the while loop walks the tree backwards to get `varname` = "a". Cannot just look at a single parent, therefore. I suspect this might be refactored away with your next comments, however..

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+        # 1. Declared non-array variables in spec.
+        # 2. Array-variables, EXCLUDING any variables that only exist in array dimensions.
+        # NOTE: Such dimension variables will be in group 1, if they are actually defined.
+        # 3. Imported variables. NOTE: Here collecting also imports that correspond to functions,
+        # but they are removed later. 
+        ldecls = FindNodes(VariableDeclaration).visit(inner.spec)
+        limports_vars = FindVariables().visit(FindNodes(Import).visit(inner.spec))
+        lvars = list(reduce(lambda x, y: x + y, (decl.symbols for decl in ldecls))) 
+        lvars = set(var.clone() for var in lvars + list(limports_vars))
+        lvar_names = [var.name.lower() for var in lvars]
```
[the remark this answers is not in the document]
skarppinen: Yes, thanks a lot. I did not really know about these fields. However, I noted that those variables were actually artifacts from some earlier implementation, not really used anymore in the code.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Find all variables (names), including also:
+        # 1. Declared non-array and array variables in spec.
+        # 2. Variables in body.
+        # 3. Variable kinds.
+        # 4. Array dimension variables.
+        all_vars_body = FindVariables().visit(inner.body)
+        all_vars_spec = FindVariables().visit(inner.spec)
+        all_vars = set(var.clone() for var in all_vars_body.union(all_vars_spec))
```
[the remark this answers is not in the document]
skarppinen: At the time of writing I recall there being some bugs with `clone`, so I guess this was just paranoia. :D 
I removed all empty `clone()` calls.

tests/test_lift_contained_subroutines.py
```diff
@@ -0,0 +1,356 @@
+from loki import (
+    Sourcefile, FindVariables, FindNodes, CallStatement,
+    Import,
+)
+from loki.transform import (
+    lift_contained_subroutines
+)
+import pytest
+
+def test_basic_scalar():
```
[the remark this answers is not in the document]
skarppinen: Ok, I renamed the tests, thanks.

loki/transform/transform_lift_contained_subroutines.py
```diff
@@ -0,0 +1,212 @@
...
+           integer, intent(inout) :: o
+           integer, intent(inout) :: y
+           integer :: x
+           x = 4
+           o = x + y ! Note, 'y' is no longer "global"
+    end subroutine inner
+    ```
+    """
+
+    def find_var_defining_import(varname: str, impos):
```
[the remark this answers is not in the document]
skarppinen: Indeed, thanks for the tip. Now using `import_map`.

skarppinen:
Thanks again for the comments. I think I addressed most of them and this is looking a lot better already.
However, I think the symbol resolution could be improved, since it is currently based on using strings. Should I be using `TypedSymbol`s or something like that?
 
I did not yet fulfil @mlange05's wish to make a `Transformation` out of this, but I think this is a good idea. However, I suggest a course of action to first fix the symbol resolution to make this more "Lokian", and after that, wrap everything into a `Transformation` that can handle `Module`s and `Sourcefile`s too.

skarppinen:
I made a reimplementation of the proposed functionality that I think is a lot better. Instead of referring to variables using strings, I now use the variable objects and scope information provided by Loki to do the symbol/variable resolution. I also removed `resolve_associates` as it seems it is not needed. Furthermore, I added a few new tests to handle cornercases I had overlooked before.

loki/transform/transform_lift_contained_subroutines.py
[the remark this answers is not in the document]
reuterbal: In fact, this is exactly what I understood and proposed. Note the difference between `vartmp.parent` (which will be None if it isn't a derived type member) and `vartmp.parents[0]`, which yields all parents and take the outermost - which is exactly the chasing you implemented.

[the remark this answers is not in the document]
skarppinen: Apologies, I must have read your code in haste. (somehow I missed that it was `parents[0]`, not `parent`, as you explained) I see your point now. Still, I did run the tests with your suggestion, and recall getting an error (maybe it was `IndexError` for `parents` being empty in some special case), or then it was something unrelated.

However, note that this bit got removed in my latest refactoring, and there is no need to "walk to the parent" anymore, since it appears I can just drop all variables that have a parent.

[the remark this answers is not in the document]
reuterbal: Right, sorry, I had been making my way through the comments only and hadn't realized this bit had gone entirely. Ignore my ramblings then ;-) I'll try to get through the rest of it today.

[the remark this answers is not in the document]
skarppinen: It is good you mentioned it, I had not noticed there is also the `parents` field.

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+        msg += f"in both '{inner.name}' and the parent procedure '{procedure.name}': "
+        for u in undefined:
+            msg += f"{u.name}, "
+        raise RuntimeError(msg)
+
+    ## PRODUCING VARIABLES TO INTRODUCE AS DUMMY ARGUMENTS TO `inner`.
+    # Produce a list of variables defined in the scope of `procedure` that need to be resolved in `inner`'s scope
+    # by introducing them as dummy arguments.
+    # The second line drops any derived type fields, don't want them, since want to resolve the derived type itself.
+    vars_to_resolve = [v for v in FindVariables().visit(inner.body) if v.scope == procedure]
```
reuterbal, line 108: Minor point but I think this should technically check for object identity via `is`
```suggestion
    vars_to_resolve = [v for v in FindVariables().visit(inner.body) if v.scope is procedure]
```

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+    vars_to_resolve = [v for v in vars_to_resolve if not v.parent]
+
+    # Save any `DeferredTypeSymbol`s for later, they are in fact defined through imports in `procedure`,
+    # and therefore not to be added as arguments to `inner`. (the next step removes them from `vars_to_resolve`)
+    var_imports_to_add = tuple(v for v in vars_to_resolve if isinstance(v, DeferredTypeSymbol))
+
+    # Lookup the definition of the variables in `vars_to_resolve` from the scope of `procedure`.
+    # This provides maximal information on them.
+    vars_to_resolve = [procedure.variable_map[v.name] for v in vars_to_resolve \
+        if v.name in procedure.variable_map.keys()]
```
reuterbal, line 118: Just to point out that `keys()` is redundant here (and sometimes flagged by linters) and you can mangle the contains check and lookup into one via the walrus operator:

```suggestion
    vars_to_resolve = [proc_var for v in vars_to_resolve if (proc_var := procedure.variable_map.get(v.name))]
```

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # For each array in `vars_to_resolve`, append any non-literal shape variables to `vars_to_resolve`,
+    # if not already there.
+    arr_shapes = []
+    for shape in (v.shape for v in vars_to_resolve if isinstance(v, Array)):
+        for s in shape:
+            if not isinstance(s, IntLiteral):
```
reuterbal, line 125: Note that this will likely cause problems if you have a shape like `KLEV+1` - because then `shape` is `Sum`, i.e., neither a literal nor a variable. I think something like this might work maybe:
```suggestion
    for var in vars_to_resolve:
        if isinstance(var, Array):
            arr_shapes += list(FindVariables().visit(var.shape)):
```

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+                arr_shapes.append(s)
+    for v in arr_shapes:
+        if not v.name in (var.name for var in vars_to_resolve):
```
reuterbal, line 128: Note that you can rely on the automatic case-insensitive string comparison here:
```suggestion
        if v.name not in vars_to_resolve:
```

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+            vars_to_resolve.append(v)
+
+    # 1. Rescope `vars_to_resolve` into scope of `inner`.
+    # 2. Set the intent of `vars_to_resolve` to "inout" unless specified otherwise in `procedure` scope.
+    vars_to_resolve = [v.rescope(inner) for v in vars_to_resolve]
+    newtypes = map(lambda x: x.type if x.type.intent else x.type.clone(intent = "inout"), vars_to_resolve)
+    vars_to_resolve = tuple(v.clone(type = typ) for v, typ in zip(vars_to_resolve, newtypes))
```
reuterbal, line 135: I think this could be removed if you modify the arguments update further down as:
```python
    inner.arguments += tuple(
        v.clone(type=v.type.clone(intent=v.type.intent or 'inout'), scope=inner)
        for v in vars_to_resolve
    )
```

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+            # Have not encountered the module yet, add new one.
+            newsyms = tuple(s for s in imp.symbols if s.name == val.name)
+            newimport = imp.clone(symbols = newsyms)
+        imports_to_add.append(newimport)
+
+    ## MAKING THE CHANGES TO `inner`
+    # Change `inner` to take `vars_to_resolve` as dummy arguments and add all necessary imports.
+    # After these lines, `inner` should be self-contained or there is a bug.
+    inner.arguments += tuple(vars_to_resolve)
+    inner.variables += tuple(vars_to_resolve)
```
reuterbal, line 175: This is unnecessary, the arguments update will automatically insert declarations

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+        'call_finder': FindInlineCalls() if inner.is_function else FindNodes(CallStatement),
+        'argname': 'parameters' if inner.is_function else 'arguments',
+        'transformer': SubstituteExpressions if inner.is_function else Transformer
+    }
+    call_transmap = {}
+    all_calls = m['call_finder'].visit(procedure.body)
+    inner_calls = (call for call in all_calls if call.routine == inner)
+    for call in inner_calls:
+        newargs = getattr(call, m['argname']) + tuple(map(lambda x: x.clone(dimensions = None), vars_to_resolve))
+        call_transmap[call] = call.clone(**{m['argname']: newargs})
```
reuterbal, line 193: There are a few issues with this, I think:
* vars_to_resolve variables have scope `inner`, not `procedure`
* If the existing calls have keyword-arguments, this will fail
* While I appreciate the elegance of this, I think the control flow would be much easier to follow if you had separate loop for Subroutines and Functions instead

Without having tested it, something like this might work:

```suggestion
    call_map = {}
    if inner.is_function:
        for call in FindInlineCalls().visit(procedure.body):
            if call.routine is inner:
                newkwargs = tuple((v.name, v.clone(dimensions=None, scope=procedure)) for v in vars_to_resolve)
                call_map = call.clone(kw_parameters=call.kwarguments + newkwargs)
        procedure.body = SubstituteExpressions(call_map).visit(procedure.body)
    else:
        for call in FindNodes(CallStatement).visit(procedure.body):
            if call.routine is inner:
                newkwargs = tuple((v.name, v.clone(dimensions=None, scope=procedure)) for v in vars_to_resolve)
                call_map = call.clone(kwarguments=as_tuple(call.kwarguments) + newkwargs)
    procedure.body = Transformer(call_map).visit(procedure.body)
```

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,502 @@
...
+        end subroutine outer 
+    """
+    src = Sourcefile.from_source(fcode)
+    routines = extract_contained_procedures(src.routines[0])
+    assert len(routines) == 1
+    assert routines[0].name == "inner"
+    inner = routines[0]
+    outer = src.routines[0]
+    assert 'x' in (var.name for var in FindVariables().visit(inner.spec))
+    assert 'x' in (var.name for var in inner.arguments)
```
reuterbal, line 46: You can make your life a lot easier if you just rely on the automatic string comparison feature of variables:
```suggestion
    assert 'x' in FindVariables().visit(inner.spec)
    assert 'x' in inner.arguments
```
Similar changes can be made to most tests

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,502 @@
...
+from loki.ir import (
+    CallStatement, Import,
+)
+from loki.visitors import FindNodes
+from loki.transform import (
+    extract_contained_procedures
+)
+from loki.subroutine import Subroutine
+
+def test_extract_contained_procedures_basic_scalar():
```
reuterbal, line 20: To ensure compatibility across frontends, I suggest to parametrize the tests in the usual way, see for example: 
https://github.com/ecmwf-ifs/loki/blob/955a9da0bed33ec04e9578650164842543652cf6/tests/test_transform_inline.py#L34-L35

And then specify the frontend in the Sourcefile constructor, as is done for the module constructor here:
https://github.com/ecmwf-ifs/loki/blob/955a9da0bed33ec04e9578650164842543652cf6/tests/test_transform_inline.py#L66

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+    FindVariables, FindInlineCalls, SubstituteExpressions,
+    DeferredTypeSymbol, Array, IntLiteral
+)
+from loki.ir import (
+    CallStatement, DerivedType
+)
+from loki.visitors import (
+    Transformer, FindNodes,
+)
+
```
reuterbal, line 19: To explicitly list symbols that should be imported from this module, please add
```python
__all__ = ['extract_contained_procedures', 'extract_contained_procedure']
```

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,502 @@
...
+    routines = extract_contained_procedures(src.routines[0])
+    assert len(routines) == 1
+    assert routines[0].name == "inner"
+    inner = routines[0]
+    outer = src.routines[0]
+    assert 'x' in (var.name for var in FindVariables().visit(inner.spec))
+    assert 'x' in (var.name for var in inner.arguments)
+
+    call = FindNodes(CallStatement).visit(outer.body)[0]
+    assert 'x' in (var.name for var in call.arguments)
```
reuterbal, line 49:
```suggestion
    assert 'x' in call.arguments
```
Note that with the changes suggested above (i.e., always treating the injected arguments as keyword-arguments), this would have to look differently, e.g.:
```python
assert dict(call.kwarguments)['x'] == 'x'
```

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,502 @@
...
+                z = x + y + c2
+            end subroutine inner
+        end subroutine outer 
+    """
+    src = Sourcefile.from_source(fcode)
+    routines = extract_contained_procedures(src.routines[0])
+    assert len(routines) == 1
+    inner = routines[0]
+    assert "c2" in inner.import_map.keys()
+    assert not "c1" in inner.import_map.keys()
```
reuterbal, line 146: '.keys()' is redundant for checking if a key exists in a dict

reuterbal (changes requested):
Many thanks for the changes, I agree that the implementation looks much better now. I very much like how you've structured the routine. It appears very logical to me and the control flow was easy to read and follow!

I have left a number of remarks, some only stylistic, others pointing out a few potential problems. Please have a look if you agree with them.

Canonical style recommendation (PEP-8) is also to have two empty lines between free function definitions.

Testing is also very extensive, which is great. To ensure this works also for slightly more complex dummy argument interfaces, would you mind adding something where the `inner` function and subroutine expect also by default some argument? This can be just an addition to an existing test, no need to add a new one. E.g.
```
function inner(i) result(z)
  integer, intent(in) :: i
  integer :: y
  integer :: z
  y = i
  z = x + y
end function inner
```

And then test with various uses, like
```
y1 = inner(1)
y2 = inner(i=1)
y3 = inner(y1)
```

Since we want to ensure that all transformation utilities work independently of the specific Loki frontend that has been used, would you mind adding test parameterisations for the frontend? See the in-line comment for this.

Wrapping this in a transformation would indeed be nice, but I think this one has already enough complexity. So, that should ideally come as a separate PR.

tests/test_extract_contained_procedures.py
skarppinen: Thanks, indeed it is. I removed all instances of this.

loki/transform/transform_extract_contained_procedures.py
skarppinen: Added this.

tests/test_extract_contained_procedures.py
skarppinen: Thanks, I did not know about this feature. I now use this in the tests.

loki/transform/transform_extract_contained_procedures.py
skarppinen: We discussed this, I removed the unnecessary updating of `inner.variables`, but updating `inner.arguments` needs to remain.

loki/transform/transform_extract_contained_procedures.py
skarppinen: Yes, fixed it.

loki/transform/transform_extract_contained_procedures.py
skarppinen: Great catch. 

There is actually a futher corner case, which occurs when the dimension of a variable to be resolved is determined via the field of a derived type. To fix the error you pointed out, and this further case, I tweaked your suggestion to:
`arr_shapes += list(v for v in FindVariables().visit(var.shape) if not v.parent)`
which correctly introduces the derived type itself as an argument.

Furthermore, I added tests for both of these cases.

loki/transform/transform_extract_contained_procedures.py
skarppinen: True.

loki/transform/transform_extract_contained_procedures.py
skarppinen: Yes. Thanks for spotting this. I was not aware of the difference between `==` and `is` due to my Python-noobishness.

loki/transform/transform_extract_contained_procedures.py
skarppinen: Yes, using `or` is more handy here, thanks.

As a side note, I wish `intent` wasn't a subtype of `var.type`, but rather `var` itself, which I think would make sense and allow for `v.clone(intent = ...)`.

loki/transform/transform_extract_contained_procedures.py
skarppinen: Thanks, I had not even thought about keyword arguments. 

After our discussion, we came to the conclusion that indeed it is best to introduce all resolved variables as keyword arguments to calls. This is because if the call already contains keyword arguments, Fortran requires that all arguments that come after them in the call are also keyword arguments. Therefore, introducing resolved variables as positional arguments would require special processing (of the existing call) in case that the call happens to contain keyword arguments.

I changed all tests to check that resolved variables are added to kwargs instead and I added a few new tests for cases with existing arguments present in the call.

tests/test_extract_contained_procedures.py
skarppinen: Thanks. Tests now run for all frontends.

tests/test_extract_contained_procedures.py
skarppinen: Yes, did this change.

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,594 @@
...
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_extract_contained_procedures_intent(frontend):
+    """
+    This test is just to document the current behaviour: when a global is
+    introduced as an argument to the extracted contained procedure,
+    its intent will be 'inout', unless the intent is specified in the parent procedure.
+    """
+    fcode = """
+        subroutine outer(v)
```
reuterbal, line 432:
```suggestion
        subroutine outer(v, p)
```

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,594 @@
...
+    assert inner.variable_map['v'].type.intent == "in"
+    assert inner.variable_map['x'].type.intent == "inout"
+    assert inner.variable_map['p'].type.intent == "out"
+
+    # Also check that the intents don't change in the parent.
+    assert outer.variable_map['v'].type.intent == "in"
+    assert outer.variable_map['x'].type.intent is None
+    assert outer.variable_map['p'].type.intent == "out"
+
+@pytest.mark.parametrize('frontend', available_frontends())
```
reuterbal, line 463: For this test we have to mark OMNI as xfail (which is fine, it's a limitation of the frontend) because it is refusing to parse code with undefined symbols:
```suggestion
@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Parser fails on undefined symbols')]))
```

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,594 @@
...
+    assert routines[0].name == "inner"
+    inner = routines[0]
+    outer = src.routines[0]
+    assert 'x' in inner.arguments
+
+    call = list(FindInlineCalls().visit(outer.body))[0]
+    assert 'x' in call.kw_parameters
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_extract_contained_procedures_basic_scalar_function_both(frontend):
```
reuterbal, line 565: This test fails with OFP and I'm not quite clear why that is.

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,594 @@
...
+                z = y%a
+            end subroutine inner
+        end subroutine outer
+    """
+    src = Sourcefile.from_source(fcode, frontend = frontend)
+    routines = extract_contained_procedures(src.routines[0])
+    inner = routines[0]
+    assert "mytype" in inner.import_map
+
+@pytest.mark.parametrize('frontend', available_frontends())
```
reuterbal, line 375: For this test we have to mark OMNI as xfail (which is fine, it's a limitation of the frontend) because it is missing the `types` module:
```suggestion
@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Parser fails on missing types module')]))
```

reuterbal (changes requested):
Thanks so much for implementing the requested changes.
On the implementation side, I have only one more remark about style, which I would appreciate if we could follow to ensure consistent appearance.

Unfortunately, the other (more strict) Fortran frontends seem unhappy with some of the source code in the tests. OMNI refuses to parse incomplete code where it doesn't have the definition of an imported module. This is a known limitation and it is permissible to skip that frontend in these tests - I have pointed out which those are and how to mark this in the parameterisation.

Also, OMNI has found a genuine bug in the Fortran source for one of the tests.

FInally, one test fails with OFP at the end and I'm not clear what the cause is for that.

tests/test_extract_contained_procedures.py
skarppinen: Thanks, my mistake.

loki/transform/transform_extract_contained_procedures.py
```diff
@@ -0,0 +1,198 @@
...
+    ## TRANSFORMING CALLS TO `inner` in `procedure`.
+    # The resolved variables are all added as keyword arguments to each call.
+    # (to avoid further modification of the call if it already happens to contain kwargs).
+    # Here any dimensions in the variables are dropped, since they should not appear in the call.
+    # Note that functions need different visitors and mappers than subroutines.
+    call_map = {}
+    if inner.is_function:
+        for call in FindInlineCalls().visit(procedure.body):
+            if call.routine is inner:
+                newkwargs = tuple((v.name, v.clone(dimensions = None, scope = procedure)) for v in vars_to_resolve)
```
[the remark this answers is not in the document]
skarppinen: Ah, sorry. I will leave these unresolved from now on.

tests/test_extract_contained_procedures.py
```diff
@@ -0,0 +1,598 @@
...
+            end subroutine inner
+        end subroutine outer
+    """
+    src = Sourcefile.from_source(fcode, frontend=frontend)
+    routines = extract_contained_procedures(src.routines[0])
+    assert len(routines) == 1
+    inner = routines[0]
+    outer = src.routines[0]
+    assert 'x' in inner.arguments
+    assert 'arr(3)' in inner.arguments
```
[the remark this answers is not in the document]
skarppinen: Thanks @reuterbal, I fixed this in the latest commit.

reuterbal (approved):
Many thanks for the patience and endurance getting this over the line. This looks great now!

skarppinen:
Thanks @reuterbal, this was a good learning experience.

mlange05 (approved):
Indeed, just wanted to add that this looks really great now, and the rigorous test coverage is much appreciated. I only skimmed the final version, but test look solid, so GTG from me too. 

And also, another big thank you for the patient and rigorous PR and review process. The final result looks fantastic, and a great full first contribution. Many, many thanks @skarppinen and @reuterbal

PR #183, opened 2023-10-27
Draft: Enable ir_graph representation while dataflow analysis

joscao:
Enhances the IR graph representation by also showing lives,uses and defines when `dataflow_analysis` is attached

uses_defines_live

- [ ] Provide a reasonable test case

### DIFF
--- a/loki/visitors/ir_graph.py
+++ b/loki/visitors/ir_graph.py
@@ -185,7 +185,19 @@ class GraphCollector(Visitor):
         """
         label = kwargs.get("label", "")
         if label == "":
-            label = self.format_node(repr(node))
+            try:
+                comment = "\nlive: " + ", ".join(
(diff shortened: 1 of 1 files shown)


loki/visitors/ir_graph.py
```diff
@@ -184,9 +184,33 @@ def __add_node(self, node, **kwargs):
...
+                label,
+                "\n",
+                live_symbols,
+                "], ",
+                defines_symbols,
+                "], ",
+                uses_symbols,
+                "]",
+            )
+        except (RuntimeError, KeyError, AttributeError) as _:
```
reuterbal, line 211: I think you can simply skip the `as _` if you're not planning to use the exception

reuterbal (approved):
Useful addition, many thanks!. Implementation looks good to me

PR #185, opened 2023-11-07
add 'kernel_only=True' to RemoveCallsTransformation

MichaelSt98:
add `kernel_only=True` to RemoveCallsTransformation to avoid DrHook Calls being removed from driver routines. 

This was part of the `loki_transform.py` `ecphys()` entry point, but not yet part of the `convert()` entry point. Tested with/for ecphys.


reuterbal (approved):
Uncontroversial, I think. Looks good to me, many thanks!

PR #188, opened 2023-11-15
Minor transformation fixes

reuterbal:
Lifting some changes from the Scheduler overhaul that are purely targeted at decoupling transformation functionality from scheduler API/behaviour.


mlange05 (approved):
Visually all good for me, and should all be covered by CLOUDSC regression, so GTG from me.

PR #191, opened 2023-11-20
Transformation configuration and SchedulerConfig update

mlange05:
**Note: This change is not backward compatible for our regression test suite!** We therefore temporarily point to custom CLOUDSC (ecWAM) branches - the respective branches need to be merged, and **the respective commit(s) removed before merging this PR.** 

This change is part of the wider plan to make the transformation composition more configurable. It primarily brings in the ability to add configuration entries for transformations to the scheduler config file, and then allow the `SchedulerConfig` to directly instantiate them. This instantiation process allows two things that are important for further refactoring:
* Remote instantiation of `Transformation` classes from any module (even external!)
* Resolution of `Dimension` objects via a custom placeholder notation that allows us to reuse the ones defined in the `SchedulerConfig` already.

In order to provide a more streamlined configuration file, this PR also removes some backward hand-holding, where we would generate dicts of Dimension/Routine objects with custom handling. Instead, we now utilise the native TOML dict/table nesting to use the name of a Dimension/Routine/Transformation directly form the processed TOML output.

_And finally, I'm filing this as a draft PR first, to see what breaks in the regression tests, etc. Please feel free to chime in though already, as this is somewhat invasive and might need some careful marshalling._


mlange05:
@reuterbal, apologies for the large diff in the tests due to the name restructuring. I can move this into a separate PR if that makes things easier. I'll also file PRs for the corresponding upstream changes in CLODUSC / ecWAM and reference this one here.

loki/batch/configure.py
```diff
@@ -0,0 +1,177 @@
...
+        self.path = path
+        self.options = dict(options)
+
+    def resolve_dimensions(self, dimensions):
+        """
+        Substitute :any:`Dimension` objects for placeholder strings.
+
+        The format of the string replacement matches the TOML
+        configuration.  It will attempt to replace ``%dimensions.dim_name%``
+        with a :any:`Dimension` found in :param dimensions:
```
reuterbal, line 142:
```suggestion
        with a :any:`Dimension` found in :data:`dimensions`:
```

loki/batch/configure.py
```diff
@@ -0,0 +1,177 @@
...
+        If not provided, ``name`` will be used instead.
+    path : str or Path, optional
+        Path to add to the sys.path before attempting to load the ``module``
+    options : dict
+        Dicts of options that define the transformation behaviour.
+        These options will be passed as constructor arguments using
+        keyword-argument notation.
+    """
+
+    _re_dimension = re.compile(r'\%dimensions?\.(.*)\%')
```
reuterbal, line 127: Two questions about the regex:
- is the `s` of `dimensions` really intended to be optional?
- I think the capture group should be lazy, otherwise something like `%dimensions.vertical% + %dimensions.horizontal%` would be matched as one rather than two occurences (for whichever reason one might write that)

```suggestion
    _re_dimension = re.compile(r'\%dimensions\.(.*?)\%')
```

tests/test_scheduler.py
```diff
@@ -2190,3 +2176,61 @@ def test_scheduler_unqualified_imports(config):
...
+    assert transformation.module_suffix == '_roll'
+    assert not transformation.replace_ignore_items
+
+
+def test_transformation_config_with_dimension(config):
+    """
+    Test instantiation of :any:`Transformation` from config with
+    :any:`Dimension` argument.
+    """
+    from transformations.single_column_coalesced import SCCBaseTransformation  # pylint: disable=import-outside-toplevel
```
reuterbal, line 2212: Not sure I'm a fan of using a transformation from the "transformations" package within the core Loki testbase - This introduces a circular dependency between the two packages.
Could we not write a dummy transformation to file on the fly and load that instead?

tests/test_scheduler.py
```diff
@@ -2190,3 +2176,61 @@ def test_scheduler_unqualified_imports(config):
 
     assert item.enable_imports
     assert item.children == ('other_routine',)
+
```
reuterbal, line 2179: I think it would be useful to add here also negative tests that verify behaviour and error messages when the transformation cannot be found or instantiation fails. This will likely be a reasonable common occurence, if only because of typos in the configuration file, faulty installations, or similar mistakes.

tests/test_scheduler.py
```diff
@@ -2190,3 +2176,61 @@ def test_scheduler_unqualified_imports(config):
...
+
+def test_transformation_config(config):
+    """
+    Test the correct instantiation of :any:`Transformation` objecst from config
+    """
+    my_config = config.copy()
+    my_config['transformations'] = {
```
reuterbal, line 2186: Can we parametrise this test to provide `my_config['transformations']` in the two supported formats? I'm not quite clear what the second is meant to be here:
https://github.com/ecmwf-ifs/loki/blob/5899bf00a6f969c762e6c7551b7141b30dbaaf02/loki/batch/configure.py#L57-L60

reuterbal (changes requested):
First of all, this is a vast improvement in the appearance of the Scheduler config and unlocks an entirely new level of configurability and separation of transformation recipes from target code-specific hand-holding. I'm really pleased about this!

Also, conceptually I don't have much to add - this looks all sound and is in line with what we discussed offline. It needs a rebase, though, and Codecov seems a little confused right now as to what is covered by tests and what not. But I'd like at least additional testing for when the instantiation of transformations from config file entries fails, as I expect this to be a not too rare occurence.

Also, there is now a dependency on the transformations package in the tests of the Loki core library, which I would like to avoid.

loki/batch/configure.py
mlange05: Hehe, the optional `s` was lazyness (I kept getting it wrong...). Removed now, as we should aim for consistency. :smirk:

tests/test_scheduler.py
mlange05: Agreed. Added a new dummy test to pick this up properly, which incidentally now also covers the custom import-path case. :smile:

tests/test_scheduler.py
mlange05: Nah, I think I just forgot to remove this. Fixed now.

tests/test_scheduler.py
```diff
@@ -2333,3 +2318,79 @@ def test_scheduler_disable_wildcard(here, config):
...
+
+    worse_config = config.copy()
+    worse_config['transformations'] = {
+        'DependencyTransform': {
+            'module': 'loki.transformats',  # <= typo
+            'options': {}
+        }
+    }
+    with pytest.raises(ModuleNotFoundError):
+        cfg = SchedulerConfig.from_dict(worse_config)
```
reuterbal, line 2367: Just to point this out: For full coverage we would need a `worst_config` with invalid options provided, that's the only bit currently not covered in the tests.

reuterbal (approved):
Many thanks, this looks very good now and we could include it as is. For 100% gold-star test coverage, there's one combination that could be tested additionally but I'll leave it up to you if you think it's necessary.

However, to reduce the impact on the parallel scheduler rewrite, delaying the bulk->batch rename to a separate, later PR would help. Generally very much in favor of this, though!

And lastly, as discussed, I'll hold off merging until we have rolled a release as this incurs a breaking change.

mlange05:
Ok, great, many thanks @reuterbal ! I've rebased this without the bulk->batch rename (please double-check), and pushed again (including the final test addition). Should be fine now, but waiting for CI to confirm. Please also note that there's a `[DO NOT MERGE]` commit in there that temporarily re-pointed the branches for regression testing.

reuterbal (approved):
Fantastic, many thanks, very happy with this now!

This has currently a conflict, so needs a rebase over latest version of main. I would suggest to remove the `[DO NOT MERGE]` commit in the process and accept the failing regression tests when merging.
Afterwards, we should see the checks turn green in the pending PRs on ecWAM and CLOUDSC and then merge them:
https://github.com/ecmwf-ifs/ecwam/pull/7
https://github.com/ecmwf-ifs/dwarf-p-cloudsc/pull/61

After both are merged, we can trigger a re-run of the regression tests on main to turn them green again.

reuterbal:
As discussed: Will merge this now and hope for PRs on regression test repositories to turn green, before merging those and re-running regression on Loki.

Issue #182, opened 2023-10-27
SCC-stack: UnboundLocalError when a field of a derived type is used to declare a temporary array


skarppinen:
This is bug occurs when running "loki-transform.py" with mode = scc-stack.
The error message is "UnboundLocalError: local variable 'new_type' referenced before assignment", coming from expression/mappers.py:600

Here is a reproducer of the issue, note the comments in the Fortran code:
```
module mod3
    implicit none 
    type :: mytype
            integer :: i
    end type mytype
contains
    subroutine wrapper
        integer :: h_start = 1
        integer, parameter :: h_dim = 10
        integer :: h_end 
        integer :: v_start = 1
        integer, parameter :: v_dim = 20
        integer, parameter :: block_size = 30
        integer :: block_index
        
        type(mytype) :: m 
        m%i = 2

        real :: arr(h_dim, v_dim, block_size)

        h_end = h_dim
        do block_index = 1, block_size 
        call depth1(h_start, h_end, h_dim, v_start, v_dim, &
            m, arr(:, :, block_index))
        end do
        print *, "Sum of array is ", sum(arr)
    end subroutine wrapper

    subroutine depth1(h_start, h_end, h_dim, v_start, v_dim, m, array)
        ! Arguments.
        integer, intent(in) :: h_start
        integer, intent(in) :: h_end
        integer, intent(in) :: h_dim
        integer, intent(in) :: v_start
        integer, intent(in) :: v_dim
        real, intent(inout) :: array(h_dim, v_dim)
        type(mytype), intent(in) :: m

        ! Non-array local variables.
        integer :: v_index
        integer :: h_index

        ! Temporary arrays.
        real :: tmp1(h_dim, v_dim), tmpx(h_dim, v_dim)
        real :: tmpy(h_dim, v_dim, m%i) ! NOTE: This temporary is declared using m%i, which is the root of the issue.

        do v_index = v_start, v_dim
            do h_index = h_start, h_end
                tmpx(h_index, v_index) = 4.0 
                ! NOTE: Here we write to the offending array. If this is omitted, the crash doesn't occur.
                tmpy(h_index, v_index, 1) = 13.0
                tmp1(h_index, v_index) = exp(log(real(h_index)) + log(real(v_index)) - 1.0)
            end do
        end do

        do v_index = v_start, v_dim
            do h_index = h_start, h_end
                array(h_index, v_index) = exp(tmp1(h_index, v_index) + 0.25)
            end do
        end do


[...]

PR #192, opened 2023-11-22
Nested derived type bug fix

rolfhm:
Possibly a solution to issue https://github.com/ecmwf-ifs/loki/issues/182

### DIFF
--- a/loki/expression/mappers.py
+++ b/loki/expression/mappers.py
@@ -597,7 +597,7 @@ class LokiIdentityMapper(IdentityMapper):
         if expr.scope is None:
             if parent is expr.parent and not is_type_changed:
                 return expr
-            elif not is_type_changed:
+            if not is_type_changed:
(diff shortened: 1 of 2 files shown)


mlange05:
Hi @rolfhm I just found this, and have recently stumbled across a slightly better fix. I really like your test though, so I've just triggered a rebase, and will bring this in with the other fix on top, I think. Hope you don't mind?

rolfhm:
No problem. Glad it was of some use.
Just noticed there's a double 'replace' in the test docstring

mlange05 (approved):
Apologies for the delay in picking this up. The test is neat and the problem had been fixed in another branch already. I've brought this all up to speed with latest main and consolidated (probably should be a squash-merge). GTG from me.

reuterbal (approved):
Thanks all, fix looks great and test is much appreciated. GTG!

PR #193, opened 2023-11-24
pool_allocator handles range indices

rolfhm:
Added a bit of code to handle range indices in array declarations and modified tests accordingly.


reuterbal (approved):
Many thanks, this is great! I have manually verified regression tests offline and everything looks ok. GTG! 
(I suggest a squash merge for this.)

mlange05 (approved):
Only checked this visually, but the test coverage looks convincing. GTG from me too.

PR #194, opened 2023-11-27
Allow wildcards in `disable` list for scheduler

mlange05:
A very small and pragmatic change that allows us to generically exclude groups of calls from `Scheduler` traversal by using wildcards in the `disable` method. A more robust and generic solution would probably extend wildcard notation to all path/source name resolution mechanisms, but as the `Scheduler` is still being refactored heavily I chose the "quick and easy" option. Opinions and comments very welcome, of course!

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -6,6 +6,7 @@
 # nor does it submit to any jurisdiction.
 
 from abc import abstractmethod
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Very useful addition, well tested. Uncontroversial and GTG from me!

PR #197, opened 2023-12-08
Create separate ModuleWrapTransformation from DependencyTransformation

reuterbal:
The DependencyTransformation is responsible for an increasing number of workarounds and special casing in the scheduler traversals. Part of this has been identified to be related to the fact that two concepts are mixed together here:
1. Renaming of scopes (modules/subroutines get new names, which requires renaming on the caller side)
2. Changing scopes (subroutines get embedded into modules)

This PR splits the two into separate transformation passes but tries to be minimally invasive otherwise. There may undoubtedly be capacity in rewriting some of the transformation logic itself but that can now be a separate change.

The original `strict` mode behaviour is retained by plainly calling the DependencyTransformation, the `module` mode is recovered by applying the new ModuleWrapTransformation before the DependencyTransformation.

A caveat is the fact that the current scheduler items loose the ability to identify inline calls as targets after the module wrapping. This is because these links are only identified for functions that are declared via explicit interfaces, and after the module wrap these are regular fortran imports via `use`. Previously this wasn't a problem because the wrapping and renaming happened in the same step, and afterwards in the transformation pipeline this relation was no longer required. 

To retain this piece of information across the now two independent transformation passes I converted the targets property to a cached property. This seems to be ok for all tests and regression tests but needs careful verification on ecphys! It's important to note that this is only a temporary measure because the scheduler rewrite will introduce inline call dependencies independently from an interface declaration.

### DIFF
--- a/docs/source/transform.rst
+++ b/docs/source/transform.rst
@@ -82,6 +82,7 @@ to grow in the future:
    loki.transform.build_system_transform.CMakePlanner
    loki.transform.build_system_transform.FileWriteTransformation
    loki.transform.dependency_transform.DependencyTransformation
+   loki.transform.dependency_transform.ModuleWrapTransformation
    loki.transform.fortran_c_transform.FortranCTransformation
    loki.transform.fortran_max_transform.FortranMaxTransformation
    loki.transform.fortran_python_transform.FortranPythonTransformation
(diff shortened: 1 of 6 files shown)


mlange05 (approved):
Looks good to me and works with current EC-physics regression. I think the cached `target` property was the ~hack~ trick I was missing when I tried this recently. As this will be removed soon, I'm happy for this to go ahead as is. GTG from me.

PR #198, opened 2023-12-12
Pragma-driven subroutine inlining and associated utilities

mlange05:
The primary purpose of this PR is to introduce pragma-driven source inlining for Suboutine where `CallStatement` nodes annotated with the `!$loki inline` pragma are replaced by the body of the called routine, with resolved argument mappings. As I developed this on the EC-physics regression suite, I needed to include a few additional utilities that deal with the fallout of the resolution of optional arguments (simplistic logic evaluation in `simplify` and primitive dead code elimination), but these could easily be factored into a separate PR if that is desirable (@reuterbal, please shout!). 

Otherwise, the PR is primarily a refactoring of the procedure inlining utility, where the remapping of a call to the called body is now completely separated from the dealing with local variable declarations (in `inline_subroutine_calls`). This allows us to reuse these entry points between `inline_marked_subroutines` and `inline_internal_procedures`.

The rest of the changes and fixes some more detail:
- Separation of call remapping from declaration/scope management in procedure inlining
- Renaming of `inline_member_subroutines` to the more accurate "internal" and "procedure" (no change in capability though!)
- Dealing with unresolved `optional` arguments after call remapping (resolve the `present(n)` clause, depending on the presence of `n`, to either `.true.` or `.false.`)
-  Simplistic dead code elimination that resolves explicitly unreachable branches of conditionals (eg. from resolving `if (present(arg)) <do_something_with_arg>`. This then avoids dealing with remnant uses of the optional args, in case they were not given to an inlined call.
- Extension of the above dead code elimination by resolving simple constant expressions to logical values, eg. `1 == 6 .or. 2 == 4` to simply `.false.`
- A small fix to vector section-derivation that fixes a corner case of `!$loki separator` prgamas sitting deep in a nested control flow (possibly under tested :hand_over_mouth: )

### DIFF
--- a/loki/transform/transform_dead_code.py
+++ b/loki/transform/transform_dead_code.py
@@ -64,5 +64,5 @@ class DeadCodeEliminationTransformer(Transformer):
         if condition == 'False':
             return else_body
 
-        has_elseif = isinstance(else_body, Conditional)
+        has_elseif = o.has_elseif and else_body and isinstance(else_body[0], Conditional)
         return self._rebuild(o, tuple((condition,) + (body,) + (else_body,)), has_elseif=has_elseif)
--- a/loki/transform/transform_dead_code.py
+++ b/loki/transform/transform_dead_code.py
@@ -1,0 +1,66 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""
+Collection of utilities to perform Dead Code Elimination.
+"""
+from loki.visitors import Transformer
+from loki.expression.symbolic import simplify
+from loki.tools import flatten, as_tuple
+
+
+__all__ = ['dead_code_elimination', 'DeadCodeEliminationTransformer']
+
+
+def dead_code_elimination(routine, use_simplify=True):
+    """
+    Perform Dead Code Elimination on the given :any:`Subroutine` object.
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine to which to apply dead code elimination.
+    simplify : boolean
+        Use :any:`simplify` when evaluating expressions for branch pruning.
+    """
+
+    transformer = DeadCodeEliminationTransformer(use_simplify=use_simplify)
+    routine.body = transformer.visit(routine.body)
+
+
+class DeadCodeEliminationTransformer(Transformer):
+    """
+    :any:`Transformer` class that removes provably unreachable code paths.
+
+    The pirmary modification performed is to prune individual code branches
+    under :any:`Conditional` nodes.
+
+    Parameters
+    ----------
+    simplify : boolean
+        Use :any:`simplify` when evaluating expressions for branch pruning.
+    """
+
+    def __init__(self, use_simplify=True, **kwargs):
+        super().__init__(**kwargs)
+        self.use_simplify = use_simplify
+
+    def visit_Conditional(self, o, **kwargs):
+        condition = self.visit(o.condition, **kwargs)
+        body = as_tuple(flatten(as_tuple(self.visit(o.body, **kwargs))))
+        else_body = as_tuple(flatten(as_tuple(self.visit(o.else_body, **kwargs))))
+
+        if self.use_simplify:
+            condition = simplify(condition)
+
+        if condition == 'True':
+            return body
+
+        if condition == 'False':
+            return else_body
+
+        return self._rebuild(o, tuple((condition,) + (body,) + (else_body,)))
--- a/loki/transform/transform_dead_code.py
+++ b/loki/transform/transform_dead_code.py
@@ -11,6 +11,7 @@ Collection of utilities to perform Dead Code Elimination.
 from loki.visitors import Transformer
 from loki.expression.symbolic import simplify
 from loki.tools import flatten, as_tuple
+from loki.ir import Conditional
 
 
 __all__ = ['dead_code_elimination', 'DeadCodeEliminationTransformer']
@@ -63,4 +64,5 @@ class DeadCodeEliminationTransformer(Transformer):
         if condition == 'False':
             return else_body
 
-        return self._rebuild(o, tuple((condition,) + (body,) + (else_body,)))
+        has_elseif = isinstance(else_body, Conditional)
+        return self._rebuild(o, tuple((condition,) + (body,) + (else_body,)), has_elseif=has_elseif)
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -10,6 +10,8 @@ Collection of utility routines to perform code-level force-inlining.
 
 
 """
+from collections import defaultdict
+
 from loki.expression import (
     FindVariables, FindInlineCalls, FindLiterals,
     SubstituteExpressions, LokiIdentityMapper
@@ -20,11 +22,13 @@ from loki.types import BasicType
 from loki.visitors import Transformer, FindNodes
 from loki.tools import as_tuple
 from loki.logging import warning, error
+from loki.pragma_utils import pragmas_attached, is_loki_pragma
 
 
 __all__ = [
     'inline_constant_parameters', 'inline_elemental_functions',
-    'inline_internal_procedures', 'inline_member_procedures'
+    'inline_internal_procedures', 'inline_member_procedures',
+    'inline_marked_subroutines'
 ]
 
 
@@ -364,3 +368,35 @@ def inline_internal_procedures(routine):
 
 
 inline_member_procedures = inline_internal_procedures
+
+
+def inline_marked_subroutines(routine):
+    """
+    Inline :any:`Subroutine` objects guided by pragma annotations.
+
+    When encountering :any:`CallStatement` objects that are marked with a
+    ``!$loki inline`` pragma, this utility will attempt to replace the call
+    with the body of the called procedure and remap all passed arguments
+    into the calling procedures scope.
+
+    Please note that this utility requires :any:`CallStatement` objects
+    to be "enriched" with external type information.
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine in which to look for pragma-marked procedures to inline
+    """
+
+    with pragmas_attached(routine, node_type=CallStatement):
+
+        # Group the marked calls by callee routine
+        call_sets = defaultdict(list)
+        for call in FindNodes(CallStatement).visit(routine.body):
+            if is_loki_pragma(call.pragma, starts_with='inline'):
+                call_sets[call.routine].append(call)
+
+        # Trigger per-call inlining on collected sets
+        for callee, calls in call_sets.items():
+            if callee:  # Skip the unattached calls (collected under None)
+                inline_subroutine_calls(routine, calls, callee)
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -282,7 +282,7 @@ def map_call_to_procedure_body(call, caller):
     return (comment, c_line) + as_tuple(callee_body) + (c_line, )
 
 
-def inline_subroutine_calls(routine, calls, callee):
+def inline_subroutine_calls(routine, calls, callee, allowed_aliases=None):
     """
     Inline a set of call to an individual :any:`Subroutine` at source level.
 
@@ -300,26 +300,32 @@ def inline_subroutine_calls(routine, calls, callee):
     calls : tuple or list of :any:`CallStatement`
     callee : :any:`Subroutine`
         The called target subroutine to be inlined in the parent
+    allowed_aliases : tuple or list of str or :any:`Expression`, optional
+        List of variables that will not be renamed in the parent scope, even
+        if they alias with a local declaration.
     """
+    allowed_aliases = as_tuple(allowed_aliases)
 
     # Ensure we process sets of calls to the same callee
     assert all(call.routine == callee for call in calls)
 
     # Prevent shadowing of callee's variables by renaming them a priori
     parent_variables = routine.variable_map
-    duplicate_locals = tuple(
+    duplicates = tuple(
         v for v in callee.variables
         if v.name in parent_variables and v.name.lower() not in callee._dummies
     )
+    # Filter out allowed aliases to prevent suffixing
+    duplicates = tuple(v for v in duplicates if v.name not in allowed_aliases)
     shadow_mapper = SubstituteExpressions(
-        {v: v.clone(name=f'{callee.name}_{v.name}') for v in duplicate_locals}
+        {v: v.clone(name=f'{callee.name}_{v.name}') for v in duplicates}
     )
     callee.spec = shadow_mapper.visit(callee.spec)
 
     var_map = {}
-    duplicate_locals_names = {dl.name.lower() for dl in duplicate_locals}
+    duplicate_names = {dl.name.lower() for dl in duplicates}
     for v in FindVariables(unique=False).visit(callee.body):
-        if v.name.lower() in duplicate_locals_names:
+        if v.name.lower() in duplicate_names:
             var_map[v] = v.clone(name=f'{callee.name}_{v.name}')
     callee.body = SubstituteExpressions(var_map).visit(callee.body)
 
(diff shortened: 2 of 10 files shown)


tests/test_transform_inline.py
```diff
@@ -607,3 +609,145 @@ def test_inline_member_routines_with_associate(frontend):
...
+contains
+  subroutine add_one(a)
+    real(kind=8), intent(inout) :: a
+    a = a + 1
+  end subroutine add_one
+
+  subroutine add_a_to_b(a, b, n)
+    real(kind=8), intent(inout) :: a(:), b(:)
+    integer, intent(in) :: n
+
```
MichaelSt98, line 654: missing `integer :: i`, which changes the inlined code as well 

```fortran
  DO add_a_to_b_i=1,add_a_to_b_n
    a(add_a_to_b_i) = a(add_a_to_b_i) + b(add_a_to_b_i)
  END DO
```

Thus still works and is valid, just the question whether this should be fixed for possible future changes?

MichaelSt98 (approved):
Really nice. Looking forward to use this 👍

tests/test_transform_inline.py
reuterbal: Yes, original Fortran should be valid (technically it is, because there's also no `implicit none`)

tests/test_transform_inline.py
```diff
@@ -607,3 +609,145 @@ def test_inline_member_routines_with_associate(frontend):
 
     assocs = FindNodes(Associate).visit(routine.body)
     assert len(assocs) == 2
+
+
+@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'No header information in test')]))
```
reuterbal, line 614: Why should this not work for OMNI? If you parse the module first, the missing xmod is auto-generated and the driver should parse successfully.

tests/test_transform_inline.py
```diff
@@ -607,3 +609,145 @@ def test_inline_member_routines_with_associate(frontend):
...
+    assert assign[1].lhs == 'a(i)' and assign[1].rhs == 'a(i) + b(i)'
+
+    # Check that the last call is left untouched
+    calls = FindNodes(CallStatement).visit(driver.body)
+    assert len(calls) == 1
+    assert calls[0].routine.name == 'add_one'
+    assert calls[0].arguments == ('b(i)',)
+
+
+@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'No header information in test')]))
```
reuterbal, line 686: Same comment as in the previous test

tests/test_transform_dead_code.py
```diff
@@ -0,0 +1,52 @@
...
+
+  if (flag) then
+    if (1 == 6) then
+      a = a + b
+    else
+      b = b + 2.0
+    end if
+
+    if (2 == 2) then
+      b = b + a
```
reuterbal, line 33: Would you mind adding a second test with conditionals with `elseif` branches that exercise the permutations of options which paths are to be eliminated (if, elseif, else, if&elseif, elseif&else)? I expect this to largely work as expected without further work thanks to the multi-body conditionals being converted to if/else blocks. But the issue discovered by Paddy makes me think we may lack an update of the corresponding `is_elseif` flag.

scripts/loki_transform.py
```diff
@@ -107,15 +107,19 @@ def cli(debug):
               help="Remove derived-type arguments and replace with canonical arguments")
 @click.option('--inline-members/--no-inline-members', default=False,
               help='Inline member functions for SCC-class transformations.')
+@click.option('--inline-marked/--no-inline-marked', default=True,
+              help='Inline pragma-marked subroutines for SCC-class transformations.')
 @click.option('--resolve-sequence-association/--no-resolve-sequence-association', default=False,
               help='Replace array arguments passed as scalars with arrays.')
 @click.option('--derive-argument-array-shape/--no-derive-argument-array-shape', default=False,
               help="Recursively derive explicit shape dimension for argument arrays")
+@click.option('--eliminate_dead_code/--no-eliminate_dead_code', default=True,
```
reuterbal, line 116: I'd prefer to use hyphen notation instead of mixing with underscores.

loki/transform/transform_inline.py
```diff
@@ -229,88 +227,140 @@ def _map_unbound_dims(var, val):
...
     parent_variables = routine.variable_map
-    duplicate_locals = tuple(
-        v for v in member.variables
-        if v.name in parent_variables and v.name.lower() not in member._dummies
+    duplicates = tuple(
+        v for v in callee.variables
+        if v.name in parent_variables and v.name.lower() not in callee._dummies
     )
+    # Filter out allowed aliases to prevent suffixing
+    duplicates = tuple(v for v in duplicates if v.name not in allowed_aliases)
```
reuterbal, line 329: Do we need to be careful here with regard to case-sensitivity, since comparing string `v.name`?

loki/transform/transform_dead_code.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.use_simplify:
+            condition = simplify(condition)
+
+        if condition == 'True':
+            return body
+
+        if condition == 'False':
+            return else_body
+
+        return self._rebuild(o, tuple((condition,) + (body,) + (else_body,)))
```
reuterbal, line 66: I believe here we need a check for the case that [`has_elseif`](https://github.com/ecmwf-ifs/loki/blob/37d0033ea165985fb35cec2273067024af8e1e11/loki/ir.py#L566C5-L566C15) is True, whether the `else_body` still starts with a conditional as that part could have been resolved.

reuterbal (changes requested):
Really nice and powerful addition and I'm pleasantly surprised that we can do something as reasonably advanced as the dead code elimination. Many thanks for this!

I left a few comments throughout. No showstoppers but I suspect we may run into problems with if-elseif-else style conditionals when rebuilding the nodes. Also, not quite clear on why the tests shouldn't work with OMNI.

skarppinen:
Hello @mlange05, is there a particular reason to not add `allowed_aliases` for `inline_member_procedures` as well?

skarppinen:
Here is also a reproducer for a small bug in `inline_internal_procedures`:

This code
```
import loki
from loki.transform import inline_internal_procedures
fcode = """
    subroutine outer() 
        integer :: z
        integer :: jlon
        z = 0
        jlon = 0

        !$loki inline
        call inner(z)

        jlon = z + 4
        contains
        subroutine inner(z)
            integer, intent(inout) :: z
            integer :: jlon, jg ! This declaration with jlon and jg together causes the issue!
            jlon = 1
            jg = 2
            z = jlon + jg
        end subroutine inner
    end subroutine outer
"""
src = loki.Sourcefile.from_source(fcode)
routine = src.routines[0]
inline_internal_procedures(routine, "jlon")
```
produces for `routine` the Fortran source:
```
SUBROUTINE outer ()
  INTEGER :: z
  INTEGER :: jlon
  z = 0
  jlon = 0

!$loki inline
  ! [Loki] inlined child subroutine: inner
  jlon = 1
  jg = 2
  z = jlon + jg

  jlon = z + 4
  CONTAINS
END SUBROUTINE outer
```
where `jg` is never declared. This happens, because `jlon` and `jg` are declared on the same line in the inner routine. If they are declared each on their own lines, this issue disappears. Thus, I suspect that a call to `single_variable_declaration` should be made at some point.

tests/test_transform_inline.py
mlange05: Ahh, this is what the `allowed_aliases` is for; I've amended the test to use this and retrofitted the `implicit none`.

tests/test_transform_inline.py
mlange05: Correct, wasn't thinking! Many thanks for picking this out!

tests/test_transform_dead_code.py
mlange05: Yes, I've added two small tests for nested conditionals - one to the base test that gets fully resolved, and one additional test for partially resolved nested conditionals. The bug you feared never materialised though. :wink:

scripts/loki_transform.py
mlange05: Fully agree. Copy-paste error ... fixed now.

loki/transform/transform_inline.py
mlange05: Aha, yes. But a better fix is to use `v.symbol` as our base symbols do indeed inherit case-insensitive equivalence. :grin: I've added this to the base test.

loki/transform/transform_dead_code.py
mlange05: This one is curious, as we actually do not need to check for it. The reasoning is that if we reach this line, none of the sub-branches have been resolved, so the current value of `has_elseif` is still valid. I've added tests to ensure this is covered though.

mlange05:
> Here is also a reproducer for a small bug in `inline_internal_procedures`:
> 
> This code
> 
> ```
> import loki
> from loki.transform import inline_internal_procedures
> fcode = """
>     subroutine outer() 
>         integer :: z
>         integer :: jlon
>         z = 0
>         jlon = 0
> 
>         !$loki inline
>         call inner(z)
> 
>         jlon = z + 4
>         contains
>         subroutine inner(z)
>             integer, intent(inout) :: z
>             integer :: jlon, jg ! This declaration with jlon and jg together causes the issue!
>             jlon = 1
>             jg = 2
>             z = jlon + jg
>         end subroutine inner
>     end subroutine outer
> """
> src = loki.Sourcefile.from_source(fcode)
> routine = src.routines[0]
> inline_internal_procedures(routine, "jlon")
> ```
> 
> produces for `routine` the Fortran source:
> 
> ```
> SUBROUTINE outer ()
>   INTEGER :: z
>   INTEGER :: jlon
>   z = 0
>   jlon = 0
> 
> !$loki inline
>   ! [Loki] inlined child subroutine: inner
>   ! =========================================
>   jlon = 1
>   jg = 2
>   z = jlon + jg
>   ! =========================================
> 
>   jlon = z + 4
>   CONTAINS
> END SUBROUTINE outer
> ```
> 
> where `jg` is never declared. This happens, because `jlon` and `jg` are declared on the same line in the inner routine. If they are declared each on their own lines, this issue disappears. Thus, I suspect that a call to `single_variable_declaration` should be made at some point.

Hi @skarppinen, many thanks for flagging this and providing the test. The reason this was not added was indeed just me being focused on the problem at hand, but I've now extended this option to the internal procedure inliner too.

Secondly, the test you provided indeed then causes a bug, and the fix follows closely to what you suggest (explicitly separate the "allowed alias" declarations into single-variable declarations using the utility. Please have a look and see of this fixes your issue.

mlange05:
@reuterbal and @skarppinen many thanks for the thorough review and valuable suggestions. I think I've addressed all points, so this is ready for another look.

tests/test_transform_dead_code.py
reuterbal: Yeah, because you're leaving out the case that the conditional remains but the elseif branch is eliminated. See below for a suggestion how to amend the test

tests/test_transform_dead_code.py
```diff
@@ -0,0 +1,101 @@
...
+    dead_code_elimination(routine)
+
+    conditionals = FindNodes(Conditional).visit(routine.body)
+    assert len(conditionals) == 1
+    assert conditionals[0].condition == 'flag'
+    assigns = FindNodes(Assignment).visit(routine.body)
+    assert len(assigns) == 3
+    assert assigns[0].lhs == 'b' and assigns[0].rhs == 'b + 2.0'
+    assert assigns[1].lhs == 'b' and assigns[1].rhs == 'b + a'
+    assert assigns[2].lhs == 'a' and assigns[2].rhs == 'a + b'
```
reuterbal, line 63:
```suggestion

    if (a > 2.0) then
      a = a + 5.0
    elseif (2 == 3) then
      a = a + 3.0
    else
      a = a + 1.0
    endif
  end if
end subroutine test_dead_code_conditional
"""
    routine = Subroutine.from_source(fcode, frontend=frontend)
    # Please note that nested conditionals (elseif) counts as two
    assert len(FindNodes(Conditional).visit(routine.body)) == 7
    assert len(FindNodes(Assignment).visit(routine.body)) == 10

    dead_code_elimination(routine)

    conditionals = FindNodes(Conditional).visit(routine.body)
    assert len(conditionals) == 2
    assert conditionals[0].condition == 'flag'
    assert not conditionals[0].has_elseif
    assert conditionals[1].condition == 'a > 2.0'
    assert not conditionals[1].has_elseif
    assigns = FindNodes(Assignment).visit(routine.body)
    assert len(assigns) == 5
    assert assigns[0].lhs == 'b' and assigns[0].rhs == 'b + 2.0'
    assert assigns[1].lhs == 'b' and assigns[1].rhs == 'b + a'
    assert assigns[2].lhs == 'a' and assigns[2].rhs == 'a + b'
    assert assigns[3].lhs == 'a' and assigns[3].rhs == 'a + 5.0'
    assert assigns[4].lhs == 'a' and assigns[4].rhs == 'a + 1.0'
```

This fails as expected, because the `has_elseif` flag remains for the last conditional although the conditional nested into the else branch has vanished:

```
self = Conditional::

    def __post_init__(self):
        super().__post_init__()
        assert self.condition is not None

        if self.body is not None:
            assert isinstance(self.body, tuple)
            assert all(isinstance(c, Node) for c in self.body)  # pylint: disable=not-an-iterable

        if self.has_elseif:
            assert len(self.else_body) == 1
>           assert isinstance(self.else_body[0], Conditional)  # pylint: disable=unsubscriptable-object
E           AssertionError

loki/ir.py:610: AssertionError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
[36] > /etc/ecmwf/nfs/dh1_home_a/nabr/loki/naml-routine-inline-pragma/loki/ir.py(610)__post_init__()
-> assert isinstance(self.else_body[0], Conditional)  # pylint: disable=unsubscriptable-object
   5 frames hidden (try 'help hidden_frames')
```

reuterbal (changes requested):
Many thanks, looks all good except for the `has_elseif` situation: That one I think isn't covered with the tests you added - I've proposed an addition which fails for me in the way I expected it to. Would you mind taking a look at that?

tests/test_transform_dead_code.py
mlange05: Aha, I knew I forgot something! :wink: Thanks for pointing this out, I've adjusted the test now accordingly and fixed the bug.

loki/transform/transform_dead_code.py
```diff
@@ -63,4 +64,5 @@ def visit_Conditional(self, o, **kwargs):
         if condition == 'False':
             return else_body
 
-        return self._rebuild(o, tuple((condition,) + (body,) + (else_body,)))
+        has_elseif = isinstance(else_body, Conditional)
```
reuterbal, line 67: Hm, not sure this works as intended. The `else_body` should always be a tuple, not an individual node - and it might misclassify genuinely nested (albeit equivalent) conditionals as `has_elseif`.
I think something like this might work:
```suggestion
        has_elseif = o.has_elseif and else_body and isinstance(else_body[0], Conditional)
```

transformations/transformations/single_column_coalesced.py
```diff
@@ -296,14 +312,23 @@ def process_kernel(self, routine):
         if self.resolve_sequence_association:
             transform_sequence_association(routine)
 
-        # Perform full source-inlining for member subroutines if so requested
+        # Perform full source-inlining for member subroutines
         if self.inline_members:
             inline_member_procedures(routine)
```
skarppinen, line 317: Should this call still be changed to `inline_member_procedures(routine, allowed_aliases=(self.horizontal.index,))`?

loki/transform/transform_dead_code.py
mlange05: Gotcha! Many thanks, fixed now (with additional testing).

transformations/transformations/single_column_coalesced.py
mlange05: Ah, good catch. I have no means of testing this, but seems easy enough, so added as suggested.

reuterbal (approved):
Thanks, that looks great now. Ready to merge! :shipit:

PR #199, opened 2023-12-15
introduce 'flatten_arrays()' (to overcome pointer hack)

MichaelSt98:
**Usable and tested as drop in replacement for the CLOUDSC Loki C transpilation**

* do we want to keep the original *pointer hack* as another option or just replace it?

Convert e.g., 

```fortran
  INTEGER, INTENT(INOUT) :: x1(l1)
  INTEGER, INTENT(INOUT) :: x2(l2, l1)
  INTEGER, INTENT(INOUT) :: x3(l3, l2, l1)
  INTEGER, INTENT(INOUT) :: x4(l4, l3, l2, l1)
  DO i1=1,l1
    x1(i1) = ...
    DO i2=1,l2
      x2(i2, i1) = ...
      DO i3=1,l3
        x3(i3, i2, i1) = ...
        DO i4=1,l4
          x4(i4, i3, i2, i1) = ...
        END DO
      END DO
    END DO
  END DO
```

to:

```fortran
  INTEGER, INTENT(INOUT) :: x1(l1)
  INTEGER, INTENT(INOUT) :: x2(l2*l1)
  INTEGER, INTENT(INOUT) :: x3(l3*l2*l1)
  INTEGER, INTENT(INOUT) :: x4(l4*l3*l2*l1)
  DO i1=1,l1
    x1(i1) = ...
    DO i2=1,l2
      x2(i2 + l2*(i1 + -1)) = ...
      DO i3=1,l3
        x3(i3 + l3*(i2 + l2*(i1 + -1) + -1)) = ...
        DO i4=1,l4
          x4(i4 + l4*(i3 + l3*(i2 + l2*(i1 + -1) + -1) + -1)) = ...
        END DO
      END DO
    END DO
  END DO
```

**However:**

currently not handled are arrays like 

```fortran
NTEGER, INTENT(INOUT) :: x4(0:l4, 0:l3, 0:l2, 0:l1)
```

This complicates things a lot and the question is how to handle this especially regarding the f2c transpilation... thus, whether this utility should handle that or whether another utility/transformation before flattening the arrays is more appropriate.

### DIFF
--- a/loki/transform/fortran_c_transform.py
+++ b/loki/transform/fortran_c_transform.py
@@ -11,8 +11,8 @@ from collections import OrderedDict
 from loki.transform.transformation import Transformation
 from loki.transform.transform_array_indexing import (
     shift_to_zero_indexing, invert_array_indices,
-    resolve_vector_notation, normalize_range_indexing,
-    normalize_array_access, flatten_arrays
+    resolve_vector_notation, normalize_array_shape_and_access,
+    flatten_arrays
 )
 from loki.transform.transform_associates import resolve_associates
 from loki.transform.transform_utilities import (
@@ -368,8 +368,7 @@ class FortranCTransformation(Transformation):
 
         # Clean up Fortran vector notation
         resolve_vector_notation(kernel)
-        normalize_array_access(kernel)
-        normalize_range_indexing(kernel)
+        normalize_array_shape_and_access(kernel)
 
         # Convert array indexing to C conventions
         # TODO: Resolve reductions (eg. SUM(myvar(:)))
--- a/loki/transform/fortran_c_transform.py
+++ b/loki/transform/fortran_c_transform.py
@@ -50,9 +50,8 @@ class FortranCTransformation(Transformation):
     # Set of standard module names that have no C equivalent
     __fortran_intrinsic_modules = ['ISO_FORTRAN_ENV', 'ISO_C_BINDING']
 
-    def __init__(self, header_modules=None, inline_elementals=True, order='F'):
+    def __init__(self, header_modules=None, inline_elementals=True):
         self.inline_elementals = inline_elementals
-        self.order = order
 
         # Maps from original type name to ISO-C and C-struct types
         self.c_structs = OrderedDict()
@@ -376,7 +375,7 @@ class FortranCTransformation(Transformation):
         # TODO: Resolve reductions (eg. SUM(myvar(:)))
         invert_array_indices(kernel)
         shift_to_zero_indexing(kernel)
-        flatten_arrays(kernel, order=self.order, start_index=0)
+        flatten_arrays(kernel, order='C', start_index=0)
 
         # Inline all known parameters, since they can be used in declarations,
         # and thus need to be known before we can fetch them via getters.
--- a/loki/transform/fortran_c_transform.py
+++ b/loki/transform/fortran_c_transform.py
@@ -11,7 +11,8 @@ from collections import OrderedDict
 from loki.transform.transformation import Transformation
 from loki.transform.transform_array_indexing import (
     shift_to_zero_indexing, invert_array_indices,
-    resolve_vector_notation, normalize_range_indexing
+    resolve_vector_notation, normalize_range_indexing,
+    flatten_arrays
 )
 from loki.transform.transform_associates import resolve_associates
 from loki.transform.transform_utilities import (
@@ -374,6 +375,7 @@ class FortranCTransformation(Transformation):
         # TODO: Resolve reductions (eg. SUM(myvar(:)))
         invert_array_indices(kernel)
         shift_to_zero_indexing(kernel)
+        flatten_arrays(kernel, order="C", start_index=0)
 
         # Inline all known parameters, since they can be used in declarations,
         # and thus need to be known before we can fetch them via getters.
--- a/loki/transform/fortran_c_transform.py
+++ b/loki/transform/fortran_c_transform.py
@@ -12,7 +12,7 @@ from loki.transform.transformation import Transformation
 from loki.transform.transform_array_indexing import (
     shift_to_zero_indexing, invert_array_indices,
     resolve_vector_notation, normalize_range_indexing,
-    flatten_arrays
+    normalize_array_access, flatten_arrays
 )
 from loki.transform.transform_associates import resolve_associates
 from loki.transform.transform_utilities import (
@@ -37,7 +37,6 @@ from loki.visitors import Transformer, FindNodes
 from loki.tools import as_tuple, flatten
 from loki.types import BasicType, DerivedType, SymbolAttributes
 
-
 __all__ = ['FortranCTransformation']
 
 
@@ -51,8 +50,9 @@ class FortranCTransformation(Transformation):
     # Set of standard module names that have no C equivalent
     __fortran_intrinsic_modules = ['ISO_FORTRAN_ENV', 'ISO_C_BINDING']
 
-    def __init__(self, header_modules=None, inline_elementals=True):
+    def __init__(self, header_modules=None, inline_elementals=True, order='F'):
         self.inline_elementals = inline_elementals
+        self.order = order
 
         # Maps from original type name to ISO-C and C-struct types
         self.c_structs = OrderedDict()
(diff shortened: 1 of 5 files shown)


loki/transform/transform_array_indexing.py
```diff
@@ -496,3 +497,46 @@ def demote_variables(routine, variable_names, dimensions):
...
+            var: var.clone(dimensions=new_dims(list(var.dimensions)[::-1], list(var.shape)[::-1]))
+            for var in FindVariables().visit(routine.body)
+                if isinstance(var, sym.Array) and var.shape and len(var.shape)
+        }
+    elif order == 'F':
+        array_map = {
+            var: var.clone(dimensions=new_dims(list(var.dimensions), list(var.shape)))
+            for var in FindVariables().visit(routine.body)
+                if isinstance(var, sym.Array) and var.shape and len(var.shape)
+        }
```
reuterbal, line 537: To improve our error reporting, I'd suggest not to use `assert` here but instead throw an exception in an else branch:
```suggestion
    if order == 'C':
        array_map = {
            var: var.clone(dimensions=new_dims(list(var.dimensions)[::-1], list(var.shape)[::-1]))
            for var in FindVariables().visit(routine.body)
                if isinstance(var, sym.Array) and var.shape and len(var.shape)
        }
    elif order == 'F':
        array_map = {
            var: var.clone(dimensions=new_dims(list(var.dimensions), list(var.shape)))
            for var in FindVariables().visit(routine.body)
                if isinstance(var, sym.Array) and var.shape and len(var.shape)
        }
    else:
        raise ValueError(f'Unsupported array order "{order}")
```

loki/transform/transform_array_indexing.py
```diff
@@ -496,3 +497,46 @@ def demote_variables(routine, variable_names, dimensions):
...
+            new_dim.extend(_dim)
+            return new_dims(new_dim, shape[:-1])
+        return as_tuple(dim)
+
+    assert order in ['F', 'C']
+    if order == 'C':
+        array_map = {
+            var: var.clone(dimensions=new_dims(list(var.dimensions)[::-1], list(var.shape)[::-1]))
+            for var in FindVariables().visit(routine.body)
+                if isinstance(var, sym.Array) and var.shape and len(var.shape)
```
reuterbal, line 530: It's ratehr unconventional to increase indentation here since you are not entering a new block/scope

loki/transform/transform_array_indexing.py
```diff
@@ -496,3 +497,46 @@ def demote_variables(routine, variable_names, dimensions):
...
+        }
+    elif order == 'F':
+        array_map = {
+            var: var.clone(dimensions=new_dims(list(var.dimensions), list(var.shape)))
```
reuterbal, line 534: I'd avoid the conversion to `list` here and the other uses. Instead, use a tuple for _dim and replace the call to `extend` in `new_dims` by `new_dim += _dim` which allows you to work directly with the tuples.

loki/transform/transform_array_indexing.py
```diff
@@ -496,3 +497,46 @@ def demote_variables(routine, variable_names, dimensions):
...
+        The subroutine in which the variables should be promoted.
+    order : str
+        Assume Fortran (F) vs. C memory/array order.
+    start_index : int
+        Assume array indexing starts with `start_index`.
+    """
+    def new_dims(dim, shape):
+        if len(dim) > 1:
+            assert not isinstance(shape[-1], sym.RangeIndex)
+            _dim = [sym.Sum((dim[-2], sym.Product((shape[-2], dim[-1] - start_index))))]
```
reuterbal, line 519: A little confused by this: You are checking that `shape[-1]` isn't a `RangeIndex`, yet `shape[-2]` is the dimension you are actually using here. 
I might be thinking too simple here, but to accomodate the mentioned situation of
```f90
INTEGER, INTENT(INOUT) :: x4(0:l4, 0:l3, 0:l2, 0:l1)
```
would only require to add something like
```suggestion
            if isinstance(shape[-2], sym.RangeIndex):
                _len = sym.Sum((shape[-2].upper, sym.Product((-1, shape[-2].lower))))
                _start_index = sym.Sum((start_index, shape[-2].lower))
            else:
                _len = shape[-2]
                _start_index = start_index
            if isinstance(shape[-1], sym.RangeIndex):
                _start_index = sym.Sum((start_index, shape[-1].lower))
            else:
                _start_index = start_index
            _dim = [sym.Sum((
                dim[-2],
                sym.Product((
                    _len,
                    sym.Sum((
                        dim[-1],
                        sym.Product((-1, _start_index))
                    ))
                ))
            ))]
```

I'm almost sure I'm making some mistake here between what shape to use for which, but this should take care of the possibly longer or shorter dimension due to the non-standard shape, and the offset relative to this.

tests/test_transform_array_indexing.py
```diff
@@ -315,3 +318,93 @@ def test_transform_demote_dimension_arguments(here, frontend):
...
+    normalize_range_indexing(routine) # Fix OMNI nonsense
+    filepath = here/(f'{routine.name}_{frontend}.f90')
+    function = jit_compile(routine, filepath=filepath, objname=routine.name)
+    x1, x2, x3, x4 = init_arguments(l1, l2, l3, l4)
+    function(x1, x2, x3, x4, l1, l2, l3, l4)
+    validate_arguments(x1, x2, x3, x4, l1, l2, l3, l4)
+
+    f_routine = Subroutine.from_source(fcode, frontend=frontend)
+    normalize_range_indexing(f_routine) # Fix OMNI nonsense
+    flatten_arrays(routine=f_routine, order='F', start_index=1)
```
reuterbal, line 391: I would propose to add some simple testing of the IR here as well. For example, look for all `Array` nodes in the body and make sure they all have `len(arr.dimension) == 1`.

reuterbal (changes requested):
Thanks so much, this is a great addition and clean solution to the original pointer hack. I think there's no need to retain the original pointer variant.

I have left a few remarks throughout, minor things mostly. The testing could benefit from a sanity check on the IR level directly, and I've proposed a potential solution for range-based shapes.

reuterbal (commented):
Very nice, thanks for this! I've left a few housekeeping remarks but otherwise this looks very promising.

Note that you may have to rebase over main again to avoid the regression tests from failing.

loki/transform/fortran_c_transform.py
```diff
@@ -368,12 +368,14 @@ def generate_c_kernel(self, routine, **kwargs):
 
         # Clean up Fortran vector notation
         resolve_vector_notation(kernel)
+        normalize_array_access(kernel)
         normalize_range_indexing(kernel)
```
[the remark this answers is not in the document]
MichaelSt98: correct

loki/transform/transform_array_indexing.py
```diff
@@ -496,3 +497,86 @@ def demote_variables(routine, variable_names, dimensions):
     routine.spec = Transformer(decl_map).visit(routine.spec)
 
     info(f'[Loki::Transform] Demoted variables in {routine.name}: {", ".join(variable_names)}')
+
+def normalize_array_access(routine):
```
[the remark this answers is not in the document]
MichaelSt98: Yes I agree, I'll go with `normalize_array_shape_and_access`

reuterbal (approved):
Great, everything looks fantastic now. 

@mlange05 do you want to confirm this is fine in offline regression tests?

mlange05 (approved):
Only had a brief look at code changes, but played with CLODUSC regression test. All seems good, so GTG from me. :shipit:

MichaelSt98:
* Old/"pointer cast hack" version
    * compiles and runs with Intel
    * won't compile with NVHPC
* New/"index arithmetic" version
    * compiles and runs with Intel
    * compiles with NVHPC and runs for `NPROMA=1,2` 
        * however, floating point exception for different `NPROMA`
        * works with arbitrary `NPROMA` compiling with `-O0` instead of `-O2` (haven't checked for `-O1`)

reuterbal:
>         * however, floating point exception for different `NPROMA`
>         * works with arbitrary `NPROMA` compiling with `-O0` instead of `-O2` (haven't checked for `-O1`)

This suggests to me that we see one of the recurring vectorization-related issues. Does this occur in the kernel itself or the validation routine?

MichaelSt98:
> > ```
> >     * however, floating point exception for different `NPROMA`
> >     * works with arbitrary `NPROMA` compiling with `-O0` instead of `-O2` (haven't checked for `-O1`)
> > ```
> 
> 
> This suggests to me that we see one of the recurring vectorization-related issues. Does this occur in the kernel itself or the validation routine?

Yes ... must be kernel itself, as the rest (including validation) is still (the original) Fortran?!

PR #201, opened 2023-12-19
Integer offset-based pool allocator transformation (Rawstack)

rolfhm:
Introduce the rawstack transformation

The transformation inserts a stack variable of size `(klon, stack_size[dtype][kind], kgpblock)` into the driver for each of the types of temporary arrays with lead dimension `klon` encountered in the kernels. The stacks are then passed to the kernels along with their size where all the temporary arrays are allocated as calculated offsets stored in integers. Access to the temporary arrays are then calculated by offsets from the offset integers.

Assumes that all variables used for dimensions are available when entering the driver and that they are kept constant.


rolfhm:
As I mentioned in the WP2 meeting, it can be limiting that all dimensions have to be known at the beginning of the the driver. In the acraneb2 dwarf, some of the arrays have dimension `YDML_PHY_MF%YRPHY3%N_SPBAND`, but this is set in `PREPARE_ACRANEB2`, which is called from the driver. One solution is to make the stacks allocatable and allocate right before the block loop and its pragmas. On the other hand, this is an artifact of the dwarf and I'm not sure we would find anything like it in the actual code.

rolfhm:
It looks like this thing has some pretty serious conflicts now. Are they worth the effort to try to resolve?

reuterbal:
Hi @rolfhm, thanks and apologies for taking so long with this. We are firmly committed to merging this development but are considering to integrate this with the existing pool allocator transformation. Therefore, we are possibly going to cherry-pick the code into a separate PR, I hope that's fine. So, I don't think there's a need to spend time on resolving conflicts for now.

reuterbal:
This has been merged with #314.

PR #202, opened 2023-12-20
Array shadowing bugfix for inliner

skarppinen:
Hello.

This PR fixes a bug in `inline_subroutine_calls`.
Consider the following example (here demonstrating with `inline_member_procedures`, which eventually calls the above):
```
from loki import Sourcefile
from loki.transform import (
    inline_member_procedures
)
fcode = """
    subroutine outer(klon)
        integer :: jg, jlon
        integer :: arr(3, 3)

        jg = 70000
        call inner2()

        contains

        subroutine inner2()
            integer :: jlon, jg 
            integer :: arr(3, 3)
            do jg=1,3
                do jlon=1,3
                   arr(jlon, jg) = 11 
                end do
            end do
        end subroutine inner2

    end subroutine outer
"""
src = Sourcefile.from_source(fcode)
routine = src.routines[0]
print(routine.to_fortran())
```

On the main branch, this prints:
```
SUBROUTINE outer (klon)
  INTEGER :: jg, jlon
  INTEGER :: arr(3, 3)
  INTEGER :: inner2_jlon, inner2_jg
  INTEGER :: inner2_arr(3, 3)

  jg = 70000
  ! [Loki] inlined child subroutine: inner2
  DO inner2_jg=1,3
    DO inner2_jlon=1,3
      inner2_arr(jlon, jg) = 11 ! Wrong 'jlon' and 'jg'
    END DO
  END DO

  CONTAINS

END SUBROUTINE outer
```
Inside the double loop, note the wrong `jlon` and `jg`. 
This occurs because the inliner renames shadowing variables only by "name", i.e `arr(jlon, jg)` is transformed only to `inner_arr(jlon, jg)` even though `jlon` and `jg` should also be renamed as they are shadowed as well.
This occurs here (`var_map` only changes name):
https://github.com/ecmwf-ifs/loki/blob/556db2ea0ae5d8475bfa64eb2d935265dba4aeb8/loki/transform/transform_inline.py#L337-L342

This PR fixes the above issue by first building a transformation map for renaming nonarray variables, and then a transformation map for renaming array variables. In the construction of the latter, the `dimensions` attribute is transformed using the transformation map for nonarray variables. There may be a more elegant way to achieve this.

There is also a new test to check that this remains fixed.

### DIFF
--- a/loki/transform/transform_inline.py
+++ b/loki/transform/transform_inline.py
@@ -14,7 +14,7 @@ from collections import defaultdict
 
 from loki.expression import (
     FindVariables, FindInlineCalls, FindLiterals,
-    SubstituteExpressions, LokiIdentityMapper,
+    SubstituteExpressions, LokiIdentityMapper
 )
 from loki.ir import Import, Comment, Assignment, VariableDeclaration, CallStatement
 from loki.expression import symbols as sym
@@ -24,8 +24,10 @@ from loki.tools import as_tuple
 from loki.logging import warning, error
 from loki.pragma_utils import pragmas_attached, is_loki_pragma
 
-from loki.transform.transform_utilities import single_variable_declaration
-
+from loki.transform.transform_utilities import (
+    single_variable_declaration,
+    recursive_expression_map_update
+)
 
 __all__ = [
     'inline_constant_parameters', 'inline_elemental_functions',
@@ -208,9 +210,6 @@ def map_call_to_procedure_body(call, caller):
          Procedure (scope) into which the callee's body gets mapped
     """
 
-    # pylint: disable=import-outside-toplevel,cyclic-import
-    from loki.transform import recursive_expression_map_update
-
     def _map_unbound_dims(var, val):
         """
         Maps all unbound dimension ranges in the passed array value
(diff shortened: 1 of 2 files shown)


reuterbal:
Thank you very much for finding this and supplying test and fix. However, I think all that's missing here is a call to [recursive_expression_map_update](https://github.com/ecmwf-ifs/loki/blob/556db2ea0ae5d8475bfa64eb2d935265dba4aeb8/loki/transform/transform_utilities.py#L417) before applying the `SubstituteExpressions`. Would you mind adding just that at the place you have flagged, and seeing whether your test still passes?

skarppinen:
Thanks @reuterbal, that's a lot simpler and works! 

Note that there was an additional local import for `recursive_expression_map_update` in another function. I couldn't think of why it was introduced as a local import rather than global. When I removed it, the tests seemed to still pass, and everything seemed to be fine, so I removed the local import and just added a global import for the function to avoid duplicate imports.

reuterbal (approved):
Fantastic, many thanks!

I don't remember what the history of the local import was, but the pylint annotations next to it suggest this may have been to break a cyclic import in the past. However, this does not seem to be a problem any more. Thanks for changing this!

mlange05 (approved):
Many thanks for catching this @skarppinen and apologies, I must have dropped this in the recent refactoring. GTG from me. :shipit:

Issue #176, opened 2023-10-19
Recursive SCC-hoist: invalid indexing for recursively hoisted variables right below wrapper level


skarppinen:
Hello, I believe I have found a bug in the recent recursive SCC hoist implementation.
The issue is that in the first "kernel" below "wrapper", the recursively hoisted arrays are still indexed with the "block index".
Below is a reproducer. 

Consider the following Fortran code with a call tree wrapper -> depth1 -> depth2:
```
module mod3
    implicit none 
contains
    subroutine wrapper
        integer :: h_start = 1
        integer, parameter :: h_dim = 10
        integer :: h_end 
        integer :: v_start = 1
        integer, parameter :: v_dim = 20
        integer, parameter :: block_size = 30
        integer :: block_index

        real :: arr(h_dim, v_dim, block_size)

        h_end = h_dim
        do block_index = 1, block_size 
        call depth1(h_start, h_end, h_dim, v_start, v_dim, &
            arr(:, :, block_index))
        end do
        print *, "Sum of array is ", sum(arr)
    end subroutine wrapper

    subroutine depth1(h_start, h_end, h_dim, v_start, v_dim, array)
        ! Arguments.
        integer, intent(in) :: h_start
        integer, intent(in) :: h_end
        integer, intent(in) :: h_dim
        integer, intent(in) :: v_start
        integer, intent(in) :: v_dim
        real, intent(inout) :: array(h_dim, v_dim)

        ! Non-array local variables.
        integer :: v_index
        integer :: h_index

        ! Temporary arrays.
        real :: tmp1(h_dim, v_dim)
        real :: tmpx(h_dim, v_dim)

        do v_index = v_start, v_dim
            do h_index = h_start, h_end
                tmp1(h_index, v_index) = exp(log(real(h_index)) + log(real(v_index)) - 1.0)
            end do
        end do

        do v_index = v_start, v_dim
            do h_index = h_start, h_end
                array(h_index, v_index) = exp(tmp1(h_index, v_index) + 0.25)
            end do
        end do

        do v_index = v_start, v_dim
            do h_index = h_start, h_end
                array(h_index, v_index) = log(tmp1(h_index, v_index)) + array(h_index, v_index) * 2.0
            end do
        end do

        call depth2(h_start, h_end, h_dim, v_start, v_dim, array)
    end subroutine depth1

    subroutine depth2(h_start, h_end, h_dim, v_start, v_dim, array)
        integer, intent(in) :: h_start
[...]

PR #206, opened 2024-01-11
Fix for SCC-HOIST, regarding wrong (hoisted) argument(s) (indexing) i…

MichaelSt98:
…n calls to nested subroutines

Fix for #176 - "Recursive SCC-hoist: invalid indexing for recursively hoisted variables right below wrapper level".


mlange05 (approved):
Agreed. Looks neat and well tested, many thanks. GTG from me. :shipit:

PR #207, opened 2024-01-12
Refactored Global Variable Offload

reuterbal:
This is a refactoring of the Global Variable Offload in preparation for the scheduler rewrite, where items correspond to scopes (and therefore no longer each global variable is captured as an explicit dependency in the graph).

The notable change is that it splits the Transformation again into separate analysis and transformation passes. (The irony, that I was the one requesting the merge into one originally, is not lost on me ;-) )

The analysis is kept as general as possible, to provide more of a generic "inform me about the use of global variables in a call tree" view of things rather than tailored to offloading as before - the hope is that will make it useful also for other transformations in the future.
The intelligence of translating this into offloading is kept inside the transformation itself.


awnawab (approved):
All looks good to me, and the global variable offload transformation is a lot more elegant now, thanks! I've also regression tested on ecWAM with offload, validation still passes. Just as an aside, unless I've missed something, I don't think this solves the problem you brought up a few weeks ago about potentially duplicated modules. Is this laying the groundwork for that fix?

mlange05 (approved):
Looks really great to me, and the usage of dataflow analysis to deal with some degree of complexity looks very neat to me. I can't claim to understand this is full detail, test coverage is great, so GTG from me! :shipit:

PR #208, opened 2024-01-15
f2c transpile via convert

MichaelSt98:
C transpilation via `convert()`

* tested with CLOUDSC branch: [nams_f2c_transpile_via_convert](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/compare/develop...nams_f2c_transpile_via_convert)
* do we already want to get rid of `transpile()`?


loki/transform/fortran_c_transform.py
```diff
@@ -64,7 +65,10 @@ def transform_file(self, sourcefile, **kwargs):
             self.transform_subroutine(routine, **kwargs)
```
reuterbal, line 65: The `transform_file` method is redundant and should be removed. Would you mind including this here, as part of the initial house keeping?

loki/transform/fortran_c_transform.py
```diff
@@ -50,8 +50,9 @@ class FortranCTransformation(Transformation):
     # Set of standard module names that have no C equivalent
     __fortran_intrinsic_modules = ['ISO_FORTRAN_ENV', 'ISO_C_BINDING']
 
-    def __init__(self, header_modules=None, inline_elementals=True):
+    def __init__(self, header_modules=None, inline_elementals=True, path=None):
```
reuterbal, line 53: Would you mind adding a docstring for the `path` parameter? (And, while you're at it, also `header_modules` and `inline_elementals`?)

loki/transform/fortran_c_transform.py
```diff
@@ -64,7 +65,10 @@ def transform_file(self, sourcefile, **kwargs):
             self.transform_subroutine(routine, **kwargs)
 
     def transform_module(self, module, **kwargs):
-        path = Path(kwargs.get('path'))
+        if self.path is None:
```
reuterbal, line 68: The `transform_module` method should never be invoked because the `item_filter` is not changed from the default to include anything beyond `SubroutineItem`:

https://github.com/ecmwf-ifs/loki/blob/3af8578c629cdb50b1c0c9ef7446a71eafa378e9/loki/transform/transformation.py#L78-L83

I suspect it is, however, used via direct invocation on files or in the test base?

scripts/loki_transform.py
```diff
@@ -199,7 +201,7 @@ def convert(
 
     # Now we instantiate our transformation pipeline and apply the main changes
     transformation = None
-    if mode in ['idem', 'idem-stack']:
+    if mode in ['idem', 'idem-stack', "c"]:
```
reuterbal, line 204: Why is this required? `IdemTransformation` is a do-nothing, so should not be necessary for the F2C transpilation.

scripts/loki_transform.py
```diff
@@ -268,10 +270,19 @@ def transform_subroutine(self, routine, **kwargs):
...
-    # Housekeeping: Inject our re-named kernel and auto-wrapped it in a module
-    scheduler.process( ModuleWrapTransformation(module_suffix='_MOD') )
-    mode = mode.replace('-', '_')  # Sanitize mode string
-    scheduler.process( DependencyTransformation(suffix=f'_{mode.upper()}', module_suffix='_MOD') )
+    mode = mode.replace('-', '_')  # Sanitize mode string    
+    if mode in ["c"]:
+        f2c_transformation = FortranCTransformation(path=build)
+        scheduler.process(f2c_transformation)
+        for h in definitions:
+            f2c_transformation.apply(h, role='header')
```
reuterbal, line 278: Ah, here it is applied directly to a file, but the F2C Transformation is not marked as recursing to internal scopes - instead it does this internally. I think we should take this opportunity to make the Transformation compliant with the manifest parameterisation. I'll explain in the overall review comment what I mean.

Secondly, I'm not too thrilled by the fact that we provide files as headers (which, to me, implies a "read-only for enrichment" meaning) and then apply transformations to them. Since the purpose here is to deal with derived types, which we cannot capture in the dependency graph in the current Scheduler implementation, we may have to live with this for now, but that will be different with #213. To avoid that others copy this pattern, would you mind including a comment here that explains that this is temporary?
Longer term, this transformation step to generate accessor methods for derived types should become a separate Transformation that is applied to `ModuleItem` only.

reuterbal (changes requested):
Many thanks, very much looking forward to bringing this in!

However, I think we should use this opportunity to go a little further:

- [ ]  Remove the `transpile` entry point from `loki_transform.py`. Going forward, we don't want to use this anymore so better get rid of it right away. The only place that uses this is CLOUDSC. To avoid breaking the regression tests, I would convert the [`loki_transform_transpile`](https://github.com/ecmwf-ifs/loki/blob/cb411156cf1c48feda2eb8f89016e402f5b6c40d/cmake/loki_transform.cmake#L457) CMake wrapper to a compatibility layer that reshuffles the provided options, prints a big deprecation warning, and then calls `loki_transform`, similar to what [`loki_transform_convert`](https://github.com/ecmwf-ifs/loki/blob/cb411156cf1c48feda2eb8f89016e402f5b6c40d/cmake/loki_transform.cmake#L379) does.

- [ ] Make the Fortran2CTransformation compliant with the Transformation manifests. I'm not sure what would be the minimal invasive way (we can have an offline discussion to talk through the different options), but one possibility would be to remove all the explicit recursion from the transformation, and change the invocation for the headers such that it is invoked for every module in those.

I would, however, not include anything beyond that in this PR to not expand the scope too much. For example, a subsequent PR should split the F2C Transformation to generate derived type accessors ("the header transformation") in a separate Transformation class.

MichaelSt98:
`transpile()` is now removed from `scripts/loki_transform.py` as the F2C transpilation is integrated in `convert()`.

Unfortunately it was not possible to "re-route" the CMake `loki_transform_transpile` to `loki_transform`/`loki_transform_convert` because the `CONFIG` argument can't be generically reconstructed.

Depends on [CLOUDSC PR 69](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/pull/69).

reuterbal (approved):
Thanks, very happy with this now! This is a great base for the further developments.

Once @mlange05 has had a look, I suggest the following (as acted out before) as a merge strategy:

1. Remove the commit that points to the custom CLOUDSC branch
2. Merge the PR with regression tests failing
3. Tests on CLOUDSC PR should be successful now
4. Merge the CLOUDSC PR
5. Run regression tests on Loki main again, should now be successful.

mlange05 (approved):
Needs a quick rebase, but otherwise GTG I think. Merge strategy sounds good too @reuterbal . :shipit:

PR #211, opened 2024-01-17
Fixes to minor issues related to SCC HOIST

skarppinen:
Hello.

This PR contains two very small changes. I discuss them below.

### DIFF
--- a/loki/transform/transform_hoist_variables.py
+++ b/loki/transform/transform_hoist_variables.py
@@ -85,6 +85,7 @@ from loki.tools.util import is_iterable, as_tuple, CaseInsensitiveDict
 from loki.visitors import Transformer, FindNodes
 from loki.transform.transformation import Transformation
 from loki.transform.transform_utilities import single_variable_declaration
+from loki.bulk.item import SubroutineItem
 import loki.expression.symbols as sym
 
 
@@ -150,6 +151,8 @@ class HoistVariablesAnalysis(Transformation):
         call_map = CaseInsensitiveDict((str(call.name), call) for call in calls)
 
         for child in successors:
+            if not isinstance(child, SubroutineItem):
+                continue
             arg_map = dict(call_map[child.local_name].arg_iter())
             hoist_variables = []
             for var in child.trafo_data[self._key]["hoist_variables"]:
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -225,9 +225,6 @@ def convert(
         ))
         scheduler.process( SCCDemoteTransformation(horizontal=horizontal))
         scheduler.process( SCCRevectorTransformation(horizontal=horizontal))
-        scheduler.process( SCCAnnotateTransformation(
-            horizontal=horizontal, vertical=vertical, directive=directive, block_dim=block_dim
-        ))
 
     if mode == 'scc-hoist':
         # Apply recursive hoisting of local temporary arrays.
@@ -235,6 +232,10 @@ def convert(
         # direction through the call graph to gather temporary arrays.
         scheduler.process( HoistTemporaryArraysAnalysis(dim_vars=(vertical.size,)) )
         scheduler.process( SCCHoistTemporaryArraysTransformation(block_dim=block_dim) )
+
+    scheduler.process( SCCAnnotateTransformation(
+            horizontal=horizontal, vertical=vertical, directive=directive, block_dim=block_dim
+    ))
 
     if mode in ['cuf-parametrise', 'cuf-hoist', 'cuf-dynamic']:
         # These transformations requires complex constructor arguments,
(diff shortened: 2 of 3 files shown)


loki/transform/transform_hoist_variables.py
```diff
@@ -150,6 +151,8 @@ def transform_subroutine(self, routine, **kwargs):
         call_map = CaseInsensitiveDict((str(call.name), call) for call in calls)
 
         for child in successors:
+            if not isinstance(child, SubroutineItem):
```
skarppinen, line 154: If SCC HOIST is used together with `GlobalVarOffloadTransformation`, some of the successors may be `Item`s. This loop only seems to look at `SubroutineItem`s, therefore adding this if statement.

loki/transform/transform_hoist_variables.py
```diff
@@ -150,6 +151,8 @@ def transform_subroutine(self, routine, **kwargs):
...
+                continue
             arg_map = dict(call_map[child.local_name].arg_iter())
```
skarppinen, line 156: If SCC HOIST is used together with `GlobalVarOffloadTransformation`, some of `successors` may be `Item`s (corresponding to the imported global variables). Therefore, adding this if statement to skip processing for `Item`s other than `SubroutineItem`s. To me, everything here seems to indicate this loop only processes subroutines.

scripts/loki_transform.py
```diff
@@ -236,6 +233,10 @@ def convert(
         scheduler.process( HoistTemporaryArraysAnalysis(dim_vars=(vertical.size,)) )
         scheduler.process( SCCHoistTemporaryArraysTransformation(block_dim=block_dim) )
 
+    scheduler.process( SCCAnnotateTransformation(
+            horizontal=horizontal, vertical=vertical, directive=directive, block_dim=block_dim
+    ))
+
     if mode in ['cuf-parametrise', 'cuf-hoist', 'cuf-dynamic']:
```
skarppinen, line 240: Notice that this change moves the location of `SCCAnnotate` "one step" further in the SCC pipeline. The reasoning for this is that `SCCHoistTemporaryArraysTransformation` does the SCC hoisting, introducing new arguments to subroutines, that have to be marked as "present", This change ensures that `SCCAnnotate` adds these pragmas for the new arguments as well.

I am not sure whether this should actually be moved after the next if statement even.

loki/transform/transform_hoist_variables.py
reuterbal, line 154: This is currently not covered in the test base, which is why this bug existed in the first place.
I would suggest to amend the existing test [`test_single_column_coalesced_hoist_openacc`](https://github.com/ecmwf-ifs/loki/blob/c86b83aea180c6823ba0615ce395dc7b8dd51e35/transformations/tests/test_single_column_coalesced.py#L711) by moving the integer `c` to a new module and importing it in the kernel. 

Here's an untested implementation suggestion:

```python
fcode_module = """
module my_scaling_value_mod
    implicit none
    REAL :: c = 5.345
end module my_scaling_value_mod
""".strip()

[...]

module_source = Sourcefile.from_source(fcode_module, frontend=frontend)
module = module_source['my_scaling_value_mod']

[...]

kernel.enrich(module) # do this before enriching the driver

[...]

module_item = GlobalVarImportItem(name='my_scaling_value_mod#c', source=module_source)

[...]

    # This replaces the current implementation
    for transform in scc_transform:
        transform.apply(driver, role='driver', item=driver_item, targets=['compute_column'], successors=[kernel_item])
        transform.apply(kernel, role='kernel', item=kernel_item), successors=[module_item])

```

reuterbal (changes requested):
Thanks, this looks great to me.

I let @mlange05 assess whether moving the Annotate transformation is safe to do but in principle I think it would be great to have this later in the pipeline and develop it into a more generic offload annotation mechanism that will eventually be able to take on that duty for other transformations and programming models as well.

Otherwise, I have left two remarks. One of them relates to a suggestion to capture the original issue in the test base to ensure we don't re-introduce that bug in the future.

scripts/loki_transform.py
```diff
@@ -236,6 +233,10 @@ def convert(
         scheduler.process( HoistTemporaryArraysAnalysis(dim_vars=(vertical.size,)) )
         scheduler.process( SCCHoistTemporaryArraysTransformation(block_dim=block_dim) )
 
+    scheduler.process( SCCAnnotateTransformation(
+            horizontal=horizontal, vertical=vertical, directive=directive, block_dim=block_dim
+    ))
```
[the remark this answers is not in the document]
skarppinen: I agree, this is a good idea.

loki/transform/transform_hoist_variables.py
skarppinen: Thanks, this was a good idea. 

I changed the test to the best of my ability, please check to see it is OK. I had to add the successors argument to the analysis part (a bit later than you suggested). I also ran the modified test in the main branch to make sure that it fails there.

reuterbal (approved):
Thanks for the changes and the additional testing. All looks good from my side but I let @mlange05 confirm.

skarppinen:
Thanks for reviewing @reuterbal.

mlange05 (approved):
Hi, I can confirm this works as expected and is GTG from me. :shipit:

PR #212, opened 2024-01-17
Minor fixes to frontends and IR nodes

reuterbal:
This is intended to be the last maintenance PR of small fixes that are lifted from the Scheduler rewrite. Mostly centered around an improved representation of interfaces and recognizing them as an external-facing declaration in modules, and a consistent representation of procedure bindings in the type information of procedure symbols.

It includes also the stubs to perform partial (i.e., only with a subset of patterns) parsing in the REGEX frontend, including (partial) re-parsing.

### DIFF
--- a/loki/program_unit.py
+++ b/loki/program_unit.py
@@ -9,7 +9,10 @@ from abc import abstractmethod
 
 from loki import ir
 from loki.expression import Variable
-from loki.frontend import Frontend, parse_omni_source, parse_ofp_source, parse_fparser_source
+from loki.frontend import (
+    Frontend, parse_omni_source, parse_ofp_source, parse_fparser_source,
+    RegexParserClass
+)
 from loki.logging import debug
 from loki.scope import Scope
 from loki.tools import CaseInsensitiveDict, as_tuple, flatten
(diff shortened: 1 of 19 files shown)


loki/program_unit.py
```diff
@@ -54,16 +57,20 @@ class ProgramUnit(Scope):
...
-                       ast=None, source=None, rescope_symbols=False, incomplete=False):
+                       ast=None, source=None, rescope_symbols=False, incomplete=False,
+                       parser_classes=None):
         # Common properties
         assert name and isinstance(name, str)
         self.name = name
         self._ast = ast
         self._source = source
         self._incomplete = incomplete
+        self._parser_classes = parser_classes
```
mlange05, line 73: I'm not sure I understand why the `ParserClass` (a static property, no?) needs to be passed through to the object constructor all of a sudden. Should the regex parser class not be known statically from the type? What am I missing?

mlange05 (commented):
All in all looks good to me - the only thing I do not understand is why we now push the parser class through the constructors. I'll leave this on comment for now, but once the question is answered, this should be GTG.

loki/program_unit.py
reuterbal: Ah, apologies, that wasn't explained very well.

This property does not represent the `ParserClass` that is responsible for matching the specific `ProgramUnit`, but is for book-keeping which parser classes were active when the code has been parsed with the REGEX frontend. The rationale is that we do now (in fact, did before but now it's tested and exposed) support an incremental parsing with the REGEX frontend. For example:

In a first round, we parse 10 files with the REGEX frontend and only using `RegexParserClass.ProgramUnitClass`. That will match the `Module` and `Subroutine` objects inside each of them and not recurse any further.
Then, we take one of these routines ("seed") and trigger a new parse of that file with additional parser classes, say `ImportClass | CallClass`. It will now match additional IR nodes inside the previously matched ProgramUnit, and update this property to include all parser classes that have been used so far. If the re-parse is requested with a subset of already used parser classes, this is a no-op and we return directly.

There's a corresponding test that illustrates the behaviour:

https://github.com/ecmwf-ifs/loki/blob/15f0de3da52d71d0c51c643f095e8f2966070281/tests/test_frontends.py#L608-L754

The new Scheduler builds upon this behaviour when populating the dependency graph. With the IR nodes matched that constitutes dependencies, we pick out the corresponding scope objects and repeat the exercise there: match new IR nodes, find their dependencies, pick out the next scope objects.

mlange05 (approved):
Question answered, and all looking great. GTG :shipit:

PR #213, opened 2024-01-18
The new Scheduler

reuterbal:
Draft PR to trigger CI tests for now

### DIFF
--- a/docs/source/transform.rst
+++ b/docs/source/transform.rst
@@ -35,25 +35,21 @@ transformation, only one or all of the relevant methods
 :any:`Transformation.transform_module`, or :any:`Transformation.transform_file`
 need to be implemented.
 
-*Example*: A transformation that renames every module and subroutine by
-appending a given suffix:
+*Example*: A transformation that inserts a comment at the beginning of every
+module and subroutine:
 
 .. code-block:: python
 
-    class RenameTransformation(Transformation):
+    class InsertCommentTransformation(Transformation):
 
-        def __init__(self, suffix):
-            super().__init__()
-            self._suffix = suffix
-
-        def _rename(self, item):
-            item.name += self._suffix
+        def _insert_comment(self, program_unit):
+            program_unit.spec.prepend(Comment(text='! Processed by Loki'))
 
         def transform_subroutine(self, routine, **kwargs):
-            self._rename(routine)
+            self._insert_comment(routine)
 
         def transform_module(self, module, **kwargs):
-            self._rename(module)
+            self._insert_comment(module)
 
 The transformation can be applied by calling :meth:`apply` with
 the relevant object.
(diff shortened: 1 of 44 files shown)


reuterbal:
Just a little heads-up: I force-pushed after rebasing over latest main, to make this branch compatible with GlobalVarHoist and the switch to convert for F2C processing.
This branch includes also two temporary commits to point to different ecwam/cloudsc branches - these should be removed before (squash) merging the new Scheduler

loki/bulk/scheduler.py
```diff
@@ -102,97 +154,84 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
...
+            self.item_factory.item_cache.update(definition_items)
 
     @property
-    def typedefs(self):
-        return as_tuple(flatten(module.typedefs for obj in self.obj_map.values() for module in obj.modules))
+    def sgraph(self):
+        """
+        Create and return the :any:`SGraph` constructed from the :attr:`seeds` of the Scheduler.
+        """
+        return SGraph.from_seed(self.seeds, self.item_factory, self.config)
```
mlange05, line 209: As discussed offline, this automatic re-generation of the SGraph is very expensive at the moment, even though it does seem to correctly use the caching mechanism provided by the `ItemFactory`. My suspicion is the use of `Item.dependencies` triggering a complete re-parse of each edge/node of the base graph via the REGEX frontend, as this always seems to trigger `Item.concretize_dependencies`.

In this current stage this effect makes prohibitively expensive (enrichment triggers about 6 rebuilds for me at around 40s each in ec-physics), which can be mitigated by not caching the sgraph. However, I wonder if there is a way to not "concretize" things that are already concrete, and if this would alleviate the problem somewhat?

loki/bulk/item.py
```diff
@@ -118,691 +154,1278 @@ def __eq__(self, other):
...
+        only : list of :any:`Item` classes
+            Filter the generated items to include only those provided in the list
+
+        Returns
+        -------
+        tuple
+            The list of :any:`Item` nodes
         """
-        return self.config.get('expand', False)
+        if not (dependencies := self.dependencies):
```
mlange05, line 291: So, I'm not 100% that I'm understanding this right, but to illustrate my point on expensive SGraph rebuilds, I believe this lines is triggered for determining every edge in the base graph, and will cause a regex-parse of every dependency item (via `concretize_dependencies`), regardless of whether this has been done before. I've not fully confirmed this, other than by observing that this call can take up to 2s for certain items. 

If this is the case, I'm not super-sure how to fix this, but it feels like this should be fixable.

loki/bulk/scheduler.py
reuterbal: I think you might be right!
The intention is to call make_complete to ensure the relevant parser classes have been parsed already - which should be a no-op after the initial discovery. For that reason, there is this bail-out here:

https://github.com/ecmwf-ifs/loki/blob/b1018476bb6408ee3b96adb5880968b41af0ebca/loki/program_unit.py#L290

Or at least, that's what it is intended to be. We call make_complete only with the `Item._depends_class`, which ideally will be a subset of what has been parsed already. But that's not what we are checking for... If my half-asleep brain is not completely wrong, this line should instead read as:
```python
            if self._parser_classes & parser_classes:
```

Would you mind trialling this?

docs/source/transform.rst
```diff
@@ -35,131 +35,201 @@ transformation, only one or all of the relevant methods
...
-   transformations.scc_cuf.SccCufTransformation
+   source = Sourcefile(...)  # may contain modules and subroutines
+   transformation = InsertCommentTransformation()
+   transformation.apply(source)
 
-Additionally, a number of tools for common transformation tasks are provided as
-functions that can be readily used in a step of the transformation pipeline:
+Most transformations, however, will only require modifying those parts of a file
+that are part of the call tree that is to be transformed to avoid unexpected
+side-effects. For that reason,
```
mlange05, line 103: sentence not quite finished?

docs/source/transform.rst
```diff
@@ -35,131 +35,201 @@ transformation, only one or all of the relevant methods
...
-and thus derive the dependencies within this source tree.
+of paths and builds a dependency graph from a given starting point.
+This dependency graph includes all called procedures and imported modules.
 
 Calling :any:`Scheduler.process` on a source tree and providing it with a
-:any:`Transformation` applies this transformation to all modules and routines,
-making sure that routines with the relevant :any:`CallStatement` are always
-processed before their target :class:`Subroutine`.
+:any:`Transformation` applies this transformation to all files, modules, or
+routines that are appear in the dependency graph. The exact traversal
```
mlange05, line 128: that ~are~ appear

loki/bulk/item.py
mlange05: Ok for now, follow-up tracked via #254 .

mlange05 (approved):
Ok, first of all this is an impressive piece of live-wire engineering, and the improvements from the `SGraph`, and the incremental parsing infrastructure are very powerful! Obviously, this type of major change is too large to do a detailed line-by-line review, but having played with it for a little while, I'm confident that any remaining issues can be addressed in follow-up PRs. So, I've left a couple of docstring pointers, but they are truly minor, so GTG from me! :shipit: 

And just to re-iterate, this is a major breaking change, so integration strategy and corresponding point-release management should happen as discuss offline (roll minor point release first, merge this, increment major point version number in quick follow-up release. 

And again, many thanks @reuterbal for a truly impressive feature! :grin:

docs/source/transform.rst
reuterbal: Indeed, but don't remember where I was going with this...

reuterbal:
Documentation updated, will now rebase to remove pointers to temporary regression test branches, and then merge to main.

PR #215, opened 2024-01-22
introduce flag to allow removing all derived types

MichaelSt98:
from procedure signatures (via replacing by its members). 

I will almost certainly rely on this for F2LowLevelGPU transformation(s), at least until different/more sophisticated solutions exist. Anyway, I think this is a useful in general. 

Introducing: `all_derived_types : bool, optional`

> Whether to remove all derived types from procedure signatures by
        replacing the derived type arguments using its member variables or
        only the "relevant" ones, referring to derived types with array
        members or nested derived types (default: `False`).

### DIFF
--- a/transformations/tests/test_transform_derived_types.py
+++ b/transformations/tests/test_transform_derived_types.py
@@ -100,6 +100,85 @@ end module transform_derived_type_arguments_mod
             assert member.type.dtype != BasicType.DEFERRED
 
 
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Useful addition, neatly tested and well documented!

mlange05 (approved):
Fully agree, well done! GTG from me.

PR #217, opened 2024-01-24
CMake: Fix DERIVE_ARGUMENT_SHAPE_ARRAY and argument handling

mlange05:
The re-use of the _PAR prefix means we get aliasing between loki_transform_convert and loki_convert_plan. We fix this by changing the prefix in the outer loki_transform_target.

Please note, this was discussed offline and is exclusively restricted to the CMake layer only, and thus does not come with explicit test coverage. We use this in EC-physics regression, where the use of the `--derive-argument-shape-array` is needed for performance reasons in the surface sub-branches - it's been verified to work there.

### DIFF
--- a/cmake/loki_transform.cmake
+++ b/cmake/loki_transform.cmake
@@ -223,57 +223,60 @@ endfunction()
 
 function( loki_transform_target )
 
-    set( options NO_PLAN_SOURCEDIR COPY_UNMODIFIED CPP CPP_PLAN INLINE_MEMBERS RESOLVE_SEQUENCE_ASSOCIATION )
+    set( options
+         NO_PLAN_SOURCEDIR COPY_UNMODIFIED CPP CPP_PLAN INLINE_MEMBERS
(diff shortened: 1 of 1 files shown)


cmake/loki_transform.cmake
```diff
@@ -223,57 +223,60 @@ endfunction()
...
 
-    cmake_parse_arguments( _PAR "${options}" "${single_value_args}" "${multi_value_args}" ${ARGN} )
+    cmake_parse_arguments( _PAR_T "${options}" "${single_value_args}" "${multi_value_args}" ${ARGN} )
 
     if( _PAR_UNPARSED_ARGUMENTS )
         ecbuild_critical( "Unknown keywords given to loki_transform_target(): \"${_PAR_UNPARSED_ARGUMENTS}\"")
     endif()
 
-    if( NOT _PAR_TARGET )
+    if( NOT _PAR_T_TARGET )
```
reuterbal, line 239: I like the choice `_PAR_T`. Reads in my head as "party" 🥳

reuterbal (approved):
Many thanks and apologies for the oversight and mistake in the first place. Changes look good to me!

PR #218, opened 2024-01-25
assumed size array handling for 'normalize_array_shape_and_access'

MichaelSt98:
`normalize_array_shape_and_access()` didn't handle assumed size arrays correctly. 
This is a proposed fix/extension.

### DIFF
--- a/loki/transform/transform_array_indexing.py
+++ b/loki/transform/transform_array_indexing.py
@@ -504,14 +504,16 @@ def normalize_array_shape_and_access(routine):
     """
     def is_explicit_range_index(dim):
         # return False if assumed sized array or lower dimension equals to 1
-        return (isinstance(dim, sym.RangeIndex) and not dim.lower == 1 and not dim is None
-                    and not dim.lower is None and not dim.upper is None)
+        # return (isinstance(dim, sym.RangeIndex) and not dim.lower == 1 and not dim is None
+        #             and not dim.lower is None and not dim.upper is None)
+        return (isinstance(dim, sym.RangeIndex)
+                and not (dim.lower == 1 or dim.lower is None or dim.upper is None))
 
     vmap = {}
     for v in FindVariables(unique=False).visit(routine.body):
         if isinstance(v, sym.Array):
             # skip if e.g., `array(len)`, passed as `call routine(array)`
-            if len(v.shape) != len(v.dimensions):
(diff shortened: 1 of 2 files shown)


loki/transform/transform_array_indexing.py
```diff
@@ -502,15 +502,20 @@ def normalize_array_shape_and_access(routine):
...
+    def is_explicit_range_index(dim):
+        # return False if assumed sized array or lower dimension equals to 1
+        return (isinstance(dim, sym.RangeIndex) and not dim.lower == 1 and not dim is None
+                    and not dim.lower is None and not dim.upper is None)
 
     vmap = {}
     for v in FindVariables(unique=False).visit(routine.body):
         if isinstance(v, sym.Array):
+            # skip if e.g., `array(len)`, passed as `call routine(array)`
+            if len(v.shape) != len(v.dimensions):
```
reuterbal, line 514: Shouldn't the only case be that `v.dimensions is None`? Otherwise, `shape` and `dimensions` should match or there's something fishy anyway.

tests/test_transform_array_indexing.py
```diff
@@ -368,49 +380,73 @@ def test_transform_normalize_array_shape_and_access(here, frontend, start_index)
...
         x2 = np.zeros(shape=(l2,l1,), order='F', dtype=np.int32)
         x3 = np.zeros(shape=(l3,l2,l1,), order='F', dtype=np.int32)
         x4 = np.zeros(shape=(l4,l3,l2,l1,), order='F', dtype=np.int32)
-        return x1, x2, x3, x4
+        return x1, x2, x3, x4, assumed_x1
 
     def validate_routine(routine):
         arrays = [var for var in FindVariables().visit(routine.body) if isinstance(var, Array)]
         for arr in arrays:
+            # print(f"arr: {arr} | {all(not isinstance(shape, sym.RangeIndex) for shape in arr.shape)}")
```
reuterbal, line 409:
```suggestion
```

reuterbal (changes requested):
Thanks, looks good. Two suggestions where I think stuff could be simplified, otherwise GTG!

reuterbal (approved):
Thanks for making these changes, looks good now!

mlange05 (approved):
Looks great, going in now. :shipit:

Issue #216, opened 2024-01-23
issue with acc loop gang 


ecossevin:
SUBROUTINE TOTO(A,B)

IMPLICIT NONE
REAL, INTENT(IN) :: A
REAL, INTENT(INOUT) :: B

!$ACC PARALLEL LOOP GANG &
!$ACC&PRESENT(ZRDG_LCVQ,ZFLU_QSATS,ZRDG_CVGQ) &
!$ACC&PRIVATE (JBLK) &
!$ACC&VECTOR_LENGTH (YDCPG_OPTS%KLON)

END SUBROUTINE TOTO

 from loki import *
path="sub.F90"
file=Sourcefile.from_file(path)
routine=file["TOTO"]
print(fgen(routine))

**loki removes the "&" after !$ACC PARALLEL LOOP GANG statement.**

mlange05:
Hi @ecossevin I can confirm that this is indeed an issue (and wrong). The problem seem to be that the lack of whitespaces in the second pragma line `!$ACC& ...` seems to defeat the pragma parsers and doesn't register as part of the pragma (`!$ACC & ...` works). I'll investigate further and hopefully find a fix.

PR #221, opened 2024-01-26
Frontend: Fix bug for multi-line pragmas and add short test

mlange05:
Small bug-fix and test to fix erroneous parsing of multi-line pragmas with missing whitespaces after the keyword. Many thanks for @ecossevin for reporting this (see issue #216).


reuterbal (approved):
Thanks, the fix looks straightforward and is well-tested. GTG!

PR #222, opened 2024-01-29
Enabling SCC-Stack for EC-physics, part 2

mlange05:
_Please note that this is a replay of some of the changes in Loki #135 , brought together with an alternative solution for determining stack-size across mulitple libs. A lot of the credit for this excellent work goes to @MichaelSt98 (sorry, in the rebase I dropped the commit authorship when separating the individual steps)._

The main purpose of the PR is to apply the remaining fixes to the `PoolAllocatorTransformation` in unison and enable it as the "scc-stack" mode in the `loki_transform.py convert`. With this, the PR enables the SCC-stack transformation for the current EC-physics demonstrator via three types of changes:
* Enabling the SCC-Stack transformation as a "mode" in convert (bug fix to deal with hyphens in "mode")
* Allowing the analysis part of SCC-stack to run past "ignore" items in the scheduler by changing the filtering rules
* Bringing forward a few of the fixes to the `PoolAllocatorTransformation` that were part of the original PR

The scheduler changes are probably temporary in nature, as we can likely do this better after the rewrite. For now, we change the behaviour to instead of stopping parsing one level below "ignore" items, we keep extending but keep inheriting the "ignored" property. These items are then available for transformations with the explicit property `process_ignored_items`, so that inter-procedural analysis passes may reach the leaves to gather and agglomerate the necessary information. 

The various changes to the pool allocator bug-fixes, changes to the argument insertion (strictly via keyword-arguments now) and enforcing alignment. I've also tried to tidy up the test and imports a tiny bit in the process.

In a little more detail:
* Enabling the SCC-Stack transformation as a "mode" in convert (bug fix to deal with hyphens in "mode")
* Allow "ignore" items to be processed by marked analysis passes and use for `TemporariesPoolAllocatorAnalysis`
* Bug-fix for type handling in base expression mapper
* Adjusting stack size derivation to enforce alignment via `(max(c_sizeof<...>, 8)` and according test adjustments(!)
* Insertion of stack-size at a fixed location in the driver that is marked via `!$loki stack-size` - this is planned to be done more generally at a later stage)
* Strictly use keyword-argument passing for the stack variables to avoid cross-fire with the array-shape adjustment trafo 

Due to the partial replay nature of the PR, the test updates are done in bulk towards the end. @MichaelSt98 please check if the original features and intentions of your fixes are still fine.

### DIFF
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -369,12 +369,11 @@ class Scheduler:
 
             # Append child to work queue if expansion is configured
             if item.expand:
-                # Do not propagate to dependencies marked as "ignore"
-                # Note that, unlike blackisted items, "ignore" items
-                # are still marked as targets during bulk-processing,
-                # so that calls to "ignore" routines will be renamed.
-                if child.local_name in item.ignore:
-                    continue
+                # Mark children as "ignored", which means they may be
+                # used for certain analysis passes, but are not part
+                # of the injected changes for this batch-transformation
+                if item.ignored or child.local_name in item.ignore:
+                    child.ignored = True
 
                 if child not in self.item_map:
                     new_items += [child]
@@ -575,9 +574,15 @@ class Scheduler:
                             continue
 
                     _item = items[0]
+                    if _item.ignored and not transformation.process_ignored_items:
+                        continue
+
                     transformation.apply(items[0].source, items=items)
             else:
                 for item in traversal:
+                    if item.ignored and not transformation.process_ignored_items:
+                        continue
+
                     if item_filter and not isinstance(item, item_filter):
                         continue
 
@@ -709,6 +714,9 @@ class Scheduler:
         sources_to_transform = []
 
         for item in self.items:
+            if item.ignored:
+                continue
+
             sourcepath = item.path.resolve()
             newsource = sourcepath.with_suffix(f'.{mode.lower()}.F90')
             if buildpath:
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -372,8 +372,8 @@ class Scheduler:
                 # Mark children as "ignored", which means they may be
                 # used for certain analysis passes, but are not part
                 # of the injected changes for this batch-transformation
-                if item.ignored or child.local_name in item.ignore:
-                    child.ignored = True
+                if item.is_ignored or child.local_name in item.ignore:
+                    child.is_ignored = True
 
                 if child not in self.item_map:
                     new_items += [child]
@@ -574,13 +574,13 @@ class Scheduler:
                             continue
 
                     _item = items[0]
-                    if _item.ignored and not transformation.process_ignored_items:
+                    if _item.is_ignored and not transformation.process_ignored_items:
                         continue
 
                     transformation.apply(items[0].source, items=items)
             else:
                 for item in traversal:
-                    if item.ignored and not transformation.process_ignored_items:
+                    if item.is_ignored and not transformation.process_ignored_items:
                         continue
 
                     if item_filter and not isinstance(item, item_filter):
@@ -714,7 +714,7 @@ class Scheduler:
         sources_to_transform = []
 
         for item in self.items:
-            if item.ignored:
+            if item.is_ignored:
                 continue
 
             sourcepath = item.path.resolve()
--- a/tests/test_fgen.py
+++ b/tests/test_fgen.py
@@ -146,6 +146,8 @@ end subroutine test_fgen
     out = fgen(routine, linewidth=132)
     for line in out.splitlines():
         assert line.count('&') <= 2
+        if line.count('&') == 2:
+            assert len(line.split('&')[1]) > 60
 
 
 @pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Loki likes only valid code')]))
--- a/tests/test_fgen.py
+++ b/tests/test_fgen.py
@@ -148,6 +148,28 @@ end subroutine test_fgen
         assert line.count('&') <= 2
 
 
+@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Loki likes only valid code')]))
+def test_multiline_inline_conditional_long(frontend):
+    """
+    Test correct formatting of an inline :any:`Conditional` that
+    that creates a particularly long line.
+    """
+    fcode = """
+subroutine test_inline_multiline_long(array, flag)
+  real, intent(inout) :: array
+  logical, intent(in) :: flag
+
+  if (flag) call a_subroutine_with_an_exquisitely_loong_and_expertly_chosen_name_and_a_few_keyword_arguments(my_favourite_array=array)
+end subroutine test_inline_multiline_long
+    """.strip()
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+
+    out = fgen(routine, linewidth=132)
+    for line in out.splitlines():
+        assert len(line) < 132
+        assert line.count('&') <= 2
+
+
 @pytest.mark.parametrize('frontend', available_frontends())
 def test_fgen_save_attribute(frontend):
     """
--- a/transformations/tests/test_pool_allocator.py
+++ b/transformations/tests/test_pool_allocator.py
@@ -297,19 +297,15 @@ end module kernel_mod
     calls = FindNodes(CallStatement).visit(driver.body)
     assert len(calls) == 1
     if nclv_param:
-        if check_bounds:
-            expected_args = ('1', 'nlon', 'nlon', 'nz', 'field1(:,b)', 'field2(:,:,b)', 'ylstack_l', 'ylstack_u')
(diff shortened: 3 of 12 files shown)


transformations/transformations/pool_allocator.py
```diff
@@ -730,7 +764,7 @@ def inject_pool_allocator_into_calls(self, routine, targets):
         for call in FindInlineCalls().visit(routine.body):
             if call.name.lower() in [t.lower() for t in targets]:
                 parameters = call.parameters
-                call_map[call] = call.clone(parameters=parameters + new_vars)
+                call_map[call] = call.clone(parameters=parameters + new_arguments)
```
MichaelSt98, line 767: Can we have the `args` vs. `kwargs` problem for `InlineCalls` as well?

transformations/tests/test_pool_allocator.py
```diff
@@ -275,19 +297,15 @@ def test_pool_allocator_temporaries(frontend, generate_driver_stack, block_dim,
...
     else:
-        if check_bounds:
-            expected_args = ('1', 'nlon', 'nlon', 'nz', '2', 'field1(:,b)', 'field2(:,:,b)', 'ylstack_l', 'ylstack_u')
-            assert calls[0].arguments == expected_args
-        else:
-            expected_args = ('1', 'nlon', 'nlon', 'nz', '2', 'field1(:,b)', 'field2(:,:,b)', 'ylstack_l')
-            assert calls[0].arguments == expected_args
+        expected_args = ('1', 'nlon', 'nlon', 'nz', '2', 'field1(:,b)', 'field2(:,:,b)')
+    if check_bounds:
+        expected_kwargs = (('YDSTACK_L', 'ylstack_l'), ('YDSTACK_U', 'ylstack_u'))
```
MichaelSt98, line 304: minor thing: not always consistent with upper and lower case ...

MichaelSt98 (approved):
Looks great!
Haven't found anything I can flag and seems to have all the relevant features and fixes.

One thing: Can the "`args`" vs. "`kwargs`" thingy become a problem for `InlineCalls` as well?!

I don't see any problem with passing the stack always as `kwargs`. However, we should probably think about a generic solution for this or rather another utility (transformation) to solve this (and e.g., convert all `kwargs` with `args` or at least sort them) as this could become a problem e.g., for F2C transpilation?!

transformations/tests/test_pool_allocator.py
```diff
@@ -104,13 +110,13 @@ def test_pool_allocator_temporaries(frontend, generate_driver_stack, block_dim,
         integer(kind=8) :: ylstack_l
         {'integer(kind=8) :: ylstack_u' if check_bounds else ''}
 
-        {'istsz = c_sizeof(real(1,kind=jprb))*nlon/c_sizeof(real(1,kind=jprb))+c_sizeof(real(1,kind=jprb))*nlon*nz/c_sizeof(real(1,kind=jprb))+c_sizeof(real(1,kind=jprb))*nclv*nlon/c_sizeof(real(1,kind=jprb))' if nclv_param else 'istsz = 3*c_sizeof(real(1,kind=jprb))*nlon/c_sizeof(real(1,kind=jprb))+c_sizeof(real(1,kind=jprb))*nlon*nz/c_sizeof(real(1,kind=jprb))+2*c_sizeof(real(1,kind=jprb))/c_sizeof(real(1,kind=jprb))'}
+        {'istsz = max(c_sizeof(real(1,kind=jprb)), 8)*nlon/max(c_sizeof(real(1,kind=jprb)), 8)+max(c_sizeof(real(1,kind=jprb)), 8)*nlon*nz/max(c_sizeof(real(1,kind=jprb)), 8)+max(c_sizeof(real(1,kind=jprb)), 8)*nclv*nlon/max(c_sizeof(real(1,kind=jprb)), 8)' if nclv_param else 'istsz = 3*max(c_sizeof(real(1,kind=jprb)), 8)*nlon/max(c_sizeof(real(1,kind=jprb)), 8)+max(c_sizeof(real(1,kind=jprb)), 8)*nlon*nz/max(c_sizeof(real(1,kind=jprb)), 8)+2*max(c_sizeof(real(1,kind=jprb)), 8)/max(c_sizeof(real(1,kind=jprb)), 8)'}
```
reuterbal, line 113: These tests are utterly unreadable 😭 (nothing to action on here, they were like this before)

loki/bulk/scheduler.py
```diff
@@ -709,6 +714,9 @@ def write_cmake_plan(self, filepath, mode, buildpath, rootpath):
         sources_to_transform = []
 
         for item in self.items:
+            if item.ignored:
```
reuterbal, line 717: Coverage report suggests we don't have a test for the CMake plan... or am I just overlooking this somehow? Is this something we could add here maybe, also to have a baseline against which to check the new Scheduler?

transformations/transformations/pool_allocator.py
```diff
@@ -493,13 +512,18 @@ def _create_stack_allocation(self, stack_ptr, stack_end, ptr_var, arr, stack_siz
 
         # Build expression for array size in bytes
         dim = arr.dimensions[0]
+        if isinstance(dim, RangeIndex):
+            dim = Sum((dim.upper, Product((-1, dim.lower)), 1))
```
reuterbal, line 516: codecov suggests that case is currently not covered in the tests. It could be reasonably straightforward to add this to the new test by making one of the kernel arguments have a `2:klev+1` dimension for example? It shouldn't change the overall analysis result but would ensure this behaviour does not get broken in the future.

reuterbal (changes requested):
Thanks, this looks great!

I've flagged a few minor things and suggested some additional testing, but I don't consider any of it to be strictly required in this PR but would be welcome soon after.

tests/test_fgen.py
```diff
@@ -146,6 +146,30 @@ def test_multiline_inline_conditional(frontend):
     out = fgen(routine, linewidth=132)
     for line in out.splitlines():
         assert line.count('&') <= 2
+        if line.count('&') == 2:
+            assert len(line.split('&')[1]) > 60
+
+
+@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Loki likes only valid code')]))
```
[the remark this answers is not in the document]
mlange05: Ahh, yes, good idea. Unfortunately OMNI disagrees anyway:
```
"/etc/ecmwf/ssd/ssd1/tmpdirs/naml.503342.20240213_045707.934/loki/omni-427fad663ed467af585a9732825b1c71.f90", line 6: name is too long. 
```

transformations/tests/test_pool_allocator.py
mlange05: Yes, the reason for this is subtle: The original code is all upper case, but our symbols are case-insensitive, so whenever we're doing symbol comparison the original test style of lower case is fine. The one exception to that are the keywords when using keyword arguments, which are treated as pure (case sensitive) strings. Changing this might be a good idea in general, but quite invasive, so I'd like to keep this for a separate PR.

loki/bulk/scheduler.py
mlange05: Hate to disagree with CodeCov, but...
https://github.com/ecmwf-ifs/loki/blob/main/tests/test_scheduler.py#L860

Maybe it's not recognising it because we're testing the effect somewhat indirectly - as in reading the generated file and checking that?

transformations/transformations/pool_allocator.py
mlange05: Agreed. Should be changed now.

loki/bulk/scheduler.py
reuterbal: I have to side with Codecov on that one...
The test uses the CMakePlanner, which writes the cmake file. Could it be the case that we're not actually using that function from the `Scheduler` class anymore at all? (In which case, it should definitely be deleted!)

mlange05: OMG, that does indeed look like an outdated test and needless code duplication. Will fix...

transformations/transformations/pool_allocator.py
mlange05: Ah, this one is a bit puzzling: the dimension tested is strictly the first one (horizontal) and thus never included in the stack (I think - please shout if I'm wrong on this). As such it seems redundant and I opted to remove that offending check instead.

mlange05:
Ok, I _think_ I addressed all comments and (re-)verified ec-phys regression. This still needs a rebase for integration, but should now be ready for another look. @reuterbal

transformations/tests/test_pool_allocator.py
mlange05: Agreed. Some refactoring of the expressions can alleviate this (and I've done a bit of that in this PR), but I think there's a bit more that could be don on test clean-up for this trafo. However, I also agree that this would be best left for a separate PR.

reuterbal (approved):
Thanks, looks great now! Nothing further to add from my side.

mlange05:
Confirmed clean with latest EC-physics regression (just now, after latest rebase).

PR #223, opened 2024-01-29
C/C++ De/Reference

MichaelSt98:
Should be `Dereference(<derived_type>%member)`. However, it would be more correct to have `Dereference(<derived_type>)%member`. But, this won't work unless `Reference`/ `Dereference` get e.g., the following methods:

```py
@property
def name(self):
    return self.expression.name

@property
def type(self):
    return self.expression.type

@property
def scope(self):
    return self.expression.scope

@property
def initial(self):
    return self.expression.initial
```

Moreover, the insertion of parentheses is a bit hacky rn ...

**Tested** currently via the *transpile* tests and tested with CLOUDSC Loki C.

___________

**EDIT:**: using the `Dereference(<derived_type>)%member` approach with the above mentioned methods implemented. This seems to be the best solution for now, but it is **experimental** and may be refactored at some point.

### DIFF
--- a/loki/expression/symbols.py
+++ b/loki/expression/symbols.py
@@ -1477,26 +1477,8 @@ class Reference(pmbl.Expression):
         assert isinstance(expression, pmbl.Expression)
         self.expression = expression
 
-    """
-    @property
-    def name(self):
-        return self.expression.name
-
-    @property
-    def type(self):
-        return self.expression.type
-
-    @property
-    def scope(self):
-        return self.expression.scope
-
-    @property
-    def initial(self):
-        return self.expression.initial
-
     mapper_method = intern('map_c_reference')
     make_stringifier = loki_make_stringifier
-    """
 
 
 class Dereference(pmbl.Expression):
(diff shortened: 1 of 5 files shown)


loki/expression/symbols.py
```diff
@@ -1457,3 +1460,80 @@ class StringSubscript(StrCompareMixin, pmbl.Subscript):
...
+    def type(self):
+        return self.expression.type
+
+    @property
+    def scope(self):
+        return self.expression.scope
+
+    @property
+    def initial(self):
+        return self.expression.initial
```
reuterbal, line 1497: I think it would be useful to leave a comment here that points out why this is necessary (e.g., something like that we allow the compound `[De]Reference(Variable(name))` to appear  with behaviour akin to a symbol itself for easier processing in mappers). Same in `Dereference`.

reuterbal (changes requested):
Thanks, conceptually this looks good to me and is in-line with what we discussed.

Although it serves a very specific use case I think it would benefit from additional testing - mostly to avoid that other, unrelated changes, e.g., to the expression tree mappers, break the expected behaviour later on.

You could maybe add a second test where, e.g., the variable symbol inside the reference or dereference is renamed? That should increase the test coverage and corresponds to one of the edge cases that we work around with the additional properties.

reuterbal (approved):

mlange05 (approved):
Very nice and neat! GTG, and thanks both! :ship:

PR #225, opened 2024-02-01
Allow for limiting `resolve_sequence_association` to procedures that are inlined in loki-transform convert

skarppinen:
This PR makes the `resolve_sequence_association` option of `loki-transform.py convert` accept one of three options (instead of on/off before):

- 'off': sequence association is turned off (default)
- 'all': all calls found are checked and resolved for sequence association
- 'inlined-only': sequence association is checked and resolved only for procedures that get inlined (member procedures or inline pragma-marked procedures only)

The main changes here are:

1. changing the argument `resolve_sequence_association` to be a `click.Choice` among the three above values. I also did my best to make changes to the respective cmake files. I am not an expert with cmake, but I hope the tests will catch any errors I might have made. Basically I moved `resolve_sequence_association` from an "option" to a "oneValueArgument" in one of the cmake files.

2. adding logic to the `convert` function to pass the correct `resolve_sequence_association` choice to `SanitiseTransformation` and `InlineTransformation`.

3. I reintroduced `resolve_sequence_association` as a constructor argument for `InlineTransformation`.

4. In `transform_inline.py`, I added a new function `resolve_sequence_association_for_inlined_calls` (we can discuss if it should actually be in `transform_sanitise.py`). This function implements the functionality of the "inlined-only" option, making changes in `transform_subroutine` of `InlineTransformation`, only if all of the following are true: 
- user requests `resolve_sequence_association` (for `InlineTransformation`)
- either of `inline_internals` or `inline_marked` is set to `True` (for `InlineTransformation`)
- the procedure to be processed contains calls that will be inlined

5. slight refactoring of `transform_sequence_association` in `transform_sanitise.py`. This function processes all calls of a procedure simultaneously. I rewrote this in terms of a new function `transform_sequence_association_append_map(call_map, call)` that builds up the transformation map for resolving sequence assiociation one call at a time. This allows for resolving sequence association on a "call by call" basis elsewhere, as opposed to every call in a procedure. I use this in the implementation of `resolve_sequence_association_for_inlined_calls` above.

6. I modified the help of `loki-transform.py convert` accordingly and added new tests for `InlineTransformation` that end up calling `resolve_sequence_association_for_inlined_calls`.


scripts/loki_transform.py
```diff
@@ -112,8 +112,18 @@ def cli(debug):
...
+                ['off', 'all', 'inlined-only']
+              ),
+              help=(
+                'Option to replace array arguments passed as scalars with arrays. ' +
+                'Choice `all` does replacement for all procedures as part of a global sanitisation transform. ' +
+                'Choice `inlined-only` applies replacement only for procedures ' +
+                'that will be inlined as a result of specifying options ' +
+                '`--inline-members` or `--inline-marked`. ' + 
+                'If inlining is deactivated, choice `inlined-only` has no effect.'
+              ))
```
reuterbal, line 126: I'm against overloading the option in such a fashion. This will break existing API.
Instead, we should keep the original option as is and do one of the two following (or ideally both):
1. Add a new bool option `--resolve-sequence-association-inlined-calls/--no-resolve-sequence-association-inlined-calls`
2. Parametrize the inline transformation via the Scheduler config (which is already possible). This will be the preferred option in the future anyway.

tests/test_transform_inline.py
```diff
@@ -945,3 +945,187 @@ def test_inline_transformation(frontend):
...
+        inline_marked=True, inline_internals=True, resolve_sequence_association=True
+    )
+    outer = module["outer"]
+    trafo.apply(outer)
+    callnames = [call.name for call in FindNodes(CallStatement).visit(outer.body)]
+    assert 'plusone' not in callnames
+    assert 'inner' not in callnames
+    assert 'minusone_second' in callnames
+
+    # Test case that a crash occurs if marked routine with sequence association is
```
reuterbal, line 1062: I would suggest splitting the below into a separate test

tests/test_transform_inline.py
```diff
@@ -945,3 +945,187 @@ def test_inline_transformation(frontend):
...
+    module = Module.from_source(fcode, frontend=frontend)
+    trafo = InlineTransformation(
+        inline_constants=True, external_only=True, inline_elementals=True,
+        inline_marked=True, inline_internals=False, resolve_sequence_association=True
+    )
+    outer = module["outer"]
+    trafo.apply(outer)
+    assert len(FindNodes(CallStatement).visit(outer.body)) == 0
+
+    # Testing that ValueError is thrown if sequence association is requested with inlining,
```
reuterbal, line 1106: Make this another separate test, please.

reuterbal (changes requested):
Thanks, the changes in the transformation look great and the test is comprehensive!

However, the integration via redefining the CLI option breaks the current interface and is counter-intuitive as it encodes different behaviour. I left a suggestion to instead introduce an additional argument or rely on the config-file parameterisation of the transformation entirely instead.

skarppinen:
Many thanks @reuterbal for the review. I have made the changes you requested. It appears that some of the regression tests failed, but it seems like they fail due to some FIELD API related issue.

reuterbal (changes requested):
Thank you for taking care of the flagged points, all addressed!

Don't worry about the failing regression test, this is due to the ecwam regression test that depends on an internal repository still - and the relevant secret is only available to CI runners if triggered from a branch on the same repository. This should be resolved soon.

I apologise for overlooking the redundant exclusion of OFP in the tests, would you mind removing that? Other than that, I'll let @mlange05 have a look as well since he had been involved in this, too.

reuterbal (approved):
Thanks!

Waiting for @mlange05 before merging.

mlange05 (approved):
Sorry for the delay, and yes, this looks good to me too. 

One note I have is that we should indeed soon start shifting towards using the new config files for setting these particular parameters rather than adding more CLI options, but that is beyond this PR. So, GTG from me too. :shipit:

skarppinen:
Thanks, @mlange05. I agree with what you said about configuration files.

PR #226, opened 2024-02-02
Global var hoisting

MichaelSt98:
Global/Imported variable hoisting.

Convert e.g., 

```fortran
module moduleB
   real :: var2
   real :: var3
end module moduleB

module moduleC
   real :: var4
   real :: var5
end module moduleC

subroutine driver()
implicit none

call kernel()

end subroutine driver

subroutine kernel()
use moduleB, only: var2,var3
use moduleC, only: var4,var5
implicit none

var4 = var2
var5 = var3

end subroutine kernel
```

to:

```fortran
module moduleB
   real :: var2
   real :: var3
end module moduleB

module moduleC
   real :: var4
   real :: var5
end module moduleC

subroutine driver()
use moduleB, only: var2,var3
use moduleC, only: var4,var5
implicit none

call kernel(var2, var3, var4, var5)

end subroutine driver

subroutine kernel(var2, var3, var4, var5)
implicit none
real, intent(in) :: var2
real, intent(in) :: var3
real, intent(inout) :: var4
real, intent(inout) :: var5

var4 = var2
var5 = var3

end subroutine kernel
```

Tested locally with CLOUDSC Loki C transpilation. However, this requires PR #208.

### DIFF
--- a/transformations/tests/test_data_offload.py
+++ b/transformations/tests/test_data_offload.py
@@ -12,11 +12,11 @@ import pytest
 from loki import (
     Sourcefile, FindNodes, Pragma, PragmaRegion, Loop,
     CallStatement, pragma_regions_attached, get_pragma_parameters,
-    gettempdir, Scheduler, OMNI, Import, fgen
+    gettempdir, Scheduler, OMNI, Import
 )
 from conftest import available_frontends
 from transformations import (
-        DataOffloadTransformation, GlobalVariableAnalysis, 
+        DataOffloadTransformation, GlobalVariableAnalysis,
         GlobalVarOffloadTransformation, GlobalVarHoistTransformation
 )
 
@@ -647,9 +647,10 @@ def test_transformation_global_var_import_derived_type(here, config, frontend):
 
 @pytest.mark.parametrize('frontend', available_frontends())
 @pytest.mark.parametrize('hoist_parameters', (False, True))
-def test_transformation_global_var_hoist(here, config, frontend, hoist_parameters):
+@pytest.mark.parametrize('ignore_modules', (None, ('moduleb',)))
+def test_transformation_global_var_hoist(here, config, frontend, hoist_parameters, ignore_modules):
     """
-    Test the generation of offload instructions of global variable imports.
+    Test hoisting of global variable imports.
     """
     config['default']['enable_imports'] = True
     config['routines'] = {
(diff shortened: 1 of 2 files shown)


transformations/transformations/data_offload.py
```diff
@@ -613,3 +614,279 @@ def process_driver(self, routine, successors):
...
+        for var, module in chain(kernel_uses_symbols, kernel_defines_symbols):
+            # filter modules that are supposed to be ignored
+            if module.lower() in self.ignore_modules:
+                continue
+            symbol_map[module].add(var.parents[0] if var.parent else var)
+        import_map = CaseInsensitiveDict()
+        scope = routine
+        while scope:
+            import_map.update(scope.import_map)
+            scope = scope.parent
```
reuterbal, line 813: Mind you: this poses the risk of overwriting a shadowed symbol - for example:
```f90
module a_mod
    use b_mod, only: a, b, c
contains
    subroutine a_routine
        use c_mod, only c
        ! symbol_map will now contain c => b_mod 
        ! because the parent scope is processed after the inner scope
    end subroutine a_routine
end module a_mod
```

One way to avoid this would be to use
```py
import_map = CaseInsensitiveDict(
    (s.name, imprt) for imprt in routine.all_imports[::-1] for s in imprt.symbols
)
```

reuterbal (changes requested):
Thanks, this looks very good to me!

I've left a few suggestions for stylistic improvements and flagged a potential issue when an import shadows an import in a parent scope. Arguably, this is an almost academic edge case and bad practice anyway, but the fix is easy.

transformations/tests/test_data_offload.py
```diff
@@ -634,3 +651,133 @@ def test_transformation_global_var_import_derived_type(here, config, frontend):
...
+    var_intent_map = {'var0': 'in', 'var1': 'in', 'var2': 'in',
+            'var3': 'in', 'var4': 'inout', 'var5': 'inout', 'tmp': None}
+    # DRIVER
+    imports = FindNodes(Import).visit(driver.spec)
+    import_names = [_import.module.lower() for _import in imports]
+    # check driver imports
+    expected_driver_modules = ['modulec']
+    expected_driver_modules += ['moduleb'] if ignore_modules is None else []
+    if frontend != OMNI:
+        expected_driver_modules += ['modulea'] if hoist_parameters else []
```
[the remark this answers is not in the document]
reuterbal: Thanks, got it!

transformations/transformations/data_offload.py
```diff
@@ -613,3 +614,274 @@ def process_driver(self, routine, successors):
...
+            if module.lower() in self.ignore_modules:
+                continue
+            symbol_map[module].add(var.parents[0] if var.parent else var)
+        import_map = CaseInsensitiveDict(
+                (s.name, imprt) for imprt in routine.all_imports[::-1] for s in imprt.symbols
+                )
+        scope = routine
+        while scope:
+            import_map.update(scope.import_map)
+            scope = scope.parent
```
reuterbal, line 808: Sorry for being pedantic, but this bit needs to go now otherwise you're still overwriting the values with parent scope entries ;-)
```suggestion
```

reuterbal (changes requested):
Thanks! Unfortunately, there's still a bit left related to the imports, which may just have been an oversight. Otherwise good to go!

transformations/transformations/data_offload.py
MichaelSt98: ... yes ... fair enough 😅

reuterbal (approved):

mlange05 (approved):
Only had a brief look, but looks very neat and well tested! GTG from me. :shipit:

PR #219, opened 2024-01-25
allow F2C transpilation using c_ptr (switch for old behaviour or new …

MichaelSt98:
including switch/flag `use_c_ptr` for old behaviour or new one.

"Old"/original f2c interface looked like:

```fortran
MODULE KERNEL_FC_MOD
CONTAINS
  SUBROUTINE KERNEL_FC(LEN, ARRAY)
    INTEGER :: LEN
    REAL :: ARRAY(LEN)
    INTERFACE
      SUBROUTINE KERNEL_ISO_C(LEN, ARRAY) BIND(C, name="KERNEL_C")
        INTEGER(KIND=c_int), VALUE :: LEN
        REAL(KIND=c_double) :: ARRAY(LEN)
      END SUBROUTINE KERNEL_ISO_C
    END INTERFACE
    CALL KERNEL_ISO_C(LEN, ARRAY)
  END SUBROUTINE KERNEL_FC
END MODULE KERNEL_FC_MOD
```

New interface uses `c_ptr` and `c_loc()`

```fortran
MODULE KERNEL_FC_MOD
CONTAINS
  SUBROUTINE KERNEL_FC(LEN, ARRAY)
    INTEGER :: LEN
    REAL, TARGET :: ARRAY(:)
    INTERFACE
      SUBROUTINE KERNEL_ISO_C(LEN, ARRAY) BIND(C, name="KERNEL_C")
        INTEGER(KIND=c_int), VALUE :: LEN
        TYPE(c_ptr) :: ARRAY(LEN)
      END SUBROUTINE KERNEL_ISO_C
    END INTERFACE
    CALL KERNEL_ISO_C(LEN, c_loc(ARRAY))
  END SUBROUTINE KERNEL_FC
END MODULE KERNEL_FC_MOD
```

Regression CLOUDSC test works with both `FortranCTransformation(use_c_ptr=True/False)`

However, for `use_c_ptr=True` PR #218 is necessary!
Tested with NVHPC and Intel with `use_c_ptr=True` and `use_c_ptr=False` (Remember to use nproma=1,2 for NVHPC or lower optimisation).

PR #228, opened 2024-02-08
Github actions: Only load SSH_KEY from secrets if available

reuterbal:
Nothing more than that. That should allow to run the regression tests (and fail then) even if private repositories are not accessible.


reuterbal:
Added a fix for the duplicate import regression in the linter checks, that appeared after merging #219

mlange05 (approved):
LGTM :shipit:

awnawab (approved):
Looks good 👌

reuterbal:
Putting this in first to fix the code checks

PR #231, opened 2024-02-26
Alternative stack/pool allocator implementation based on Cray pointers compatible with Cray+AMD stack

MichaelSt98:
… pointers but which works on Cray + AMD

Some problems with OMNI related to the unit tests ...

### DIFF
--- a/transformations/transformations/pool_allocator.py
+++ b/transformations/transformations/pool_allocator.py
@@ -60,6 +60,88 @@ class TemporariesPoolAllocatorTransformation(Transformation):
     * Assign stack base pointer and end pointer for each block (identified via :data:`block_dim`)
     * Pass the stack argument(s) to kernel calls
 
+
+    With ``cray_ptr_loc_rhs=False`` the following stack/pool allocator will be generated: 
+    
+    .. code-block:: fortran
+
+        SUBROUTINE DRIVER (...)
+          ...
+          INTEGER(KIND=8) :: ISTSZ
+          REAL, ALLOCATABLE :: ZSTACK(:, :)
+          INTEGER(KIND=8) :: YLSTACK_L
+          INTEGER(KIND=8) :: YLSTACK_U
+          ISTSZ = (MAX(C_SIZEOF(REAL(1, kind=jprb)), 8)*<array dim1>*<array dim2> + ...) / & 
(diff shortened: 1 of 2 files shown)


transformations/transformations/pool_allocator.py
```diff
@@ -93,6 +175,10 @@ class TemporariesPoolAllocatorTransformation(Transformation):
     check_bounds : bool, optional
         Insert bounds-checks in the kernel to make sure the allocated stack size is not
         exceeded (default: `True`)
+    cray_ptr_loc_rhs : bool, optional
+        Whether to only pass the stack variable as integer to the kernel(s) or 
+        whether to pass the whole stack array to the driver and the calls to ``LOC()`` 
+        within the kernel(s) itself (default: `False`)
```
reuterbal, line 181:
```suggestion
        When enabled, pass the full stack array from the driver to the kernel and perform
        the calls to ``LOC()`` within the kernel(s) to determine the "pointer address"
        of the temporary arrays. By default, ``LOC()`` will only be called in the
        driver and "pointer addresses" are passed as arguments to the kernel
        (default: ``False``).
```

transformations/transformations/pool_allocator.py
```diff
@@ -725,7 +857,7 @@ def create_pool_allocator(self, routine, stack_size):
         if loop_map:
             routine.body = Transformer(loop_map).visit(routine.body)
 
-    def inject_pool_allocator_into_calls(self, routine, targets, ignore):
+    def inject_pool_allocator_into_calls(self, routine, targets, ignore, driver=False):
```
reuterbal, line 860: The `driver` argument seems unnecessary here?

reuterbal (commented):
Many thanks!

I have rebased the branch and cannot find any specific OMNI failures anymore. Disregard the currently failing CLOUDSC test, which will be fixed with ecmwf-ifs/dwarf-p-cloudsc#82. I have locally verified this.

The implementation looks good to me apart from two minor comments. I still need to test this on NVIDIA and AMD platforms, though, and assess the impact. Therefore, only a comment for now.

reuterbal (approved):
Tested on Nvidia, too. Performance seems unharmed.

MichaelSt98:
Looks good to me. Thanks for fixing the OMNI related stuff and for testing!

PR #232, opened 2024-02-27
Add ExternalItem to allow building dependency graph with missing nodes

reuterbal:
This adds the missing `ExternalItem` item type for the new Scheduler, that allows to parse incomplete source trees and adds missing dependencies as `ExternalItem`. 

The missing dependency's type (i.e., the Item type that we would have created if the respective source file was available) is stored as `origin_cls` on the `ExternalItem`.

`SFilter` has been amended to ignore externals or include those externals that would have matched the `item_filter`. That allows to identify missing items with the `strict` config option during processing.

### DIFF
--- a/loki/bulk/item.py
+++ b/loki/bulk/item.py
@@ -23,7 +23,7 @@ from loki.visitors import FindNodes
 
 __all__ = [
     'Item', 'FileItem', 'ModuleItem', 'ProcedureItem', 'TypeDefItem',
-    'InterfaceItem', 'ProcedureBindingItem', 'ItemFactory'
+    'InterfaceItem', 'ProcedureBindingItem', 'ExternalItem', 'ItemFactory'
 ]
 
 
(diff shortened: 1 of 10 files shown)


reuterbal:
Rebased to resolve conflict

reuterbal:
As discussed offline: piggy-backed three changes to this:
- Removed the blanket dependency on `ModuleItem` for every import. Instead, where we can, we now rely on the dependency directly to the items in that module, e.g. subroutines, functions, interfaces, typedefs. The dependency on the ModuleItem is only added for imports that don't fall into this category, which should only be global variables.
- The dependency on Subroutines is no longer incurred through the import statement but instead via Call statements. That ensures we don't introduce dependencies for unused imports. Note that this does not apply to functions as their use cannot easily be determined via REGEX matching.
- The ModuleWrapTransformation now updates the `block` property of the item on the caller side for previously `ignore` items. This is because the ignored items are not actually transformed and therefore the dependency as a module procedure cannot be found.

mlange05 (approved):
Looks very good and is a sincere quality-of-life improvement. I've worked with this for a few days now, and managed to revert a lot of workarounds. GTG from me! :shipit:

PR #233, opened 2024-02-28
Overwrite frontend options on a per-file basis in Scheduler config

reuterbal:
This adds the ability to overwrite frontend options on a per-file basis in the Scheduler config. This allows for example to switch on preprocessing for individual files only (e.g., for those that are not valid Fortran without preprocessing), or to overwrite the frontend parser. The latter was the missing piece to successfully run CLOUDSC regression with OMNI.

### DIFF
--- a/loki/bulk/configure.py
+++ b/loki/bulk/configure.py
@@ -192,6 +192,22 @@ class SchedulerConfig:
 
         The resulting `dict` contains overwrites that have been provided
         in the :attr:`frontend_args` of the config.
+
+        Parameters
+        ----------
+        path : str or pathlib.Path
+            The file path for which to create the frontend arguments. This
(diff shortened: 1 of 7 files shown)


loki/bulk/configure.py
```diff
@@ -179,6 +185,23 @@ def create_item_config(self, name):
...
         return item_conf
 
+    def create_frontend_args(self, path, default_args):
+        """
+        Create bespoke ``frontend_args`` to pass to the constructor
+        or ``make_complete`` method for a file
+
+        The resulting `dict` contains overwrites that have been provided
+        in the :attr:`frontend_args` of the config.
+        """
```
mlange05, line 195: Might be useful to list arguments here. Was briefly wondering here what `path` pertained to.

PR #235, opened 2024-03-04
kwargs utilities

MichaelSt98:
Utilities for `CallStatement`s

* check whether `kwargs` are correctly ordered
* sort `kwargs`
* convert `kwargs` to args

necessary for e.g., F2C transpilation

### DIFF
--- a/loki/ir.py
+++ b/loki/ir.py
@@ -953,10 +953,17 @@ class CallStatement(LeafNode, _CallStatementBase):
         new_kwarguments = tuple((arg_name, kwargs[arg_name]) for arg_name in r_arg_names)
         return new_kwarguments
 
+    def check_kwarguments_order(self):
+        """
+        Check whether kwarguments are correctly ordered
+        in respect to the arguments (``self.routine.arguments``).
+        """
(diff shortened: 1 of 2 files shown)


loki/ir.py
```diff
@@ -941,6 +941,41 @@ def arg_map(self):
...
+
+    def sort_kwarguments(self):
+        """
+        Sort and update the kwarguments according to the order of the
+        arguments (``self.routine.arguments``).
+        """
+        new_kwarguments = self._sort_kwarguments()
+        self._update(kwarguments=new_kwarguments)
+
+    def kwargs_to_args(self):
```
reuterbal, line 971: Maybe use the "leading verb" naming here, too?

```suggestion
    def convert_kwargs_to_args(self):
```

reuterbal (changes requested):
Thanks, this looks very useful! A little suggestion to improve naming, otherwise happy.

reuterbal (approved):

mlange05 (approved):
Looks great! Many thanks for you patience. GTG from me. :shipit:

PR #236, opened 2024-03-04
Allow to ignore specific dimensions in "shift to zero indexing"

MichaelSt98:
allow to ignore specific dimensions ...

### DIFF
--- a/loki/transform/transform_array_indexing.py
+++ b/loki/transform/transform_array_indexing.py
@@ -33,7 +33,7 @@ __all__ = [
 ]
 
 
-def shift_to_zero_indexing(routine):
+def shift_to_zero_indexing(routine, ignore=()):
     """
     Shift all array indices to adjust to 0-based indexing conventions (eg. for C or Python)
     """
@@ -47,7 +47,14 @@ def shift_to_zero_indexing(routine):
                     # no shift for stop because Python ranges are [start, stop)
                     new_dims += [sym.RangeIndex((start, d.stop, d.step))]
                 else:
-                    new_dims += [d - sym.Literal(1)]
+                    shift = True
+                    for var in FindVariables().visit(d):
+                        if var in ignore:
+                            shift = False
+                    if shift:
+                        new_dims += [d - sym.Literal(1)]
+                    else:
+                        new_dims += [d]
             vmap[v] = v.clone(dimensions=as_tuple(new_dims))
     routine.body = SubstituteExpressions(vmap).visit(routine.body)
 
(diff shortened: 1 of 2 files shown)


loki/transform/transform_array_indexing.py
```diff
@@ -47,7 +54,14 @@ def shift_to_zero_indexing(routine):
...
                 else:
-                    new_dims += [d - sym.Literal(1)]
+                    shift = True
+                    for var in FindVariables().visit(d):
+                        if var in ignore:
+                            shift = False
+                    if shift:
+                        new_dims += [d - sym.Literal(1)]
+                    else:
+                        new_dims += [d]
```
reuterbal, line 64: Just to propose a slightly streamlined alternative to avoid the explicit loop:
```suggestion
                    if ignore and any(var in ignore for var in FindVariables().visit(d)):
                        new_dims += [d]
                    else:
                        new_dims += [d - sym.Literal(1)]
```

loki/transform/transform_array_indexing.py
```diff
@@ -33,9 +33,16 @@
 ]
 
 
-def shift_to_zero_indexing(routine):
+def shift_to_zero_indexing(routine, ignore=()):
```
reuterbal, line 36: It's generally preferred to assign `None` as default value for optional arguments, and then convert it to a tuple in the function body, e.g. `ignore = as_tuple(ignore)`.
That also adds the benefit that you can do something like `shift_to_zero_indexing(routine, ignore='myvar')` and wouldn't get the unexpected behaviour that this is treated as `ignore=('m', 'y', 'v', 'a', 'r')`.

loki/transform/transform_array_indexing.py
```diff
@@ -33,9 +33,16 @@
...
     """
     Shift all array indices to adjust to 0-based indexing conventions (eg. for C or Python)
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine in which the array dimensions should be shifted
+    ignore : list of str
+        Dimensions (or rather variables being dimensions) to be ignored
```
reuterbal, line 45: Maybe something like this would be more descriptive?
```suggestion
        List of variable names for which, if found in the dimension expression of an array subscript, that dimension is not shifted to zero.
```

reuterbal (changes requested):
Thanks, looks good. I've left two remarks about implementation and I think the docstring could be slightly improved maybe.

reuterbal (approved):
Thanks for addressing the comments, all good now!

mlange05 (approved):
Looks good. Many thanks for you patience. GTG :shipit:

PR #234, opened 2024-03-02
Further frontend performance optimisations

mlange05:
Another small round of tricks to improve the performance of the `sanitize_ir` utility and the `FP` frontend to reduce the frontend processing time for large input subroutines. Primarily, this PR focuses on using fast in-place transformers for the necessary IR sanitisation and avoid using the explicit rescoping for the FParser frontend by attaching the scopes on first IR node creation.

In more detail:
* Use an on-the-fly `Transformer to do inline-comment attachment
* Make all transformers used during sanitisation use `inplace=True`
* Avoid calling `rescope_symbols` on the finalised IR when creating subroutines

On the pathological test case we the following improvements:
```
# On main
In [2]: with config_override({'log-level': 'perf'}):
   ...:     Sourcefile.from_file("apl_arpege_parallel.F90")
[Loki::Frontend] Executed sanitize_input in 0.12s
[Loki::FP] Executed parse_fparser_source in 7.41s
[Loki::Frontend] Executed sanitize_ir in 0.25s
[Loki::Frontend] Executed sanitize_ir in 5.70s
[Loki::Frontend] Executed sanitize_ir in 0.94s
[Loki::FP] Executed parse_fparser_ast in 19.30s
[Loki::Sourcefile] Constructed from apl_arpege_parallel.F90 in 26.83s

# with this PR
In [2]: with config_override({'log-level': 'perf'}):
   ...:     Sourcefile.from_file("apl_arpege_parallel.F90")
[Loki::Frontend] Executed sanitize_input in 0.12s
[Loki::FP] Executed parse_fparser_source in 7.76s
[Loki::Frontend] Executed sanitize_ir in 0.09s
[Loki::Frontend] Executed sanitize_ir in 2.13s
[Loki::Frontend] Executed sanitize_ir in 0.47s
[Loki::FP] Executed parse_fparser_ast in 9.32s
[Loki::Sourcefile] Constructed from apl_arpege_parallel.F90 in 17.27s
```

_Integration note: PR is currently passing ECphysics regression, but not local test base. Will attempt to fix local tests ASAP, but filing draft now to trigger test base. Initial bulk-parse in EC-physics also seems to improve with this._

PR #241, opened 2024-03-09
Frontend Transformer optimisation

mlange05:
_Note: To manage the complexity of what became quite a complex operation, I'm separating the simple steps of PR #234 into a separate PR._ 

This PR improves the way we deal with comments and multi-line pragmas in the frontend sanitization steps. Following the previous example, we now also do comment inlining with a one-pass transformer and make all the respective transformers use `inplace=True`. 

The PR also includes a small fix to the previously introduced utility transformers, where we forgot to strip `None` and empty tuples from the resulting IR. I also needed to adjust the IR graph test - not quite sure why, but it seems safe to me.


mlange05:
Confirmed clean with ec-physics regression.

loki/frontend/util.py
```diff
@@ -52,23 +52,23 @@ def __str__(self):
...
+        pairs += PatternFinder(pattern=(VariableDeclaration, Comment)).visit(o)
+        pairs += PatternFinder(pattern=(ProcedureDeclaration, Comment)).visit(o)
+
+        for pair in pairs:
+            # Comment is in-line and can be merged
+            if pair[0].source and pair[1].source:
+                if pair[1].source.lines[0] == pair[0].source.lines[1]:
+                    new = pair[0]._rebuild(comment=pair[1])
+                    o = replace_windowed(o, pair, new)
+        return o
```
reuterbal, line 71: Slightly confused by the implementation:

Undoubtedly, the performance gain comes here from not using `NestedTransformer` and instead doing in-place updates for the found nodes. However, I don't understand why this is implemented as a `Transformer`, and not convinced it does always behave as expected?

It shortcuts the recursion in the `Transformer` upon the first tuple - which will in the frontend use-case usually be the `Section.body` for something like spec, body in a subroutine. But the PatternFinder is then the only place to actually recurse into the IR. Does the replacement really work as expected for inline comments further nested into the IR than the outermost tuple?

Unless we're explicitly testing for this, it wouldn't upset anything because it just means the inline comments continue to live as standalone comments. So, I'm wondering whether the effect is merely that we don't convert the comments to inline comments at all anymore?

reuterbal (changes requested):
Thanks for looking into this! In principle uncontroversial and the use of inplace transformers is very sensible here.

I'm a bit puzzled about the comments inlining implementation - I'm not convinced it still does what it is intended to do. Just want to make sure we don't break this functionality accidentally.

loki/frontend/util.py
mlange05: Wow, very good spot! You are totally right that 
a) this isn't tested properly!
b) We recurse in `PatternFinder`, which we don't need anymore, but not in `visit_tuple`

I've tried to address both in two commits I just pushed, and with those the behaviour seems correct, and performance benefits stay as before. Please have another look (once CI agrees)

mlange05: Funny, this also fixed the `test_ir_graph`, so that I could remove the commit that adjusts the node/edge counting test (hence recent force-push).

reuterbal (approved):
Thanks, sanity on my end restored 😆 
Looks great, many thanks for making this change and adding the test!

PR #245, opened 2024-03-12
Subroutine: Only clone symbol when inferring from allocatable

mlange05:
Another small frontend parsing optimisation. This utility (deriving conceptual shape from `allocate` statements in the routine body, is ancient and predates the symbol table mechanism, hence it was never updated. The now unnecessary symbol substitution it used to propagate the information is not necessary anymore, so we now merely clone the symbol once with the necessary type update - et voila.

This brings the overall parsing time of a large auto-generated kernel down to
```
[Loki::Frontend] Executed sanitize_input in 0.12s
[Loki::FP] Executed parse_fparser_source in 7.72s
[Loki::Frontend] Executed sanitize_ir in 0.11s
[Loki::Frontend] Executed sanitize_ir in 2.92s
[Loki::Frontend] Executed sanitize_ir in 0.48s
[Loki::FP] Executed parse_fparser_ast in 5.93s
[Loki::Sourcefile] Constructed from apl_arpege_parallel.F90 in 13.80s
```

Previously this was (assuming PR #242 and 241):
```
In [2]: with config_override({'log-level': 'perf'}):
   ...:     Sourcefile.from_file("apl_arpege_parallel.F90")
   ...: 
[Loki::Frontend] Executed sanitize_input in 0.12s
[Loki::FP] Executed parse_fparser_source in 7.62s
[Loki::Frontend] Executed sanitize_ir in 0.11s
[Loki::Frontend] Executed sanitize_ir in 2.81s
[Loki::Frontend] Executed sanitize_ir in 0.48s
[Loki::FP] Executed parse_fparser_ast in 9.84s
[Loki::Sourcefile] Constructed from apl_arpege_parallel.F90 in 17.59s
```


loki/subroutine.py
```diff
@@ -131,31 +131,21 @@ def __setstate__(self, s):
...
-        for v in FindVariables().visit(spec):
-            if v.name.lower() in alloc_map:
-                vtype = v.type.clone(shape=alloc_map[v.name.lower()])
-                smap[v] = v.clone(type=vtype)
-        return (SubstituteExpressions(smap, invalidate_source=False).visit(spec),
-                SubstituteExpressions(vmap, invalidate_source=False).visit(body))
+                        new_shape = v.dimensions
+
+                    # Clone the symbol with the updated type to inject shape info
+                    v.clone(type=v.type.clone(shape=new_shape))
```
reuterbal, line 148: Any reason for not using `v.type = v.type.clone(shape=new_shape)` here? Seems like the canonical API use

reuterbal (approved):
Many thanks! This looks like a well-hung but low-hanging fruit :)
I left a comment that's purely syntactical

loki/subroutine.py
mlange05: Good catch! Updated now (sorry, habitual rebase).

PR #247, opened 2024-03-12
Loki-lint: First 3 rules for new IFS-Arpégé coding standards

reuterbal:
This adds an implementation of the first 3 (arbitrarily chosen) rules from the new coding standards as a new rules module and makes this the default rules module in loki-lint.py.

Piggy-backed is a small fix for the expression dimensions inference, which would fail with an exception when the dimension expression contains deferred type symbols. However, this exception is actually unnecessary and the deferred type symbol should just be used as-is.

The second fix is just in the output formatting for rule violations to a yaml file. So far, even without any violations it would have inserted an empty line for every file, and skips now over empty reports.

### DIFF
--- a/lint_rules/lint_rules/ifs_arpege_coding_standards.py
+++ b/lint_rules/lint_rules/ifs_arpege_coding_standards.py
@@ -28,7 +28,7 @@ import loki.ir as ir
 
 
 __all__ = [
-    'MissingImplicitNoneRule', 'MissingIntfbRule',
+    'MissingImplicitNoneRule', 'OnlyParameterGlobalVarRule', 'MissingIntfbRule',
 ]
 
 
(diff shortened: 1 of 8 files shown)


lint_rules/lint_rules/ifs_arpege_coding_standards.py
```diff
@@ -0,0 +1,196 @@
...
+            if not decl.symbols[0].type.parameter:
+                msg = f'Global variable(s) declared that are not parameters: {", ".join(s.name for s in decl.symbols)}'
+                rule_report.add(msg, decl)
+
+
+class MissingIntfbRule(GenericRule):
+    """
+    Calls to subroutines and functions that are provided neither by a module
+    nor by a CONTAINS statement, must have a matching explicit interface block.
+    """
```
mlange05, line 119: No action: For practicality this rule might need to be able to handle "exceptions", like for example "gstats" to make the output more usable. But this is up for debate and could be added later.

mlange05 (approved):
Very nice! Looks great and makes for a great demonstrator! GTG from me. :shipit: 

Only left one side comment, but no need to action that now.

lint_rules/lint_rules/ifs_arpege_coding_standards.py
reuterbal: The coding standard rule does not currently allow for any exceptions: https://sites.ecmwf.int/docs/ifs-arpege-coding-standards/fortran/rules/L9.html
For gstats in particular, the fix should likely be the addition of the interface block - it's peculiar that this was missing in the first place and I don't see any advantage in omitting it.

From a practical point of view I agree, this would be something that can be easily added to the rule.

PR #251, opened 2024-03-18
Preserve imported statement functions

awnawab:
An imported statement function (😭 ) with the same root as a target subroutine is overwritten by the dependency transformation. This PR fixes the above.

### DIFF
--- a/loki/transform/dependency_transform.py
+++ b/loki/transform/dependency_transform.py
@@ -319,7 +319,7 @@ class DependencyTransformation(Transformation):
         for im in imports:
             if im.c_import:
                 target_symbol = im.module.split('.')[0].lower()
-                if targets and target_symbol.lower() in targets:
+                if targets and target_symbol.lower() in targets and 'intfb' in im.module.lower():
                     # Modify the the basename of the C-style header import
                     s = '.'.join(im.module.split('.')[1:])
                     im._update(module=f'{target_symbol}{self.suffix}.{s}')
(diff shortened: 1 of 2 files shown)


loki/transform/dependency_transform.py
```diff
@@ -319,7 +319,7 @@ def rename_imports(self, source, imports, targets=None):
         for im in imports:
             if im.c_import:
                 target_symbol = im.module.split('.')[0].lower()
-                if targets and target_symbol.lower() in targets:
+                if targets and target_symbol.lower() in targets and 'intfb' in im.module.lower():
```
reuterbal, line 322: Are we already relying on `.intfb.h` as a naming convention for these imports? Otherwise we're implying a new restriction here.

awnawab: Yes we are relying on subroutine interface imports being named with an `.intfb.h` suffix. That felt like a safer assumption rather than stripping any c-style import containing the subroutine name.

mlange05: Yeah, I don't like hard-baking IFS conventions into this either, but we already have this in place at the moment, unfortunately. @reuterbal Maybe we should file a separate issue to un-hook this assumption from the `DependencyTransformation` and write a custom "IFS-specific" transformation that deals with those and makes the import chain coherent with a more generic import mechanism?

mlange05 (approved):
Aside from the `intfb.h` convention issues this looks GTG to me. :shipit:

PR #252, opened 2024-03-18
Updates for EC-physics regression

mlange05:
This PR brings in a set of small changes (mostly discussed offline) that aligns the new scheduler with prior behaviour and thus gets EC-physics to pass again (in a reasonable amount of time). The main updates are concerned with the CMake-planning stage, where we need to re-introduce the `enable_imports` option and honour it for default parse traversals, as well as the disabling of the SGraph caching (which otherwise makes things prohibitively slow).

In some more detail:
* Reduce the logging of `make_complete` to only trigger on non-REGEX frontends
* Avoid self-reference cycles when building the SGraph (no longer needed after `ExternalItem` addition, but thought I'd keep it anyways)
* (Temporarily) stop caching the SGraph, and instead re-build it after each use of discovery and item-parsing. Note that this is meant as temporary until we can improve the re-build speed of large graphs via caching.
* Bugfix: Fix the bailout mechanism for parser_classes in `make_complete` with REGEX frontend (as suggested by @reuterbal )
* Re-introduce the `enable_imports` options for the `Scheduler`, which allows us to prune the search tree for EC-physics significantly

### DIFF
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -748,7 +748,8 @@ class SGraph:
         if new_items:
             self.add_nodes(new_items)
 
-        self.add_edges((item, item_) for item_ in dependencies)
+        # Careful not to include cycles (from recursive TypeDefs)
+        self.add_edges((item, item_) for item_ in dependencies if not item == item_)
         return new_items
 
     def _populate(self, seed, item_factory, config):
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -576,7 +576,8 @@ class Scheduler:
         sources_to_remove = []
         sources_to_transform = []
 
-        for item in self.items:
+        # Filter the SGraph to get a pure call-tree
+        for item in SFilter(self.sgraph, item_filter=ProcedureItem):
             if item.is_ignored:
                 continue
 
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -618,7 +618,6 @@ class Scheduler:
                     sources_to_append += [newsource]
                     sources_to_remove += [sourcepath]
 
-        info(f'[Loki] CMakePlanner writing plan: {filepath}')
         with Path(filepath).open('w') as f:
             s_transform = '\n'.join(f'    {s}' for s in sources_to_transform)
             f.write(f'set( LOKI_SOURCES_TO_TRANSFORM \n{s_transform}\n   )\n')
--- a/loki/bulk/scheduler.py
+++ b/loki/bulk/scheduler.py
@@ -250,7 +250,10 @@ class Scheduler:
         :any:`SGraph`
             A dependency graph containing only :any:`FileItem` nodes
         """
-        return self.sgraph.as_filegraph(self.item_factory, self.config)
+        item_filter = None if self.config.enable_imports else ProcedureItem
+        return self.sgraph.as_filegraph(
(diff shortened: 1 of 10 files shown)


loki/bulk/scheduler.py
```diff
@@ -448,7 +457,12 @@ def _get_definition_items(_item, sgraph_items):
 
         if transformation.creates_items:
             self._discover()
+            # Rebuild after discovery
+            self._sgraph = SGraph.from_seed(self.seeds, self.item_factory, self.config)
+
             self._parse_items()
+            # And rebuild after parsing
+            self._sgraph = SGraph.from_seed(self.seeds, self.item_factory, self.config)
```
reuterbal, line 465: Given that `_discover()` and `_parse_items()` are now the two patterns which mandate a rebuild, I'm wondering if the `SGraph` construction and caching should not just be put in there?

loki/bulk/scheduler.py
```diff
@@ -576,7 +590,9 @@ def write_cmake_plan(self, filepath, mode, buildpath, rootpath):
         sources_to_remove = []
         sources_to_transform = []
 
-        for item in self.items:
+        # Filter the SGraph to get a pure call-tree
+        item_filter = None if self.config.enable_imports else ProcedureItem
+        for item in SFilter(self.sgraph, item_filter=item_filter):
```
reuterbal, line 595: Thinking loud here, but could we not do this directly via the filegraph? After all, we're after files here... That would automatically apply the filtering.

reuterbal (changes requested):
Thanks, looks good in principle and happy to see this is beginning to converge to a useable solution.

This exposes `_discover()` and `_parse_items()` as the two conceptual steps that mandate a rebuild of the "master" graph, which sounds sensible (and the latter only because we fail/Fortran makes it impossible to detect some dependencies via regex'ing alone). 
To restore the clean algorithmic view of this and make the caching become an implementation detail, I think it might make sense to defer triggering the sgraph (re-)build directly to these two methods.

The change in default behaviour w.r.t. `enable_imports` causes a lot of tests to fail - I guess you will need to amend these tests in test_batch.py & co with the `enable_imports=True` constructor argument. We may also want to have at least one test for the `enable_imports=False` case - maybe some of these tests can be parameterised accordingly?

Same is true for the ecwam regression tests, where we need to enable imports again in the config. Do you want me to apply the change on the paired staging branch there?

Finally, it would be good to briefly document this option in the relevant section in the docs, probably somewhere in here:

https://github.com/ecmwf-ifs/loki/blob/31717002dd07368b24127d622f8bf9e9198dd082/docs/source/transform.rst?plain=1#L151-L216

loki/bulk/scheduler.py
reuterbal: Ah, yes, in that case leave as is and maybe flag this in an issue? Probably more generally the concept of propagating item properties to file items. Just out of curiosity, how is that property specified in the first place?

mlange05:
@reuterbal As discussed offline, I needed to revert the performance "fix" for parser_class comparison; I've left the revert commit in to get the test base green for now, but feel free to rebase out the respective pair of commits before merging into the main scheduler PR.

loki/bulk/scheduler.py
reuterbal: Filed #255 to track this

reuterbal (approved):
Thanks, looks good to me now. I'll put this in and mop up the documentation in the Scheduler branch directly.

Issue #253, opened 2024-03-19
Repository reorganisation


mlange05:
As we are maturing the outward-facing components of Loki, we have been contemplating a slight repository reorganisation to better encapsulate the different layers of the API and organise the re-use within these. I'm filing this as an issue here to draw up a first draft of the overall shape and contours of the sub-packages that we would long-term envision. This is primarily meant for discussion and is open to input from all.

Moreover, I think @reuterbal had also suggested in offline discussions to potentially split up the test base and move the specific tests into the respective sub-packages, allowing us to tighten up the interfaces a little.

Here a first draft organisation that we should probably evolve over time:

```
-loki
    - analyse
    - backend         | fgen, cgen, pygen, etc
    - batch              | scheduler and associated items, configs etc
        - tests
        - scheduler
        - items
        - config
        - filter
    - expression     | The expression/symbol layer alongside it's pymbolic interfaces and visitors
        - tests
        - symbols
        - symbolic
        - expr_visitor
        - 
    - frontend
    - ir_tree             | I'd like to merge the middle-layer (IR plus IR-visitors); better name needed!
        - tests
        - ir
        - visitor
        - transformer
        - pragma_utils
    - scopes            | Arguable could be merged with ir_tree
        - tests
        - scope
        - program_unit
        - subroutine/module
        - sourcefile
    - transformations
        - tests
        - general
        - inline_outline
        - single_column
        - transpile
        - temporaries
        - offload

``` 

Just a few general ideas, open for comments... @reuterbal @awnawab @MichaelSt98

reuterbal:
Just a few ideas off the top of my head:

* We should also add a `cli` module that could provide common CLI functionality that is currently replicated between the scripts, similar to what has been done in ifsbench. This could also provide a common "loki" CLI command and retain the "loki-transform.py" and "loki-lint.py" as additional aliases. That way, the `scripts` directory could be eliminated
* Naming proposals:
```
- expression
  - tests
  - symbols                    | Only variables / literals?
  - operations                | I'd like to separate operations (inlinecall, subscript, sum, ...) from the symbols/literals
  - symbolic
  - expr_visitor
- ir
  - tests
  - nodes
  - visitor
  - transformer
  - pragma_utils
```
[...]

PR #258, opened 2024-03-22
Rename `bulk`->`batch` and create `ir` sub-package

mlange05:
First steps towards the big spring clean (issue #253), which renames two sub-packages. Since this is a tedious process (re-wiring all imports) I'll propose these in small doses.

In this one we do the two uncontroversial first steps:
* Rename `bulk` => `batch`
* Separate SGraph and SFilter into separate files
* Create `ir` sub-package and move visitors next to nodes

_Triggering this early for test coverage..._


reuterbal (approved):
Thanks, looks good! Despite the large impact I would expect potential fallout would reliably be caught via missing imports.

reuterbal:
OMNI failures in regression tests have vanished. Could be due to #261 or just lucky...

PR #259, opened 2024-03-25
SingleColumn: Demote arrays that are not used at all in the body

mlange05:
Pretty much what it says on the tin: somehow we've missed demoting arrays that are unused in the body but are otherwise perfectly fine to demote in the SCC-family demotion step. This PR fixes this.


reuterbal (approved):
Thanks, looks good. Waiting for tests to confirm, otherwise GTG

PR #263, opened 2024-03-27
Scheduler: Fix handling of external module procedures

reuterbal:
Procedures in missing modules (that are recognized as ExternalItem) were not recognized as ExternalItems and instead created as ProcedureItem with missing source. This fixes that erroneous behaviour.


reuterbal:
Piggy-backed a small fix to the regex frontend: functions with a kind-specifier and return-type declaration in the prefix (i.e., `=` in the function prefix), wouldn't match if embedded in a module. modified an existing test accordingly

mlange05 (approved):
Looks good to me. GTG :shipit:

PR #265, opened 2024-04-02
Skip driver routine in `GlobalVariableAnalysis`

awnawab:
Global variables imported at the driver layer passed down to the compute kernel as a subroutine argument should not be included in the `GlobalVariableAnalysis`. Rather, they should be handled via the `DataOffloadTransformation`.

### DIFF
--- a/transformations/tests/test_data_offload.py
+++ b/transformations/tests/test_data_offload.py
@@ -388,24 +388,20 @@ def test_global_variable_analysis(frontend, key, config, global_variable_analysi
         nval_dim = '1:5'
         nfld_data = set()
         nval_data = set()
-        nval_offload = set()
-        nfld_offload = set()
     else:
         nfld_dim = 'nfld'
         nval_dim = 'nval'
         nfld_data = {('nfld', 'global_var_analysis_header_mod')}
         nval_data = {('nval', 'global_var_analysis_header_mod')}
-        nval_offload = {'nval'}
-        nfld_offload = {'nfld'}
 
     expected_trafo_data = {
         'global_var_analysis_header_mod': {
             'declares': {f'iarr({nfld_dim})', f'rarr({nval_dim}, {nfld_dim})'},
-            'offload': {f'iarr({nfld_dim})', f'rarr({nval_dim}, {nfld_dim})'} | nval_offload | nfld_offload,
+            'offload': {}
         },
         'global_var_analysis_data_mod': {
             'declares': {'rdata(:, :, :)', 'tt'},
-            'offload': {'rdata(:, :, :)', 'tt', 'tt%vals'}
+            'offload': {}
         },
         'global_var_analysis_data_mod#some_routine': {'defines_symbols': set(), 'uses_symbols': set()},
         'global_var_analysis_kernel_mod#kernel_a': {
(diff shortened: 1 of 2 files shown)


transformations/tests/test_data_offload.py
```diff
@@ -435,7 +435,7 @@ def test_global_variable_analysis(frontend, key, config, global_variable_analysi
 
     assert set(scheduler.items) == set(expected_trafo_data) | {'global_var_analysis_data_mod#some_type'}
     for item in scheduler.items:
-        if item == 'global_var_analysis_data_mod#some_type':
+        if item == 'global_var_analysis_data_mod#some_type' or item.config['role'] == 'driver':
```
MichaelSt98, line 438: So we could remove `'#driver'` in `expected_trafo_data`?

transformations/transformations/data_offload.py
```diff
@@ -274,6 +274,9 @@ def transform_subroutine(self, routine, **kwargs):
         if 'successors' not in kwargs:
             raise RuntimeError('Cannot apply GlobalVariableAnalysis without successors to store offload analysis data')
 
+        if kwargs['role'] == 'driver':
```
MichaelSt98, line 277: Maybe I misunderstood something, but why shouldn't we apply the `GlobalVariableAnalysis` to the driver?

I can see that we don't use this information right now for anything, but I don't understand why we should discard this information rather than not using this information in the subsequent transformation pass(es)?!

awnawab: A global variable that is imported only at the driver layer would have to be passed down to the kernel as a subroutine argument. In this case, from the point of view of the kernel, it stops being a global variable and behaves like any other argument, and we wouldn't need to add `!$acc declare create` clauses. Such variables should therefore be handled by the `DataOffloadTransformation` instead.

MichaelSt98: Sure, but from the point of view of the driver it is a global variable and I think the `Analysis` part should be as general as possible for potential future `Transformation` s that possibly need to know global variables from the point of view of the driver?!

awnawab: So on the driver we could imagine such a scenario:
```
subroutine driver()
use some_mod, only : some_var

!$acc data
some_var = ...
!$acc end data

end subroutine driver
```
The only "device" code in a driver routine would be inside an `!$openacc/openmp` data region, and we would have to apply data movement directives to `some_var` for that data region, just like we would for any other variable passed in to the data region. Module imports at the driver layer really do behave like any other variable passed in to the openacc data region/ device kernel. If a future `Transformation` were to use driver layer module imports in a way that is not covered by the `DataOffloadTransformation`, then that should be functionality we explicitly add.

reuterbal: In principle I agree with Michael's comment: An analysis pass shouldn't make unnecessary assumptions about the intended use of the data it collects. 

Unfortunately, the current analysis implementation performs a reduction of the data collected by successors, and therefore the information what has been used in the driver and what stems from kernels called in the driver is no longer separable directly in the `trafo_data`. However, the data offload transformation already works only on the `successors` data, therefore I'm currently not seeing how this change actually has any practical implications?

awnawab: The`GlobalVariableAnalysis` on ecwam with scc-hoist fails at the driver layer in quite an obscure way.

I understood the `GlobalVariableAnalysis` to be targeted at variables imported from within _device_ code. Variables imported only at the driver layer don't really fit this model, as they would be variables imported in _host_ code that might potentially be used in device code.

That's why I think driver layer only imports fit the conceptual model of the `DataOffloadTransformation` much better than the `GlobalVarAnalysis/Transformation`, but of course there's no harm in keeping the analysis at the driver layer.

I'll investigate the ecwam issues further and report back once I have a fix/workaround. We can then potentially close this PR.

awnawab: Update: the ecwam scc-hoist problem is partially solved in the bug fix PR #279. For the remaining issue, let's please discuss offline when I back on Thursday.

awnawab:
As discussed offline, I have moved the propagation of the offload requirements from the `GlobalVarAnalysis` to the `GlobalVarOffloadTransformation`. This combined with PR #279 works on ecWAM.

reuterbal (approved):
Thanks, very happy with this solution!

PR #266, opened 2024-04-02
improved `replace_intrinsics` and added `rename_variables`

MichaelSt98:
`replace_intrinsics()` wasn't working properly, especially for used symbols/functions within the `routine.spec` e.g., for initialising parameters ...

`rename_variables()` as additional utility to rename variables/arguments e.g. `const` within ecwam ...

### DIFF
--- a/loki/expression/expr_visitors.py
+++ b/loki/expression/expr_visitors.py
@@ -142,8 +142,12 @@ class ExpressionFinder(Visitor):
         expressions = ()
         for v in o.symbols:
             if v.type.initial is not None:
-                expressions += (self.retrieve(v.type.initial),)
-        return self._return(o, expressions)
+                retrieved = self.retrieve(v.type.initial)
+                if retrieved:
+                    expressions += as_tuple(retrieved)
+        if expressions:
+            return self._return(o, expressions)
+        return super().visit(o.children, **kwargs)
 
 
 class FindExpressions(ExpressionFinder):
--- a/loki/expression/expr_visitors.py
+++ b/loki/expression/expr_visitors.py
@@ -138,6 +138,13 @@ class ExpressionFinder(Visitor):
         """
         return self._return(o, ())
 
+    def visit_VariableDeclaration(self, o, **kwargs):
+        expressions = ()
+        for v in o.symbols:
+            if v.type.initial is not None:
+                expressions += (self.retrieve(v.type.initial),)
+        return self._return(o, expressions)
+
 
 class FindExpressions(ExpressionFinder):
     """
--- a/loki/expression/expr_visitors.py
+++ b/loki/expression/expr_visitors.py
@@ -139,15 +139,11 @@ class ExpressionFinder(Visitor):
         return self._return(o, ())
 
     def visit_VariableDeclaration(self, o, **kwargs):
-        expressions = ()
+        expressions = as_tuple(super().visit(o.children, **kwargs))
         for v in o.symbols:
             if v.type.initial is not None:
-                retrieved = self.retrieve(v.type.initial)
-                if retrieved:
-                    expressions += as_tuple(retrieved)
-        if expressions:
-            return self._return(o, expressions)
-        return super().visit(o.children, **kwargs)
+                expressions += as_tuple(self.retrieve(v.type.initial))
+        return self._return(o, expressions)
 
 
 class FindExpressions(ExpressionFinder):
--- a/loki/transform/transform_utilities.py
+++ b/loki/transform/transform_utilities.py
@@ -28,7 +28,7 @@ from loki.types import SymbolAttributes, BasicType, DerivedType, ProcedureType
 
 
 __all__ = [
-    'convert_to_lower_case', 'replace_intrinsics', 'sanitise_imports',
+    'convert_to_lower_case', 'replace_intrinsics', 'rename_variables', 'sanitise_imports',
     'replace_selected_kind', 'single_variable_declaration', 'recursive_expression_map_update'
 ]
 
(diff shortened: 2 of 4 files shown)


loki/expression/expr_visitors.py
```diff
@@ -138,6 +138,17 @@ def visit_TypeDef(self, o, **kwargs):
         """
         return self._return(o, ())
 
+    def visit_VariableDeclaration(self, o, **kwargs):
+        expressions = ()
+        for v in o.symbols:
+            if v.type.initial is not None:
+                retrieved = self.retrieve(v.type.initial)
+                if retrieved:
+                    expressions += as_tuple(retrieved)
```
reuterbal, line 147: Since `as_tuple` returns an empty tuple even for `None`, the following should suffice:
```suggestion
                retrieved = as_tuple(self.retrieve(v.type.initial))
```

loki/expression/expr_visitors.py
```diff
@@ -138,6 +138,17 @@ def visit_TypeDef(self, o, **kwargs):
...
+        if expressions:
+            return self._return(o, expressions)
+        return super().visit(o.children, **kwargs)
```
reuterbal, line 150: I'm not sure I'm following the logic here. The real purpose of this visitor method is to include the `initial` attribute in the lookup, no? However, the implementation here reads as if it has the side-effect of not searching the other expressions in the declaration anymore (e.g., a `dimensions`/`shape` attribute) if any `initial` attribute yielded a relevant return value.

Wouldn't something like this be an equivalent and more straightforward implementation?
```py
expressions = super().visit(o, **kwargs)
for v in o.symbols:
    if v.type.initial is not None:
        expressions += as_tuple(self.retrieve(v.type.initial))
return self._return(o, expressions)
```

loki/transform/transform_utilities.py
```diff
@@ -144,6 +144,36 @@ def replace_intrinsics(routine, function_map=None, symbol_map=None, case_sensiti
     routine.spec = SubstituteExpressions(callmap).visit(routine.spec)
     routine.body = SubstituteExpressions(callmap).visit(routine.body)
 
+def rename_variables(routine, symbol_map=None):
+    """
+    Replace symbols/variables including (routine) arguments.
```
reuterbal, line 149:
```suggestion
    Rename symbols/variables including (routine) arguments.
```

loki/transform/transform_utilities.py
```diff
@@ -144,6 +144,36 @@ def replace_intrinsics(routine, function_map=None, symbol_map=None, case_sensiti
...
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine object in which to replace intrinsic calls
```
reuterbal, line 154:
```suggestion
        The subroutine object in which to rename variables.
```

loki/transform/transform_utilities.py
```diff
@@ -144,6 +144,36 @@ def replace_intrinsics(routine, function_map=None, symbol_map=None, case_sensiti
...
+    # rename variable declarations and usages
+    var_map = {}
+    for var in FindVariables(unique=False).visit(routine.ir):
+        if var.name in symbol_map:
+            new_var = symbol_map[var.name]
+            if new_var is not None:
+                var_map[var] = var.clone(name=symbol_map[var.name])
+    if var_map:
+        routine.spec = SubstituteExpressions(var_map).visit(routine.spec)
+        routine.body = SubstituteExpressions(var_map).visit(routine.body)
```
reuterbal, line 176: Maybe we should update here also the symbol table to make sure we don't have leftover entries under the previous name? Something along the lines of
```py
for name in var_map:
    if name in routine.symbol_attrs:
        del routine.symbol_attrs[name]
```

tests/test_transform_utilities.py
```diff
@@ -193,3 +194,69 @@ def test_transform_utilities_recursive_expression_map_update(frontend):
     assert fgen(routine.body.body[0]).lower() == 'my_obj%a = my_obj%my_add(my_obj%a(1:my_obj%m, 1:my_obj%n), 1.)'
     routine.body = SubstituteExpressions(expr_map).visit(routine.body)
     assert fgen(routine.body.body[0]) == 'obj%a = obj%my_add(obj%a(1:obj%m, 1:obj%n), 1.)'
+
+@pytest.mark.parametrize('frontend', available_frontends(skip=[(OMNI, 'Argument mismatch for "min"')]))
+def test_tranform_utilites_replace_intrinsics(frontend):
```
reuterbal, line 199:
```suggestion
def test_transform_utilites_replace_intrinsics(frontend):
```

tests/test_transform_utilities.py
```diff
@@ -193,3 +194,69 @@ def test_transform_utilities_recursive_expression_map_update(frontend):
...
+    assert inline_calls[1].name == 'fmax'
+    assert inline_calls[2].name == 'fmin'
+    variables = FindVariables(unique=False).visit(routine.ir)
+    assert 'DBL_EPSILON' in variables
+    assert 'epsilon' not in variables
+    # check wether it really worked for variable declarations or rather parameters
+    assert 'DBL_EPSILON' in FindVariables().visit(routine.variable_map['param'].initial)
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_tranform_utilites_rename_variables(frontend):
```
reuterbal, line 228:
```suggestion
def test_transform_utilites_rename_variables(frontend):
```

reuterbal (changes requested):
Thanks, a very useful utility and valuable improvement. The renaming is something that's annoyingly complex right now and having this as a utility is extremely useful.

There are some typos in the test, and the FindExpression method seems not quite right to me. Also, I have some doubts about the behaviour in some more complex cases and suggested some additional testing.

loki/expression/expr_visitors.py
MichaelSt98: `expressions = super().visit(o, **kwargs)` gives me a problem with recursion ...

loki/transform/transform_utilities.py
MichaelSt98: Yes! Suggested solution not working for `DerivedType` s.

tests/test_transform_utilities.py
```diff
@@ -193,3 +194,69 @@ def test_transform_utilities_recursive_expression_map_update(frontend):
...
+    fcode = """
```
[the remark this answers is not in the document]
MichaelSt98: Fair enough. Expanded the tests and fixed/improved functionality of `rename_variables()`.

reuterbal (approved):
Thanks, this looks very good now!

PR #267, opened 2024-04-02
`cgen`: multiconditional/switch/select case statement

MichaelSt98:
allow transpilation of 

```fortran
select case (...)
    ...
end select
```

### DIFF
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -10,6 +10,7 @@ from pymbolic.mapper.stringifier import (
     PREC_UNARY, PREC_LOGICAL_OR, PREC_LOGICAL_AND, PREC_NONE, PREC_CALL
 )
 
+from loki.tools import as_tuple
 from loki.ir import Import, Stringifier, FindNodes
 from loki.expression import LokiStringifyMapper, Array, symbolic_op, Literal
 from loki.types import BasicType, SymbolAttributes, DerivedType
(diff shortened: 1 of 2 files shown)


loki/backend/cgen.py
```diff
@@ -364,6 +365,35 @@ def visit_TypeDef(self, o, **kwargs):
...
+            [...body...]
+          [default:]
+            [...body...]
+          }
+        """
+        header = self.format_line('switch (', self.visit(o.expr, **kwargs), ') {')
+        cases = []
+        end_cases = []
+        for value in o.values:
+            case = self.visit_all(as_tuple(value), **kwargs)
```
reuterbal, line 384: Note that you'll likely run into issues when the cases are ranges (which is allowed in Fortran: https://j3-fortran.org/doc/year/10/10-007.pdf#page=200). These are likely not directly transferable to C, but maybe it would be useful to throw an exception if a `RangeIndex` (I presume that's what we would use to represent them here) is encountered?

reuterbal (approved):
Thanks, this looks very good.

I've pointed out a limitation of the Fortran-to-C mapping, which is not easily resolvable, but maybe you want to test for it and throw an error if it is encountered.

loki/backend/cgen.py
MichaelSt98: Lovely ... thank you for pointing this out to me

tests/test_transpile.py
```diff
@@ -1003,3 +1003,113 @@ def test_transpile_expressions(here, builder, frontend, use_c_ptr):
...
+        function(in_var, out_var)
+        assert out_var == expected_results[i]
+
+    clean_test(filepath)
+
+    # apply F2C trafo
+    # TODO: RangeIndex as case is not yet implemented!
+    #  'NotImplementedError' is raised
+    f2c = FortranCTransformation()
+    f2c.apply(source=routine, path=here)
```
reuterbal, line 1115: Instead of marking the test as xfail, it might be better to use
```suggestion
    with pytest.raises(NotImplementedError):
        f2c.apply(source=routine, path=here)
```

mlange05 (approved):
LGTM :shipit:

PR #268, opened 2024-04-02
SCCHoist: hoist inline call temporaries and don't hoist statically declared arrays

awnawab:
This PR adds the capability to hoist temporaries declared in inline function calls. It also changes the default behaviour so that statically declared arrays are no longer hoisted.

### DIFF
--- a/loki/transform/transform_hoist_variables.py
+++ b/loki/transform/transform_hoist_variables.py
@@ -364,9 +364,9 @@ class HoistVariablesTransformation(Transformation):
         """
 
         if self.as_kwarguments:
-            new_args = dict((a.name, v.clone(dimensions=None)) for (a, v) in variables)
-            _call_clone = call.clone(kwparameters=call.kw_parameters)
-            _call_clone.kw_parameters.update(new_args)
+            kw_params = call.kw_parameters
+            kw_params.update(dict((a.name, v.clone(dimensions=None)) for (a, v) in variables))
+            _call_clone = call.clone(kw_parameters=kw_params)
             vmap = {call: _call_clone}
         else:
             new_args = tuple(v.clone(dimensions=None) for v in variables)
--- a/loki/transform/transform_hoist_variables.py
+++ b/loki/transform/transform_hoist_variables.py
@@ -87,6 +87,7 @@ from loki.tools.util import is_iterable, as_tuple, CaseInsensitiveDict
 from loki.transform.transformation import Transformation
 from loki.transform.transform_utilities import single_variable_declaration
 from loki.batch.item import ProcedureItem
+from loki.expression.expr_visitors import FindInlineCalls
 import loki.expression.symbols as sym
 
 
@@ -141,6 +142,7 @@ class HoistVariablesAnalysis(Transformation):
             item.trafo_data[self._key]["hoist_variables"] = []
 
         calls = FindNodes(CallStatement).visit(routine.body)
+        calls += FindInlineCalls().visit(routine.body)
         call_map = CaseInsensitiveDict((str(call.name), call) for call in calls)
 
         for child in successors:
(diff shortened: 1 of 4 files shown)


loki/transform/transform_hoist_variables.py
```diff
@@ -356,6 +364,30 @@ def kernel_call_argument_remapping(self, routine, call, variables):
...
+        if self.as_kwarguments:
+            new_args = dict((a.name, v.clone(dimensions=None)) for (a, v) in variables)
+            _call_clone = call.clone(kwparameters=call.kw_parameters)
+            _call_clone.kw_parameters.update(new_args)
+            vmap = {call: _call_clone}
+            routine.body = SubstituteExpressions(vmap).visit(routine.body)
+        else:
+            new_args = tuple(v.clone(dimensions=None) for v in variables)
+            vmap = {call: call.clone(parameters=call.parameters + new_args)}
+            routine.body = SubstituteExpressions(vmap).visit(routine.body)
```
MichaelSt98, line 390: You could move `routine.body = SubstituteExpressions(vmap).visit(routine.body)` outside the `if`/`else` statement ...

MichaelSt98 (approved):
Thanks, the "inline call part" seems to have been missing indeed.
Further, the solution/new default via `is_dimension_constant` seems reasonable.

loki/transform/transform_hoist_variables.py
awnawab: Ah yes, good spot, thanks!

loki/transform/transform_hoist_variables.py
```diff
@@ -341,6 +349,30 @@ def kernel_call_argument_remapping(self, routine, call, variables):
...
+        call : :any:`InlineCall`
+            ProcedureSymbol to which hoisted variables will be added.
+        variables : tuple of :any:`Variable`
+            The tuple of variables to be declared.
+        """
+
+        if self.as_kwarguments:
+            new_args = dict((a.name, v.clone(dimensions=None)) for (a, v) in variables)
+            _call_clone = call.clone(kwparameters=call.kw_parameters)
+            _call_clone.kw_parameters.update(new_args)
```
reuterbal, line 369: I find this pattern of in-place updating the call attribute a little weird, particularly since we expect to only add new kwargs to the end. Wouldn't something more similar to the other argument_remapping method be a little more intuitive?

Also, the clone call has a typo (`kwparameters` vs. `kw_parameters`), which suggests we don't have a test with existing `kw_parameters` :) Could you add one for that?

```suggestion
            kw_params = call.kw_parameters
            kw_params += tuple((a.name, v.clone(dimensions=None)) for (a, v) in variables)
            _call_clone = call.clone(kwparameters=dict(kw_params))
```

reuterbal (changes requested):
Thanks, nice find and great fix. There seems to be a typo in the implementation, though, which suggests a shortfall in testing. I think it would be good to expand testing to cover that.

reuterbal (approved):
Many thanks, this looks great now!

mlange05 (approved):
Only had a brief looks, but LGTM. :shipit:

loki/transform/transform_hoist_variables.py
```diff
@@ -240,7 +243,7 @@ def transform_subroutine(self, routine, **kwargs):
             routine.arguments += hoisted_temporaries
 
         call_map = {}
-        for call in FindNodes(CallStatement).visit(routine.body):
+        for call in FindNodes(CallStatement).visit(routine.body) + list(FindInlineCalls().visit(routine.body)):
```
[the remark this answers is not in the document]
reuterbal: We have a bit of a mismatch there, something we should clean up separately:
- `FindNodes` has indeed always returned lists. Since these are built incrementally and can be quite long, there could indeed a performance impact from using tuples
- Expression visitors return by default sets, unless `unique=False` is used.

PR #269, opened 2024-04-02
`cgen`: return type and var for function(s)

MichaelSt98:
`cgen`: return type and var for function(s)

### DIFF
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -173,14 +173,10 @@ class CCodegen(Stringifier):
                      for a, p in zip(o.arguments, aptr)]
 
         # check whether to return something and define function return type accordingly
-        return_var = None
         if o.is_function:
-            # Determine function result variable name
-            if not (result_name := o.result_name):
-                result_name = o.name.replace('_c', '')
-            if result_name in o.variable_map:
-                return_var = o.variable_map[result_name]
-        return_type = c_intrinsic_type(return_var.type) if return_var is not None else 'void'
+            return_type = c_intrinsic_type(o.return_type)
+        else:
+            return_type = 'void'
 
         header += [self.format_line(f'{return_type} ', o.name, '(', self.join_items(arguments), ') {')]
 
@@ -191,11 +187,10 @@ class CCodegen(Stringifier):
 
         # Fill the body
         body += [self.visit(o.body, **kwargs)]
-        # body += [self.format_line('return 0;')]
 
         # if something to be returned, add 'return <var>' statement
-        if return_var is not None:
-            body += [self.format_line(f'return {return_var.name.lower()};')]
+        if o.result_name is not None:
+            body += [self.format_line(f'return {o.result_name.lower()};')]
 
         # Close everything off
         self.depth -= 1
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -171,7 +171,18 @@ class CCodegen(Stringifier):
                 aptr += ['']
         arguments = [f'{self.visit(a.type, **kwargs)} {p}{a.name.lower()}'
                      for a, p in zip(o.arguments, aptr)]
-        header += [self.format_line('int ', o.name, '(', self.join_items(arguments), ') {')]
+
+        # check whether to return something and define function return type accordingly
+        return_var = None
+        if o.is_function:
+            # Determine function result variable name
+            if not (result_name := o.result_name):
+                result_name = o.name.replace('_c', '')
+            if result_name in o.variable_map:
+                return_var = o.variable_map[result_name]
+        return_type = c_intrinsic_type(return_var.type) if return_var is not None else 'void'
+
+        header += [self.format_line(f'{return_type} ', o.name, '(', self.join_items(arguments), ') {')]
 
         self.depth += 1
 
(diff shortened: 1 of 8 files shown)


loki/backend/cgen.py
```diff
@@ -174,7 +174,18 @@ def visit_Subroutine(self, o, **kwargs):
...
+
+        # check whether to return something and define function return type accordingly
+        return_var = None
+        if o.is_function:
+            # Determine function result variable name
+            if not (result_name := o.result_name):
+                result_name = o.name.replace('_c', '')
+            if result_name in o.variable_map:
+                return_var = o.variable_map[result_name]
+        return_type = c_intrinsic_type(return_var.type) if return_var is not None else 'void'
```
reuterbal, line 186: This reimplements the logic in https://github.com/ecmwf-ifs/loki/blob/31eabbee02531a7e5f25030cdf60b27ff0a7cfbd/loki/subroutine.py#L324-L332
but handles the special case that the function name no longer matches the name of the return variable.  But that seems like an inconsistent IR state.

I would suggest to instead handle that renaming in the f2c transformation before this line:
https://github.com/ecmwf-ifs/loki/blob/31eabbee02531a7e5f25030cdf60b27ff0a7cfbd/loki/transform/fortran_c_transform.py#L413
with something like this:
```py
if kernel.is_function and kernel.result_name is None:
    kernel.result_name = kernel.name
```

I think that should allow you to simply write here then
```suggestion
        if o.is_function:
            return_type = c_intrinsic_type(o.return_type)
        else:
            return_type = 'void'
```

loki/backend/cgen.py
```diff
@@ -183,7 +194,11 @@ def visit_Subroutine(self, o, **kwargs):
 
         # Fill the body
         body += [self.visit(o.body, **kwargs)]
-        body += [self.format_line('return 0;')]
+        # body += [self.format_line('return 0;')]
```
reuterbal, line 197: debug leftover?

loki/backend/cgen.py
```diff
@@ -183,7 +194,11 @@ def visit_Subroutine(self, o, **kwargs):
...
+
+        # if something to be returned, add 'return <var>' statement
+        if return_var is not None:
+            body += [self.format_line(f'return {return_var.name.lower()};')]
```
reuterbal, line 201: With the above change, this would then become:
```suggestion
        if o.result_name is not None:
            body += [self.format_line(f'return {o.result_name.lower()};')]
```

reuterbal (changes requested):
Thanks, in principle this looks very good. But the slight back-bending in `cgen` to determine return type and return variable suggests to me that we are actually leaving the IR in an inconsistent state for functions, and try to work around this now instead of fixing the original problem.

I've added a suggestion how this might be remedied. Of course it's untested and might unearth other issues, but let me know what you think.

loki/backend/cgen.py
```diff
@@ -183,7 +190,10 @@ def visit_Subroutine(self, o, **kwargs):
 
         # Fill the body
         body += [self.visit(o.body, **kwargs)]
-        body += [self.format_line('return 0;')]
+
+        # if something to be returned, add 'return <var>' statement
+        if o.result_name is not None:
```
MichaelSt98, line 195: Do we want to double-check wether `o.is_function`?

MichaelSt98:
Within `FortranCTransformation`

```py
if kernel.is_function and kernel.result_name is None:
    kernel.result_name = kernel.name
```

not necessary anymore, since `result_name` is now set by default if `is_function=True`

loki/backend/cgen.py
reuterbal: Good point, since we can now rely on `o.result_name` to always have a sensible value for functions, we don't have to check for result_name but can rather check for is_function

loki/subroutine.py
```diff
@@ -104,6 +104,10 @@ def __initialize__(
         self.result_name = result_name
         self.is_function = is_function
 
+        # Make sure 'result_name' is defined if it's a function
+        if self.result_name is None and self.is_function:
+            self.result_name = name
```
reuterbal, line 109: Thanks! With this, we can now simplify the code here:
https://github.com/ecmwf-ifs/loki/blob/6b58fe7fef3e151bbf74df18e034a1bc7924e023/loki/subroutine.py#L330-L332

Would you mind replacing this simply with the following?
```py
return self.symbol_attrs.get(self.result_name) 
```

reuterbal (changes requested):

reuterbal (approved):
Many thanks, this looks great now!

@mlange05: This implies a small IR change in the sense that functions will always record the name of the variable that contains the return value in `result_name`. That makes accessing this information much easier without the need to figure out which of the multiple variants of defining the result variable was used in the original Fortran.
For language compliance, the backend will only add the corresponding `RESULT(<...>)` appendix to the function statement if the name differs from the function name, thus retaining the original behaviour.
A positive side-effect of this is that renaming of functions can now happen without the need to update the result variable's name throughout the entire function body.
Let us know if you're happy with this!

mlange05 (approved):
Yes, very happy with the IR / behaviour change (was planning on something like that for the eventual procedure overhaul). Great stuff, GTG from me, and thanks both! 🙏 :shipit:

PR #271, opened 2024-04-04
Pipeline configuration from file

mlange05:
_Note this PR is built on top of and depends on #260._

This PR extends the `Pipeline` abstraction to make it composable and allow configuration from via the scheduler config files (in .yml). In particular, it adds `prepend`/`append` methods and support for a native `pipeline += trafo` notation to allow building custom pipelines from instantiated pipelines and transformations. 

It then also adds a new entry to the scheduler config where custom, complex transformation pipelines can be defined that combine internally provided, as well as externally defined transformations to allow full customisation outside of the usual top-level entry points. The use of custom transformation pipelines is then exposed via `loki-transform.py convert`.

This new generic interface for custom pipelines allows us to re-create the thus far internally defined behaviour entirely via config files. For an optional example of how Loki-SCC can be recreated, please see:
https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/naml-loki-scc-pipeline-config/src/cloudsc_loki/cloudsc_loki.config#L75

In more detail:
* Add intrinsic `.prepend()` and `.append()` methods to `Pipeline` objects for full instantiated `Transformation` and `Pipeline` objects.
* Add support for native "addition" of transformations and pipelines that behaves as suggested.
* Add pretty-printing that allows a readable representation of complex pipelines under `str(pipeline)`
* Allow instiating `Pipeline` objects in via the `SchedulerConfig` and add entry point to `convert` for these custom pipelines

### DIFF
--- a/loki/batch/configure.py
+++ b/loki/batch/configure.py
@@ -16,7 +16,10 @@ from loki.types import ProcedureType, DerivedType
 from loki.logging import error, warning
 
 
-__all__ = ['SchedulerConfig', 'TransformationConfig', 'ItemConfig']
+__all__ = [
+    'SchedulerConfig', 'TransformationConfig', 'PipelineConfig',
+    'ItemConfig'
+]
 
 
 class SchedulerConfig:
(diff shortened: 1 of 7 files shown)


reuterbal (approved):
Thanks, this is truly fantastic and not much to add to this. If anything, a different exception type might be more appropriate when composing the pipeline and some additional test coverage for a gold star.

PR #273, opened 2024-04-05
F2C: `DeReferenceTrafo`

MichaelSt98:
`DeReferenceTrafo` for and within `FortranCTransformation` to insert apply/insert `Dereference` = `*` and `Reference`/*address-of* = `&` operators.

### DIFF
--- a/loki/transform/fortran_c_transform.py
+++ b/loki/transform/fortran_c_transform.py
@@ -41,6 +41,54 @@ from loki.types import BasicType, DerivedType, SymbolAttributes
 __all__ = ['FortranCTransformation']
 
 
+class DeReferenceTrafo(Transformer):
+    """
+    Transformation to apply/insert Dereference = `*` and
+    Reference/*address-of* = `&` operators.
+
+    Parameters
+    ----------
+    vars2dereference : list
+        Variables to be dereferenced. Ususally the arguments except
+        for scalars with `intent=in`.
+    """
+    # pylint: disable=unused-argument
+
+    def __init__(self, vars2dereference):
+        super().__init__()
+        self.retriever = ExpressionRetriever(self.is_dereference)
+        self.vars2dereference = vars2dereference
+
+    @staticmethod
+    def is_dereference(symbol):
+        return isinstance(symbol, (DerivedType, Array, Scalar)) and not (
+            isinstance(symbol, Array) and symbol.dimensions is not None
+            and not all(dim == sym.RangeIndex((None, None)) for dim in symbol.dimensions)
+        )
+
+    def visit_Expression(self, o, **kwargs):
+        symbol_map = {
+            symbol: Dereference(symbol.clone()) for symbol in self.retriever.retrieve(o)
+            if symbol.name.lower() in self.vars2dereference
+        }
+        return SubstituteExpressionsMapper(symbol_map)(o)
+
+    def visit_CallStatement(self, o, **kwargs):
+        new_args = ()
+        call_arg_map = dict((v,k) for k,v in o.arg_map.items())
+        for arg in o.arguments:
+            if not self.is_dereference(arg) and (isinstance(call_arg_map[arg], Array)\
+                    or call_arg_map[arg].type.intent.lower() != 'in'):
+                new_args += (Reference(arg.clone()),)
+            else:
+                if isinstance(arg, Scalar) and call_arg_map[arg].type.intent.lower() != 'in':
+                    new_args += (Reference(arg.clone()),)
+                else:
+                    new_args += (arg,)
+        o._update(arguments=new_args)
+        return o
+
+
 class FortranCTransformation(Transformation):
     """
     Fortran-to-C transformation that translates the given routine
@@ -402,6 +450,19 @@ class FortranCTransformation(Transformation):
         header_module.rescope_symbols()
         return header_module
 
+    @staticmethod
+    def apply_de_reference(routine):
+        """
+        Utility method to apply/insert Dereference = `*` and
+        Reference/*address-of* = `&` operators.
+        """
+        to_be_dereferenced = []
+        for arg in routine.arguments:
+            if not(arg.type.intent.lower() == 'in' and isinstance(arg, Scalar)):
+                to_be_dereferenced.append(arg.name.lower())
+
+        routine.body = DeReferenceTrafo(to_be_dereferenced).visit(routine.body)
+
     def generate_c_kernel(self, routine):
         """
         Re-generate the C kernel and insert wrapper-specific peculiarities,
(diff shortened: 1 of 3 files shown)


loki/transform/fortran_c_transform.py
```diff
@@ -477,17 +478,51 @@ def generate_c_kernel(self, routine):
...
+                super().__init__()
+                self.retriever = ExpressionRetriever(lambda e: isinstance(e, (DerivedType, Array, Scalar))\
+                        and e.name.lower() in vars2dereference)
+
+            def visit_Expression(self, o, **kwargs):
+                symbols = self.retriever.retrieve(o)
+                symbol_map = {}
+                for symbol in symbols:
+                    if isinstance(symbol, Array) and symbol.dimensions is not None\
+                            and not all(dim == sym.RangeIndex((None, None)) for dim in symbol.dimensions):
```
reuterbal, line 504: Tthe creation of the symbol map is then simplified to 
```py
symbol_map = {
    symbol: Dereference(symbol.clone()) for symbol in self.retriever.retrieve()
    if symbol.name.lower() in self.vars2dereference
}
```

loki/transform/fortran_c_transform.py
```diff
@@ -477,17 +478,51 @@ def generate_c_kernel(self, routine):
...
                 if isinstance(arg.type.dtype, DerivedType):
                     # Lower case type names for derived types
                     typedef = _type.dtype.typedef.clone(name=_type.dtype.typedef.name.lower())
                     _type = _type.clone(dtype=typedef.dtype)
-                var_map[arg] = Dereference(arg)
+                to_be_dereferenced.append(arg.name.lower())
                 kernel.symbol_attrs[arg.name] = _type
-        kernel.body = SubstituteExpressions(var_map).visit(kernel.body)
+
+        class DeReferenceTrafo(Transformer):
```
reuterbal, line 492: The criterion for a symbol that needs (de)referencing is the same throughout, so maybe it would make sense to capture this as a static method in the transformer:

```py
@staticmethod
def is_reference(symbol):
    return (
        isinstance(symbol, Array) 
        and not (symbol.dimensions is None or all(dim == ':' for dim in symbol.dimensions))
    )
```

loki/transform/fortran_c_transform.py
```diff
@@ -477,17 +478,51 @@ def generate_c_kernel(self, routine):
...
+
+            def __init__(self, vars2dereference):
+                super().__init__()
+                self.retriever = ExpressionRetriever(lambda e: isinstance(e, (DerivedType, Array, Scalar))\
+                        and e.name.lower() in vars2dereference)
```
reuterbal, line 497: With the above method, you could use this directly in the retriever:
```py
self.retriever = ExpressionRetriever(self.is_reference)
self.vars2dereference = vars2dereference
```

loki/transform/fortran_c_transform.py
```diff
@@ -477,17 +478,51 @@ def generate_c_kernel(self, routine):
...
+                            and not all(dim == sym.RangeIndex((None, None)) for dim in symbol.dimensions):
+                        continue
+                    symbol_map[symbol] = Dereference(symbol.clone())
+                return SubstituteExpressionsMapper(symbol_map)(o)
+
+            def visit_CallStatement(self, o, **kwargs):
+                new_args = ()
+                call_arg_map = dict((v,k) for k,v in o.arg_map.items())
+                for arg in o.arguments:
+                    if isinstance(arg, Array) and arg.dimensions\
```
reuterbal, line 513: This conditional would then also become more readable as
```py
if self.is_reference(arg) and (isinstance(call_arg_map[arg], Array) or call_arg_map[arg].type.intent.lower() != 'in'):
```

loki/transform/fortran_c_transform.py
```diff
@@ -477,17 +478,51 @@ def generate_c_kernel(self, routine):
         convert_to_lower_case(kernel)
 
         # Force pointer on reference-passed arguments
-        var_map = {}
+        to_be_dereferenced = []
```
reuterbal, line 481: Readability of the method might benefit from putting the reference/dereference handling into a utility method maybe, and then simply calling `self.convert_args_to_references(kernel)`

tests/test_transpile.py
```diff
@@ -1004,6 +1004,60 @@ def test_transpile_expressions(here, builder, frontend, use_c_ptr):
...
+    fcode_module = """
+module transpile_call_kernel_mod
+  implicit none
+contains
+
+  subroutine transpile_call_kernel(a, b, c, arr1, len)
+    integer, intent(inout) :: a, c
+    integer, intent(in) :: b
+    integer, intent(in) :: len
+    integer, intent(inout) :: arr1(len, len)
```
reuterbal, line 1020: For my education: What is the intended outcome if we had (another) argument `arr(:,:)` here? Would this pre-empt the F2C transformation until we figure out the actual shape, or would it simply not be passed as a reference?

reuterbal (changes requested):
Thanks, this looks good. I've made some suggestions how to improve the readability of the implementation, but fundamentally no objections here.

tests/test_transpile.py
MichaelSt98: Since we need the actual `shape` information for flattening (see `flatten_arrays()` - *TypeError: Resolve shapes being of type RangeIndex, e.g., ":" before flattening!*, raised if `isinstance(shape[-2], sym.RangeIndex)`)arrays having more than one dimension, the `DeReferenceTrafo` wouldn't be executed.

However, 

```fortran
subroutine transpile_call_driver(a)
  use transpile_call_kernel_mod, only: transpile_call_kernel
    integer, parameter :: len = 5
    integer, intent(inout) :: arr1(len)
    integer, intent(inout) :: arr2(len)
    call transpile_call_kernel(arr1, arr2, len)
end subroutine transpile_call_driver

  subroutine transpile_call_kernel(arr1, arr2, len)
    integer, intent(in) :: len
    integer, intent(inout) :: arr1(len)
    integer, intent(inout) :: arr2(:)

    arr1(1) = 1
    arr2(1) = 1
  end subroutine transpile_call_kernel
```

is transformed/transpiled to:

```c
int transpile_call_driver_c() {
  int len = 5;
  transpile_call_kernel(arr1, arr2, len);
  return 0;
}

int transpile_call_kernel_c(int * restrict arr1, int * restrict arr2, int len) {

  int arr1[len];
  int arr2[len];
  arr1[1 - 1] = 1;
  arr2[1 - 1] = 1;
  return 0;
}
```

if that answers your question?!

reuterbal: It does, many thanks!

reuterbal (approved):
Many thanks, nothing further from me!

PR #274, opened 2024-04-08
REGEX frontend: white space and nesting bugfix

reuterbal:
Discovered with IFS source files.

Two small fixes for the REGEX frontend patterns, with tests that triggered before the corresponding changes:
- Multiple or no white space around the `end` statement for procedures
- Matching of module procedures is now lazy to match internal subprograms in module procedures


mlange05 (approved):
Looks good to me and behaves correctly with ec-physics regression. GTG :shipit:

PR #276, opened 2024-04-08
Marked region removal and general code removal transformation

mlange05:
This PR combines two closely related features:
* The addition of a new feature that allows pragma-marked code regions to be removed
* A clean-up and reshuffling to provide marked region removal, call removal and dead code removal all under a one umbrella `Transformation` for inclusion in more complex pipelines.

The new pragma-based region removal will pretty much do as it says: Upon encountering a region marked with `!$loki remove` / `!$loki end remove` it will simply remove the entire code region.

The new `loki.transfom.transform_remove_code` basically includes the previous dead code elimination (renamed), and combines it with the new region removal and a re-write of the `RemoveCallsTransformation`. The latter is now a as a single `Transformer` with a wrapper routine that allows it to be applied to `Subroutine` objects. All three utilities are then provided as a single `RemoveCodeTransformation` for use with the `Scheduler` and batch-processing engine. 

Since the wrapper routines for these utilities would otherwise create name clashes with the respective flags of the overarching `Transformation`, we adopt the `do_remove_xxx` naming convention, as discussed offline.

In some more detail:
* Renamed `loki.transform.transform_dead_code` => `loki.transform.transform_remove_code`
* Add `RemoveMarkedRegionsTransformer` and `do_remove_marked_regions` wrapper
* Add `RemvoeCallsTransformer` and `do_remove_calls` wrapper
* Rename `eleminate_dead_code` to `RemoveDeadCodeTransformer` and `do_remove_dead_code`
* Use new `RemoveCodeTransformation` in loki-transform and remove the old `RemoveCallsTransformation` from `transformations`
* Rename `transformations.utility_rouines` => `transformations.drhook`, as it's the last custom utility left in there.


mlange05:
Clean with EC-physics regression.

loki/transform/transform_remove_code.py
```diff
@@ -0,0 +1,333 @@
...
+class RemoveCodeTransformation(Transformation):
+    """
+    A :any:`Transformation` that provides named call and import
+    removal, code removal of pragma-marked regions and Dead Code
+    Elimination for batch processing vis the :any:`Scheduler`.
+
+    The transformation will apply the following methods in order:
+    * :method:`do_remove_calls`
+    * :method:`do_remove_marked_regions`
+    * :method:`do_remove_dead_code`
```
reuterbal, line 37: The markup is not producing the right appearance: https://sites.ecmwf.int/docs/loki/276/loki.transform.transform_remove_code.html#module-loki.transform.transform_remove_code

I think the keyword to create the link is `:meth:`, or simply `:any:` should work as well.
Not sure why the list isn't rendered as such, maybe requires an empty line before the first item?

loki/transform/transform_remove_code.py
```diff
@@ -0,0 +1,333 @@
...
+
+    transformer = RemoveDeadCodeTransformer(use_simplify=use_simplify)
+    routine.body = transformer.visit(routine.body)
+
+
+class RemoveDeadCodeTransformer(Transformer):
+    """
+    :any:`Transformer` class that removes provably unreachable code paths.
+
+    The pirmary modification performed is to prune individual code branches
```
reuterbal, line 132: Typo: pirmary -> primary

loki/transform/transform_remove_code.py
```diff
@@ -0,0 +1,333 @@
...
+        else_body = as_tuple(flatten(as_tuple(self.visit(o.else_body, **kwargs))))
+
+        if self.use_simplify:
+            condition = simplify(condition)
+
+        if condition == 'True':
+            return body
+
+        if condition == 'False':
+            return else_body
```
reuterbal, line 157: I realise that this has just been moved from the old module, but this comparison against a string put me off for a second... This is using the implicit String comparison to evaluate the condition? Would we want to account for something like `IF(1)`?

loki/transform/transform_remove_code.py
```diff
@@ -0,0 +1,333 @@
...
+    routine.body = transformer.visit(routine.body)
+
+
+class RemoveCallsTransformer(Transformer):
+    """
+    A :any:`Transformer` that removes all :any:`CallStatement` nodes
+    to specific named subroutines.
+
+    This :any:`Transformer` will by default also remove the enclosing
+    inline-conditional when encountering calls of the form ```if
```
reuterbal, line 260: A stray '`' here and after the stop.

reuterbal (approved):
Thanks, this looks very good and is a valuable utility. Also, appreciate the reshuffling code in more sensible places.

I've left a few nitpick comments about docstrings, which would be useful to clean up, but implementation looks good.

loki/transform/transform_remove_code.py
mlange05: Yes, this is actually intentional, as `IF(1)` is basically illegal in a Fortran-context (strong typing; `1` is not a `logical`) The visual oddness comes from `"True"` actually being our internal representation of `.true.` (the fortran intrinsic logical value), and that is the only thing we should match in a condition, since everything else should evaluate to that via `simplify` (eg. result of a logical operator will be a logical, never a numerical value).

mlange05:
Hi @reuterbal just a quick heads-up: there was some cross-fire between PR #271 and this one, causing a test failure for scheduler / pipeline configs. I've rebased over latest main and adjusted accordingly. Only the top two commits should be new. Hope this works now.

reuterbal:
Many thanks for catching the rebase fall-out! As discussed offline, I've done a little touch-up on the docstrings, otherwise this should be good to go!

PR #277, opened 2024-04-08
F2C: optional case-sensitivity for variables/symbols

MichaelSt98:
`convert_to_lower_case()` if not variable/symbol `.case_sensitive=True`

Imported for things like: 

```c
... = threadIdx.x
cudaMalloc(...)
```

### DIFF
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -172,7 +172,7 @@ class CCodegen(Stringifier):
                 aptr += ['*']
             else:
                 aptr += ['']
-        arguments = [f'{self.visit(a.type, **kwargs)} {p}{a.name.lower()}'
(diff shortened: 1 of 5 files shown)


reuterbal (changes requested):
Thanks. That's a good solution to the specific problem. 

I think `MetaSymbol` should derive the `case_sensitive` value directly from the wrapped `symbol`. Otherwise this looks good to me.

reuterbal (approved):
Thanks, looks good to me now!

reuterbal:
There seems some test failures after the latest rebase - hold off on merging until this is fixed.

mlange05 (approved):
Yes, makes sense and looks good. GTG from me. :shipit:

PR #278, opened 2024-04-08
Fix codecov by adding CODECOV_TOKEN

reuterbal:


mlange05 (approved):
Looks like it's working again with this. :+1: :shipit:

PR #279, opened 2024-04-08
`SCCBaseTransformation.get_integer_variable` now also checks module imports

awnawab:
The `get_integer_variable` utility in `SCCBaseTransformation` now also checks if the integer has been imported via a module.

### DIFF
--- a/transformations/tests/test_single_column_coalesced.py
+++ b/transformations/tests/test_single_column_coalesced.py
@@ -706,9 +706,16 @@ def test_single_column_coalesced_hoist_openacc(frontend, horizontal, vertical, b
     when hoisting array temporaries to driver.
     """
 
+    fcode_mod = """
+  MODULE BLOCK_DIM_MOD
+     INTEGER :: nb
+  END MODULE BLOCK_DIM_MOD
+  """
+
     fcode_driver = """
-  SUBROUTINE column_driver(nlon, nz, q, nb)
-    INTEGER, INTENT(IN)   :: nlon, nz, nb  ! Size of the horizontal and vertical
+  SUBROUTINE column_driver(nlon, nz, q)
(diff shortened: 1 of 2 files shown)


transformations/tests/test_single_column_coalesced.py
```diff
@@ -796,6 +804,9 @@ def test_single_column_coalesced_hoist_openacc(frontend, horizontal, vertical, b
     analysis = HoistTemporaryArraysAnalysis(dim_vars=(vertical.size,))
     synthesis = SCCHoistTemporaryArraysTransformation(block_dim=blocking)
 
+    # Check that blocking size has not been redefined
+    assert driver.symbol_map[blocking.size].type.module
```
reuterbal, line 808: Is this missing some kind of comparison maybe? E.g., `driver.symbol_map[blocking.size].type.module.lower() == 'block_dim_mod'`?

transformations/transformations/single_column_coalesced.py
```diff
@@ -116,8 +116,8 @@ def get_integer_variable(cls, routine, name):
         name : string
             Name of the variable to find the in the routine.
         """
-        if name in routine.variable_map:
-            v_index = routine.variable_map[name]
+        if name in routine.symbol_map:
+            v_index = routine.symbol_map[name]
         else:
```
reuterbal, line 121: Since constructing these maps is expensive, could we maybe do it like this?
```suggestion
        v_index = routine.symbol_map.get(name)
        if v_index is None:
```

reuterbal (changes requested):
Thanks, nice find! A question on the test and a suggestion how to maybe save a few cycles in the transformation.

reuterbal (approved):
Many thanks, I think this is uncontroversial and good to go!

mlange05:
Confirmed clean with ec-physics.

PR #280, opened 2024-04-09
[F2C] refactoring `cgen` and *c-like* codegen

MichaelSt98:
and introducing *c-like* codegen (`cppgen', 'cudagen`)


reuterbal:
This PR should be made redundant by #328, right?

MichaelSt98:
> This PR should be made redundant by #328, right?

That is correct, yes!

PR #281, opened 2024-04-10
Fixing nested associate scope-parentage tracking after inlining

mlange05:
A curious corner case when inlining procedures that would result in nested associates, as the newly created parentage cannot be inferred via the usual pin-hole insertion approach. So, instead we now ensure that we update the parent in `AttachScopes` and call this after the call-inlining is finalised. A test is provided that illustrates the problem.


reuterbal (approved):
Thanks, looks like a clean-enough approach to fix this and valuable addition to the rescoping functionality in general.

PR #282, opened 2024-04-10
Revert "DEPENDENCY TRAFO: statement functions included via c-style imports preserved" (#251)

reuterbal:
This reverts commit 86c2f97fa7bd39b833f69a9b535fcda2b314b494 as it was found to break ec-phys regression due to the intfb assumption.

We will need to find a different solution to resolve this naming clash.


mlange05 (approved):
Yes, I can confirm this fixes ec-physics regression. Apologies @awnawab for missing this earlier! GTG from me.

PR #283, opened 2024-04-11
Preserve import statement functions - take II

awnawab:
A second take on preserving statement function imports via c-style headers. This time, rather than only renaming imports with `intfb` in their name, we now skip all imports with `func` in their name. I do realise that this introduces another IFS specific naming convention to the `DependencyTransformation`, and this probably needs generalising.


loki/transform/dependency_transform.py
```diff
@@ -300,7 +300,7 @@ def rename_imports(self, source, imports, targets=None):
         for im in imports:
             if im.c_import:
                 target_symbol = im.module.split('.')[0].lower()
-                if targets and target_symbol.lower() in targets:
+                if targets and target_symbol.lower() in targets and not 'func' in im.module.lower():
```
reuterbal, line 303: Sorry for nitpicking further but just realized that this may fail in pretty trivial cases, e.g., `#include "myfunc.intfb.h"` will now not be processed... 😅 

Should we apply the IFS-specific knowledge a bit more explicitly, e.g. like this?
```py
target_symbol, *suffices = im.module.lower().split('.', maxsplit=1)
if targets and target_symbol in targets and 'func.h' not in suffices:
```
Note that this works even if no `.` is in the module name at all, returning `suffices` as an empty list.
That way this becomes a very loose restriction specific to the problem of statement functions and in-line with the coding standards.

reuterbal (changes requested):
Thanks, as discussed offline this seems like a justifiable variant of baking in IFS-specific knowledge but minimal impact. The blanket check for `func` in the import name might create problems, though, so I've suggested a slightly different approach.

loki/transform/dependency_transform.py
awnawab: Ah of course, well spotted and apologies for the clumsy oversight 😅

reuterbal (approved):
Thanks for taking care of the requested changes and bonus points for capturing this edge case in the tests :)

mlange05 (approved):
Looks great and is confirmed clean with ec-physics regression. GTG :shipit:

PR #284, opened 2024-04-11
MaskedTransformer: Fix in-place rebuilding of scoped nodes

mlange05:
The offending line was accidentally packing the `associations` tuple and `body` tuple of `Associate` object together on rebuild, which inserted rogue symbols into the body of the associate when it is rebuilt in-place. This happens, for example, during pragma-region attachment.


reuterbal (approved):
Thanks! Very subtle bug, nice find!

PR #285, opened 2024-04-11
Avoid variable_map in TypedSymbol.get_derived_type_member and verify type information is derived correctly

reuterbal:
I played around a little with this because I wanted to see if we can get away without using the symbol's `variable_map`, which can be costly for large and nested derived types. For this, I've also added an additional test that validates that type information is derived correctly.


awnawab (approved):
So the type information is looked in the `Variable` factory method via the parent? Very clever! many thanks for this!

PR #286, opened 2024-04-12
Improve performance of pragma-region attach/detach by using transformers

mlange05:
A subtle but significant performance bug meant that attaching/detaching pragma-regions was becoming ver expensive for large code regions, eg. when finding marked regions to remove (PR #276). This PR fixes this by re-implementing the matching and resolution of `PragmaRegion` objects with custom in-place transformers that perform the matching at tuple-level.

One important consideration here is that we now only match pragma-pairs at tuple level, and thus have an implicit check for misplaced `!$loki end <marker>` pragmas, that will now trigger warnings. We also need to be very careful about tuple-insertion when resolving the regions objects during detach, as we do not want nested tuples in our IR tree, but cannot use `flatten` on tuples generically. This is marked via comments in the respective places.

And finally, some anecdotal speed-up evidence using an experimental ec-physics inlined control flow routine:
Before
```
(loki_env) $ loki-ecphys-gen.py inline --source ./input  --build .
[Loki::Sourcefile] Constructed from input/ec_phys_drv.F90 in 0.89s
[Loki::Sourcefile] Constructed from input/ec_phys.F90 in 1.37s
[Loki::Sourcefile] Constructed from input/callpar.F90 in 2.56s
[Loki] Inlined EC_PHYS in 2.78s
[Loki] Inlined CALLPAR in 4.76s
[Loki] Remove marked regions in 136.73s   <= this uses `with pragma_regions_attached`
```

after:
```
(loki_env) $ loki-ecphys-gen.py inline --source ./input  --build .
[Loki::Sourcefile] Constructed from input/ec_phys_drv.F90 in 0.88s
[Loki::Sourcefile] Constructed from input/ec_phys.F90 in 1.48s
[Loki::Sourcefile] Constructed from input/callpar.F90 in 2.53s
[Loki::EC-Physics] Inlined EC_PHYS in 2.78s
[Loki::EC-Physics] Inlined CALLPAR in 4.90s
[Loki::EC-Physics] Remove marked regions in 2.62s
```

### DIFF
--- a/loki/pragma_utils.py
+++ b/loki/pragma_utils.py
@@ -8,19 +8,21 @@
 import re
 from collections import defaultdict
 from contextlib import contextmanager
+from codetiming import Timer
 
 from loki.expression import symbols as sym
 from loki.ir import (
-    VariableDeclaration, Pragma, PragmaRegion,
-    FindNodes, Visitor, Transformer, MaskedTransformer
+    VariableDeclaration, Pragma, PragmaRegion, FindNodes, Visitor,
+    Transformer
 )
-from loki.tools.util import as_tuple, flatten
+from loki.tools.util import as_tuple
 from loki.types import BasicType
+from loki.logging import debug, warning
 
 
 __all__ = [
     'is_loki_pragma', 'get_pragma_parameters', 'process_dimension_pragmas',
-    'attach_pragmas', 'detach_pragmas', 'extract_pragma_region',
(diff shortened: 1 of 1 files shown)


reuterbal (approved):
Many thanks! Another long-standing utility done right 😄 
Implementation looks good to me and the speed-up is extremely impressive!

PR #287, opened 2024-04-12
Reorganising test directories

mlange05:
The next step in the repository spring clean (issue #253 ) is to move the test battery inside the actual loki package and reorganise them to move the respective test tests into the appropriate sub-packages (eg. `loki.batch.tests` or `loki.transform.tests`. 

This PR does this for all obvious cases, to get the bulk out of the way, leaving the more subtle cases for later and a potentially more detailed refactoring. Importantly, I have resolved the `tests/conftest.py` and moved all the utilities into the respective sub-packages, to ensure imports for multi-sub-directory testing works.
 As a result, we can now do cute things like this:
```

(loki_env) $ py.test --pyargs loki.batch
======================================================================================================== test session starts ========================================================================================================
platform linux -- Python 3.8.8, pytest-7.3.1, pluggy-1.0.0
rootdir: /lus/h2resw01/hpcperm/naml/loki
configfile: pyproject.toml
plugins: anyio-3.6.2, cov-4.0.0
collected 162 items                                                                                                                                                                                                                 

loki/batch/tests/test_batch.py ..................................................................                                                                                                                             [ 40%]
loki/batch/tests/test_scheduler.py ............................................x.x.................................................                                                                                           [100%]

================================================================================================== 160 passed, 2 xfailed in 51.56s ==================================================================================================

```

This PR is very invasive, but most of the changes are mechanical file moves and the necessary import changes. I used this opportunity to do a general import clean-up across the test battery, trying to enforce a few common conventions and doing imports mainly from the respective sub-packages (eg. `from loki.expression import symbols as sym` and `from loki.ir import nodes as ir`. This causes the second source of a large amount of generic line changes. 

The third set of source changes pertains to the location of the (sometimes shared) raw source files in the test directory. Where necessary, I introduced a fixture `testdir` that points to the root test package `loki.tests`. Eventually we can selectively move some of these into the respective sub-package test directories, but that is beyond this PR.

Importantly, I've also not touched the `transformations` sub-package, as the consolidation with `loki.transform` will be more subtle in many way. Restructuring and re-integrating that is also left to a follow-on PR.

In a little more detail:
* Move `pragma_utils` into `loki.ir`
* Moved all tests into `loki.tests`
* Removed `conftests.py` and moved `available_frontends` into `loki.frontend`, the JIT and test-cleaning utilities into `loki.build.jit` and the remainder into `loki.tools.util`
* Moved respective tests into subpackages for `loki.ir`, `loki.expression`, `loki.batch`, `loki.frontend`, `loki.backend`, `loki.build`, `loki.analysis`, `loki.tools` and, of course, `loki.transform`
* Introduced `testdir` fixture to point to the root testdir for external source file paths where needed
* Refactored import lists across tests touched (apologies, I might have I missed a few) to use a roughly common import convention


reuterbal (changes requested):
Thanks, this is excellent!

Given the guarantee for this to introduce conflicts I would like to bring this in asap. I left a remark about one of the utilities coming from conftest, otherwise nothing that needs changing from my side.

mlange05:
@reuterbal Ok, I pushed the changes as a single top-level commit and then rebased to remove the conflicts. :crossed_fingers:

reuterbal (approved):
Thanks, looks good to me now!

PR #288, opened 2024-04-15
`DataflowAnalysis` bugfix: preserve body nesting in `visit_MaskedStatement`

awnawab:
The `_visit_body` method in the `DataflowAnalysis` flattens the bodies, which can sometimes lead to invalid IR. This PR provides a fix.

### DIFF
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -189,8 +189,14 @@ class DataflowAnalysisAttacher(Transformer):
     def visit_MaskedStatement(self, o, **kwargs):
         live = kwargs.pop('live_symbols', set())
         conditions = self._symbols_from_expr(o.conditions)
-        body, defines, uses = self._visit_body(o.bodies, live=live, uses=conditions, **kwargs)
-        body = tuple(as_tuple(b,) for b in body)
+
+        body = ()
+        defines = set()
+        uses = set(conditions)
+        for b in o.bodies:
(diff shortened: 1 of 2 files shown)


reuterbal (changes requested):
Thanks, nice find!

While the fix produces the correct result, it is only working around the original bug. In `visit_MaskedStatement`, we are calling the routine with multiple bodies instead of separately per body, as in the other visitor methods:
https://github.com/ecmwf-ifs/loki/blob/88d865682b6a65342544fcb840257054bf3c69fe/loki/analyse/analyse_dataflow.py#L196

Would you mind trying if 
- a simple loop over the `bodies` property in `visit_MaskedStatement`
- and only removing the `flatten` in `_visit_body`
would yield the same result?

awnawab:
Thanks for the feedback 😄  I was trying to make `_visit_body` respect the nesting of tuples regardless of what any `visit_Node` method might do, but you are right, it is better to fix `visit_MaskedStatement` rather than let `_visit_body` mop-up any potential mistakes.

tests/test_analyse_dataflow.py
```diff
@@ -496,6 +496,7 @@ def test_analyse_maskedstatement(frontend):
 
   where (mask(:) < -5)
     vec1(:) = -5.0
```
reuterbal, line 498: If we removed this line, would `vec1` then also appear in the uses set?

reuterbal (approved):
Thanks, looks good to me now!

It needs another rebase, unfortunately, to resolve conflicts before we can merge.

PR #289, opened 2024-04-15
RemoveCodeTransformation: make call_names and intrinsic_names configurable

awnawab:
This PR makes the `call_names` and `intrinsic_names` constructor args in the `RemoveCodeTransformation` configurable.

### DIFF
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -207,7 +207,8 @@ def convert(
     if not remove_code_trafo:
         remove_code_trafo = RemoveCodeTransformation(
             remove_marked_regions=True, remove_dead_code=False,
-            call_names=('ABOR1', 'DR_HOOK'), intrinsic_names=('WRITE(NULOUT',)
+            call_names=scheduler.config.default.get('call_names', None) or ('ABOR1', 'DR_HOOK'),
+            intrinsic_names=scheduler.config.default.get('intrinsic_names', None) or ('WRITE(NULOUT',)
         )
     scheduler.process(transformation=remove_code_trafo)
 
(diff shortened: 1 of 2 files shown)


scripts/loki_transform.py
```diff
@@ -207,7 +207,8 @@ def convert(
     if not remove_code_trafo:
         remove_code_trafo = RemoveCodeTransformation(
             remove_marked_regions=True, remove_dead_code=False,
-            call_names=('ABOR1', 'DR_HOOK'), intrinsic_names=('WRITE(NULOUT',)
+            call_names=scheduler.config.default.get('call_names', None) or ('ABOR1', 'DR_HOOK'),
```
mlange05, line 210: @awnawab I think you can just configure the transformation itself with all of it's constructor arguments in the config file as something like (untested):
```
transformations.RemoveCodeTransformation
  module = "loki.transform"
transformations.RemoveCodeTransformation.options
  remove_marked_regions = true
  remove_dead_code = false
  call_names =['dr_hook', 'abor1']
  intrinsic_names = ['write(nulout']
```
This way, no custom assumptions or behaviour are added to the `convert` step, which is current direction of travel (removing all the custom logic). Ultimately the plan was to remove all of this and configure the entire pipeline in config, but I've no full-scale example of this yet. Does this make sense?

reuterbal: Agreed - transformation-specific configuration should not be added to `defaults`

awnawab: Aah, I had completely missed the addition of this functionality! Thanks for pointing this out, of course the above suggestion works and I'll close this PR 👍

PR #290, opened 2024-04-16
[Bugfix] available_frontends: Import pytest locally to make dependency optional

reuterbal:
This is a hotfix following merge of #287


mlange05 (approved):
Oops. Thanks for fixing. Yes, looks good to me. GTG :shipit:

PR #291, opened 2024-04-16
Transformation to hoist temporaries in kernel language transpilation

MichaelSt98:
hoisting arrays via pragma offload for low-level GPU implementations (no block index, which makes it different to `SCCHoistTemporaryArraysTransformation`)

### DIFF
--- a/transformations/transformations/scc_cuf.py
+++ b/transformations/transformations/scc_cuf.py
@@ -78,8 +78,7 @@ class HoistTemporaryArraysPragmaOffloadTransformation(HoistVariablesTransformati
         var: :any:`Variable`
             The variable to be declared
         """
-        for var in variables:
-            routine.variables += tuple([var.clone(scope=routine)])
+        routine.variables += tuple(var.clone(scope=routine) for var in variables)
 
         vnames = ', '.join(v.name for v in variables)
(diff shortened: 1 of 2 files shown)


transformations/transformations/scc_cuf.py
```diff
@@ -60,6 +61,35 @@ def driver_variable_declaration(self, routine, variables):
...
+
+        Parameters
+        ----------
+        routine: :any:`Subroutine`
+            The subroutine to add the variable declaration
+        var: :any:`Variable`
+            The variable to be declared
+        """
+        for var in variables:
+            routine.variables += tuple([var.clone(scope=routine)])
```
reuterbal, line 82: Note that editing the variables list multiple times can be very costly, because it will compile the list of variables from scratch from the declarations on every invocation, and then check and insert declarations. I think it would be much faster to do it like this:
```suggestion
        routine.variables += tuple(var.clone(scope=routine) for var in variables)
```

reuterbal (changes requested):
Thanks.

I think the list of variables to be added should be compiled in full before adding it to the routine, instead of adding them 1-by-1, otherwise good to go from me.

reuterbal (approved):

PR #293, opened 2024-04-19
fix scoping for global var hoisting

MichaelSt98:
fix scoping for `GlobalVarHoistTransformation'`


reuterbal (approved):
Thanks, nice find!

PR #294, opened 2024-04-19
Consistent, environment-configurable use of Compiler class in JIT compilation

reuterbal:
JIT compilation was hard coded to `gfortran` and not using the `Compiler` class in the `compile_and_load` method used throughout the test base. This prevented testing with other gfortran versions or different compilers.

The first change in this PR is to make Loki use consistently the `Compiler` class for JIT compilation. This unlocks the use of other compilers when executing tests.

The second change is to inspect the environment for common variables setting the compiler (`CC`, `FC`, `F90`) and choose an appropriate compiler. This consists of two steps:
1. First, we try to identify the compiler "family", for which now "GNU" and "Nvidia" exist (I used the opportuinity to delete the ancient "ESCAPE" compiler definition). This is done by matching the value in the respective env variable against a set of common names, allowing also for absolute/relative paths (e.g. to test user-supplied compilers). A Fortran compiler specification takes precedence over a C-compiler specification in this case.
2. Second, the executable name and compiler flags are overwritten with any values supplied in the environment.

As a side-effect, this enables tests that require compilation also on MacOS. With a user-installed GNU compiler suite (e.g. via homebrew) and setting `CC=gcc-13 FC=gfortran-13`, the tests use now correctly the user-supplied compilers instead of the system-default "gcc-relabeled" clang compiler.

### DIFF
--- a/loki/build/compiler.py
+++ b/loki/build/compiler.py
@@ -45,7 +45,7 @@ def clean(filename, pattern=None):
             delete(f)
 
 
-def compile_and_load(filename, cwd=None, use_f90wrap=True, f90wrap_kind_map=None):  # pylint: disable=unused-argument
+def compile_and_load(filename, cwd=None, f90wrap_kind_map=None, compiler=None):
     """
     Just-in-time compile Fortran source code and load the respective
     module or class.
@@ -59,12 +59,12 @@ def compile_and_load(filename, cwd=None, use_f90wrap=True, f90wrap_kind_map=None
         The source file to be compiled.
     cwd : str, optional
         Working directory to use for calls to compiler.
-    use_f90wrap : bool, optional
-        Flag to trigger the ``f90wrap`` compiler required
-        if the source code includes module or derived types.
     f90wrap_kind_map : str, optional
         Path to ``f90wrap`` KIND_MAP file, containing a Python dictionary
         in f2py_f2cmap format.
+    compiler : :any:`Compiler`, optional
+        Use the specified compiler to compile the Fortran source code. Defaults
+        to :any:`_default_compiler`
     """
     info(f'Compiling: {filename}')
     filepath = Path(filename)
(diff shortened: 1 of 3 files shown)


mlange05 (commented):
Looks good to me, although I don't think the NVHPC compiler description is functional.  From a quick test it looks like the problem is in f90wrap, though, and I'm not sure we can override the flags that sets. That is, unless I've missed something?

The only question I have is if we then want to formally "support" this, or leave adding the `NvidiaCompiler` class until we know whether this actually works? I'd be happy with this either way. For now, I'll leave this as "comment", but happy otherwise.

Also, I don't have a Mac-local setup, so didn't test this, but assuming @reuterbal has checked, I think this is ok.

reuterbal:
Thanks, and apologies for not actually checking this properly...

I had a go at this and there is indeed a simple hook in f2py to specify the Fortran compiler. I've pushed two commits to fix this. We're now consistently using the corresponding methods from the `Compiler` class to invoke f2py and f90wrap, and provide the compiler selector argument. The only caveat seems to be, that f2py does not seem to work with any other C compiler than gcc...

With this, loading the `nvidia` module on Atos HPCF (which sets `FC`, `F90` etc. accordingly) and then running the tests with 
```
CC=gcc pytest loki
```
works correctly in principle (i.e., tests that use JIT compilation, and even some transpilation tests, validate). There's a fair number of tests failing with compiler errors, but that seems to be the Nvidia compiler being genuinely unhappy with some of the files in the tests. 

So, I'd be comfortable leaving the `NvidiaCompiler` class in and claim that the build module supports its use, but not claim full compatibility of the test base with nvfortran. Instead, I'd suggest we defer resolving that to a separate PR.

reuterbal:
After a slightly longer glimpse, it seems most failing tests are due to the use of Fortran 2008's iso_fortran_env and particularly `real64` from that. I'm finding only conflicting information but it looks like nvfortran's support of that is limited.

mlange05 (approved):
Ok, many, many thanks @reuterbal that is a fantastic feature, and the follow-up is very good to know! Happy to leave the fixing of non-NV compatible tests for later, so GTG from my side for this one! :shipit:

PR #295, opened 2024-04-22
Derived-type inheritance

awnawab:
This PR adds the ability to propagate all the non-overridden symbols from a parent to a child type. It should be noted that per the Fortran standard, only type-bound procedures can be shadowed by children and data/procedure members cannot, but I have kept the implementation agnostic to the type of symbol being inherited.


awnawab:
Many thanks for the feedback @reuterbal and for spotting the potentially empty declarations 🙏

reuterbal (approved):
Many thanks, all good from my side!

PR #296, opened 2024-04-23
Repo reorganisation: Moving transformations

mlange05:
The next step in our spring clean (issue #253) is to move all the internally support transformations into the core package, alongside the associated tests. This them also merges the `loki.transform` general utilities into the more custom `loki.transformations` packages. 

I've tried to restrain myself to only move files and fix-up import headers, with a few small exceptions listed in the details below:
* `Transformation` and `Pipeline` class have been moved into `loki.batch` as they are the interface to the scheduler
* `transformations` has been moved into `loki.transformations`
  * the associated test have been moved in the same fashion
  * the additional conftest.py has been resolved and the runner utilities for CLOUDSC and ecWAM added to `loki.tools`
* The subpackages `loki.transformations.single_column`, `loki.transformations.trnaspile` and `loki.transformations.build_system` have been separated
  * `DependencyTransformation` and `ModuleWrapTransformation` have been mode to separate files inside `trnasformations.build_system`
  * CI And packaging has been changes as needed (please double-check this)
  * `loki.transform` has been moved into `loki.transformations`, often removing the `transform_xxx` prefix

And finally, apologies to @MichaelSt98 , as this one might conflict quite a bit with his current stuff... :grimacing:


MichaelSt98:
Thanks, really nice! I'm happy with it and haven't discovered anything problematic. 
Nevertheless some comments:

I think in order to be consistent:

* `transformations/build_system/tests/test_transform_dependency.py` --> `transformations/build_system/test_dependency.py`
* `transformations/tests/test_transform_derived_types.py` --> `transformations/tests/test_derived_types.py`

Why not?

*  `transformations/transform_loop.py` --> `transformations/loop.py`
* `transformations/transform_region.py` --> `transfrormations/region.py`

_________

Me thinking out loud:

* introduce subdirectories within transformations **?**
    * `transformations/temporaries/` (or something like that) for *hoist-variables*, *pool_allocator*, ...
    * `transformations/control_flow/` (or something like that) for *loop*, *inline*, *region*, ...
    * `transformations/utilities/` for *utilities*, *remove_code*, *drhook*, ...
* it makes sense for `transformation` and `pipeline` to be within `batch` as they are interfacing the `scheduler`, however, I somehow would have expected them to be in `transformations/`

I guess that further repo reorganisation like "separate operations (inlinecall, subscript, sum, ...) from the symbols/literals" will be done in a future/separate PR?!

mlange05:
@MichaelSt98 Many thanks for the review - to answer your questions:
1. 
```
transformations/build_system/tests/test_transform_dependency.py --> transformations/build_system/test_dependency.py
transformations/tests/test_transform_derived_types.py --> transformations/tests/test_derived_types.py
```
Yes, absolutely, that was an oversight! Will fix in the needed rebase!

2. 
```
transformations/transform_loop.py --> transformations/loop.py
transformations/transform_region.py --> transfrormations/region.py
```
Mostly because I actually thought "transform" to be a good high-level description for "fusion/fission/interchange" and "hoist/extract". Open for suggestions though, as the grouping by target "node type" seemed not scalable to me.

3. 
Agree in general on the proposed further sub-division or the `transformations` package; this will potentially also require some re-jigging of the stuff insied some of those transformation (eg. lots in SCCBase could go into `utility`) and I simply wanted to keep this PR manageable.

With regards to `Transformation`/`Pipeline` moving, this simply creates a clean import chain, as everything in `loki.transformations` should be "do something to code", and thus depend on the `ir`/`expression` and `batch` sub-packages. Hope this makes sense. :wink:

loki/tools/files.py
```diff
@@ -242,3 +244,39 @@ def load_module(module, path=None):
         # If module caching interferes, try again with clean caches
         invalidate_caches()
         return import_module(module)
+
+
+def write_env_launch_script(here, binary, args):
+    # Write a script to source env.sh and launch the binary
```
reuterbal, line 250: It took me a while to figure out why we have these because they didn't show up with test coverage. So, I would suggest adding short docstrings:
```suggestion
def write_env_launch_script(here, binary, args):
    """
    Utility method that is used for regression tests that require
    activating an environment file before running :data:`binary`.
    
    This writes a simple script of the form
    
    .. code-block::
       source env.sh
       bin/<binary> <args>
       exit $?
       
    Parameters
    ----------
    here : pathlib.Path or str
        The directory in which the script is created
    binary : str
        The name of the binary
    args : list of str
        List of arguments to pass to the binary
        
    Returns
    -------
    pathlib.Path
        The path to the created script file
    """
```

loki/tools/files.py
```diff
@@ -242,3 +244,39 @@ def load_module(module, path=None):
...
+source env.sh >&2
+bin/{binary} {' '.join(args)}
+exit $?
+    """.strip())
+    script.chmod(0o750)
+
+    return script
+
+
+def local_loki_setup(here):
```
reuterbal, line 264:
```suggestion
def local_loki_setup(here):
    """
    Utility method that is used to determine paths for injecting the currently running
    source code of Loki into an [ecbundle](https://github.com/ecmwf/ecbundle)-based
    worktree

    This is used for regression tests to facilitate the use of a local Loki source copy
    in the build. In particular, any existing Loki source copy in the bundle worktree is
    moved to a backup location.
    
    .. warning::
       If a backup copy exists already at the backup location, this is removed before
       moving the existing Loki copy to the backup location.
       
    Note that injecting the currently running Loki installation only works if it 
    has been installed in editable mode.  However, this utility also does not
    take care of the actual injection of the currently running installation, therefore
    making this also useful if the purpose is to trigger a Loki download via the
    bundle create mechanism.
    
    The companion utility :any:`local_loki_cleanup` can be used to revert these changes.
    
    Parameters
    ----------
    here : pathlib.Path
        The root path of the bundle worktree.
        
    Returns
    -------
    tuple of (str, pathlib.Path, pathlib.Path)
        The absolute path to the base directory of the currently running Loki
        installation, the ``target`` path where Loki needs to be injected in the
        bundle directory, and the ``backup`` path where an existing Loki
        copy in the bundle has been moved to.
    """
```

loki/tools/files.py
```diff
@@ -242,3 +244,39 @@ def load_module(module, path=None):
...
+    # Do not overwrite any existing Loki copy
+    if target.exists():
+        if backup.exists():
+            shutil.rmtree(backup)
+        shutil.move(target, backup)
+
+    return str(lokidir.resolve()), target, backup
+
+
+def local_loki_cleanup(target, backup):
```
reuterbal, line 278:
```suggestion
def local_loki_cleanup(target, backup):
    Companion utility to :any:`local_loki_setup` to revert the changes
    
    This removes a symlink at :data:`target`, if it exists, and moves the
    :data:`backup` path in its original location.
    
    Parameters
    ---------
    target : pathlib.Path
        The target injection path as returned by :any:`local_loki_setup`
    backup : pathlib.Path
        The backup path as created by :any:`local_loki_setup`
```

reuterbal (approved):
Sorry for taking so long to review this and many thanks, this looks fantastic.

I agree with all what's been said so far. Regarding the renaming to just `loop` & co I share your assessment that this might not scale well and the `transform` adds information despite being somewhat redundant at first look - since it is the commonly used terminology for the kind of changes applied by the utilities in there.

I've added some comments where I spotted missing docstrings or possible clean-up opportunities, but overall this looks very good.

The documentation scripts need to be updated, too, but since we have already dangling pointers in there and Sphinx steps over this gracefully, we may as well do a sweep in a separate PR to rectify this.

loki/tools/files.py
mlange05: Thanks for these! Added as suggested.

mlange05:
Ok, apologies, this took a bit longer than anticipated. @reuterbal I think I've addressed all comments - many thanks for the suggestions! 

I've also tried to address the naming inconsistencies pointed out by @MichaelSt98, but those generated some name clashes in the test base (`test_derived_type` existed twice). So instead I've opted to rename `loki.transformations.derived_type` to `loki.transformations.transform_derived_types` to match the test. I hope this works now, and I think we'll get to tidy this up a little further in a follow-on PR.

install
```diff
@@ -512,7 +511,7 @@ print_step "Installation finished"
 echo
 echo "Activate the Loki environment with"
 echo
-echo "    source loki-activate"
+echo "    source loki-activate"xf
```
reuterbal, line 514: Something's gone wrong here
```suggestion
echo "    source loki-activate"
```

loki/tools/files.py
```diff
@@ -242,3 +244,116 @@ def load_module(module, path=None):
...
+    script.chmod(0o750)
+
+    return script
+
+
+def local_loki_setup(here):
+    """
+    Utility method that is used to determine paths for injecting the
+    currently running source code of Loki into an
+    [ecbundle](https://github.com/ecmwf/ecbundle)-based worktree This
```
reuterbal, line 293: Sorry, got caught out by the annoying RST vs. MD syntax differences for link formatting...

Also, empty line after the first sentence improves formatting in the HTML output.
```suggestion
    `ecbundle <https://github.com/ecmwf/ecbundle>`_-based worktree.
    
     This
```

reuterbal (approved):
Thanks for this. I found another oopsie and my docstring suggestion was faulty (sorry!), but otherwise good to go!

PR #302, opened 2024-04-26
Fix installation without git checkout

reuterbal:
We are using setuptools_scm to determine the version number from the latest tag. This prevents installation if there is no .git worktree information, e.g., when installing from a tarball. The recommended workaround is to provide the variable `SETUPTOOLS_SCM_PRETEND_VERSION`, which this PR adds to the `install` script as well as the CMake.

Apparently, I forgot to update the `VERSION` file for a while, so that's fixed with this as well...


awnawab:
Many many thanks for looking into this and implementing this so quickly 🙏  Whilst this does allow the installation to complete successfully, I get a `ModulenotFoundError` when trying to import the frontend:
```
>>> from loki import frontend
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/naan/Documents/ifs-source/loki/loki/__init__.py", line 13, in <module>
    from loki.frontend import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/naan/Documents/ifs-source/loki/loki/frontend/__init__.py", line 8, in <module>
    from loki.frontend.preprocessing import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/naan/Documents/ifs-source/loki/loki/frontend/preprocessing.py", line 15, in <module>
    import pcpp
ModuleNotFoundError: No module named 'pcpp'
```

awnawab:
I've done a bit more digging. When doing a cmake install without the `.git` directory, the contents of `build/loki_env/lib/python3.10/site-packages/loki` are as follows:
```
config.py  dimension.py  __init__.py  logging.py  module.py  program_unit.py  __pycache__  scope.py  sourcefile.py  subroutine.py  types.py
```

When installing with the `.git` directory, we get the following instead:
```
analyse  batch	config.py     expression  __init__.py  lint module.py	    __pycache__  sourcefile.py	tests  transform backend  build	dimension.py  frontend	  ir logging.py  program_unit.py  scope.py	 subroutine.py	tools  types.py
```

The former is missing all the subdirectories that one would expect in the`loki` dir e.g. `frontend, backend`.

This can however easily be worked around by installing loki as an editable package, and I think for the purposes of achieving ecmwf-actions/build-package that would be enough and this bug can be mopped up in a subsequent pull-request.

awnawab:
As loki is not yet compatible with python3.12, for macos cmake installations we often want to specify `Python3_ROOT_DIR` so that we can use the correct python rather than the system installation or even the homebrew version which is now at 3.12 because of other projects that depend on it. As such, the `find_python_venv` function can fail to find the newly created virtual env and keeps finding the python installation pointed to by `Python3_ROOT_DIR`. This can be remedied by adding `set( Python3_ROOT_DIR "${VENV_PATH}" )` just after `unset( Python3_EXECUTABLE )` in `find_python_venv`. Please let me know if you agree and think that's not too hacky.

awnawab (approved):
Many thanks again for this 🙏 After *a lot* of trial and error, I have a working macos + loki configuration for the ecwam CI suite. Can you please merge this when ready and tag a hotfix release?

PR #305, opened 2024-04-30
Fetch tags in Github workflows

reuterbal:
Small bugfix PR: The GH workflows did not fetch tags in the clone, therefore the Loki version number was always wrong (including in the published docs).


mlange05 (approved):
Looks good. :shipit:

PR #308, opened 2024-05-02
Fix: import of private symbols affects the type inference

quepas:
As described in #309, imports of private symbols destroys the type inference. This pull request propose a simple check of symbols access specificators to filter out the private symbols.

TODO:
- [x] Support for other frontends
- [x] Unit test

### DIFF
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -465,6 +465,9 @@ class FParser2IR(GenericVisitor):
             if module is not None:
                 # Import symbol attributes from module, if available
                 for k, v in module.symbol_attrs.items():
+                    # Don't import private module symbols
+                    if v.private == True or (module.default_access_spec == "private" and v.public is None):
+                        continue
                     if k in rename_list:
                         local_name = rename_list[k].name
                         scope.symbol_attrs[local_name] = v.clone(imported=True, module=module, use_name=k)
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -466,7 +466,7 @@ class FParser2IR(GenericVisitor):
                 # Import symbol attributes from module, if available
                 for k, v in module.symbol_attrs.items():
                     # Don't import private module symbols
-                    if v.private == True or (module.default_access_spec == "private" and v.public is None):
+                    if v.private is True or (module.default_access_spec == "private" and v.public is None):
                         continue
                     if k in rename_list:
                         local_name = rename_list[k].name
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -466,7 +466,7 @@ class FParser2IR(GenericVisitor):
                 # Import symbol attributes from module, if available
                 for k, v in module.symbol_attrs.items():
                     # Don't import private module symbols
-                    if v.private is True or (module.default_access_spec == "private" and v.public is None):
+                    if v.private or (module.default_access_spec == "private" and not v.public):
                         continue
                     if k in rename_list:
                         local_name = rename_list[k].name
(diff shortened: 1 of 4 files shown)


loki/frontend/tests/test_frontends.py
```diff
@@ -2031,3 +2031,46 @@ def to_str(expr):
...
+    mod_public = Module.from_source(code_mod_public, frontend=frontend)
+    mod_main = Module.from_source(code_mod_main, frontend=frontend, definitions=[mod_private, mod_public])
+    var = mod_main.subroutines[0].body.body[0].rhs
+    # Check if this is really our symbol
+    assert var.name == "var"
+    assert var.scope == mod_main
+    # Check if the symbol is imported
+    assert var.type.imported is True
+    # Check if the symbol comes from the mod_public module
+    assert var.type.module == mod_public
```
reuterbal, line 2076: It's better to use `var.type.module is mod_public` here to ensure it really links to the correct module and not just to one that compares equal.

loki/frontend/tests/test_frontends.py
```diff
@@ -2031,3 +2031,46 @@ def to_str(expr):
...
+end module mod_main
+    """
+
+    mod_private = Module.from_source(code_mod_private, frontend=frontend)
+    mod_public = Module.from_source(code_mod_public, frontend=frontend)
+    mod_main = Module.from_source(code_mod_main, frontend=frontend, definitions=[mod_private, mod_public])
+    var = mod_main.subroutines[0].body.body[0].rhs
+    # Check if this is really our symbol
+    assert var.name == "var"
+    assert var.scope == mod_main
```
reuterbal, line 2072: Similarly, better to use `var.scope is mod_main` here

loki/frontend/omni.py
```diff
@@ -215,6 +215,9 @@ def visit_FuseDecl(self, o, **kwargs):
         if module is not None:
             # Import symbol attributes from module, if available
             for k, v in module.symbol_attrs.items():
+                # Don't import private module symbols
+                if v.private is True or (module.default_access_spec == "private" and v.public is None):
```
reuterbal, line 219: A suggestion to make the expression a little shorter and more pythonic:
```suggestion
                if v.private or (module.default_access_spec == "private" and not v.public):
```

loki/frontend/ofp.py
```diff
@@ -1416,6 +1416,9 @@ def visit_use(self, o, **kwargs):
             if module is not None:
                 # Import symbol attributes from module, if available
                 for k, v in module.symbol_attrs.items():
+                    # Don't import private module symbols
+                    if v.private is True or (module.default_access_spec == "private" and v.public is None):
```
reuterbal, line 1420:
```suggestion
                    if v.private or (module.default_access_spec == "private" and not v.public):
```

loki/frontend/fparser.py
```diff
@@ -465,6 +465,9 @@ def visit_Use_Stmt(self, o, **kwargs):
             if module is not None:
                 # Import symbol attributes from module, if available
                 for k, v in module.symbol_attrs.items():
+                    # Don't import private module symbols
+                    if v.private is True or (module.default_access_spec == "private" and v.public is None):
```
reuterbal, line 469:
```suggestion
                    if v.private or (module.default_access_spec == "private" and not v.public):
```

reuterbal (changes requested):
Many thanks for finding this and providing the fix, this looks fantastic!

I've just a few stylistic pointers but otherwise very happy to merge this as soon as possible.

loki/frontend/tests/test_frontends.py
quepas: Applied! :)

quepas:
> Many thanks for finding this and providing the fix, this looks fantastic!
> 
> I've just a few stylistic pointers but otherwise very happy to merge this as soon as possible.

Thank you for great suggestions! I have applied them. 

No problem, the pleasure is on our side (IPSL), as we are actively using your tool in our work.

mlange05 (approved):
Fully agree, this looks great. Many thanks @quepas

reuterbal:
ecwam regression fails due to missing permissions from an external contribution. Verified compatibility offline.

PR #311, opened 2024-05-03
Transformations: Remap declaration symbols and adjust imports when inlining

mlange05:
This PR brings two fixes to the `InlineTransformation`:
1. Additional symbol-remapping of hoisted declarations, so that renaming of array declaration sizes is adjusted appropriately. 
2. In addition to removing the inlined procedure symbol from the caller's imports, we now also hoist any needed imports from the inlined callee into the spec of the caller. Subsequently, I've renamed the `remove_imports` constructor flag to `adjust_imports` to toggle this behaviour.

As always, appropriate tests have been added or extended.


loki/transformations/inline.py
```diff
@@ -597,4 +604,27 @@ def inline_marked_subroutines(routine, allowed_aliases=None, remove_imports=True
                 )
                 # Remove import if no further symbols used, otherwise clone with new symbols
                 import_map[impt] = impt.clone(symbols=new_symbols) if new_symbols else None
+
+        # Now move any callee imports we might need over to the caller
+        new_imports = set()
+        imported_module_map = {im.module: im for im in routine.imports}
```
reuterbal, line 610: Are we case-safe here? Or should this be a `CaseInsensitiveDict` maybe, because we're subsequently comparing imports from the callee against imports from the caller, where module names might have different casing.

loki/transformations/inline.py
```diff
@@ -597,4 +604,27 @@ def inline_marked_subroutines(routine, allowed_aliases=None, remove_imports=True
...
+            for impt in callee.imports:
+
+                # Add any callee module we do not yet know
+                if impt.module not in imported_module_map:
+                    new_imports.add(impt)
+
+                # If we're importing the same module, check for missing symbols
+                if m := imported_module_map.get(impt.module):
+                    if not all(s in m.symbols for s in impt.symbols):
+                        import_map[m] = m.clone(symbols=tuple(set(m.symbols + impt.symbols)))
```
reuterbal, line 621: Is scoping correct here? `impt.symbols` should have scope `callee`, `m.symbols` should have scope `routine`.

reuterbal (approved):
Thanks, looks great. Two small questions to double-check scoping and case-insensitivity, but otherwise looks good to me!

loki/transformations/inline.py
mlange05: Very good catch! Yes, this needs fixing, will push a fix on top of rebase.

loki/transformations/inline.py
mlange05: Also good point! Added fix on top of rebase.

reuterbal:
Looks great! Merge after tests are complete

PR #313, opened 2024-05-07
IR: Fix `get_pragma_params` for multiline pragmas

mlange05:
The recent addition of `parse_expr` for reading pragma parameters does not account for multi-line pragmas with `&` linebreaks. This PR fixes this and adds a small test for this.

I've also piggy-backed a small fix to loki-transform.py, where the recent introduction of `RemoveCodeTransformation` was not told to keep `DR_HOOK` calls in the drivers via the `kernel_only` option. With these two changes, EC-physics regression should be fine again.


reuterbal (approved):
Many thanks, let's wait for CI tests to confirm but this looks great.

PR #315, opened 2024-05-09
JIT compilation updates and compatibility with f90wrap v0.2.14

reuterbal:
The latest release of f90wrap includes module names in the name of generated interface routines (introduced in https://github.com/jameskermode/f90wrap/pull/215). This lets them grow quickly beyond the Fortran length limit of 63 characters for names, and gfortran (not even 13.2) does not allow to lift that limitation.

Therefore, I have shorted the module and routine names in the test base where the character limit was exceeded.
In the process, I also introduced the use of `tempdir` fixtures more widely to ensure clean-up of test files is handled gracefully and the littering of the source directory is reduced.

### DIFF
--- a/loki/batch/tests/test_batch.py
+++ b/loki/batch/tests/test_batch.py
@@ -14,7 +14,7 @@ import networkx as nx
 import pytest
 
 from loki import (
-    Sourcefile, Subroutine, as_tuple, gettempdir, RawSource, TypeDef,
+    Sourcefile, Subroutine, as_tuple, RawSource, TypeDef,
     Scalar, ProcedureSymbol
 )
 from loki.batch import (
@@ -632,7 +632,7 @@ def test_procedure_item_with_config2(testdir, disable):
 
 
 @pytest.mark.parametrize('enable_imports', [False, True])
-def test_procedure_item_external_item(enable_imports, default_config):
+def test_procedure_item_external_item(tmp_path, enable_imports, default_config):
     """
     Test that dependencies to external module procedures are marked as external item
     """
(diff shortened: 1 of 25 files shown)


reuterbal:
This turned into quite a rabbit hole. Something in the original set of changes seems to have slightly skewed timings in the test execution, presumably due to the increased use of test-local temporary directories. The symptom was then random segfaults when garbage collection was triggered.

The subsequent investigation unearthed a few glaring holes in our JIT infrastructure for tests:

- The tests of the parametrise transformation accidentally rebuilt the same files at least twice per test, and both loaded as C extensions into the running Python process. Some unfortunate interplay between GC, address offsets and name clashes resulted in segfaults. This has been mitigated by ensuring each file is only compiled once per test.
- In the process of debugging, I applied the `tmp_path` fixture more widely to curtail the creation of in-source test files further and avoid aliasing surprises.
- This has helped to discover a new issue - Surprise! The `Obj` nodes that are used to build the dependency graph for library builds are cached. The cache index is, however, simply a "name", which defaults to the filepath stem. With our various parametrised tests, that generate variants of files with the same name in different temporary directories, we now hit aliasing issues because the full file path is not taken into account. This needs to be cleaned up properly, potentially using Scheduler infrastructure to derive these dependency graphs. For now, I introduced a cache cleaning mechanism, which is now triggered as part of the `builder` fixtures after completing a test.
- To stress-test this further, I have now replaced the `tempdir` fixture in all tests by `tmp_path`

I'm afraid, the end result is now a fairly sizeable diff. Sorry!

mlange05 (approved):
Indeed quite a handful of changes in the end... :sob: Many thanks for grinding this through, and many apologies for the JIT issues (how did that ever work? :thinking: In any case, very much appreciated and overall reduction in boilerplate is still very nice!

GTG from me! :shipit:

PR #319, opened 2024-05-28
`HoistVariablesAnalysis`: remove unused explicit interfaces after inlining

awnawab:
This small PR implements two bug fixes:
- `HoistVariablesAnalysis` now skips successors that might no longer be present (e.g. they were inlined)
- Unused explicit interfaces of inlined subroutines are now removed

### DIFF
--- a/loki/transformations/hoist_variables.py
+++ b/loki/transformations/hoist_variables.py
@@ -153,10 +153,6 @@ class HoistVariablesAnalysis(Transformation):
             if not isinstance(child, ProcedureItem):
                 continue
 
-            # skip successors that might have been inlined
-            if not call_map.get(child.local_name, None):
-                continue
-
             arg_map = dict(call_map[child.local_name].arg_iter())
             hoist_variables = []
             for var in child.trafo_data[self._key]["hoist_variables"]:
--- a/loki/transformations/hoist_variables.py
+++ b/loki/transformations/hoist_variables.py
@@ -152,6 +152,11 @@ class HoistVariablesAnalysis(Transformation):
         for child in successors:
             if not isinstance(child, ProcedureItem):
                 continue
+
+            # skip successors that might have been inlined
+            if not call_map.get(child.local_name, None):
+                continue
+
             arg_map = dict(call_map[child.local_name].arg_iter())
             hoist_variables = []
             for var in child.trafo_data[self._key]["hoist_variables"]:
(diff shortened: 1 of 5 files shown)


awnawab:
Hi. Can I push a small update here?

loki/transformations/hoist_variables.py
```diff
@@ -152,6 +152,11 @@ def transform_subroutine(self, routine, **kwargs):
         for child in successors:
             if not isinstance(child, ProcedureItem):
                 continue
+
+            # skip successors that might have been inlined
+            if not call_map.get(child.local_name, None):
```
reuterbal, line 157: Shouldn't that simply be
```suggestion
            if child.local_name not in call_map:
```

loki/transformations/inline.py
```diff
@@ -606,6 +606,18 @@ def inline_marked_subroutines(routine, allowed_aliases=None, adjust_imports=True
                 # Remove import if no further symbols used, otherwise clone with new symbols
                 import_map[impt] = impt.clone(symbols=new_symbols) if new_symbols else None
 
+        # Remove explicit interfaces of inlined routines
+        for intf in FindNodes(Interface).visit(routine.ir):
```
reuterbal, line 610: Do we really need to go over `routine.ir` here? Or couldn't we simply use 
```suggestion
        for intf in routine.interfaces:
```

loki/transformations/inline.py
```diff
@@ -606,6 +606,18 @@ def inline_marked_subroutines(routine, allowed_aliases=None, adjust_imports=True
...
+            if not intf.spec:
+                _body = []
+                for s in intf.symbols:
+                    if s.name not in callees or s.name in not_inlined:
+                        _body += [s.type.dtype.procedure,]
```
reuterbal, line 615: This could be done with a list comprehension:
```suggestion
                _body = tuple(
                    s.type.dtype.procedure for s in intf.symbols
                    if s.name not in callees or s.name in not_inlined
                )
```
Also allows to directly build a tuple, avoiding the `as_tuple` further down.

reuterbal (changes requested):
Thanks, this looks really great! Three small comments/questions regarding implementation details but nothing too important really.

loki/transformations/inline.py
awnawab: As discussed offline, this can be dropped now because `InlineTransformation.creates_items = True` if `inline_marked = True`.

loki/transformations/hoist_variables.py
awnawab: As discussed offline, this can be dropped now because `InlineTransformation.creates_items = True` if `inline_marked = True`.

reuterbal (approved):
Many thanks, all good for me now!

mlange05 (approved):
Very nice! Clean and well tested, GTG from me. :shipit:

PR #320, opened 2024-05-31
`DataFlowAnalysis` bug fixes

awnawab:
This PR fixes the following bugs in the `DataFlowAnalysis`:
1. Module imports are now only marked as  `defined` if they are `ProcedureSymbol`s
2. `CallStatement` argument arrays indexed indirectly now no longer add the array representing the indirection to the `defined` set
3. `VariableDeclaration` nodes now add symbols used to define array sizes to the `uses` set

The `GlobalVariableAnalysis` now also adds symbols used in `subroutine.spec` to the `uses_symbols` set.

### DIFF
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -275,7 +275,9 @@ class DataflowAnalysisAttacher(Transformer):
 
     def visit_VariableDeclaration(self, o, **kwargs):
         defines = self._symbols_from_expr(o.symbols, condition=lambda v: v.type.initial is not None)
-        return self.visit_Node(o, defines_symbols=defines, **kwargs)
+        arrays = [a for a in o.symbols if isinstance(a, Array)]
+        uses = set(v for a in arrays for v in FindVariables().visit(a.dimensions))
+        return self.visit_Node(o, defines_symbols=defines, uses_symbols=uses, **kwargs)
 
 
 class DataflowAnalysisDetacher(Transformer):
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -10,7 +10,7 @@ Collection of dataflow analysis schema routines.
 """
 
 from contextlib import contextmanager
-from loki.expression import FindVariables, Array, FindInlineCalls
+from loki.expression import FindVariables, Array, FindInlineCalls, ProcedureSymbol, FindTypedSymbols
 from loki.tools import as_tuple, flatten
 from loki.types import BasicType
 from loki.ir import Visitor, Transformer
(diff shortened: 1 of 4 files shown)


loki/analyse/analyse_dataflow.py
```diff
@@ -269,12 +269,15 @@ def visit_Deallocation(self, o, **kwargs):
     visit_Nullify = visit_Deallocation
 
     def visit_Import(self, o, **kwargs):
-        defines = self._symbols_from_expr(o.symbols or ())
+        defines = set(s.clone(dimensions=None) for s in FindTypedSymbols().visit(o.symbols or ())
+                      if isinstance(s, ProcedureSymbol))
```
reuterbal, line 273: Let's stick to the commonly used abstraction:
```suggestion
        defines = self._symbols_from_expr(o.symbols or (), condition=lambda s: isinstance(s, ProcedureSymbol))
```

loki/analyse/analyse_dataflow.py
```diff
@@ -269,12 +269,15 @@ def visit_Deallocation(self, o, **kwargs):
...
         return self.visit_Node(o, defines_symbols=defines, **kwargs)
 
     def visit_VariableDeclaration(self, o, **kwargs):
         defines = self._symbols_from_expr(o.symbols, condition=lambda v: v.type.initial is not None)
-        return self.visit_Node(o, defines_symbols=defines, **kwargs)
+        arrays = [a for a in o.symbols if isinstance(a, Array)]
+        uses = set(v for a in arrays for v in FindVariables().visit(a.dimensions))
```
reuterbal, line 279:
```suggestion
        uses = {v for a in o.symbols if isinstance(a, Array) for v in self._symbols_from_expr(a.dimensions)}
```

loki/transformations/tests/test_data_offload.py
```diff
@@ -428,6 +433,8 @@ def test_global_variable_analysis(frontend, key, config, global_variable_analysi
         if item == 'global_var_analysis_data_mod#some_type':
             continue
         for trafo_data_key, trafo_data_value in item.trafo_data[key].items():
+            print(item)
+            print(trafo_data_value)
```
reuterbal, line 437: debug leftover?

reuterbal (changes requested):
Thanks! I would prefer to keep using the existing abstraction for extracting symbols in the dataflow analysis, see inline comments.

Regarding the second point about a behaviour change for CallStatements: I haven't been able to find where the corresponding change in the code is. Maybe I'm too hungry to understand this, but could you give me a pointer?

loki/analyse/analyse_dataflow.py
```diff
@@ -232,9 +232,9 @@ def visit_CallStatement(self, o, **kwargs):
             outvals = [val for arg, val in o.arg_iter() if str(arg.type.intent).lower() in ('inout', 'out')]
             invals = [val for arg, val in o.arg_iter() if str(arg.type.intent).lower() in ('inout', 'in')]
 
+            arrays = [v for v in FindVariables().visit(outvals) if isinstance(v, Array)]
+            dims = set(v for a in arrays for v in self._symbols_from_expr(a.dimensions))
```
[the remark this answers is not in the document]
reuterbal: Ah, noted. Thanks for the explainer!

loki/analyse/analyse_dataflow.py
awnawab: Ah I remember now. I intentionally used `FindTypedSymbols` here because a `ProcedureSymbol` is not a `Variable` and so it's not picked up by `FindVariables`.

reuterbal (approved):
Looks great, many thanks!

mlange05 (approved):
Yes, looks good to me. GTG :shipit:

PR #321, opened 2024-06-10
Add transformation generated imports to driver or after inlining

awnawab:
This PR provides the following fixes:
1. `TemporariesPoolAllocator` transformation now adds global variable imports to the driver for variables used to define array sizes.
2. `HoistArraysTransformation` adds imports used to define the sizes of hoisted variables.
3.  The addition of module imports from inlined children in the `InlineTransformation` is fixed.
4. `InlineTransformation` now also adds any explicit interfaces from inlined children.

### DIFF
--- a/loki/transformations/hoist_variables.py
+++ b/loki/transformations/hoist_variables.py
@@ -142,8 +142,9 @@ class HoistVariablesAnalysis(Transformation):
             variables = self.find_variables(routine)
             item.trafo_data[self._key]["to_hoist"] = variables
             dims = flatten([getattr(v, 'shape', []) for v in variables])
+            import_map = routine.import_map
             item.trafo_data[self._key]["imported_sizes"] = [(d.type.module, d) for d in dims
-                                                            if str(d) in routine.import_map]
+                                                            if str(d) in import_map]
             item.trafo_data[self._key]["hoist_variables"] = [var.clone(name=f'{routine.name}_{var.name}')
                                                              for var in variables]
         else:
@@ -281,8 +282,9 @@ class HoistVariablesTransformation(Transformation):
 
         # Add imports used to define hoisted
         missing_imports_map = defaultdict(set)
+        import_map = routine.import_map
         for module, var in item.trafo_data[self._key]["imported_sizes"]:
-            if not var.name in routine.import_map:
+            if not var.name in import_map:
                 missing_imports_map[module] |= {var}
 
         if missing_imports_map:
--- a/loki/transformations/hoist_variables.py
+++ b/loki/transformations/hoist_variables.py
@@ -80,15 +80,17 @@ are provided to create derived classes for specialisation of the actual hoisting
         scheduler.process(transformation=HoistTemporaryArraysTransformationAllocatable(key=key))
 """
 
+from collections import defaultdict
+
 from loki.batch import Transformation, ProcedureItem
 from loki.expression import (
     symbols as sym, FindVariables, FindInlineCalls,
     SubstituteExpressions, is_dimension_constant
 )
 from loki.ir import (
-    CallStatement, Allocation, Deallocation, Transformer, FindNodes
+    CallStatement, Allocation, Deallocation, Transformer, FindNodes, Comment, Import
 )
-from loki.tools.util import is_iterable, as_tuple, CaseInsensitiveDict
+from loki.tools.util import is_iterable, as_tuple, CaseInsensitiveDict, flatten
 
 from loki.transformations.utilities import single_variable_declaration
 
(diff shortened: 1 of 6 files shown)


loki/transformations/hoist_variables.py
```diff
@@ -139,9 +141,13 @@ def transform_subroutine(self, routine, **kwargs):
         if role != 'driver':
             variables = self.find_variables(routine)
             item.trafo_data[self._key]["to_hoist"] = variables
+            dims = flatten([getattr(v, 'shape', []) for v in variables])
+            item.trafo_data[self._key]["imported_sizes"] = [(d.type.module, d) for d in dims
+                                                            if str(d) in routine.import_map]
```
reuterbal, line 146: This rebuilds the import_map in every iteration. Better to cache it before the list comprehension. 

```suggestion
            import_map = routine.import_map
            item.trafo_data[self._key]["imported_sizes"] = [(d.type.module, d) for d in dims
                                                            if d.name in import_map]
```

loki/transformations/hoist_variables.py
```diff
@@ -272,6 +279,24 @@ def transform_subroutine(self, routine, **kwargs):
                         routine=routine, call=call, variables=hoisted_variables
                     )
 
+        # Add imports used to define hoisted
+        missing_imports_map = defaultdict(set)
+        for module, var in item.trafo_data[self._key]["imported_sizes"]:
+            if not var.name in routine.import_map:
```
reuterbal, line 285: Same comment as above about caching `import_map`

loki/transformations/inline.py
```diff
@@ -194,8 +194,8 @@ def map_inline_call(self, expr, *args, **kwargs):
 
 def resolve_sequence_association_for_inlined_calls(routine, inline_internals, inline_marked):
     """
-    Resolve sequence association in calls to all member procedures (if `inline_internals = True`) 
-    or in calls to procedures that have been marked with an inline pragma (if `inline_marked = True`). 
+    Resolve sequence association in calls to all member procedures (if `inline_internals = True`)
+    or in calls to procedures that have been marked with an inline pragma (if `inline_marked = True`).
```
reuterbal, line 198:
```suggestion
    Resolve sequence association in calls to all member procedures (if ``inline_internals = True``)
    or in calls to procedures that have been marked with an inline pragma (if ``inline_marked = True``).
```

RST syntax...

loki/transformations/inline.py
```diff
@@ -618,13 +618,27 @@ def inline_marked_subroutines(routine, allowed_aliases=None, adjust_imports=True
...
         # Finally, apply the import remapping
         routine.spec = Transformer(import_map).visit(routine.spec)
 
+        # Add missing explicit interfaces from inlined subroutines
+        new_intfs = []
+        intf_symbols = routine.interface_symbols
+        for callee in call_sets.keys():
+            for intf in callee.interfaces:
+                for b in intf.body:
+                    s = getattr(b, 'procedure_symbol', None)
```
reuterbal, line 635: Can we not use `intf.symbols` here?

reuterbal (changes requested):
Thanks, looks great! Some small implementation pointers but otherwise all good.

loki/transformations/inline.py
awnawab: Oops, not sure how that ended up here, thanks for spotting!

loki/transformations/inline.py
awnawab: `Interface.body` comprises of `Subroutine` objects and not just the symbols, so I did it like this to capture the whole `Subroutine` object. If using `intf.symbols`, I would have to add `symbol.type.dtype.procedure` to the body rather than the symbol. Either is fine by me, please let me know if you would like me to update to the latter.

reuterbal: Yes, I'm aware - but you're only inspecting the procedure symbols of these subroutines not the objects themselves, which is what `intf.symbols` should give you. Have a look at the implementation of `Interface.symbols`, which should be essentially the same as what you do here, but also includes the case of named interfaces etc: https://github.com/ecmwf-ifs/loki/blob/77114a954c0d774831d5ab7d4138774adaa20aba/loki/ir/nodes.py#L700-L710

awnawab:
Many thanks @reuterbal for the super prompt review and feedback 🙏 I've addressed all your comments, please have another look whenever you can!

reuterbal (approved):
Thanks for the quick implementation of changes! Looks good to me

reuterbal:
Apologies, this has accrued conflicts now - @awnawab, would you mind rebasing this over main?

PR #322, opened 2024-06-10
InlineTransformation: Update Scheduler SGraph if marked_inline is activated

awnawab:
Certain transformations (e.g. `InlineTransformation`) might create or delete callgraph edges without renaming or creating new items. Updating the `sgraph` in such cases requires only `scheduler._discover` to be triggered. This PR adds such an option to the `Transformation` class.

### DIFF
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -503,8 +503,10 @@ class Scheduler:
 
         if transformation.creates_items:
             self._discover()
-
             self._parse_items()
+
+        if transformation.creates_edges:
+            self._discover()
 
     def callgraph(self, path, with_file_graph=False, with_legend=False):
         """
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -505,9 +505,6 @@ class Scheduler:
             self._discover()
             self._parse_items()
 
-        if transformation.creates_edges:
-            self._discover()
-
     def callgraph(self, path, with_file_graph=False, with_legend=False):
         """
         Generate a callgraph visualization and dump to file.
--- a/loki/transformations/inline.py
+++ b/loki/transformations/inline.py
@@ -74,20 +74,20 @@ class InlineTransformation(Transformation):
         inlining constants (default: True)
     resolve_sequence_association: bool
         Resolve sequence association for routines that contain calls to inline (default: False)
-    creates_edges: bool
-        Run the scheduler discovery mechanism to re-compute the callgraph edges (default: False)
     """
 
     # Ensure correct recursive inlining by traversing from the leaves
     reverse_traversal = True
+
+    # This transformation will potentially change the edges in the callgraph
+    creates_items = False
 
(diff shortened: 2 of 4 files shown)


loki/batch/scheduler.py
```diff
@@ -503,9 +503,11 @@ def _get_definition_items(_item, sgraph_items):
 
         if transformation.creates_items:
             self._discover()
-
             self._parse_items()
 
+        if transformation.creates_edges:
+            self._discover()
```
reuterbal, line 509: If no items need to be parsed, I don't expect calling `_parse_items` to be particularly expensive. Wouldn't re-using the `creates_items` manifest entry work here, too?

The reason is that I don't want to make the Transformation manifest unnecessarily complex.

loki/transformations/inline.py
```diff
@@ -84,7 +86,8 @@ def __init__(
             inline_internals=False, inline_marked=True,
             remove_dead_code=True, allowed_aliases=None,
             adjust_imports=True, external_only=True,
-            resolve_sequence_association=False
+            resolve_sequence_association=False,
+            creates_edges=False
```
reuterbal, line 90: Manifest properties shouldn't be a runtime option but instead need to be specified as a class property on the transformation class implementation. But I would vote for setting `creates_items = True` in the manifest instead, as explained above.

reuterbal (changes requested):
Thanks, that's a very useful addition and key ingredient that we hacked around for a while. However, I would like to limit the complexity of the Transformation manifest - see inline comments.

loki/batch/scheduler.py
awnawab: Yes `_parse_items` isn't too expensive; `_discover` and `_parse_items` both cost around ~14s for ecWAM. And yes, `creates_items` would also do the job and I am happy to just use that rather than adding this new option.

loki/transformations/inline.py
awnawab: Yes I agree it would be cleaner to just add it to the manifest. I was just trying to make it optional so we only incur this cost if we have to, I am happy to set `creates_items = True` to the manifest instead.

reuterbal (approved):
Many thanks, looks great!

mlange05 (approved):
Looks great and is neatly solved and tested. I've left one small comment for future consideration, but this need not be addressed now (or soon). GTG from me. :shipit:

loki/transformations/inline.py
```diff
@@ -95,6 +98,8 @@ def __init__(
         self.adjust_imports = adjust_imports
         self.external_only = external_only
         self.resolve_sequence_association = resolve_sequence_association
+        if self.inline_marked:
+            self.creates_items = True
```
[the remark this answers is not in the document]
reuterbal: That pattern is likely my fault, aiming to be conservative in switching this on to avoid the discovery overhead when inlining is limited to member routines only

Issue #317, opened 2024-05-23
acc seq bug


ecossevin:
I'm trying to add **acc seq** to a routine spec, and the "seq" disappears : 

**test.py :**

```
from loki import *
src_file = "sub.F90"
src_name = "TEST"
source = Sourcefile.from_file(src_file)
routine = source[src_name]

routine.spec.insert(0,ir.Pragma(keyword='acc', content='routine ('+    routine.name+') seq'))
routine.spec.insert(1,ir.Comment(text=''))

print(fgen(routine))
```

**sub.F90 :** 

```
SUBROUTINE TEST()

IMPLICIT NONE

END SUBROUTINE TEST
```

**result :** 

```
[Loki::Sourcefile] Constructed from sub.F90 in 0.02s
SUBROUTINE TEST ()
  
!$acc routine( TEST )
  
  IMPLICIT NONE
  
END SUBROUTINE TEST
```

reuterbal:
Confirmed and fixed in #324. There was a problem in the get_pragma_parameters utility that would ignore clauses without parentheses that occur after a clause with parentheses.

PR #324, opened 2024-06-10
get_pragma_parameters: Fix parsing clauses without parentheses in the tail string

reuterbal:
This fixes the issue reported in #317


mlange05 (approved):
Looks good, and test is very neat. GTG from me. :shipit:

Issue #161, opened 2023-10-02
Definitions not attached if multiple scopes defined in the same `Sourcefile`


rommelDB:
Regarding the execution of the following code:

```
from loki import Sourcefile, Subroutine, FindVariables
from pprint import pprint

fcode = """
module foo
  implicit none
  integer, parameter :: j = 16

  contains
    integer function SUM(v)
      implicit none
      integer, intent(in) :: v
      SUM = v + 1
    end function SUM
end module foo

module test
  use foo
  implicit none
  integer, parameter :: rk = selected_real_kind(12)
  integer, parameter :: ik = selected_int_kind(9)

  contains
    subroutine calc (n, res)
      integer, intent(in) :: n
      real(kind=rk), intent(inout) :: res
      real(kind=rk), dimension(n) :: vec

      integer(kind=ik) :: i

      do i = 1, n
         vec(i) = 1.0_rk/(real(n - i + 1, kind=rk))
      end do

      do i = 1, n
        res = res + SQRT(real(i, kind=rk)) + SUM(j)
      end do
    end subroutine calc

end module test
"""

source = Sourcefile.from_source(fcode)
calc_subroutine = source['calc']
vars = FindVariables().visit(calc_subroutine.body)
pprint(vars)
```

Actual output:
```
{Array('vec', None, None, None, (Scalar('i', None, None, None),)),
 DeferredTypeSymbol('REAL', None, None, <SymbolAttributes BasicType.DEFERRED>),
 DeferredTypeSymbol('SQRT', None, None, <SymbolAttributes BasicType.DEFERRED>),
 DeferredTypeSymbol('SUM', None, None, <SymbolAttributes BasicType.DEFERRED>),
 DeferredTypeSymbol('j', None, None, <SymbolAttributes BasicType.DEFERRED>), 
 Scalar('i', None, None, None),
 Scalar('rk', None, None, None),
 Scalar('n', None, None, None),
 Scalar('res', None, None, None)}
```
I wanted to print variables only, however, I didn't expect to REAL, SQRT, and SUM be in the list.

I'm aware that it is the current definition, though: https://github.com/ecmwf-ifs/loki/blob/19fd1c439250e951744cfe941efea71df85b5a7a/loki/expression/expr_visitors.py#L169

So, in order to obtain the true type of deferred symbol. Would a new pass or something like that be required? I would appreciate any feedback.

reuterbal:
Hi Rommel,
[...]

PR #325, opened 2024-06-10
ProgramUnit.resolve_typebound_var: raise error if top-level parent is not declared

### LINKED ISSUE #307: resolve_typebound_var warning
Hi,
routine.resolve_typebound_var("A%B") 
resolve_typebound_var should complain when A isn't in the routine. 

Erwan

reuterbal:
This fixes #307.

Piggy-backed is a small fix with test for missing enrichment between modules within the same source file, reported in #161.

### DIFF
--- a/loki/tests/test_subroutine.py
+++ b/loki/tests/test_subroutine.py
@@ -2270,7 +2270,6 @@ end subroutine
     """.strip()
 
     source = Sourcefile.from_source(fcode, frontend=frontend, xmods=[tmp_path])
-    # routine = Subroutine.from_source(fcode_routine, frontend=frontend, xmods=[tmp_path], definitions=source.definitions) #source['some_routine']
     routine = source['some_routine']
 
     tt_some = routine.resolve_typebound_var('tt%some')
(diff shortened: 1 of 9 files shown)


loki/tests/test_subroutine.py
```diff
@@ -2238,3 +2239,75 @@ def test_call_args_kwargs_conversion(frontend):
...
+
+subroutine some_routine
+    use header_mod, only: third_type
+    implicit none
+    type(third_type) :: tt
+end subroutine
+    """.strip()
+
+    source = Sourcefile.from_source(fcode, frontend=frontend, xmods=[tmp_path])
+    # routine = Subroutine.from_source(fcode_routine, frontend=frontend, xmods=[tmp_path], definitions=source.definitions) #source['some_routine']
```
mlange05, line 2273: Debug leftover?

reuterbal: Ooops.

loki/tests/test_subroutine.py
```diff
@@ -2238,3 +2239,75 @@ def test_call_args_kwargs_conversion(frontend):
...
+
+    tt = routine.resolve_typebound_var('tt')
+    assert tt == 'tt'
+    assert tt.type.dtype.name == 'third_type'
+    assert tt.type.dtype.typedef is source['header_mod']['third_type']
+
+    # This does not throw an error as the use-case of incomplete type definitions
+    # may well require working with incomplete type definitions
+    tt_invalid_val = routine.resolve_typebound_var('tt%invalid%val')
+    assert tt_invalid_val == 'tt%invalid%val'
```
mlange05, line 2295: I'm not sure I understand the reasoning here. The type of `tt` is known so it can be safely inferred that `tt%invalid` is in deed invalid. If the type of `tt` were deferred, I'd agree, but it is attached in this case, no?

mlange05 (commented):
Ok, this looks great and is indeed very useful and makes things safer. I've left one question on the test, because I don;t quite understand when/what is invalid, and what is not. There's also a rogue commented line still in test - but once those are addressed, this is GTG, I think.

loki/tests/test_subroutine.py
reuterbal: Right, I see your point. I was thinking of situations where we don't have the type definition and overlooked the fact that we actually have the complete definition here. Let me see if I can quickly add an error in this case without harming the situation where we don't have the full type definition.

mlange05: Looks great. Many thanks for addressing this.

mlange05 (approved):
Perfect! Looks great now, GTG from me. :shipit:

PR #327, opened 2024-06-11
Fparser frontend: Fix wrong classification as StatementFunction in translation to Loki IR

reuterbal:
_Sigh_....

Found with scan2m.F90, correctly removing all misclassifications now.


reuterbal:
I think this is ready for another look @mlange05
As discussed offline, the type attribute is gone and the heuristic is handled differently - in the process I fixed some OFP issues.

mlange05 (approved):
Excellent! Many thanks for grinding through the extra changes and the respective fixes. Looks great now! GTG from me. :shipit:

PR #329, opened 2024-06-12
Split reads and writes for certain accumulation patterns

awnawab:
Lukas M. demonstrated that for accumulation patterns of the type:

```
do jlon=1,nproma
   var(jlon, n1) = var(jlon, n1) + 1.
   var(jlon, n2) = var(jlon, n2) + 1.
enddo
```

a compiler cannot rule out the possibility that `n1` and `n2` do not in fact alias the same location. In such a case, it is unable to run the loads and stores out-of-order.

This PR contributes a pragma assisted transformation to split the above into separate reads and writes, thereby removing the dependency between subsequent loads and stores and allowing the compiler to optimise more effectively.

### DIFF
--- a/loki/transformations/split_read_write.py
+++ b/loki/transformations/split_read_write.py
@@ -19,6 +19,16 @@ class SplitReadWriteWalk(Transformer):
     """
     A :any:`Transformer` class to traverse the IR, in-place replace read-write
     assignments with reads, and build a transformer map for the corresponding writes.
+
+    Parameters
+    ----------
+    dimensions : list
+       A list of :any:`Dimension` objects corresponding to all :any:`Loop`s in the ``!$loki split-read-write`` region.
+    variable_map : dict
+       The variable_map of the parent :any:`Subroutine`.
+    count : int
+       A running count of the newly created temporaries in the parent :any:`Subroutine` so that
+       temporaries created by previous ``!$loki split-read-write`` regions are not redefined.
     """
 
     def __init__(self, dimensions, variable_map, count=-1, **kwargs):
@@ -116,6 +126,11 @@ class SplitReadWriteTransformation(Transformation):
            var(jlon, n2) = loki_temp_1(jlon)
         enddo
         !$loki end split-read-write
+
+    Parameters
+    ----------
+    dimensions : list
+       A list of :any:`Dimension` objects corresponding to all :any:`Loop`s in the ``!$loki split-read-write`` region.
     """
 
     item_filter = (ProcedureItem,)
--- a/loki/transformations/split_read_write.py
+++ b/loki/transformations/split_read_write.py
@@ -7,14 +7,79 @@
 
 
 from loki.batch import Transformation, ProcedureItem
-from loki.expression import FindVariables, Array, SubstituteExpressions
-from loki.tools import as_tuple, flatten
+from loki.expression import Array, SubstituteExpressions
+from loki.tools import as_tuple
 from loki.ir import (
-    pragma_regions_attached, is_loki_pragma, nodes as ir, FindNodes,
-    Transformer, FindScopes
+    pragma_regions_attached, is_loki_pragma, nodes as ir, FindNodes, Transformer
 )
 
 __all__ = ['SplitReadWriteTransformation']
+
+class SplitReadWriteWalk(Transformer):
+    """
+    A :any:`Transformer` class to traverse the IR, in-place replace read-write
+    assignments with reads, and build a transformer map for the corresponding writes.
+    """
+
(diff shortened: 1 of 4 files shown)


loki/transformations/split_read_write.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # cache variable_map for fast lookup later
+        variable_map = routine.variable_map
+
+        # find split read-write pragmas
+        with pragma_regions_attached(routine):
+            for region in FindNodes(ir.PragmaRegion).visit(routine.body):
+                if is_loki_pragma(region.pragma, starts_with='split read-write'):
+
+                    # find assignments inside pragma region
```
reuterbal, line 76: I feel like this requires quite a lot of IR traversals for something that should be doable in a single pass. For small regions that shouldn't make a big difference, but if there are many then it might pile up. I'm counting at least 6 passes over the region body (in a quick eyeball-estimate).

How about a single tree walk that in-place updates the assignments that need splitting, and collects all assignments at the end?

Drafting here a bit of pseudo code to illustrate what I mean:

```py
class SplitReadWriteWalk(Transformer):

    def __init__(self, **kwargs):
        self.write_assignments = []
        kwargs['inplace'] = True
        super().__init__(**kwargs)

    def visit_Loop(self, o, **kwargs):
        # This collects the relevant loop dimensions in kwargs['dim_nest'] during the
        # tree walk
        dim = [d for d in self.dimensions if d.index == o.variable]
        dim_nest = kwargs.pop('dim_nest', [])
        return super().visit_Node(o, dim_nest=dim_nest + dim, **kwargs)

    def visit_Assignment(self, o, **kwargs):
        if isinstance(o.lhs, Array) and a.lhs.name in o.rhs:
            parent_loop_dims = kwargs.pop('dim_nest', [])
            # <do the logic for shape etc here>
            tmp_var = <...>
            self.write_assignments += [ir.Assignment(lhs=o.lhs, rhs=tmp_var)
            o._update(lhs=tmp_var)
        return o

for region in ...:
    transformer = SplitReadWriteWalk()
    region.body = transformer.visit(region.body)
    if transformer.write_assignments:
        region.body.append(transformer.write_assignments)
```

Not sure if I'm overlooking something, but it feels to me like that should be sufficient and much more efficient.

reuterbal (changes requested):
Thanks, this is really neat (and many thanks to @lukasm91 for pointing this out in the first place!). However, the implementation appears a bit too complicated and costly and I'm suggesting an alternative in the comments. Please have a look if that makes sense.

loki/transformations/split_read_write.py
awnawab: Yes, tree-walk is definitely the way to do it! The use of `FindScopes` to discover a parent loop nest felt a bit clunky to me too but the idea of implementing this as a visitor never occured to me. Many thanks for the suggestion 🙏

awnawab:
Many thanks again @reuterbal for the great tip! I've now implemented the `SplitReadWriteWalk` that you suggested. I've retained a second pass to insert the newly split write assignments. These have to be nested correctly within the control flow, and the simplest way I could think of achieving this was to copy the entire `PragmaRegion` and amend the leaf nodes as necessary.

@mlange05 could you also please review the PR? Thanks!

mlange05 (changes requested):
Great utility and very neatly done with the single-pass `Transformer`. I've only left a few small nit-pick comments about docstrings (and an "out of interest" question), but otherwise this is ready, I think. Well done, and many thanks to both, @awnawab and @reuterbal

scripts/loki_transform.py
```diff
@@ -186,7 +186,9 @@ def convert(
         mode = mode.replace('-', '_')  # Sanitize mode string
 
         # Write out all modified source files into the build directory
-        file_write_trafo = FileWriteTransformation(builddir=build, mode=mode)
+        file_write_trafo = scheduler.config.transformations.get('FileWriteTransformation', None)
```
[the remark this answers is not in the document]
awnawab: For ecWAM I need to enable the `include_module_var_imports` option in the `FileWriteTransformation`. I'm assuming `global_var_offload` will eventually be removed as an argument from `convert` as we further simplify `loki_transform.py`, so I thought I might as well make the `FileWriteTransformation` also configurable.

awnawab:
Many thanks @mlange05 for the review and the feedback 😄  The constructor arguments were indeed unclear, and I've now added documentation for them.

mlange05:
Thanks, this looks great now. Going in...

mlange05 (approved):
Looking good now, going in.

reuterbal (approved):

Issue #330, opened 2024-06-12
Derived types with inheritance: parsing fails if parent typedef not within same module


reuterbal:
Consider the following reproducer:
```py
@pytest.mark.parametrize('frontend', available_frontends())
def test_derived_type_inheritance_missing_parent(frontend, tmp_path):
    fcode_parent = """
module parent_mod
    implicit none
    type, abstract, public :: parent_type
        integer :: val
    end type parent_type
end module parent_mod
    """.strip()

    fcode_derived = """
module derived_mod
    use parent_mod, only: parent_type
    implicit none
    type, public, extends(parent_type) :: derived_type
        integer :: val2
    end type derived_type
contains
    subroutine do_something(this)
        class(derived_type), intent(inout) :: this
        this%val = 1
        this%val2 = 2
    end subroutine do_something
end module derived_mod
    """.strip()

    parent = Module.from_source(fcode_parent, frontend=frontend, xmods=[tmp_path])
    derived = Module.from_source(fcode_derived, frontend=frontend, xmods=[tmp_path])
```

This fails with 
```
self = TypeDef:: derived_type

    @property
    def parent_type(self):
        if not self.extends:
            return None
        if not self.parent:
            return BasicType.DEFERRED
>       return self.parent.symbol_attrs[self.extends].dtype.typedef
E       AttributeError: 'BasicType' object has no attribute 'typedef'

loki/ir/nodes.py:1538: AttributeError
```

PR #331, opened 2024-06-12
Fix derived type inheritance when parent type is not available (#330)

reuterbal:
This assumed parent definition was always available in the symbol table, which it isn't if the parent is defined in a different module and no enrichment has been applied.


mlange05 (approved):
Yes, looks good - neat and well-tested. GTG. :shipit:

PR #332, opened 2024-06-12
Fix Linter warnings for inline calls with interface block imported from header with func.h suffix

reuterbal:
Another beautiful edge case but this time purely in Linter recipes: 

- apparently, interface blocks may not only be declared in `.intfb.h` headers, but also in `.func.h` headers
- inline calls with string parameters can actually be identified as inline calls (rather than array subscripts) by fparser


mlange05 (approved):
Nifty Fparser trick! Looks clean and neat as always, GTG from me. :shipit:

PR #334, opened 2024-06-17
POC - linter: missing kind specifier for real literals and possible fix via git patch

MichaelSt98:
What it does/How it works:

* read the original Fortran file
* find all real literals (including its nodes `with_ir_node=True`)
* if `literal.kind is None`: fix kind and save in dict
* apply dict via `SubstituteExpressions` to the relevant node (only!)
* substitute in the Fortran file representation/string the newly generated node with the old node (using `node.source.string`) 
* generate diff/patch using `difflib` and write to file
* (git apply this file/patch)

__________

Demonstration with CLOUDSC:

With this config file 

```
basedir: .
include:
  - src/cloudsc_fortran/*.F90

rules:
  - MissingKindSpecifierRealLiterals

junitxml_file: linter.xml
violations_file: test_new_violations.yml

max_workers: 4
```

applied to the CLOUDSC dwarf via e.g., `loki-lint.py check -c lint_config.yml --fix --worker 4` 
generates a file "loki_lint_CLOUDSC.patch" which looks like

```
--- a/src/cloudsc_fortran/cloudsc.F90
+++ b/src/cloudsc_fortran/cloudsc.F90
@@ -2048,11 +2048,11 @@
         ! otherwise faster to freeze (snow or ice pellets)
         ZQPRETOT(JL) = MAX(ZQX(JL,JK,NCLDQS)+ZQX(JL,JK,NCLDQR),ZEPSEC)
         PRAINFRAC_TOPRFZ(JL) = ZQX(JL,JK,NCLDQR)/ZQPRETOT(JL)
-        IF (PRAINFRAC_TOPRFZ(JL) > 0.8) THEN 
-          LLRAINLIQ(JL) = .True.
+        IF (PRAINFRAC_TOPRFZ(JL) > 0.8_JPRB) THEN
+          LLRAINLIQ(JL) = .true.
         ELSE
-          LLRAINLIQ(JL) = .False.
-        ENDIF
+          LLRAINLIQ(JL) = .false.
+        END IF
       ENDIF

       ! If temperature less than zero
@@ -2382,15 +2382,15 @@

       ZAPLUSB   = RCL_APB1*ZVPICE-RCL_APB2*ZVPICE*ZTP1(JL,JK)+ &
      &             PAP(JL,JK)*RCL_APB3*ZTP1(JL,JK)**3
-      ZCORRFAC  = (1.0/ZRHO(JL))**0.5
-      ZCORRFAC2 = ((ZTP1(JL,JK)/273.0)**1.5)*(393.0/(ZTP1(JL,JK)+120.0))
+      ZCORRFAC = (1.0_JPRB / ZRHO(JL))**0.5_JPRB
+      ZCORRFAC2 = ((ZTP1(JL, JK) / 273.0_JPRB)**1.5_JPRB)*(393.0_JPRB / (ZTP1(JL, JK) + 120.0_JPRB))

       ZPR02 = ZRHO(JL)*ZPRECLR*RCL_CONST1S/(ZTCG*ZFACX1S)

       ZTERM1 = (ZQSICE(JL,JK)-ZQE)*ZTP1(JL,JK)**2*ZVPICE*ZCORRFAC2*ZTCG* &
      &          RCL_CONST2S*ZFACX1S/(ZRHO(JL)*ZAPLUSB*ZQSICE(JL,JK))
-      ZTERM2 = 0.65*RCL_CONST6S*ZPR02**RCL_CONST4S+RCL_CONST3S*ZCORRFAC**0.5 &
-     &          *ZRHO(JL)**0.5*ZPR02**RCL_CONST5S/ZCORRFAC2**0.5
+      ZTERM2 = 0.65_JPRB*RCL_CONST6S*ZPR02**RCL_CONST4S + RCL_CONST3S*ZCORRFAC**0.5_JPRB*ZRHO(JL)**0.5_JPRB*ZPR02**RCL_CONST5S /  &
+      & ZCORRFAC2**0.5_JPRB

       ZDPEVAP = MAX(ZCOVPCLR(JL)*ZTERM1*ZTERM2*PTSPHY,0.0_JPRB)

```

and can be applied via `git apply --whitespace=fix loki_lint_*.patch`

resulting in `git diff`

```
diff --git a/src/cloudsc_fortran/cloudsc.F90 b/src/cloudsc_fortran/cloudsc.F90
index 17f7144..035bf07 100644
--- a/src/cloudsc_fortran/cloudsc.F90
+++ b/src/cloudsc_fortran/cloudsc.F90
@@ -2048,11 +2048,11 @@ DO JK=NCLDTOP,KLEV
         ! otherwise faster to freeze (snow or ice pellets)
         ZQPRETOT(JL) = MAX(ZQX(JL,JK,NCLDQS)+ZQX(JL,JK,NCLDQR),ZEPSEC)
         PRAINFRAC_TOPRFZ(JL) = ZQX(JL,JK,NCLDQR)/ZQPRETOT(JL)
-        IF (PRAINFRAC_TOPRFZ(JL) > 0.8) THEN 
-          LLRAINLIQ(JL) = .True.
+        IF (PRAINFRAC_TOPRFZ(JL) > 0.8_JPRB) THEN
+          LLRAINLIQ(JL) = .true.
         ELSE
-          LLRAINLIQ(JL) = .False.
-        ENDIF
+          LLRAINLIQ(JL) = .false.
+        END IF
       ENDIF
     
       ! If temperature less than zero
@@ -2382,15 +2382,15 @@ ENDIF ! on IEVAPRAIN
 
       ZAPLUSB   = RCL_APB1*ZVPICE-RCL_APB2*ZVPICE*ZTP1(JL,JK)+ &
      &             PAP(JL,JK)*RCL_APB3*ZTP1(JL,JK)**3
-      ZCORRFAC  = (1.0/ZRHO(JL))**0.5
-      ZCORRFAC2 = ((ZTP1(JL,JK)/273.0)**1.5)*(393.0/(ZTP1(JL,JK)+120.0))
+      ZCORRFAC = (1.0_JPRB / ZRHO(JL))**0.5_JPRB
+      ZCORRFAC2 = ((ZTP1(JL, JK) / 273.0_JPRB)**1.5_JPRB)*(393.0_JPRB / (ZTP1(JL, JK) + 120.0_JPRB))
 
       ZPR02 = ZRHO(JL)*ZPRECLR*RCL_CONST1S/(ZTCG*ZFACX1S)
 
       ZTERM1 = (ZQSICE(JL,JK)-ZQE)*ZTP1(JL,JK)**2*ZVPICE*ZCORRFAC2*ZTCG* &
      &          RCL_CONST2S*ZFACX1S/(ZRHO(JL)*ZAPLUSB*ZQSICE(JL,JK))
-      ZTERM2 = 0.65*RCL_CONST6S*ZPR02**RCL_CONST4S+RCL_CONST3S*ZCORRFAC**0.5 &
-     &          *ZRHO(JL)**0.5*ZPR02**RCL_CONST5S/ZCORRFAC2**0.5
+      ZTERM2 = 0.65_JPRB*RCL_CONST6S*ZPR02**RCL_CONST4S + RCL_CONST3S*ZCORRFAC**0.5_JPRB*ZRHO(JL)**0.5_JPRB*ZPR02**RCL_CONST5S /  &
+      & ZCORRFAC2**0.5_JPRB
 
       ZDPEVAP = MAX(ZCOVPCLR(JL)*ZTERM1*ZTERM2*PTSPHY,0.0_JPRB)
  
```

________

Also tested for "ifs-source" with "include: arpifs/phys_ec/**/*.F90".

**Some notes:**

* we could also directly substitute within the Fortran file and don't use the git patch approach
* just a POC to show how we could fix things and minimise the diff (as we don't use fgen() for all of the Fortran file/subroutine)
    *  diff is still relatively large for the transformed/fixed nodes (white space, indentation, new lines, ...)

**Current limitations:**

* problematic for multiple Rules/Fixes that apply to the same node (not tested but this won't work)
    * we would need to have a common dict/map per node for all the rules before applying `SubstituteExpressions`

### DIFF
--- a/lint_rules/lint_rules/ifs_arpege_coding_standards.py
+++ b/lint_rules/lint_rules/ifs_arpege_coding_standards.py
@@ -14,6 +14,7 @@ current version of the coding standards.
 
 from collections import defaultdict
 import re
+import difflib
 
 try:
     from fparser.two.Fortran2003 import Intrinsic_Name
@@ -22,14 +23,83 @@ except ImportError:
     _intrinsic_fortran_names = ()
 
 from loki import (
-    FindInlineCalls, FindNodes, GenericRule, Module, RuleType
+    FindInlineCalls, FindNodes, GenericRule, Module, RuleType,
+    ExpressionFinder, ExpressionRetriever, FloatLiteral,
+    SubstituteExpressions
 )
-from loki import ir
-
+from loki import ir, fgen
+from loki.frontend.util import read_file
 
 __all__ = [
     'MissingImplicitNoneRule', 'OnlyParameterGlobalVarRule', 'MissingIntfbRule',
+    'MissingKindSpecifierRealLiterals'
 ]
+
+
+class FindFloatLiterals(ExpressionFinder):
+    """
+    A visitor to collects :any:`FloatLiteral` used in an IR tree.
+
+    See :class:`ExpressionFinder`
+    """
+    retriever = ExpressionRetriever(lambda e: isinstance(e, (FloatLiteral,)))
+
+class MissingKindSpecifierRealLiterals(GenericRule):
+    """
+    ...
+    """
+
+    type = RuleType.SERIOUS
+    fixable = True
+
+    docs = {
+        'id': 'L0',
+        'title': (
+            'Real Literals must have a kind specifier. '
+        ),
+    }
+
+
+    @classmethod
+    def check_subroutine(cls, subroutine, rule_report, config, **kwargs):
+        """
+        ...
+        """
+        literal_nodes = FindFloatLiterals(with_ir_node=True).visit(subroutine.body)
+        for node, literals in literal_nodes:
+            for literal in literals:
+                if literal.kind is None:
+                    rule_report.add(f'Real/Float literal without kind specifier "{literal}"', node)
+
+    @classmethod
+    def fix_subroutine(cls, subroutine, rule_report, config, sourcefile=None):
+        """
+        ...
+        """
+        original_content = read_file(str(sourcefile.path))
+        content = original_content
+        literal_nodes = FindFloatLiterals(with_ir_node=True).visit(subroutine.body)
+        for node, literals in literal_nodes:
+            literal_map = {}
+            for literal in literals:
+                if literal.kind is None:
+                    literal_map[literal] = FloatLiteral(value=literal.value, kind='JPRB')
+            if literal_map:
+                fixed_node = SubstituteExpressions(literal_map).visit(node)
+                indent = int((len(node.source.string) - len(node.source.string.lstrip(' ')))/2)
+                fixed_node_str = fgen(fixed_node, depth=indent)
+                content_new = re.sub(rf'{re.escape(node.source.string)}',
+                        fixed_node_str, content, flags = re.S)
+                content = content_new
+        diff = difflib.unified_diff(original_content.splitlines(), content_new.splitlines(),
+                f'a/{sourcefile.path}', f'b/{sourcefile.path}', lineterm='')
+        diff_str = '\n'.join(list(diff))
+        # print(f"---{sourcefile.path}------")
+        # print(diff_str)
+        # print(f"--------------------------")
+        with open (f'loki_lint_{subroutine.name}.patch', 'w') as f:
+            f.write(diff_str)
+            f.write('\n')
 
 
 class MissingImplicitNoneRule(GenericRule):
--- a/lint_rules/lint_rules/ifs_arpege_coding_standards.py
+++ b/lint_rules/lint_rules/ifs_arpege_coding_standards.py
@@ -72,14 +72,161 @@ class MissingKindSpecifierRealLiterals(GenericRule):
                     rule_report.add(f'Real/Float literal without kind specifier "{literal}"', node)
 
     @classmethod
+    def fix_subroutinei_test_2(cls, subroutine, rule_report, config, sourcefile=None):
+        """
+        ...
+        """
+        for node, literals in literal_nodes:
+            literal_map = {}
+            for literal in literals:
+                if literal.kind is None and 'e' not in literal.value.lower() and 'd' not in literal.value.lower():
+                    literal_map[literal] = FloatLiteral(value=literal.value, kind='JPRB')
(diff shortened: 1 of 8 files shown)


mlange05 (commented):
Hi, so this looks great and I really like the overall concept. As discussed offline, I think there are a few "next steps" to generalise this and make this a bit more usable in practice:
* A generalisation of the diff-regex-per-node mechanism to generate patches should be provided, likely as part of the loki.lint sub-package
* The false-positive generated by the `.true.` vs. `.TRUE.` in the adjoining lines shows that we need to make the `fgen` backend a bit more configurable (eg. via something like a pre-defined "style" dict; with this we could have a dedicated IFS-style that keeps the noise in the diffs low.
* The per-node expression-substitution limits this to only local changes like the kind-identifier, but in general this should be fine for now. Further extension can come later (if needed).

As a proof-of-concept this looks great though -very cool!

reuterbal (commented):
As discussed offline: This is fantastic and I'm positively surprised how relatively straightforward the implementation appears to be. The testing on our Fortran codes also makes this look very promising.

In terms of the implementation:
- I think we might want to think a little more about the association of IR containers with the corresponding Sourcefile object. We should try to somehow provide a mechanism of providing this upward link, to make the additional argument that you had to introduce unnecessary. Moreover, the `source` object of the Sourcefile should also have the full string already, which would make it redundant to read the file again. 
- The mechanics of creating a diff from replacing individual source lines should likely become a set of independent utilities. Maybe we need to add a second use case to trial what the interface of such utilities should look like in a more generic sense?
- We should think of how we are going to embed this in the larger linter control flow. Right now you've hijacked the fixer method, which was originally intended for editing the IR and later dumping the IR to file. We could make the "generate a diff instead"-mode a generic variant of that, or create a secondary fixer workflow side-by-side, or switch the entire fixer workflow to the diff methodology. I don't have a strong opinion on this (yet?) but probably worth a discussion, also with @mlange05 .

mlange05:
As discussed offline, this has now gone stale and can be closed. However, I think the diff/patch feature is a nice idea that we should aim to implement as a generic feature for loki-lint. I've filed issue #691 to that effect and propose that we should pick this up in a separate development.

Closing the PR now.

PR #335, opened 2024-06-26
Update ecwam regression tests to use develop branch

reuterbal:


reuterbal:
Thanks to @awnawab this should be ready now!

.github/workflows/regression_tests.yml
```diff
@@ -76,22 +76,14 @@ jobs:
...
-          source loki-activate
-          pip install fypp
+          pip install pyyaml fypp
 
       - name: Run CLOUDSC and ECWAM regression tests
         env:
           CLOUDSC_DIR: ${{ github.workspace }}/cloudsc
           CLOUDSC_ARCH: ${{ github.workspace }}/cloudsc/arch/github/ubuntu/gnu/9.4.0
           ECWAM_DIR: ${{ github.workspace }}/ecwam
-          ECWAM_ARCH: ${{ github.workspace }}/ecwam/arch/github/ubuntu/gnu/9.4.0
```
mlange05, line 87: Just double-checking: Is relying on default arch safe here? @awnawab

mlange05 (approved):
Looks good to me, and indeed many thanks for figuring out all the nifty details @reuterbal and @awnawab . I left one small comment on the use of the default arch (is this intentional and safe?), but if someone can confirm that this is fine, we're GTG. :+1: :shipit:

.github/workflows/regression_tests.yml
reuterbal: My understanding is that implies no arch files are used, i.e., no specific toolchain.cmake and env.sh. Since none are required, this should be fine.

awnawab: Yes exactly, the ecwam CI also runs without any arch files and just relies on the default cmake/ecbuild flags plus those set by `ecwam_compile_flags.cmake`.

awnawab (approved):
Many thanks for this, apologies for not wrapping this up sooner myself, it was long overdue. I can now finally forget about that very old ecwam dev branch that we were previously using 😄

Issue #339, opened 2024-07-03
Proposal: resolution of concrete procedures


quepas:
**This is only a proposal**. A draft implementation has been prepared here: .

One of the useful information missing in loki is the information about the real function being called when a generic interface is used:

```fortran
...
interface swap
        module procedure swap_int, swap_real
end interface swap
...
```

Then, called somewhere else:

```fortran
real :: r1, r2
...
! Loki knows it is a "generic" procedure, but that's all 
call swap(r1, r2)
```

Currently, in the `CallStmt`, loki points to a `dtype` with a procedure of DEFERED type. This PR adds capability of figuring out the concrete function really called (in some easy cases). The resolution mechanism tries to match parameters of each concrete function with arguments passed to the generic function call. Of course, the whole resolution is conservative (as everything in compilers! :D) and additional information is added only when everything matches.

The matching is based on the `ExpressionDimensionsMapper` and a new `ExpressionTypeMapper` which takes a piece of the AST and tries to deduce it's final type. This mapper is not very capable, it works well on typed variables, literals and some easy expressions (with `sum` op).

We weren't sure if we should populate the `procedure` with the concrete routine in the `ProcedureType`, so we have added an additional field `concrete_procedure` which points to the concrete procedure when a generic is used, and the normal procedure otherwise.

I am looking forward to your comments and the discussion of this feature.

Things to do:
- [ ] Implementation for other parsers
- [ ] More capable `ExpressionTypeMapper`
- [ ] More unit tests

PR #340, opened 2024-07-03
Draft: Proposal: resolution of the concrete procedure

quepas:
This is a proposal draft implementation of resolution of the concrete procedure used behind a call to a generic interface. More in: #339

### DIFF
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -34,7 +34,7 @@ import loki.expression.symbols as sym
 from loki.expression.operations import (
     StringConcat, ParenthesisedAdd, ParenthesisedMul, ParenthesisedDiv, ParenthesisedPow
 )
-from loki.expression import ExpressionDimensionsMapper, AttachScopes, AttachScopesMapper
+from loki.expression import ExpressionDimensionsMapper, ExpressionTypeMapper, AttachScopes, AttachScopesMapper
 from loki.logging import debug, perf, info, warning, error
 from loki.tools import (
     as_tuple, flatten, CaseInsensitiveDict, LazyNodeLookup, dict_override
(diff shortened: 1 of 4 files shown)


loki/frontend/fparser.py
```diff
@@ -2483,6 +2483,42 @@ def visit_Call_Stmt(self, o, **kwargs):
...
             arguments, kwarguments = (), ()
+        # Figure out the exact procedure being called if this is a call to a generic interface
+        if name.type.dtype.is_generic:
+            # If the interface is imported, take its definition from a module
+            # TODO: handle interfaces defined in the same module
+            if name.type.imported:
+                module = name.type.module
+                interface = [i for i in module.interfaces if i.spec.name == name.name]
+                if len(interface) == 1:
+                    interface = interface[0]
```
reuterbal, line 2494: You should be able to shortcut this with
```suggestion
                interface = module.interface_map[name.name]
```

loki/frontend/fparser.py
```diff
@@ -2483,6 +2483,42 @@ def visit_Call_Stmt(self, o, **kwargs):
...
+                            new_procedure_type = ProcedureType(name=name.name,
+                                                               is_function=ptype.is_function,
+                                                               is_generic=ptype.is_generic,
+                                                               return_type=ptype.return_type,
+                                                               concrete_procedure=symbol.type.dtype.procedure)
+                            new_symbol_attribute = SymbolAttributes(dtype=new_procedure_type,
+                                                                    imported=name.type.imported,
+                                                                    module=name.type.module)
+                            # When the scope is name.scope, then the concrete_procedure takes always the same value (
+                            # bug or my incomprehension?)
```
reuterbal, line 2520: This is precisely the problem that I'm describing. Hence my alternative suggestion which ensures that you can attach the correct scope to the symbol. We will enforce a strict requirement for having a symbol scope in the future, hence this workaround would no longer work.

reuterbal (changes requested):
First of all, a huge apology for letting this hang for so long! This is a really cool extension of the type system and neatly implemented.

I left a few comments throughout with a suggestion how to resolve the problem that you were hitting and described in a comment. Let me know what you think.

Lastly, due to the long time before I took a proper look at this, we have now a few conflicts with the main branch. These will need to be resolved after merging main or rebasing over main. Sorry about that.

quepas:
Thank you for all the suggestions! I haven't had enough time to work on this problem again, but it is something we are interested in. I will try to get back to it :)

PR #343, opened 2024-07-11
Rigorous use of tmp_path in tests

reuterbal:
Our tests have been littering the source directories and current working directory with temporary files, shared libraries, mod and xmod files, etc. This practice is not only annoying in terms of unclean git worktrees after running tests, but can also potentially cause issues due to false sharing of files between tests. For example, OMNI will happily use previously generated xmod files from other tests to resolve imports, even if they are entirely unrelated.

This PR introduces the rigorous use of the `tmp_path` fixture provided by pytest to set up per-test temporary directories. That is required whenever modules are parsed or temporary files are written, e.g., as source output or builder artifacts. 

As this affects a majority of all tests, the resulting diff is very substantial, but should hopefully only consists of the same 2 or 3 changes everywhere. Only in a few isolated instances I had to actually fix tests where they were implicitly relying on other tests to provide xmod files for OMNI.


mlange05 (approved):
Fantastic, very neat and very much appreciated! This must have been some grind. :wink: 

I'd like to prioritise that to keep rebase fallout small and allow this style to be retrofitted to newt tests immediately. So, GTG from mys side! :shipit:

PR #344, opened 2024-07-11
Housekeeping on CMakeLists.txt and pyproject.toml

### LINKED ISSUE #304: Fix gitless installation with CMake
PR #302 fixes the Loki installation from a non-VCS copy (e.g. tarball) with the `install` script or manual pip installs. However, in the CMake bundled installation, it fails when not performing an `EDITABLE` install. It's unclear why that is, but the following reproduces the problem on Atos HPCF:

```sh
module reset
module load cmake python3 ecbuild

git clone https://github.com/ecmwf-ifs/loki gitless_install
cd gitless_install
rm -rf .git
mkdir build
cd build
cmake ..
loki_env/bin/python -c "import loki; print(loki.__version__); print(loki.sourcefile.Sourcefile.from_source('subroutine my_routine\ninteger i\nend subroutine my_routine').to_fortran());"
```

The symptom is that the venv will only contain the Python files in the top level directoy and not the subdirectories...

reuterbal:
We still had the installation of the transformations package hanging about in the CMake integration. I also did some housekeeping on the pyproject.toml in the same instance and this seems to have fixed #304. @awnawab please confirm when you can.


reuterbal:
Python packaging is still as badly broken as ever, but it seems I may have found a stable configuration that yields the desired outcome.

In the process I discovered that tests that use f90wrap or f2py fail if they are being run directly using the pytest binary from a venv without activating the virtual environment (e.g., `venv/bin/pytest --pyargs loki`). To fix this, I piggy-backed a little utility which ensures that the python interpreter's parent directory is in the search path.

mlange05 (approved):
All looks great, and many thanks as always! I've run through the stuff in #304 and it works very nicely! GTG from me. :shipit:

awnawab (approved):
Many thanks for fixing this @reuterbal 🙏 This fixes the problems I encountered in the ecwam CI, and with this in place I can roll back all the related workarounds.

PR #345, opened 2024-07-18
Allow inlining of Statement Functions

MichaelSt98:
* OFP excluded from test because of some problems with Statement Functions I haven't investigated further yet
* if the order or statement function declaration is "wrong" this won't work as the statement function would be interpreted as array access ... but this is a problem that has nothing to do with the inlining itself

From

```fortran
subroutine stmt_func(arr, ret)
    implicit none
    real, intent(in) :: arr(:)
    real, intent(inout) :: ret(:)
    real :: ret2
    real, parameter :: rtt = 1.0
    real :: PTARE
    real :: FOEDELTA
    FOEDELTA ( PTARE ) = PTARE + 1.0
    real :: FOEEW
    FOEEW ( PTARE ) = PTARE + FOEDELTA(PTARE)

    ret = foeew(arr) 
    ret2 = foedelta(3.0)
end subroutine stmt_func
```

to

```fortran
SUBROUTINE stmt_func (arr, ret)
  IMPLICIT NONE
  REAL, INTENT(IN) :: arr(:)
  REAL, INTENT(INOUT) :: ret(:)
  REAL :: ret2
  REAL, PARAMETER :: rtt = 1.0
  
  ret = arr + arr + 1.0
  ret2 = 3.0 + 1.0
END SUBROUTINE stmt_func
```

### DIFF
--- a/loki/transformations/inline.py
+++ b/loki/transformations/inline.py
@@ -200,7 +200,7 @@ class InlineSubstitutionMapper(LokiIdentityMapper):
             function = expr.routine
             v_result = expr.routine.variable
             # Substitute all arguments through the elemental body
-            arg_map = dict(zip(function.arguments, expr.parameters))
+            arg_map = dict(expr.arg_iter())
             fbody = SubstituteExpressions(arg_map).visit(function.rhs)
             return fbody
 
@@ -208,7 +208,7 @@ class InlineSubstitutionMapper(LokiIdentityMapper):
         v_result = [v for v in function.variables if v == function.name][0]
 
         # Substitute all arguments through the elemental body
-        arg_map = dict(zip(function.arguments, expr.parameters))
+        arg_map = dict(expr.arg_iter())
         fbody = SubstituteExpressions(arg_map).visit(function.body)
 
         # Extract the RHS of the final result variable assignment
--- a/loki/transformations/inline.py
+++ b/loki/transformations/inline.py
@@ -11,11 +11,13 @@ Collection of utility routines to perform code-level force-inlining.
 
 """
 from collections import defaultdict, ChainMap
+from pymbolic.primitives import Expression
 
 from loki.batch import Transformation
 from loki.ir import (
     Import, Comment, Assignment, VariableDeclaration, CallStatement,
-    Transformer, FindNodes, pragmas_attached, is_loki_pragma, Interface
+    Transformer, FindNodes, pragmas_attached, is_loki_pragma, Interface,
+    StatementFunction
 )
 from loki.expression import (
     symbols as sym, FindVariables, FindInlineCalls, FindLiterals,
@@ -36,7 +38,8 @@ from loki.transformations.utilities import (
 __all__ = [
     'inline_constant_parameters', 'inline_elemental_functions',
     'inline_internal_procedures', 'inline_member_procedures',
-    'inline_marked_subroutines', 'InlineTransformation'
+    'inline_marked_subroutines', 'InlineTransformation',
+    'inline_statement_functions'
 ]
 
 
@@ -54,6 +57,10 @@ class InlineTransformation(Transformation):
         Replaces :any:`InlineCall` expression to elemental functions
         with the called function's body (see :any:`inline_elemental_functions`);
         default: True.
+    inline_stmt_funcs: bool
+        Replaces  :any:`InlineCall` expression to statement functions
+        with the corresponding rhs of the statement function if
+        the statement function declaration is available; default: True.
     inline_internals : bool
         Inline internal procedure (see :any:`inline_internal_procedures`);
         default: False.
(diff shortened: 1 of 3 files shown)


MichaelSt98:
Maybe it would be better to not replicate the functionality of `recursive_expression_map_update(expr_map, max_iterations=10):` and instead introduce a new argument `mapper` to allow using the `InlineSubstitutionMapper` instead of the `SubstitutionMapper`...
What do you think?

loki/transformations/inline.py
```diff
@@ -184,6 +196,15 @@ def map_inline_call(self, expr, *args, **kwargs):
             # We still need to recurse and ensure re-scoping
             return super().map_inline_call(expr, *args, **kwargs)
 
+        # if it is an inline call to a Statement Function
+        if isinstance(expr.routine, StatementFunction):
+            function = expr.routine
+            v_result = expr.routine.variable
+            # Substitute all arguments through the elemental body
+            arg_map = dict(zip(function.arguments, expr.parameters))
```
reuterbal, line 204: Just wondering: Wouldn't `arg_map = dict(expr.arg_iter())` work here, and also below for "regular" inline calls? That should also take care of keyword arguments.
(Cc @mlange05)

loki/transformations/inline.py
```diff
@@ -339,6 +360,89 @@ def inline_elemental_functions(routine):
...
+    """
+    # Keep track of removed symbols
+    removed_functions = set()
+
+    stmt_func_decls = FindNodes(StatementFunction).visit(routine.spec)
+    exprmap = {}
+    for call in FindInlineCalls().visit(routine.body):
+        if call.procedure_type is BasicType.DEFERRED:
+            continue
+        if call.procedure_type.is_function and isinstance(call.routine, StatementFunction):
```
reuterbal, line 419: Just a little hint to avoid repeated lookup of type info:
```suggestion
        proc_type = call.procedure_type
        if proc_type is BasicType.DEFERRED:
            continue
        if proc_type.is_function and isinstance(proc_type.procedure, StatementFunction):
```

loki/transformations/inline.py
```diff
@@ -339,6 +360,89 @@ def inline_elemental_functions(routine):
...
+    # remove the statement function declarations
+    routine.spec = Transformer(stmt_func_map).visit(routine.spec)
+
+    # remove the declarations of the statement function arguments
+    decls = FindNodes(VariableDeclaration).visit(routine.spec)
+    decls = tuple(d for d in decls if any(isinstance(s, sym.ProcedureSymbol) or s in stmt_arguments for s in d.symbols))
+    decl_map = {}
+    for decl in decls:
+        decl_map[decl] = None
+    routine.spec = Transformer(decl_map).visit(routine.spec)
```
reuterbal, line 443: This boils down to removing a bunch of variable declarations and everything in `stmt_func_decls`, so that should be possible in a single `Transformer` pass. Maybe something along these lines works?
```python
vars_to_remove = {stmt_func.variable.name.lower() for stmt_func in stmt_func_decls}
vars_to_remove += {arg.name.lower() for stmt_func in stmt_func_decls for arg in stmt_func.arguments}
spec_map = {stmt_func: None for stmt_func in stmt_func_decls}
for decl in routine.declarations:
    if any(var in vars_to_remove for var in decl.symbols):
        symbols = tuple(var for var in decl.symbols if var not in vars_to_remove)
        if symbols:
            decl._update(symbols=symbols)
        else:
            spec_map[decl] = None
routine.spec = Transformer(spec_map).visit(routine.spec)
```

reuterbal (changes requested):
Thanks, this looks great. Small remarks on improving the implementation and I agree with your suggestion to merge the expression map update.

loki/transformations/tests/test_inline.py
```diff
@@ -706,6 +708,53 @@ def test_inline_member_routines_with_associate(frontend):
...
+    """.strip()
+
+    fcode = f"""
+subroutine stmt_func(arr, ret)
+    implicit none
+    real, intent(in) :: arr(:)
+    real, intent(inout) :: ret(:)
+    real :: ret2
+    real, parameter :: rtt = 1.0
+    {stmt_decls_code if stmt_decls else ''}
```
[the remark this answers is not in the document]
MichaelSt98: That is correct. However, it does not change anything here.

loki/transformations/inline.py
```diff
@@ -184,11 +195,20 @@ def map_inline_call(self, expr, *args, **kwargs):
             # We still need to recurse and ensure re-scoping
             return super().map_inline_call(expr, *args, **kwargs)
 
+        # if it is an inline call to a Statement Function
+        if isinstance(expr.routine, StatementFunction):
+            function = expr.routine
+            v_result = expr.routine.variable
```
reuterbal, line 201: This looks unused?

reuterbal (commented):
Thanks, this looks great now!
Due to other merged PRs we have a conflict now, would you mind rebasing this over latest main? I hope this will then also trigger the tests again.

reuterbal (approved):

loki/transformations/utilities.py
```diff
@@ -468,7 +468,7 @@ def replace_selected_kind(routine):
             routine.spec.prepend(imprt)
 
 
-def recursive_expression_map_update(expr_map, max_iterations=10):
+def recursive_expression_map_update(expr_map, max_iterations=10, mapper_cls=SubstituteExpressionsMapper):
```
mlange05, line 471: No action: Very neat! :clap:

mlange05 (approved):
Looks fantastic! Neatly implemented, neatly tested and very powerful! Very nice! GTG from me. :shipit:

PR #346, opened 2024-07-18
DrHookTransformation: Add explicit label renaming

mlange05:
This PR adds an renaming option to the `DrHookTransformation` that allows and explicit renaming via a string-to-string dict in addition to the current append-suffix mode. This allows a broader use for this housekeeping transformation, eg. when generating control-flow subroutines.

With this, I've also added a few minor clean-ups and fixes, as detailed below:
* Rename the argument `mode` to `suffix`, as it's a suffix we add to the DR_HOOK labels
* Use the `Transformation.recurse_to_internal_procedures` class property instead of explicit recursion calls
* Add a `kernel_only` argument (`True` by default), with which we can force it to apply to "driver" routines as well
* Add `rename` argument (dict) that gets used when/if a label is matched for explicit label renaming.

### DIFF
--- a/loki/transformations/drhook.py
+++ b/loki/transformations/drhook.py
@@ -44,22 +44,29 @@ def remove_unused_drhook_import(routine):
 
 class DrHookTransformation(Transformation):
     """
-    Re-write or remove the DrHook label markers in transformed
-    kernel routines
+    Re-write or remove the DrHook label markers either by appending a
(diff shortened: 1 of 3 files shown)


reuterbal (approved):
Thanks, looks very neat and good to see another implicit recursion gone. GTG!

PR #349, opened 2024-07-22
IR: Update to Pydantic >2.0 compatibility

mlange05:
~_Note:` For now, just for testing..._~

This PR brings a few enabling changes to the internal IR node definitions that allow us to lift the version restriction on `pydantic`. We needed this previously, as we were relying on some (broken) default behaviour that was fixed upstream in pydantic version 2. By changing to the correct type annotations, in particular for `Optional` arguments, we can now lift this restriction.

In addition, I also added a small set of rudimentary checks for the constructor and object behaviour (immutability) for three most common/interesting node types. I plan to expand on this somewhat in a follow-up PR that will require pydantic features from version 2.x. 

In some more detail:
* Use `Optional` throughout node type annotations to allow default values of `None`
* Add a rudimentary test for basic constructor behaviour and "frozeness" of our IR nodes, for `Assignment`, `Loop` and `Conditional`
* Lift version restriction on pydantic and add a few small fixes to deal with the fallout

Note: This is the first of two PRs and paves the way for more cool stuff to come (as discussed offline).

### DIFF
--- a/loki/backend/tests/test_cufgen.py
+++ b/loki/backend/tests/test_cufgen.py
@@ -81,7 +81,7 @@ end module transformation_module_cufgen
     call_map = {}
     for call in FindNodes(ir.CallStatement).visit(driver.body):
         if "kernel" in str(call.name):
-            with pytest.raises(AssertionError):
+            with pytest.raises(ValidationError):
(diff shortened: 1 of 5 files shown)


reuterbal (approved):
This looks great, many thanks for grinding this through!

PR #350, opened 2024-07-24
IR: Automatic sanitisation of tuples in IR constructors

mlange05:
_Just testing for now...._


loki/ir/nodes.py
```diff
@@ -560,7 +570,7 @@ class _ConditionalBase():
 
     condition: Expression
     body: Tuple[Node, ...]
-    else_body: Optional[Tuple[Node, ...]] = None
+    else_body: Optional[Tuple[Node, ...]] = ()
```
reuterbal, line 573: Is that a change in behaviour for the default value of `else_body`?
I think so far we used `else_body=None` as indicator that there isn't an else branch in the conditional - as opposed to an empty else branch (which has the same result but different on a string level).

As in, will `ir.Conditional(condition=<...>, body=<...>)` now be represented as the following?
```
IF (<...>) THEN
  <...>
ELSE
ENDIF
```

reuterbal (approved):
This looks fantastic, many thanks, and we finally gain the flexibility of being more forgiving in the tuple nesting in the constructor while ensuring correct IR nodes.

One question about a potential change in default behaviour for Conditional and whether this is intentional, otherwise good to go!

mlange05:
@reuterbal I just noted that we now have a hard dependency on `pydantic>=2.0` with this, so please let me quickly update the setup file....

mlange05:
Ok, fix pushed. Please note that existing local checkouts might fail after this - to remedy, a simple `pip install -e .` in the loki directory should suffice to get pydantic updated.

PR #351, opened 2024-07-25
Fix DEV_ALLOC_SIZE for ecwam regression and add SCC-HOIST variant

awnawab:
A few small changes to fix the ecwam regression tests. What was causing the runtime failures was the default `DEV_ALLOC_SIZE` in the FIELD_API buddy allocator. The change to the fypp+pyyaml installation is for a future ecwam update which relies on (in the absence of fckit) the system python installation having fypp+pyyaml.


reuterbal (approved):
Fantastic, many thanks for figuring this out 🙏

PR #355, opened 2024-08-02
Transformations: Test and fix corner case in get_local_arrays

mlange05:
_Note: This is a small bug fix to #354 that was overlooked and is currently breaking H24-physics regression._

When running get_local_arrays over the spec of a subroutine we were erroneously picking up component arrays of derived-type arguments that were used for shape definitions. This extends the test for this corner case and fixes it.


awnawab (approved):
Neatly fixed and tested, looks g2g 👌 The macos test failure seems unrelated, I don't think it should block this PR. I'll have a look at debugging that one next week.

PR #356, opened 2024-08-07
Tools: Disable timeout utility test on MacOS due to sporadic failures

mlange05:
We suspect the failures to be due to load on CI runners, so disabling this test for now.


MichaelSt98 (approved):
Thanks!

PR #357, opened 2024-08-07
extend hoist variables functionality

MichaelSt98:
to account for possibly uninitialised variables used as dimensions (if wanted via flag)

This 

```fortran
subroutine driver(NLON, NB, FIELD1)
    use kernel_mod, only: kernel
    implicit none
    INTEGER, INTENT(IN) :: NLON, NB
    integer :: b
    integer, intent(inout) :: field1(nlon, nb)
    integer :: local_nlon
    local_nlon = nlon
    do b=1,nb
        call KERNEL(local_nlon, field1(:,b))
    end do
end subroutine driver
```

would be transformed to

```fortran
SUBROUTINE driver (NLON, NB, field1)
  USE kernel_mod, ONLY: kernel
  implicit none
  INTEGER, INTENT(IN) :: NLON, NB
  INTEGER :: b
  INTEGER, INTENT(INOUT) :: field1(nlon, nb)
  INTEGER :: local_nlon
  INTEGER :: kernel_tmp1(local_nlon)
  local_nlon = nlon
  DO b=1,nb
    CALL KERNEL(local_nlon, field1(:, b), kernel_tmp1)
  END DO
END SUBROUTINE driver
```

which would be a problem as `local_nlon` is not initialised. 

Setting `remap_dimensions=True` the transformation will produce

```fortran
SUBROUTINE driver (NLON, NB, field1)
  USE kernel_mod, ONLY: kernel
  implicit none
  INTEGER, INTENT(IN) :: NLON, NB
  INTEGER :: b
  INTEGER, INTENT(INOUT) :: field1(nlon, nb)
  INTEGER :: local_nlon
  INTEGER :: kernel_tmp1(nlon)
  local_nlon = nlon
  DO b=1,nb
    CALL KERNEL(local_nlon, field1(:, b), kernel_tmp1)
  END DO
END SUBROUTINE driver
```

### DIFF
--- a/loki/transformations/hoist_variables.py
+++ b/loki/transformations/hoist_variables.py
@@ -213,7 +213,7 @@ class HoistVariablesTransformation(Transformation):
 
     _key = 'HoistVariablesTransformation'
 
-    def __init__(self, as_kwarguments=False, remap_dimensions=False):
+    def __init__(self, as_kwarguments=False, remap_dimensions=True):
         self.as_kwarguments = as_kwarguments
         self.remap_dimensions = remap_dimensions
 
@@ -343,10 +343,16 @@ class HoistVariablesTransformation(Transformation):
         variables : tuple of :any:`Variable`
             The tuple of variables for remapping.
         """
-        dim_map = {}
-        assignments = FindNodes(Assignment).visit(routine.body)
-        for assignment in assignments:
-            dim_map[assignment.lhs] = assignment.rhs
+        dim_vars = [
+            dim_var
+            for var in variables if isinstance(var, sym.Array)
+            for dim_var in FindVariables().visit(var.dimensions)
+        ]
+        dim_map = {
+            assignment.lhs: assignment.rhs
+            for assignment in FindNodes(Assignment).visit(routine.body)
+            if assignment.lhs in dim_vars
+        }
         variables = [var.clone(dimensions=SubstituteExpressions(dim_map).visit(var.dimensions))
                 if isinstance(var, sym.Array) else var for var in variables]
         return variables
--- a/loki/transformations/hoist_variables.py
+++ b/loki/transformations/hoist_variables.py
@@ -88,7 +88,8 @@ from loki.expression import (
     SubstituteExpressions, is_dimension_constant
 )
 from loki.ir import (
-    CallStatement, Allocation, Deallocation, Transformer, FindNodes, Comment, Import
+    CallStatement, Allocation, Deallocation, Transformer, FindNodes, Comment, Import,
+    Assignment
 )
 from loki.tools.util import is_iterable, as_tuple, CaseInsensitiveDict, flatten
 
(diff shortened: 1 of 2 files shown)


loki/transformations/hoist_variables.py
```diff
@@ -319,6 +330,27 @@ def driver_variable_declaration(self, routine, variables):
...
+        ----------
+        routine : :any:`Subroutine`
+            The relevant subroutine.
+        variables : tuple of :any:`Variable`
+            The tuple of variables for remapping.
+        """
+        dim_map = {}
+        assignments = FindNodes(Assignment).visit(routine.body)
+        for assignment in assignments:
+            dim_map[assignment.lhs] = assignment.rhs
```
reuterbal, line 349: Just to point out how to do this in a "single line", or single list comprehension at least. For larger kernels like CLOUDSC with hundreds of assignments I expect you might actually be able to measure a small advantage with below change.
```suggestion
        dim_map = {
            assignment.lhs: assignment.rhs
            for assignment in FindNodes(Assignment).visit(routine.body)
        }
```

Also, this map might become quite big, so we might want to build a list of candidates first and then create the mapping? Suggestion (untested) would be something like this:
```suggestion
        dim_vars = [
            dim_var
            for var in variables
            for dim_var in FindVariables().visit(var.dimensions)
            if isinstance(var, sym.Array)
        ]
        dim_map = {
            assignment.lhs: assignment.rhs
            for assignment in FindNodes(Assignment).visit(routine.body)
            if assignment.lhs in dim_vars
        }
```

loki/transformations/hoist_variables.py
```diff
@@ -204,12 +205,17 @@ class HoistVariablesTransformation(Transformation):
     ----------
     as_kwarguments : boolean
         Whether to pass the hoisted arguments as `args` or `kwargs`.
+    remap_dimensions : boolean
```
reuterbal, line 208: Is there a situation where we wouldn't want to do this? It seems to me like we might not need to make this an option and instead always do this.

reuterbal (changes requested):
Thanks, this looks great! I left two remarks, please have a look if that makes sense or point out what I'm missing there. Otherwise this is fantastic.

loki/transformations/hoist_variables.py
MichaelSt98: Good question ... I can't think of a situation we don't want to do that but on the other hand I think having a switch is not a bad idea?!
Anyway, I switched the default but if you insist we can remove this flag.

reuterbal: No, flipping the default should be enough. Thanks!

reuterbal (approved):
Thanks, looks great and gtg from me!
I'll rebase the branch and will merge once green

loki/transformations/hoist_variables.py
```diff
@@ -319,6 +330,33 @@ def driver_variable_declaration(self, routine, variables):
         """
         routine.variables += tuple(v.rescope(routine) for v in variables)
 
+    @staticmethod
+    def driver_variable_declaration_dim_remapping(routine, variables):
```
mlange05, line 334: [no action]: Just a general note: we keep hitting this remapping problem in many use cases - so eventually we might want a general utility that can do this. But for this use case this looks great.

mlange05 (approved):
Looks good to me; clean and nicely tested. I left a general comment for later development, but no action needed and GTG from me. :shipit:

PR #358, opened 2024-08-09
ecWAM regression tests: switch to develop-1.3 branch

awnawab:
Technical development in ecWAM will (at least for the short term) lag one cycle behind the latest scientific version. Therefore it makes more sense to run the loki regression tests on `develop-1.3` which will be the staging branch for further technical developments on top of 49r1 science.


mlange05 (approved):
Yes, as discussed offline, makes sense. GTG :shipit:

PR #328, opened 2024-06-12
CLOUDSC low-level GPU (transpilation) via Loki (CUF/CUDA)

MichaelSt98:
Still work in progress ... 

Corresponding CLOUDSC branch `nams-cloudsc-low-level-GPU`

Build via `cloudsc-bundle build --clean --arch=arch/ecmwf/hpc2020/nvhpc/22.11 --with-loki --with-gpu --with-cuda`

New modes/binaries: 

* `./bin/dwarf-cloudsc-loki-idem-lower` CPU variant with lowered block index
* `./bin/dwarf-cloudsc-loki-idem-lower-loop` CPU variant with lowered block index and loop
* `./bin/dwarf-cloudsc-loki-scc-parametrise-cuda` SCC-PARAMETRISE CUDA C
* `./bin/dwarf-cloudsc-loki-scc-hoist-cuda` SCC-HOIST CUDA C

Furthermore, usage of pipelines:

```
SCCLowLevelCufParametrise = partial(
    Pipeline, classes=(
        SCCBaseTransformation,
        SCCDevectorTransformation,
        SCCDemoteTransformation,
        SCCRevectorTransformation,
        LowerBlockIndexTransformation,
        InjectBlockIndexTransformation,
        LowerBlockLoopTransformation,
        SccLowLevelLaunchConfiguration,
        SccLowLevelDataOffload,
        ParametriseTransformation
    )
)

SCCLowLevelCufHoist = partial(
    Pipeline, classes=(
        SCCBaseTransformation,
        SCCDevectorTransformation,
        SCCDemoteTransformation,
        SCCRevectorTransformation,
        LowerBlockIndexTransformation,
        InjectBlockIndexTransformation,
        LowerBlockLoopTransformation,
        SccLowLevelLaunchConfiguration,
        SccLowLevelDataOffload,
        HoistTemporaryArraysAnalysis,
        HoistTemporaryArraysDeviceAllocatableTransformation
    )
)

SCCLowLevelParametrise = partial(
    Pipeline, classes=(
        InlineTransformation,
        GlobalVariableAnalysis,
        GlobalVarHoistTransformation,
        DerivedTypeArgumentsTransformation,
        SCCBaseTransformation,
        SCCDevectorTransformation,
        SCCDemoteTransformation,
        SCCRevectorTransformation,
        LowerBlockIndexTransformation,
        InjectBlockIndexTransformation,
        LowerBlockLoopTransformation,
        SccLowLevelLaunchConfiguration,
[...]

PR #359, opened 2024-08-09
SCC: Add vectorisation annotations in SCCRevector and translate in SCCAnnotate 

mlange05:
~_Note: Just testing for now..._~

This PR changes the way we add OpenACC annotations in the SCC pipelines and generally refactors `SCCAnnotate` and `SCCRevector`. The key change is that we now insert Loki-specific annotations for vector-loops, driver loops, routine annotations and vector reductions in `SCCRevector` and then let `SCCAnnotate` translate this into OpenACC directives. This is in preparation for the addition of an additional re-vectorisation scheme and the eventual support for OpenMP-offload in the SCC pipelines, as it will allow `SCCAnnotate` to easily switch between directive flavours.

There's also quite a bit of refactoring and clean-up, in particular in the upper control methods of `SCCAnnotate`, as well as the utility method `check_routine_pragmas`. Overall I hope this is now a bit cleaner and better structured for future development.

In more detail:
* `wrap_vector_section` has been made a standalone routine (will be re-used in alternative re-vector scheme).
* `SCCRevector` now annotates vector and sequential loops with respective `!$loki loop` pragmas, as well as marking kernel routines `!%loki routine seq|vector` and adding `!$loki loop vector-reductions` for vector loops in marked vector-reduction regions.
* `SCCRevector` also re-uses some of these utilities when re-vectoring kernel loops; the respective routines have been refactored to work on generic node "sections" (tuple of `Node`).
* `SCCAnnotate` converts `!$loki loop` and `!$loki routine` directives into `!$acc` equivalents and adds `private` clauses to vector loops and `!$acc data present` clauses to kernel routines.
* The `check_routine_pragmas` utility has been renamed `check_routine_sequential` and no longer inserts pragmas at all. It merely checks if a routine has been marked with `!$loki routine seq`. The second use case (check for repeated processing) is now handled explicitly in `SCCAnnotate` when annotating kernel routines (which cannot happen twice!).
* A general tidy-up of `SCCAnnotate` to make utility methods object-bound and re-use common ones in the driver processing code path. We still need the `block_dim`, but the `horizontal` is no longer needed in `SCCAnnotate`.
* `test_scc_vector.py` now also checks for the insertion of certain `!$loki loop` pragmas.

### DIFF
--- a/loki/transformations/single_column/tests/test_scc.py
+++ b/loki/transformations/single_column/tests/test_scc.py
@@ -805,7 +805,7 @@ def test_scc_annotate_empty_data_clause(frontend, horizontal, blocking):
 
        k = n
        do k=1, 3
-          n = k + 1.
+          n = k + 1
        enddo
     end subroutine some_kernel
     """
--- a/loki/transformations/single_column/tests/test_scc.py
+++ b/loki/transformations/single_column/tests/test_scc.py
@@ -790,6 +790,43 @@ def test_scc_annotate_routine_seq_pragma(frontend, horizontal, blocking):
 
 
 @pytest.mark.parametrize('frontend', available_frontends())
+def test_scc_annotate_empty_data_clause(frontend, horizontal, blocking):
+    """
+    Test that we do not generate empty `!$acc data` clauses.
+    """
+
+    fcode = """
+    subroutine some_kernel(n)
+       implicit none
+       ! Scalars should not show up in `!$acc data` clause
+       integer, intent(inout) :: n
+!$loki routine seq
+       integer :: k
+
+       k = n
+       do k=1, 3
+          n = k + 1.
+       enddo
+    end subroutine some_kernel
+    """
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+
+    pragmas = FindNodes(Pragma).visit(routine.ir)
+    assert len(pragmas) == 1
+    assert pragmas[0].keyword == 'loki'
+    assert pragmas[0].content == 'routine seq'
+
+    transformation = SCCAnnotateTransformation(directive='openacc', block_dim=blocking)
+    transformation.transform_subroutine(routine, role='kernel', targets=['some_kernel',])
+
+    # Ensure the routine pragma is in the first pragma in the spec
+    pragmas = FindNodes(Pragma).visit(routine.ir)
+    assert len(pragmas) == 1
+    assert pragmas[0].keyword == 'acc'
+    assert pragmas[0].content == 'routine seq'
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
 def test_scc_vector_reduction(frontend, horizontal, blocking):
     """
     Test for the insertion of OpenACC vector reduction directives.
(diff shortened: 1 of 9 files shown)


mlange05:
Quick note: This is ready for review, but currently still fails ec-phys regression locally and will need some small fixes.

loki/transformations/single_column/vector.py
```diff
@@ -264,48 +299,154 @@ class SCCRevectorTransformation(Transformation):
...
+        section : tuple of :any:`Node`
+            Code section in which to mark "seq loops".
+        """
+        for loop in FindNodes(ir.Loop).visit(section):
 
-        # Ensure we clone all body nodes, to avoid recursion issues
-        vector_loop = ir.Loop(variable=index, bounds=bounds, body=Transformer().visit(section))
+            # Skip loops explicitly marked with `!$loki/claw nodep`
+            if loop.pragma and any('nodep' in p.content.lower() for p in as_tuple(loop.pragma)):
+                continue
```
reuterbal, line 366: I think there's currently no test that covers this functionality.

loki/transformations/tests/test_utilities.py
```diff
@@ -537,12 +537,10 @@ def test_transform_utilites_check_routine_pragmas(frontend, tmp_path):
     i = i + 1
   end subroutine test_acc_vec
 
-end module test_check_routine_pragmas_mod
+end module test_check_routine_sequential_mod
 """
     module = Module.from_source(fcode, frontend=frontend, xmods=[tmp_path])
 
-    # TODO: This utility needs some serious clean-up, so we're just testing
-    # the bare basics here and promise to do better next time ;)
```
reuterbal, line 545: Deleting this must have felt good 😏

reuterbal (approved):
Thanks, this is a really nice clean-up of that processing step that will make subsequent developments much easier and more flexible, and improve the ways how we can interact with and guide transformation pipelines manually when necessary.

One minor comment, nothing that strictly needs an action now.

loki/transformations/tests/test_utilities.py
mlange05: No comment 😏

mlange05:
Thanks for the review @reuterbal . However, I just confirmed that this still fails H24-dev locally, so will need to have another look. I'll remove the merge-ready label and rebase. Will ping once fixed.

mlange05:
Small integration note: I've needed to rebase manually, as one of the changes in this (the explicit removal of vector-section objects/labels) was implement in a similar, but separate implementation in a recent PR (#328 ). I've also found the bugs that were causing local H24-dev to fail, and added fixes+tests for those corner cases (top two commits). I think I'm happy with this now (confirmed to work locally with H24 ecphys regression), so please have another look and merge once ready.

loki/transformations/single_column/tests/test_scc.py
```diff
@@ -776,20 +774,57 @@ def test_scc_base_routine_seq_pragma(frontend, horizontal):
...
+    subroutine some_kernel(n)
+       implicit none
+       ! Scalars should not show up in `!$acc data` clause
+       integer, intent(inout) :: n
+!$loki routine seq
+       integer :: k
+
+       k = n
+       do k=1, 3
+          n = k + 1.
```
reuterbal, line 808: Should this not be an integer literal `1`?

reuterbal (approved):
Thanks, looks all good to me!

loki/transformations/single_column/tests/test_scc.py
mlange05: Indeed, it should. Fixed now

PR #360, opened 2024-08-20
Transformations: Internalise `IdemTransformation`

mlange05:
This change move the `IdemTransformation` that we often use for testing to the `loki.transformations` sub-package and fixes a configure bug when omitting the `options` entry in TOML(!)-based configuration entries. This is in preparation for the full move to config-file transformation configuration in CLOUDSC.

### DIFF
--- a/loki/transformations/idempotence.py
+++ b/loki/transformations/idempotence.py
@@ -10,8 +10,9 @@ from loki.batch import Transformation
 
 class IdemTransformation(Transformation):
     """
-    A custom transformation pipeline that primarily does nothing,
-    allowing us to test simple parse-unparse cycles.
+    A custom transformation that does absolutely nothing!
+
+    This can be used to test simple parse-unparse cycles.
     """
 
     def transform_subroutine(self, routine, **kwargs):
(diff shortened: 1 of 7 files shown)


mlange05:
@reuterbal Ok, I added two more small tweaks that were needed for the CLOUDSC loki generalisation. With this, I can recover most of the standards behaviour, so should be ok now.

loki/transformations/idempotence.py
```diff
@@ -0,0 +1,18 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from loki.batch import Transformation
+
+
+class IdemTransformation(Transformation):
+    """
+    A custom transformation pipeline that primarily does nothing,
```
reuterbal, line 13: Even secondarily it does nothing, I think ;-P

reuterbal (approved):
Thanks, this looks great. Looking forward to the end result of this...

Coverage reports that the call to `normalize_range_indexing` from the `SanitiseTransformation` is not invoked by any tests. Maybe it would be easy to amend one of the existing tests to apply this?

mlange05:
Ok, apologies, this last piggy-backed change has turned this into a bit of a mess. I've now stripped the change regarding `normalize_range_indexing` and will solve this another way - and in another PR. @reuterbal Please double-check that this is sane again.

reuterbal:
Yes, looks all good. Thanks for this!

PR #361, opened 2024-08-22
fix for 'resolve_vector_notation' utility

MichaelSt98:
problem occurred whenever two dimensions had the same size

e.g.,

```fortran
integer :: ret1(param1, param1)
...
ret1(:, :) = 0
```

resulted in:

```fortran
  DO i_ret1_1=1,param1
    DO i_ret1_0=1,param1
      ret1(i_ret1_1, i_ret1_1) = 0
    END DO
  END DO
```

instead of the correct result:

```fortran
  DO i_ret1_1=1,param1
    DO i_ret1_0=1,param1
      ret1(i_ret1_0, i_ret1_1) = 0
    END DO
  END DO
```


reuterbal (approved):
Thanks! What an annoying little oversight... Fix looks great, good to go from me!

mlange05 (approved):
Yes! Good catch, clean fix, neat test. GTG from me. :shipit:

PR #364, opened 2024-08-26
Loki-transform: Pass `cuf` option to FilewriteTrafo

mlange05:
This is a small workaround for enabling SCC-CUF pipelines from config file in CLOUDSC. As we still push static config info alongside dynamic build system settings, this is needed for now if we don't want to preprocess the loki config via cmake.


reuterbal (approved):
Looks good, thanks! Will merge as soon as CI is complete

PR #366, opened 2024-08-27
String-based expression substitution and moar expression tests!

mlange05:
This change separates some of the `loki.expression` tests into separate sub-packages and adds a new `SubstituteStringExpressions` utility that builds on the recently added `parse_expr` utility for expression substitution.

The test separation is meant to reduce the size of the general expression test, which is growing slightly out of proportions. I've also added some explicit test of the much used `SubstituteExpressions` transformer.

Steps in detail:
* A new `loki.expression.test.test_expr_visitors` sub-package with some existing tests and new tests for expression substitution; this includes an added test for `FindVariables` which was previously only covered implicitly (it seems)
* Separating out the tests of for the new expression parser into `loki.expression.tests.test_parser`
* New `SubstituteStringExpressions` transformer that works pretty much as the `SubstituteExpressions` but uses `parse_expr` to use string mapping as input.


loki/expression/tests/test_parser.py
reuterbal, line 1: I've not gone through this in detail but I'm trusting this is largely just a source move.

loki/expression/tests/test_expr_visitors.py
```diff
@@ -0,0 +1,222 @@
...
+implicit none
+contains
+subroutine rick(dave, never)
+  real(kind=8), intent(inout) :: dave, never
+end subroutine rick
+end module external_mod
+    """
+    fcode = """
+module test_mod
+  use external_mod, only: rick
```
reuterbal, line 76:
```suggestion
  use external_mod, only: rick
  implicit none
```

missing also from a few other places in this file, 'cause, wouldn't want to risk Rick letting me down due to implicit typing after all.

reuterbal (approved):
The `SubstituteStringExpressions` utility is a really nice one! Obviously, loads of ways how someone can use this in dangerous ways, so we need to watch out for this.

The test clean-up and extended explicit coverage is much appreciated!

Good to go from me!

loki/expression/tests/test_parser.py
mlange05: Yes, it is. I tried to be careful here. 😉

loki/expression/tests/test_expr_visitors.py
mlange05: Rick would never tell a lie and hurt you. Fixed now 😉

PR #370, opened 2024-08-30
Filter out globals in `get_local_arrays`

awnawab:
A small PR to filter out global variables from `get_local_arrays`; these may be included if `get_local_arrays` is applied to `routine.spec`.


reuterbal (approved):
Thanks, nice catch! Implementation looks straightforward to me and is well tested.

mlange05 (approved):
Yes, looks good to me. GTG :shipit:

PR #371, opened 2024-09-02
Change/fix pipeline for mode 'scc-raw-stack'

MichaelSt98:
Not sure whether I am missing something, but shouldn't the `SCCRawStackPipeline` be called for mode `scc-raw-stack`?


mlange05 (approved):
Yes, it should! Thanks for finding this! 🙏

PR #372, opened 2024-09-05
IR: Move `expr_visitors` to `loki.ir`

mlange05:
~Just testing for now...~

_Note: This sits on top of #366, because it moves the new tests alongside it._

Another refactoring change that moves the definition of various IR visitors (`expr_visitors.py`) to the `loki.ir` package. This means that now, at least in principle, we only have a one-way dependency between `loki.ir` => `loki.expression` - and where this is broken, it is marked as such via local imports with pylint markers.

~Integration note: I'll leave this as draft until #366 is merged, so that the diff is not too overwhelming.~


reuterbal (approved):
Thank you very much for cleaning this up and the tedious work of updating all the imports accordingly. This is a fundamentally sensible thing to do and has been recognized also by others.

Changes look good to me! :shipit:

PR #374, opened 2024-09-10
Vertical loop fusion and demotion of temporaries

MichaelSt98:
Vertical loop fusion and demotion of temporaries in the vertical dimension if possible based on
pragmas in the input source code and heuristics.

This is intended to be a starting point for a more automated and sophisticated approach (including dataflow analysis) as well as further optimisations that require this as a first step.

**To test with CLOUDSC**, checkout branch `nams-vertical-loop-fusion`.
This results in the following performance improvements (executed with `1 262144 128`):

| VARIANT | ORIG/WITHOUT | WITH NEW TRAFO |
| --------- | ---------------- | ------------------- |
| SCC | ca. 200 GF/s | ca. 460 GF/s |
| SCC-HOIST | ca. 320 GF/s  | ca. 500 GF/s |
| SCC-STACK | ca. 320 GF/s | ca. 520 GF/s |

### DIFF
--- a/loki/transformations/single_column/vertical.py
+++ b/loki/transformations/single_column/vertical.py
@@ -42,10 +42,14 @@ class SCCFuseVerticalLoops(Transformation):
     vertical : :any:`Dimension`
         :any:`Dimension` object describing the variable conventions used in code
         to define the vertical data dimension and iteration space.
+    apply_to : list of str, optional
+        list of routines to apply this transformation to, if not provided or None 
+        apply to all routines (default: None)
     """
 
-    def __init__(self, vertical=None):
+    def __init__(self, vertical=None, apply_to=None):
         self.vertical = vertical
+        self.apply_to = apply_to or ()
 
     def transform_subroutine(self, routine, **kwargs):
         """
@@ -63,6 +67,8 @@ class SCCFuseVerticalLoops(Transformation):
             return
         role = kwargs['role']
         if role == 'kernel':
+            if self.apply_to and routine.name.lower() not in self.apply_to:
+                return
             self.process_kernel(routine)
 
     def process_kernel(self, routine):
--- a/loki/transformations/single_column/vertical.py
+++ b/loki/transformations/single_column/vertical.py
@@ -14,9 +14,10 @@ from loki.ir import (
     is_loki_pragma, pragmas_attached,
     get_pragma_parameters
 )
-from loki.tools import as_tuple
+from loki.tools import as_tuple, CaseInsensitiveDict
 from loki.transformations.transform_loop import loop_fusion, loop_interchange
 from loki.transformations.array_indexing import demote_variables
+from loki.transformations.utilities import get_local_arrays
 from loki.logging import info
 
 __all__ = ['SCCFuseVerticalLoops']
@@ -28,7 +29,13 @@ class SCCFuseVerticalLoops(Transformation):
 
     .. note::
         This transfomation currently relies on pragmas being inserted in the input
-        source files.
+        source files. Relevant pragmas are `!$loki loop-interchange` to expose the
+        vertical loops (in case vertical loops are nested) and `!$loki loop-fusion`
+        possibly grouped via `group(<group name>)`. Further, if there are loops
+        that initialize multilevel arrays (`jk +/- 1`) it is possible to mark those
+        loops as `!$loki loop-fusion group(<group-name>-init)`. This allows to split
+        the relevant node and moves the initialization of those arrays to the top of
+        the group.
 
     Parameters
     ----------
@@ -50,8 +57,6 @@ class SCCFuseVerticalLoops(Transformation):
         routine : :any:`Subroutine`
             The subroutine in the vertical loops should be fused and
             temporaries be demoted.
-        horizontal : :any:`Dimension`
-            The dimension specifying the horizontal vector dimension
         """
         if self.vertical is None:
             info('[SCCFuseVerticalLoops] is not applied as the vertical dimension is not defined!')
@@ -97,27 +102,22 @@ class SCCFuseVerticalLoops(Transformation):
         Current heuristic: If the candidate is used in more than one vertical loop, assume it is NOT safe
         to demote!
         """
-        loop_var_map = {}
+        loop_var_map = CaseInsensitiveDict()
         with pragmas_attached(routine, ir.Loop):
             for loop in FindNodes(ir.Loop).visit(routine.body):
-                if loop.variable.name.lower() == self.vertical.index.lower():
-                    ignore = False
+                if loop.variable == self.vertical.index:
                     if is_loki_pragma(loop.pragma, starts_with='fused-loop'):
                         parameters = get_pragma_parameters(loop.pragma, starts_with='fused-loop')
                         group = parameters.get('group', 'default')
                         if group == 'ignore':
-                            ignore = True
-                    if not ignore:
-                        loop_var_map[loop] = as_tuple(var.name.lower() for var in FindVariables().visit(loop.body)
-                                if isinstance(var, sym.Array))
+                            continue
+                        for var in FindVariables().visit(loop.body):
+                            if isinstance(var, sym.Array):
+                                loop_var_map.setdefault(var.name, set()).add(group)
 
         safe_to_demote = ()
         for var in demote_candidates:
-            count = 0
-            for _, var_names in loop_var_map.items():
-                if var in var_names:
-                    count += 1
-            if count <= 1:
+            if var in loop_var_map and len(loop_var_map[var]) <= 1:
                 safe_to_demote += (var,)
 
         return safe_to_demote
@@ -127,12 +127,10 @@ class SCCFuseVerticalLoops(Transformation):
         Find local arrays/temporaries that do have the vertical dimension.
         """
         # local/temporary arrays
-        arrays = [var for var in FindVariables(unique=False).visit(routine.body) if isinstance(var, sym.Array)]
-        argument_names = [arg.name.lower() for arg in routine.arguments]
-        local_arrays = [arr for arr in arrays if arr.name.lower() not in argument_names]
+        local_arrays = get_local_arrays(routine, routine.body)
         # only those with the vertical size within shape
         relevant_local_arrays = [arr for arr in local_arrays if self.vertical.size.lower()
-                in [var.name.lower() for var in FindVariables().visit(arr.shape)]]
+                in FindVariables().visit(arr.shape)]
         # filter arrays to be ignored (for whatever reason)
         ignore_names = self.find_local_arrays_to_be_ignored(routine)
         if ignore_names:
@@ -153,8 +151,7 @@ class SCCFuseVerticalLoops(Transformation):
         # look for 'loki k-caching ignore(var1, var2, ...)' pragmas within routine and ignore those vars
         for pragma in pragmas:
             if is_loki_pragma(pragma, starts_with='k-caching'):
-                pragma_ignore = get_pragma_parameters(pragma, starts_with='k-caching').get('ignore', None)
-                if pragma_ignore:
+                if pragma_ignore := get_pragma_parameters(pragma, starts_with='k-caching').get('ignore', None):
                     ignore += as_tuple(v.strip() for v in pragma_ignore.split(','))
         ignore_names = set(var.lower() for var in ignore)
         return ignore_names
@@ -175,7 +172,7 @@ class SCCFuseVerticalLoops(Transformation):
 
     def correct_init_of_multilevel_arrays(self, routine, multilevel_local_arrays):
         """
-        Possibly handle initaliztion of those multilevel local arrays via
+        Possibly handle initialization of those multilevel local arrays via
         splitting relevant loops or rather creating a new node with the relevant
         nodes moved to the newly created loop.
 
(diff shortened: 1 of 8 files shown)


loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+        loop_var_map = {}
+        with pragmas_attached(routine, ir.Loop):
+            for loop in FindNodes(ir.Loop).visit(routine.body):
+                if loop.variable.name.lower() == self.vertical.index.lower():
+                    ignore = False
+                    if is_loki_pragma(loop.pragma, starts_with='fused-loop'):
+                        parameters = get_pragma_parameters(loop.pragma, starts_with='fused-loop')
+                        group = parameters.get('group', 'default')
+                        if group == 'ignore':
+                            ignore = True
```
mlange05, line 109: Instead of tracking `ignore`, I think you could simply `continue` here, no?

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+                    if not ignore:
+                        loop_var_map[loop] = as_tuple(var.name.lower() for var in FindVariables().visit(loop.body)
```
mlange05, line 111: Would building the inverse of this map not make the logic much simple? As in `var_loop_map[var.name] = loop` would allow you to then simply check for `len(var_loop_map[var.name]) > 1` in the next part? Or am I misreading this? 

Also, a `CaseInsensitiveDict` might be useful here for `var_loop_map`.

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+        return safe_to_demote
+
+    def find_relevant_local_arrays(self, routine):
+        """
+        Find local arrays/temporaries that do have the vertical dimension.
+        """
+        # local/temporary arrays
+        arrays = [var for var in FindVariables(unique=False).visit(routine.body) if isinstance(var, sym.Array)]
+        argument_names = [arg.name.lower() for arg in routine.arguments]
+        local_arrays = [arr for arr in arrays if arr.name.lower() not in argument_names]
```
mlange05, line 132: There's a common utility for this now: https://github.com/ecmwf-ifs/loki/blob/main/loki/transformations/utilities.py#L628

Could this be used here?

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+        # only those with the vertical size within shape
+        relevant_local_arrays = [arr for arr in local_arrays if self.vertical.size.lower()
```
mlange05, line 134: Again, please check if expr-string comparison works and if you can remove the `.lower()`. Should be easy to amend this in one of the tests.

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+
+        * explicitly marked to be ignored via pragmas within the input source file, e.g.,
+          'loki k-caching ignore(var1, var2, ...)'
+        """
+        ignore = ()
+        pragmas = FindNodes(ir.Pragma).visit(routine.body)
+        # look for 'loki k-caching ignore(var1, var2, ...)' pragmas within routine and ignore those vars
+        for pragma in pragmas:
+            if is_loki_pragma(pragma, starts_with='k-caching'):
+                pragma_ignore = get_pragma_parameters(pragma, starts_with='k-caching').get('ignore', None)
```
mlange05, line 156: Minor nitpick: I wonder if the `is_loki_pragma` and `get_pragma_parameters` can be combined to something like `if pragma_ignore := get_pragma_parameters(pragma, starts_with='k-caching').get(...)`. Not strictly necessary, just wondering...

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+            for dim in arr.dimensions:
+                if self.vertical.index in FindVariables().visit(dim):
+                    # dim is not equal to vertical.index e.g., vertical.index +/- 1
+                    if dim != self.vertical.index:
+                        multilevel_local_arrays.append(arr)
+        return multilevel_local_arrays
+
+    def correct_init_of_multilevel_arrays(self, routine, multilevel_local_arrays):
+        """
+        Possibly handle initaliztion of those multilevel local arrays via
```
mlange05, line 178: typo: "initialization"

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+        splitting relevant loops or rather creating a new node with the relevant
+        nodes moved to the newly created loop.
+
+        .. note::
+            This relies on pragmas being inserted in the input source code!
+        """
+        loop_map = {}
+        # find/identify loops with pragma 'loop-fusion group(<group-name>-init)'
+        with pragmas_attached(routine, ir.Loop):
```
mlange05, line 187: Should we make `pragmas_attached` a prerequisite for all the utilities and do it once, in `process_kernel`, maybe? If so, this should be mentioned in the utility method docstrings though.

loki/transformations/single_column/vertical.py
```diff
@@ -0,0 +1,218 @@
...
+                            else:
+                                node_map_init[node] = None
+                        # split the loop/create a new node to move those nodes with
+                        # multilevel arrays to the new node
+                        if nodes_to_be_moved:
+                            pragmas = loop.pragma
+                            new_pragmas = [pragma.clone(content=pragma.content.replace('-init', '')) if '-init'
+                                    in pragma.content else pragma for pragma in pragmas]
+                            loop_map[loop] = (ir.Comment('! Loki generated loop for init ...'),
+                                    Transformer(node_map_init).visit(loop.clone(\
```
mlange05, line 213: Two inline `Transformer` invocations makes this very hard to read. Can we apply them in sequence to build the tuple, before putting them into the `loop_map`?

loki/transformations/transform_loop.py
```diff
@@ -294,140 +294,145 @@ def loop_fusion(routine):
...
-                    conditions += [sym.Comparison(variable, '>=', loop_range.start)]
-                if symbolic_op(loop_range.stop, op.ne, fusion_range.stop):
-                    conditions += [sym.Comparison(variable, '<=', loop_range.stop)]
-            if conditions:
-                if len(conditions) == 1:
-                    condition = conditions[0]
-                else:
-                    condition = sym.LogicalAnd(as_tuple(conditions))
-                body = Conditional(condition=condition, body=as_tuple(body), else_body=())
+        if not fusion_groups:
```
mlange05, line 297: [no action] This entire chunk looks like it's just been unindented. Is that right, or are there any other changes in here? Tests look fine, so I'm not worried, just curious.

mlange05 (changes requested):
Overall this looks really great, and is a first step towards some very advanced optimisations. I've left a few comments, but they are mostly cosmetic. Once addressed, I think this should be good to go.

The direct inclusion in SCC also ok for now, as it will basically become a no-op unless any pragma markers are inserted. In the future, we might want to add this stuff to a more generic non-SCC loop-change transformation that can be prepended to SCC in config, or run in it's own right (eg. for CPU opts). But that is beyond this PR.

loki/transformations/single_column/vertical.py
MichaelSt98: Not sure what you mean.
I need to apply the Transformer twice with two different maps since the key would be overwritten otherwise ...
If you propose to introduce temporary variables: sure this is possible!

loki/transformations/single_column/vertical.py
mlange05: Yeah, that's true too. As the `Transformer` object behind them are reasonably fast, I'm happy to keep this as is (and as you say, it probably won't stay like this for long).

loki/transformations/single_column/vertical.py
mlange05: Oh, I just meant the cosmetics of it - the use of two transformers is fine, it just gets really hard to read if everything is applied in a single line. If you build the tuple incrementally, each `Transformer.visit(..)` invocation could be a single line - but this is really just a cosmetic suggestion and not strictly necessary.

MichaelSt98:
Thanks for reviewing @mlange05. I addressed all your requested changes.
I am thinking about introducing a flag to activate or deactivate that transformation. It is not completely true that this is a no-op unless any pragma markers are inserted. Without pragma markers no loop-interchange and loop-fusion is performed, however, it would still look for temporaries that could be demoted (so in principle this transformation could optimise even without inserted pragmas if some temporaries unnecessarily have a vertical dimension). I don't know how much time it would take to execute this performance for e.g., ecwam or ecphys and/or whether the result is correct. Therefore, I should at least check whether it works with ecwam/ecphys and/or introduce a flag to activate or deactivate. What do you think?

mlange05:
Ok, nitpicks and code stye comments have all been addressed. Many thanks for this and over to @reuterbal for final approval.

As for the on/off switch, I see you argument, and I think having a safety switch built in is a good idea. Would you mind adding this please to the constructor? I'll check offline if having it enabled affects EC-physics. Once that switch is added, I think this is ready.

On a wider note, I think that once we refactor this a little, we can have the loop-pragma steps as a separate "pre-processing Trafo" that can be added to the pipeline in the config files, rather than be part of SCC. Demotion of the vertical would then simply require additional parameters to the SCCDemote step - but all of this if beyond this initial PR.

reuterbal (changes requested):
Thanks for a very comprehensive and nicely structured implementation and the round of review and fixes already performed! This looks already very good to me. I've left a few (minor) questions and suggestions, and a comment for one potential source of problems.

loki/transformations/transform_loop.py
```diff
@@ -294,140 +294,145 @@ def loop_fusion(routine):
...
+
+        # Merge loops in each group and put them in the position of the group's first loop
+        #  UNLESS 'insert' location is specified for at least one of the group's fusion
+        #  pragmas, in this case the position is the first occurence of 'insert' for each group
+        for group, loop_parameter_lists in fusion_groups.items():
+            loop_list, parameters = zip(*loop_parameter_lists)
+
+            # First, determine the collapse depth and extract user-annotated loop ranges from pragmas
+            collapse = [param.get('collapse', None) for param in parameters]
+            insert_locs = [param.get('insert', False) for param in parameters]
```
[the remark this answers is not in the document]
MichaelSt98: Hmm `anchor` is probably a good alternative or `insert-loc`/`insert-location`?!

loki/transformations/transform_loop.py
MichaelSt98: I am a little bit confused to be honest.
It used to be that loop fusion (and early exit) was outside of `pragmas_attached`. Currently, it is within `pragmas_attached`.
If I put the loop fusion outside the `pragmas_attached` (through removing indentation) I get spurious Loki pragmas in the transformed code, but like it is now I don't get any `!$loki loop-fusion group ...` in the transformed source?!

loki/transformations/transform_loop.py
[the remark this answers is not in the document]
reuterbal: `insert-loc` is also good! I'll leave the decision to you (including whether you want to change it in the first place)

loki/transformations/transform_loop.py
```diff
@@ -294,140 +294,145 @@ def loop_fusion(routine):
...
-            fusion_bodies += [body]
-
-        # Create the nested fused loop and replace original loops
-        fusion_loop = flatten(fusion_bodies)
-        for fusion_variable, fusion_range in zip(reversed(fusion_variables), reversed(fusion_ranges)):
-            fusion_loop = Loop(variable=fusion_variable, body=as_tuple(fusion_loop), bounds=fusion_range)
+                    if len(lower_bounds) == 1:
+                        lower_bounds = lower_bounds[0]
+                    else:
+                        fct_symbol = sym.DeferredTypeSymbol(name='min', scope=routine)
```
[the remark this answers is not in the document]
MichaelSt98: Reason being when applying this to CLOUDSC this results in:

```
...
    raise RuntimeError(f'Procedure {item_name} not found in self.item_cache.')
RuntimeError: Procedure #min not found in self.item_cache.
```

Further, executing

```py
from loki import Subroutine, FindInlineCalls

fcode_kernel = """
SUBROUTINE kernel()
INTEGER :: A
A = MIN(0, 1)
END SUBROUTINE kernel
"""

kernel = Subroutine.from_source(fcode_kernel)
inline_calls = FindInlineCalls().visit(kernel.body)
for inline_call in inline_calls:
    print(f"inline_call: {inline_call} | function: {inline_call.function} ({inline_call.function.type})")
```

yields:

```
inline_call: MIN(0, 1) | function: MIN (<SymbolAttributes BasicType.DEFERRED>)
```

or similarly:

```py
>>> from loki.expression.parser import parse_expr
>>> expr = parse_expr('min(0, 1)')
>>> expr.function
DeferredTypeSymbol('min', None, None, <SymbolAttributes BasicType.DEFERRED>, False)
```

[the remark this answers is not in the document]
reuterbal: Well, that's two separate things. 
1. You introduce a function call but in CLOUDSC we seem to also tell the Scheduler to be strict and fail for any items it cannot find in its cache. So, I'd say works as expected?
2. When parsing Fortran code we don't inject procedure information for any functions provided by the standard. It's a good but unrelated question whether we should start doing that.

Regardless of the above, we do know that the symbol here is a `ProcedureSymbol` and I would argue we should therefore mark the symbol accordingly?

The question is what we do about the above. Would we want to introduce a lookup mechanism for functions defined by the standard? Fparser would give us that information, at least for F2003, and we would only need to be careful to not assign that if intfb imports exist.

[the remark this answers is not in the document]
MichaelSt98: I don't think that's two separate things, I'd say like this it is consistent as 

> When parsing Fortran code we don't inject procedure information for any functions provided by the standard ...

so why inject procedure information if I know that `min`/`max` is provided by the standard?

Of course like you said we could start doing that in general and introduce a lookup mechanism for intrinsics.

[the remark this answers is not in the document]
MichaelSt98: Anyway, I am happy to change it back to `ProcedureSymbol`.
Only question I have: just disable `strict` for CLOUDSC to make it work again or how to resolve that?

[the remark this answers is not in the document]
reuterbal: Hmm, on second thought the issue is unrelated to the scope of this PR. So, I'd say let's not expand this unnecessarily and instead:
* Create an issue about the provision of type information for functions 
* leave it as DeferredTypeSymbol here as it is consistent with what the frontends produce. Ideally, add a comment next to each, referencing the issue
That allows to keep CLOUDSC as is and unblocks this PR. Sound OK?

mlange05 (approved):
Looking great now, and confirmed safe and fast with ecphys-regression. GTG. :shipit:

PR #375, opened 2024-09-10
IR: Symbol management on scoped nodes

mlange05:
~_Note: This PR sits on top of #372, so I'm only filing a draft before anticipated rebase._~

This PR brings a set of convenience methods and expression constructors to the `ScopedNode` and `ProgramUnit` classes. The idea is that eventually, `Subroutine` and `Module` classes should become scoped nodes and the expression management API should be equivalent between them.

The idea is to add the following three utilities to the scope objects themselves, to enable quick(er) expression generation:
* `.get_symbol(name)` - utility to look up a symbol by it's name. This will first find it's true owning scope and attach the symbol.
* `.Variable(name, ....)` - a shortcut to the general `Variable` constructor that will attempt to build array and scalar symbols, depending on available type information
* `.parse_expr` - a shortcut to the general expression parser that can generate nested expression trees from pure strings.

### DIFF
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -22,10 +22,11 @@ from pymbolic.primitives import Expression
 from pydantic.dataclasses import dataclass as dataclass_validated
 from pydantic import model_validator
 
+from loki.expression import Variable
+from loki.frontend.source import Source
 from loki.scope import Scope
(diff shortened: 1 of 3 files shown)


reuterbal (approved):
Many thanks, this is fantastic!

(And I'm already dreading the follow-on PR that applies the strict Scope-local construction of symbols throughout... ;-)

:shipit:

PR #376, opened 2024-09-13
Transformations: Re-organise `inline` and `extract` sub-packages

mlange05:
This change creates two new separate transformation sub-packages `loki.transformations.inline` and `loki.transformations.extract` that are conceptual complements. For the inline sub-package, this allows us to separate the different types of inlining (constants, small functions and full procedures), and the associated tests. For "extract", this change brings together the two disparate current implementations (formerly `extract_contained` and `region_to_call`), and provides the counter-passes to the respective inlining operations as `extract_internals` and `extract_marked`. I've also added a wrapper `ExtractTransformation` class that provides both functionalities for batch processing. 

Please note that, other than a minor bug-fix, I've left the implementations untouched. Further consolidation and additional features are planned, but I wanted to get the diff-heavy stuff out of the way first.

In some more detail:
* Created separate `loki.transformations.inline` sub-package with `constants`/`functions`/`mapper`/`procedures` modules. Importantly, I also split the test accordingly and localised it!
* Renamed `extract_contained` to `extract_internals` for consistency and created new `loki.transformations.extract` sub-package.
* Renamed `region_to_call` to `extract_marked` and moved to `loki.transformations.extract`.
* Added a new `ExtractTransformation` as batch-processing entry point. Importantly, this works on `Module` and `Sourcefile` objects only, as that defines where the newly created procedures go. Also added tests for this
* A small bug-fix for `extract_internals` and a few linter-compliance fixes for the transformation standalone tests. (Turns out pylint does not recurse into test directories, as there is no automatic package import 😉 )


loki/transformations/tests/test_parametrise.py
```diff
@@ -356,7 +356,7 @@ def stop_execution(**kwargs):
 
     abort_callbacks = (error_stop, stop_execution)
 
-    for i, abort_callback in enumerate(abort_callbacks):
+    for _, abort_callback in enumerate(abort_callbacks):
```
reuterbal, line 359: Would removing the `enumerate` not be the simpler fix?

loki/transformations/extract/marked.py
```diff
@@ -0,0 +1,132 @@
...
+)
+from loki.logging import info
+from loki.subroutine import Subroutine
+from loki.tools import as_tuple, CaseInsensitiveDict
+
+
+__all__ = ['extract_marked_subroutines']
+
+
+def extract_marked_subroutines(routine):
```
reuterbal, line 23: I find the naming a bit counterintuitive. 

1. "extract_marked_subroutines" implies to me that subroutines with some kind of marker are extracted from somewhere. Wouldn't something that makes it clear that this acts on pragma regions be more appropriated?
2. Any particular reason for calling this "extract" instead of outlining, which seems to be the more common term for this in compiler theory? Would also more directly highlight the complementary nature to inlining. The module could still be called "extract".

reuterbal (commented):
Thanks, this is great! Makes for a much cleaner separation of functionality and the clean-up on the side in tests is much appreciated.

I left two minor comments: very minor nitpick and a question about naming.

loki/transformations/extract/marked.py
mlange05: Yes, agreed. I simply mimicked what was there, but calling it "outlining" is indeed clearer. So, we would have `loki.transformations.extract.internals` and `loki.transformations.extract.outline`, where the latter provides `outline_pragma_regions`?

mlange05: Should we rename the pragmas then too? `!$loki outline` / `!$loki end outline`?

reuterbal: Yes, your suggested module naming sounds good to me.
I hadn't thought about the pragmas, but your suggestion does indeed make sense. Reading that in a Fortran source file also makes (imho) immediately clear what is going to happen there.

mlange05:
@reuterbal Ok, I've rebased this over recent master and renamed the outline method, flags and pragma names. The rebase was quite tricky (esp. splitting the new `inline` sub-package), so I'd appreciate a quick double-check that everything is still as it's supposed to be.

loki/transformations/inline/functions.py
```diff
@@ -0,0 +1,342 @@
...
+
+    # Find and filter inline calls and corresponding nodes
+    function_calls = {}
+    # Find inline calls but skip/ignore inline calls being parameters of other inline calls
+    #  to ensure correct ordering of inlining. Those skipped/ignored inline calls will be handled
+    #  in the next call to this function.
+    retriever = ExpressionRetrieverSkipInlineCallParameters(lambda e: isinstance(e, sym.InlineCall),
+            inline_elementals_only=inline_elementals_only, functions=functions)
+    # override retriever ...
+    FindInlineCallsSkipInlineCallParameters.retriever = retriever
```
reuterbal, line 141: I must have missed this before: This shouldn't be necessary, since the `FindInlineCallsSkipInlineCallParameters` is declared inside the function, we should be able to directly instantiate the retriever with the provided values for `inline_elementals_only` and `functions`.

reuterbal (commented):
Thanks, I can't spot anything that may have come from the rebasing.

While looking through the changes again, I spotted an outdated docstring and a potentially unnecessary way of instantiating a visitor. Both don't come from this PR but maybe it's easy to fix them here?

Also, I've noticed that the `InlineSubstitutionMapper` is very much undertested (https://app.codecov.io/gh/ecmwf-ifs/loki/pull/376/blob/loki/transformations/inline/mapper.py), probably because it is only being used when inlining statement functions (and therefore the other code parts are never hit). I'm wondering if it would be easy to expand the test coverage, e.g., by adding a test with a statement function that uses a scalar and an inline call? I've left a suggestion that uses the existing test and modifies it accordingly, but have not actually tested that this works as expected.

loki/transformations/inline/tests/test_functions.py
```diff
@@ -0,0 +1,241 @@
...
+    assert not FindNodes(ir.StatementFunction).visit(routine.spec)
+    if stmt_decls:
+        assert not FindInlineCalls().visit(routine.body)
+        assignments = FindNodes(ir.Assignment).visit(routine.body)
+        assert assignments[0].lhs  == 'ret'
+        assert assignments[0].rhs  ==  "arr + arr + 1.0"
+        assert assignments[1].lhs  == 'ret2'
+        assert assignments[1].rhs  ==  "3.0 + 1.0"
+    else:
+        assert FindInlineCalls().visit(routine.body)
```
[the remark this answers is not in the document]
mlange05: Sorry, @reuterbal I had pushed this in quickly, but it turns out the tests keeps triggering on some finer details of the function inliner, that seems way beyond this PR. I've stripped the commit back out in the last push, and I propose to file this as a separate bug issue or bug-fix PR, as this thing is getting quite convoluted now.

loki/transformations/inline/functions.py
mlange05: Yes, now I understand. Pushed a fix to this extend.

loki/transformations/inline/tests/test_functions.py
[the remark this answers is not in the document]
reuterbal: That seems reasonable, many thanks for checking. I will file an issue

reuterbal (approved):
Many thanks!

Issue #377, opened 2024-09-19
Function inlining


MichaelSt98:
Currently only (single-line) elemental functions are supported.

Currently, this

```fortran
  elemental function multiply(a, b)
    real(kind=real64) :: multiply
    real(kind=real64), intent(in) :: a, b

    multiply = a * b
  end function multiply
```

can be inlined correctly. However, something like

```fortran
  elemental function multiply(a, b)
    real(kind=real64) :: multiply
    real(kind=real64), intent(in) :: a, b
    real(kind=real64) :: temp

    temp = a * b
    multiply = temp
  end function multiply
```

will be inlined but **not correctly!** (since only `multiply = temp` gets inlined).

Further, inlining of member functions is not supported as well as functions that have return value based on a conditional (assignment)

```fortran
  elemental function multiply_if(a, b)
    real(kind=real64) :: multiply_if
    real(kind=real64), intent(in) :: a, b

    if (a .gt. b) then
      multiply_if = a * b
    else 
      multiply_if = a * a
    endif
  end function multiply_if
```

PR #378, opened 2024-09-23
Inline functions (including multi-line and non-elemental functions)

MichaelSt98:
Extending and improving Inline utilities as discussed in #377

### DIFF
--- a/loki/transformations/tests/test_inline.py
+++ b/loki/transformations/tests/test_inline.py
@@ -466,13 +466,14 @@ end subroutine member_routines
 
 
 @pytest.mark.parametrize('frontend', available_frontends(skip={
-    OMNI: "OMNI can't parse the code"}))
+    OMNI: "OMNI has a problem with function return variables being arrays!"}))
 def test_inline_member_functions(tmp_path, frontend):
     """
     Test inlining of member subroutines.
     """
     fcode = """
 subroutine member_functions(a, b, c)
+  implicit none
   real(kind=8), intent(inout) :: a(3), b(3), c(3)
   integer :: i
 
--- a/loki/transformations/tests/test_inline.py
+++ b/loki/transformations/tests/test_inline.py
@@ -49,8 +49,11 @@ contains
   elemental function multiply(a, b)
     real(kind=real64) :: multiply
     real(kind=real64), intent(in) :: a, b
+    real(kind=real64) :: temp
 
-    multiply = a * b
+    ! simulate multi-line function
+    temp = a * b
+    multiply = temp
   end function multiply
 end module multiply_mod
 """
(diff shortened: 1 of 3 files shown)


MichaelSt98:
The whole thing got a bit out of hand again and turned out to be a rabbit hole ...

Currently (or rather with this PR), nested functions are handled via an iterative approach. We could try to solve this recursively instead. However, I don't think this would be straight-forward and questionable whether this is beneficial. Anyway, I am open for discussing that and other implementation details.

Moreover, it would be possible to check whether functions are single-line (or could be reduced to a single line) and inline those at an expression level.

mlange05 (approved):
Looks great, and is very, very useful! Many thanks, this one has been on the todo list for a while! 🙏 🙏 🙏 🙏 🙏 🙏  GTG :shipit:

loki/transformations/tests/test_inline.py
```diff
@@ -367,6 +465,81 @@ def test_inline_member_routines(tmp_path, frontend):
     assert (b == [3., 3., 3.]).all()
 
 
+@pytest.mark.parametrize('frontend', available_frontends(skip={
+    OMNI: "OMNI can't parse the code"}))
```
reuterbal, line 469: Why? I can't see anything that would usually upset OMNI, and a quick parse with `F_Front` was successful for me.

loki/transformations/tests/test_inline.py
```diff
@@ -367,6 +465,81 @@ def test_inline_member_routines(tmp_path, frontend):
...
+def test_inline_member_functions(tmp_path, frontend):
+    """
+    Test inlining of member subroutines.
+    """
+    fcode = """
+subroutine member_functions(a, b, c)
```
reuterbal, line 475:
```suggestion
subroutine member_functions(a, b, c)
  implicit none
```

reuterbal (changes requested):
Very nicely done, this is a really fantastic addition and nicely executed!

A bunch of nitpick comments but no showstoppers.

loki/transformations/tests/test_inline.py
MichaelSt98: If you execute

```py
from loki import Subroutine, fgen, OMNI, FP
fcode = """
subroutine member_functions
    implicit none
    integer :: i
    real(kind=8) :: a(3)
    contains
    function add_to_a(b, n)
      integer, intent(in) :: n
      real(kind=8), intent(in) :: b(n)
      real(kind=8) :: add_to_a(n)

      do i = 1, n
        add_to_a(i) = a(i) + b(i)
      end do
    end function
end subroutine member_functions
    """
routine = Subroutine.from_source(fcode, frontend=OMNI)
print(fgen(routine))
```

```fortran
SUBROUTINE member_functions ()
  IMPLICIT NONE
  INTEGER :: i
  REAL(KIND=8) :: a(3)
  CONTAINS
  FUNCTION add_to_a (b, n)
    IMPLICIT NONE
    INTEGER, INTENT(IN) :: n
    REAL(KIND=8), INTENT(IN) :: b(n)
    ! HERE: MISSING DIMENSION !!!
    REAL(KIND=8) :: add_to_a
    DO i=1,n
      add_to_a(i) = a(i) + b(i)
    END DO
  END FUNCTION add_to_a
END SUBROUTINE member_functions
```

Same thing for `real(kind=8), dimension(n) :: add_to_a`. This doesn't happen for `FP` and `OFP`.

reuterbal: #391 should fix the problem

reuterbal (approved):
I have rebased and removed the pylintrc commit. Ready from my side now, many thanks again!

PR #382, opened 2024-09-24
Fix edge case for vector section mapping

### LINKED ISSUE #380: Edge case: vector section mapping conflicts with (rebuild of) Conditional
Reproducer:

```py
from loki import Subroutine, Dimension, SCCDevectorTransformation

fcode_kernel = """
SUBROUTINE kernel(start, end, nlon, q, flag1, flag2)
INTEGER, INTENT(IN) :: start, end
INTEGER, INTENT(IN) :: nlon
REAL, INTENT(INOUT) :: q(nlon)
INTEGER :: jl, i
LOGICAL, INTENT(IN) :: FLAG1, FLAG2

  IF (FLAG1) THEN
    call some_nested_kernel(q)
  ELSEIF (FLAG2) THEN
    DO jl = start, end
      q(jl, jk) = q(jl, jk-1)
    END DO
  ENDIF
END SUBROUTINE kernel
"""

kernel = Subroutine.from_source(fcode_kernel)
horizontal = Dimension(name='horizontal', index='jl', bounds=('start', 'end'), size='nlon')
SCCDevectorTransformation(horizontal=horizontal, trim_vector_sections=True).process_kernel(kernel)
```

Executing `LOKI_DEBUG=1 python3 reproducer.py` yields

```
...
    routine.body = NestedTransformer(section_mapper).visit(routine.body)
...
pydantic_core._pydantic_core.ValidationError: 1 validation error for Conditional
  Assertion failed,  [type=assertion_error, input_value=ArgsKwargs((), {'conditio...': None, 'label': None}), input_type=ArgsKwargs]
-> s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
```

which is due to

```
-> s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
(Pdb) p kwargs['else_body'][0]
Section::
(Pdb) p kwargs['else_body'][0].body[0]
Conditional::
```

being in conflict with Loki's `Conditional` post init assertion `assert isinstance(self.else_body[0], Conditional)`!

Thus, this bug can be resolved by

* either changing the assertion to allow the first element of the `else_body` to be a `Section` (and the first element of it's body to be a `Conditional`)
* or changing the logic of extracting the *vector sections*

MichaelSt98:
conflicting with (rebuild of) Conditional.

Fixes #380

### DIFF
--- a/loki/transformations/single_column/vector.py
+++ b/loki/transformations/single_column/vector.py
@@ -137,7 +137,9 @@ class SCCDevectorTransformation(Transformation):
                 subsec_body = cls.extract_vector_sections(separator.body, horizontal)
                 if subsec_body:
                     subsections += subsec_body
-                subsec_else = cls.extract_vector_sections(separator.else_body, horizontal)
+                subsec_else = cls.extract_vector_sections(separator.else_body[0].body, horizontal)\
+                        if separator.else_body and isinstance(separator.else_body[0], ir.Conditional)\
+                        else cls.extract_vector_sections(separator.else_body, horizontal)
                 if subsec_else:
                     subsections += subsec_else
 
--- a/loki/transformations/single_column/vector.py
+++ b/loki/transformations/single_column/vector.py
@@ -137,6 +137,9 @@ class SCCDevectorTransformation(Transformation):
                 subsec_body = cls.extract_vector_sections(separator.body, horizontal)
                 if subsec_body:
                     subsections += subsec_body
+                # we need to prevent that the whole 'else_body' is wrapped in a section,
+                # as 'Conditional's rely on the fact that the first element of the 'else_body'
+                # (if 'has_elseif') is a Conditional itself
                 subsec_else = cls.extract_vector_sections(separator.else_body[0].body, horizontal)\
                         if separator.else_body and isinstance(separator.else_body[0], ir.Conditional)\
                         else cls.extract_vector_sections(separator.else_body, horizontal)
--- a/loki/transformations/single_column/vector.py
+++ b/loki/transformations/single_column/vector.py
@@ -140,9 +140,10 @@ class SCCDevectorTransformation(Transformation):
                 # we need to prevent that the whole 'else_body' is wrapped in a section,
                 # as 'Conditional's rely on the fact that the first element of the 'else_body'
                 # (if 'has_elseif') is a Conditional itself
-                subsec_else = cls.extract_vector_sections(separator.else_body[0].body, horizontal)\
-                        if separator.else_body and isinstance(separator.else_body[0], ir.Conditional)\
-                        else cls.extract_vector_sections(separator.else_body, horizontal)
+                if separator.has_elseif and separator.else_body:
+                    subsec_else = cls.extract_vector_sections(separator.else_body[0].body, horizontal)
+                else:
+                    subsec_else = cls.extract_vector_sections(separator.else_body, horizontal)
                 if subsec_else:
                     subsections += subsec_else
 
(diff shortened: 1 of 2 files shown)


loki/transformations/single_column/vector.py
```diff
@@ -138,7 +138,12 @@ def extract_vector_sections(cls, section, horizontal):
                 subsec_body = cls.extract_vector_sections(separator.body, horizontal)
                 if subsec_body:
                     subsections += subsec_body
-                subsec_else = cls.extract_vector_sections(separator.else_body, horizontal)
+                # we need to prevent that the whole 'else_body' is wrapped in a section,
+                # as 'Conditional's rely on the fact that the first element of the 'else_body'
+                # (if 'has_elseif') is a Conditional itself
+                subsec_else = cls.extract_vector_sections(separator.else_body[0].body, horizontal)\
+                        if separator.else_body and isinstance(separator.else_body[0], ir.Conditional)\
+                        else cls.extract_vector_sections(separator.else_body, horizontal)
```
reuterbal, line 146: Note that this is only required if the `Conditional` actually has `has_elseif == True`.
The following two version will create the same IR, with the subtle difference that the outer conditional has `has_elseif == True`
```
IF (A > 0) THEN
  A = A + 1
ELSE
  IF (A < 0) THEN
    A = A - 1
  ENDIF
ENDIF
```
vs.
```
IF (A > 0) THEN
  A = A + 1
ELSEIF (A < 0) THEN
  A = A - 1
ENDIF
```

For readability, I would also prefer to use proper if/else blocks rather than putting everything in a single line with line breaks. I would suggest rewriting this as
```suggestion
                if separator.has_elseif and separator.else_body:
                    subsec_else = cls.extract_vector_sections(separator.else_body[0].body, horizontal)
                else:
                    subsec_else = cls.extract_vector_sections(separator.else_body, horizontal)
```

reuterbal (changes requested):
Nice find! The fix is correct but I think we need to distinguish between the case where this is an actual else-if and where this is just an if nested into an else.

loki/transformations/single_column/vector.py
MichaelSt98: Yes! There is just a problem with OMNI as the parsing of Conditionals is a bit different to FP and OFP and `has_elseif` is always `False` for OMNI. Thus I could do what you suggest and disable OMNI or "fix" Conditional parsing/construction for OMNI.
What do you think?

mlange05: I don't think there is much you can do to "fix OMNI" in this case, as it automatically expands every `ELSEIF` into a nested `ELSE => IF` code structure in its frontend, even before we get the AST. So, unless we reconstruct the correct ELSEIF by inspecting the source string, there is no catching this. Usually, I'd say this leaves only special-casing the test, but since this quickly becomes cumbersome, I think we can just disable this test (with an explicit mention of OMNI's elseif-expansion).

reuterbal: Agreed! The transformation's behaviour should be governed by the IR it receives, and not try to restore what OMNI has expanded previously. Importantly, OMNI does not change the meaning, simply the appearance. So, disabling the test for Omni (or hacking the IR by setting the flag before calling the transformation if you think that's better) seems sensible to me.

reuterbal (approved):
Thanks, looks great now! Will also do a rebase for sanity.

mlange05 (approved):
Looks good. GTG. :shipit:

PR #383, opened 2024-09-25
handle modulo operator/function for c-like-backends

MichaelSt98:
[TRANSPILATION] handle modulo operator/function for c-like-backends ...

convert

* `MOD(a, b)` to `a % b` for integer parameters/arguments
* `MOD(x, y)` to `fmod(x, y)` for floating point parameters/arguments

### DIFF
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -11,11 +11,13 @@ from pymbolic.mapper.stringifier import (
 )
 
 from loki.tools import as_tuple
-from loki.ir import Import, Stringifier, FindNodes
+from loki.ir import (
+        Import, Stringifier, FindNodes,
+        FindVariables, ExpressionFinder
+)
 from loki.expression import (
         LokiStringifyMapper, Array, symbolic_op, Literal,
-        symbols as sym, FindVariables, ExpressionFinder,
-        ExpressionRetriever
+        symbols as sym, ExpressionRetriever
 )
 from loki.types import BasicType, SymbolAttributes, DerivedType
 
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -13,11 +13,11 @@ from pymbolic.mapper.stringifier import (
 from loki.tools import as_tuple
 from loki.ir import (
         Import, Stringifier, FindNodes,
-        FindVariables, ExpressionFinder
+        FindVariables, FindRealLiterals
 )
 from loki.expression import (
         LokiStringifyMapper, Array, symbolic_op, Literal,
-        symbols as sym, ExpressionRetriever
+        symbols as sym
 )
 from loki.types import BasicType, SymbolAttributes, DerivedType
 
@@ -145,9 +145,6 @@ class CCodeMapper(LokiStringifyMapper):
 
     def map_inline_call(self, expr, enclosing_prec, *args, **kwargs):
 
-        class FindFloatLiterals(ExpressionFinder):
-            retriever = ExpressionRetriever(lambda e: isinstance(e, sym.FloatLiteral))
-
         if expr.function.name.lower() == 'mod':
             parameters = [self.rec(param, PREC_NONE, *args, **kwargs) for param in expr.parameters]
             # TODO: this check is not quite correct, as it should evaluate the
@@ -157,7 +154,7 @@ class CCodeMapper(LokiStringifyMapper):
             #  as an example: 'celing(3.1415)' got an floating point value in it, however it evaluates/returns
             #  an integer, in that case the wrong modulo function/operation is chosen
             if any(var.type.dtype != BasicType.INTEGER for var in FindVariables().visit(expr.parameters)) or\
-                    FindFloatLiterals().visit(expr.parameters):
+                    FindRealLiterals().visit(expr.parameters):
                 return f'fmod({parameters[0]}, {parameters[1]})'
             return f'({parameters[0]})%({parameters[1]})'
         return super().map_inline_call(expr, enclosing_prec, *args, **kwargs)
--- a/loki/backend/cgen.py
+++ b/loki/backend/cgen.py
@@ -14,7 +14,8 @@ from loki.tools import as_tuple
 from loki.ir import Import, Stringifier, FindNodes
 from loki.expression import (
         LokiStringifyMapper, Array, symbolic_op, Literal,
-        symbols as sym
+        symbols as sym, FindVariables, ExpressionFinder,
+        ExpressionRetriever
 )
 from loki.types import BasicType, SymbolAttributes, DerivedType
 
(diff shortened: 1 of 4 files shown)


loki/backend/cgen.py
```diff
@@ -14,7 +14,8 @@
 from loki.ir import Import, Stringifier, FindNodes
 from loki.expression import (
         LokiStringifyMapper, Array, symbolic_op, Literal,
-        symbols as sym
+        symbols as sym, FindVariables, ExpressionFinder,
+        ExpressionRetriever
```
reuterbal, line 18: This will need an update after #372 has been merged

loki/transformations/transpile/tests/test_transpile.py
```diff
@@ -19,6 +19,8 @@
 from loki.transformations.transpile import FortranCTransformation
 from loki.transformations.single_column import SCCLowLevelHoist, SCCLowLevelParametrise
 
+# pylint: disable=too-many-lines
+
```
reuterbal, line 23: We should start splitting the tests into logical units (but that can be a separate PR)

loki/transformations/transpile/tests/test_transpile.py
```diff
@@ -1249,6 +1251,77 @@ def test_transpile_multiconditional_range(tmp_path, frontend):
...
+  implicit none
+  {dtype}{'(kind=real64)' if dtype == 'real' else ''}, intent(in) :: in
+  {dtype}{'(kind=real64)' if dtype == 'real' else ''}, intent(inout) :: out
+
+  if (mod(in{'+ 2._real64' if add_float else ''}, 2{'._real64' if dtype == 'real' else ''}{'+ 0._real64' if add_float else ''}) .eq. 0) then
+    out = 42{'._real64' if dtype == 'real' else ''}
+  else
+    out = 11{'._real64' if dtype == 'real' else ''}
+  endif
+end subroutine transpile_special_functions
```
reuterbal, line 1275: A suggestion for a slightly more readable code:

```suggestion
    if dtype == 'real':
        decl_type = f'{dtype}(kind=real64)'
        kind = f'._real64'
    else:
        decl_type = dtype
        kind = ''
     
    fcode = f"""
subroutine transpile_special_functions(in, out)
  use iso_fortran_env, only: real64
  implicit none
  {decl_type}, intent(in) :: in
  {decl_type}, intent(inout) :: out

  if (mod(in{'+ 2._real64' if add_float else ''}, 2{kind}{'+ 0._real64' if add_float else ''}) .eq. 0) then
    out = 42{kind}
  else
    out = 11{kind}
  endif
end subroutine transpile_special_functions
```

loki/transformations/transpile/tests/test_transpile.py
```diff
@@ -1249,6 +1251,77 @@ def test_transpile_multiconditional_range(tmp_path, frontend):
...
+        if dtype == 'real':
+            return np.float64([val])
+        return np.int_([val])
+
+    # for testing purposes
+    in_var = init_var(dtype) # np.float64([0]) # 0
+    test_vals = [2, 10, 5, 3]
+    expected_results = [42, 42, 11, 11]
+    # out_var = np.int_([0])
+    out_var = init_var(dtype) # np.float64([0])
```
reuterbal, line 1288: Can you remove the commented debug leftovers please?

loki/transformations/transpile/tests/test_transpile.py
```diff
@@ -1249,6 +1251,77 @@ def test_transpile_multiconditional_range(tmp_path, frontend):
...
+
+    clean_test(filepath)
+
+    # apply F2C trafo
+    f2c = FortranCTransformation()
+    f2c.apply(source=routine, path=tmp_path)
+
+    # check whether correct modulo was inserted
+    with open(f2c.c_path, 'r') as f:
+        ccode = f.read()
```
reuterbal, line 1308: FYI: pathlib provides these days a convenient shortcut
```suggestion
    ccode = pathlib.Path(f2c.c_path).read_text()
```

reuterbal (changes requested):
Thanks, I think this needs a rebase and subsequent import fix, the rest is just small nitpicks.

loki/backend/cgen.py
```diff
@@ -140,6 +143,25 @@ def map_c_reference(self, expr, enclosing_prec, *args, **kwargs):
     def map_c_dereference(self, expr, enclosing_prec, *args, **kwargs):
         return self.format(' (*%s)', self.rec(expr.expression, PREC_NONE, *args, **kwargs))
 
+    def map_inline_call(self, expr, enclosing_prec, *args, **kwargs):
+
+        class FindFloatLiterals(ExpressionFinder):
```
mlange05, line 148: Just a minor nitpick, but I think this is general enough to live in `loki.ir.expr_visitor.py`

loki/backend/cgen.py
```diff
@@ -140,6 +143,25 @@ def map_c_reference(self, expr, enclosing_prec, *args, **kwargs):
...
+            retriever = ExpressionRetriever(lambda e: isinstance(e, sym.FloatLiteral))
+
+        if expr.function.name.lower() == 'mod':
+            parameters = [self.rec(param, PREC_NONE, *args, **kwargs) for param in expr.parameters]
+            # TODO: this check is not quite correct, as it should evaluate the
+            #  expression(s) of both arguments/parameters and choose the integer version of modulo ('%')
+            #  instead of the floating-point version ('fmod')
+            #  whenever the mentioned evaluations result in being of kind 'integer' ...
+            #  as an example: 'celing(3.1415)' got an floating point value in it, however it evaluates/returns
+            #  an integer, in that case the wrong modulo function/operation is chosen
```
mlange05, line 158: [no action] This is indeed a tricky corner case. Solution looks good for now, but might need some more thought when this gets problematic.

mlange05 (approved):
Looks great to me.

loki/backend/cgen.py
reuterbal: Likely the correct approach would be an `ExpressionTypeMapper` that attempts to determine the return type of an expression (similar to [`ExpressionDimensionsMapper`](https://github.com/ecmwf-ifs/loki/blob/cc179b3b86613af86df6dabb0be250cde4414fc9/loki/expression/mappers.py#L362)) - but it will require baking in a lot of knowledge about Fortran intrinsics. Fparser might provide some of that, so it could be feasible.

Definitely way beyond the scope of this PR, though.

reuterbal (commented):
Many thanks, looks good to me. 
@MichaelSt98: Since it'll take a little bit until the backlog of PRs is merged anyway, do you want to move the `FindFloatLiterals` to the expr_visitor module and add a small test? Let me know, otherwise I'll green-light and merge.

MichaelSt98:
@reuterbal Yes I'll move it to expr_visitor module and add a small test. Do we want to keep the name `FindFloatLiterals` or name it `FindRealLiterals`?

reuterbal:
Good point! Real is better

loki/backend/cgen.py
MichaelSt98: Yes! We could also think about implementing a C++ (templated) function `mymod()` which either calls `fmod` or uses the modulo operator in dependence of the arguments types.

reuterbal (approved):
That looks great now, many thanks!

PR #386, opened 2024-10-02
Extend 'resolve_vector_notation' to look for available and appropriate loops

MichaelSt98:
Imagine having:

```fortran
integer :: arr1(n), arr2(n)
arr1(:) = 0
do j=1,n
  arr2(j) = 0
enddo
```

Instead of transforming this to:

```fortran
integer :: arr1(n), arr2(n)
do i_arr1_0=1,n
  arr1(j) = 0
enddo
do j=1,n
  arr2(j) = 0
enddo
```

transform it to:

```fortran
integer :: arr1(n), arr2(n)
do j=1,n
  arr1(j) = 0
enddo
do j=1,n
  arr2(j) = 0
enddo
```

If there is no loop with appropriate bounds introduce new loop variable (like it was before in general).

### DIFF
--- a/loki/transformations/array_indexing.py
+++ b/loki/transformations/array_indexing.py
@@ -157,6 +157,10 @@ def resolve_vector_notation(routine):
     loop_map = {}
     index_vars = set()
     vmap = {}
+    # find available loops and create map {(lower, upper, step): loop_variable}
+    loops = FindNodes(Loop).visit(routine.body)
+    loop_map = {(loop.bounds.lower, loop.bounds.upper, loop.bounds.step if loop.bounds.step is not None else 1):
+                loop.variable for loop in loops}
     for stmt in FindNodes(Assignment).visit(routine.body):
         # Loop over all variables and replace them with loop indices
         vdims = []
@@ -176,6 +180,12 @@ def resolve_vector_notation(routine):
                     # Create new index variable
                     vtype = SymbolAttributes(BasicType.INTEGER)
                     ivar = sym.Variable(name=f'{ivar_basename}_{i}', type=vtype, scope=routine)
+                    # create tuple to test whether an appropriate loop is already available
+                    test_range = (sym.IntLiteral(1), s, 1) if not isinstance(s, sym.RangeIndex)\
+                            else (s.lower, s.upper, 1)
+                    # actually test for it
+                    if test_range in loop_map:
+                        ivar = loop_map[test_range]
                     shape_index_map[(i, s)] = ivar
                     index_range_map[ivar] = s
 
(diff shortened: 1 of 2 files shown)


loki/transformations/array_indexing.py
```diff
@@ -159,6 +159,10 @@ def resolve_vector_notation(routine):
     loop_map = {}
     index_vars = set()
     vmap = {}
+    # find available loops and create map {(lower, upper, step): loop_variable}
+    loops = FindNodes(Loop).visit(routine.body)
+    loop_map = {(loop.bounds.lower, loop.bounds.upper, loop.bounds.step if loop.bounds.step is not None else 1):
+                loop.variable for loop in loops}
```
reuterbal, line 165: slightly shorter that makes it fit into one line:
```suggestion
    loop_map = {(loop.bounds.lower, loop.bounds.upper, loop.bounds.step or 1): variable for loop in loops}
```

loki/transformations/array_indexing.py
```diff
@@ -178,6 +182,12 @@ def resolve_vector_notation(routine):
                     # Create new index variable
                     vtype = SymbolAttributes(BasicType.INTEGER)
                     ivar = sym.Variable(name=f'{ivar_basename}_{i}', type=vtype, scope=routine)
+                    # create tuple to test whether an appropriate loop is already available
+                    test_range = (sym.IntLiteral(1), s, 1) if not isinstance(s, sym.RangeIndex)\
+                            else (s.lower, s.upper, 1)
+                    # actually test for it
+                    if test_range in loop_map:
+                        ivar = loop_map[test_range]
```
reuterbal, line 190: We should test first before creating `ivar` and then overwriting it, because the instantiation creates already an entry in the symbol table:
```suggestion
                    # create tuple to test whether an appropriate loop is already available
                    test_range = (sym.IntLiteral(1), s, 1) if not isinstance(s, sym.RangeIndex)\
                            else (s.lower, s.upper, 1)
                    # actually test for it
                    if test_range in loop_map:
                        ivar = loop_map[test_range]
                    else:
                        vtype = SymbolAttributes(BasicType.INTEGER)
                        ivar = sym.Variable(name=f'{ivar_basename}_{i}', type=vtype, scope=routine)
```

reuterbal (changes requested):
Thanks, very nice improvement to the appearance of transformed code and ensuring it matches the naming conventions we rely on later.

I left two small remarks

reuterbal (approved):
Looks great, many thanks! I'll rebase to make sure we don't have any unexpected fallout and merge once green

mlange05 (approved):
Looks fantastic and is super useful! Many thanks, GTG! :shipit:

PR #387, opened 2024-10-02
Transformations: ResolveAssociateTransformer re-write to in-place substitution

mlange05:
~_Note: This PR sits on top of PR #372 and requires this to be merged first._~

This PR re-writes the `ResolveAssociateTransformer` and switches it to acting entirely in-place. For this, instead of finding all symbols and substituting them with their inverse-association, we now simply run over each symbol, find its inverse from the `.scope` and apply in-place. This requires the substitution to be written as a symbol mapper.

As an addition feature, we can now do partial resolution in associate bodies, where only a select sub-region of the code has its associated symbols resolved. I've added a test to this extend.

And on top of that I've refactored the imports in the associated test.

Edit: I've also had a to introduce "auto-fixer" behaviour for `CallStatement` nodes and a corresponding test. This now ensure that `call.arguments` and `call.kwarguments` are always (possibly empty) tuples and can thus safely be iterated. This fixed some fallout in the transpile tests.

### DIFF
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -501,7 +501,14 @@ class Associate(ScopedNode, Section, _AssociateBase):  # pylint: disable=too-man
         """
         An :any:`collections.OrderedDict` of associated expressions.
         """
-        return OrderedDict(self.associations)
+        return CaseInsensitiveDict((str(k), v) for k, v in self.associations)
+
+    @property
+    def inverse_map(self):
+        """
+        An :any:`collections.OrderedDict` of associated expressions.
+        """
+        return CaseInsensitiveDict((str(v), k) for k, v in self.associations)
 
     @property
     def variables(self):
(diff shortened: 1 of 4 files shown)


loki/ir/nodes.py
```diff
@@ -501,7 +501,14 @@ def association_map(self):
...
         """
-        return OrderedDict(self.associations)
+        return CaseInsensitiveDict((str(k), v) for k, v in self.associations)
+
+    @property
+    def inverse_map(self):
+        """
+        An :any:`collections.OrderedDict` of associated expressions.
+        """
+        return CaseInsensitiveDict((str(v), k) for k, v in self.associations)
```
reuterbal, line 511: Can we simply "stringify" `v` here, considering that an association can in theory be a rather complex expression?

reuterbal (approved):
Very neat trick and nicely implemented and tested!

I left a question but that's likely fine, so happy to merge as is.

PR #388, opened 2024-10-02
Utilities to merge associate blocks and restrict depth of associate resolution

mlange05:
~_Note: This sits on top of PR #387 and requires this to be merged first._~

This PR adds a new utility to merge `Associate` blocks and adds a feature to the existing `resolve_associations` method to restrict the starting depth of the association resolution. Combining the two allows for a neat look and feel, where all except the first layer of (non-nested) associations get resolved after all non-nested ones have been merged into the outermost associate block. Tests have been added for both features.

Edit: This one took on a life of its own after the interior mechanics of `ResolveAssociates` were changed ( PR #387 ). As the new resolver relies very strictly on identifying the "depth" via finding the defining scope, the new merge mechanism needs to carefully update the changes scopes to re-build the local symbol table of the `Associate` nodes. This is done by moving the derivation mechanics from the frontend into the IR node and triggering it specifically (comment and suggestions welcome). 

I've also added a rudimentary node-level test for `Associate` nodes and improved the associate mappings to allow symbols as keys, as well as strings.

### DIFF
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -35,7 +35,7 @@ import loki.expression.symbols as sym
 from loki.expression.operations import (
     StringConcat, ParenthesisedAdd, ParenthesisedMul, ParenthesisedDiv, ParenthesisedPow
 )
-from loki.expression import ExpressionDimensionsMapper, AttachScopesMapper
+from loki.expression import AttachScopesMapper
 from loki.logging import debug, detail, info, warning, error
 from loki.tools import (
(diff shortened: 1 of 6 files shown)


mlange05:
@reuterbal This one took a few little detours, but is now ready I think. Please have a look.

reuterbal (approved):
Many thanks, this is very nice! The implementation looks very clean and neat. Just a few suggestions to harden the test a little further, but I'd consider them optional.

PR #389, opened 2024-10-04
Skip privatization of arrays with existing data declarations

awnawab:
Arrays with existing data declarations should not be privatized in the driver gang loop. This PR adds the functionality to gather information from encompassing `!$acc data` regions and filter out the relevant arrays.

### DIFF
--- a/loki/transformations/single_column/annotate.py
+++ b/loki/transformations/single_column/annotate.py
@@ -5,11 +5,13 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+from collections import defaultdict
 from loki.batch import Transformation
 from loki.expression import symbols as sym, is_dimension_constant
 from loki.ir import (
     nodes as ir, FindNodes, FindVariables, Transformer,
-    pragmas_attached, is_loki_pragma, get_pragma_parameters
+    pragmas_attached, is_loki_pragma, get_pragma_parameters,
+    pragma_regions_attached
 )
 from loki.logging import info
 from loki.tools import as_tuple, flatten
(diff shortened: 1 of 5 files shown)


loki/transformations/single_column/annotate.py
```diff
@@ -194,11 +195,55 @@ def transform_subroutine(self, routine, **kwargs):
             # Mark all non-parallel loops as `!$acc loop seq`
             self.annotate_sequential_loops(routine)
 
+            # Find variables with existing OpenACC data declarations
+            acc_vars = self.find_acc_vars(routine, targets)
+
             with pragmas_attached(routine, ir.Loop, attach_pragma_post=True):
-                driver_loops = find_driver_loops(routine=routine, targets=targets)
+                driver_loops = find_driver_loops(section=routine.body, targets=targets)
```
reuterbal, line 202: Can't we move `find_acc_vars` inside the pragma context, to save doing this twice?

```py
with pragma_regions_attached(routine):
    with pragmas_attached(routine, ir.Loop, attach_pragma_post=True):
        acc_vars = self.find_acc_vars(routine, targets)
        driver_loops = find_driver_loops(section=routine.body, targets=targets)
        for loop in driver_loops:
         ...
```
And then remove all pragma attaching from `find_acc_vars`?

loki/transformations/single_column/annotate.py
```diff
@@ -221,14 +266,16 @@ def device_alloc_column_locals(cls, routine, column_locals):
...
 
-    def annotate_driver_loop(self, loop):
+    def annotate_driver_loop(self, loop, acc_vars):
         """
         Annotate driver block loop with ``'openacc'`` pragmas.
 
         Parameters
         ----------
         loop : :any:`Loop`
             Driver :any:`Loop` to wrap in ``'opencc'`` pragmas.
```
reuterbal, line 276: Not in this PR, but since you're here maybe fix the typo: `openacc` instead of `opencc` ;-)

loki/transformations/single_column/annotate.py
```diff
@@ -221,14 +266,16 @@ def device_alloc_column_locals(cls, routine, column_locals):
             routine.body.prepend((ir.Comment(''), pragma, ir.Comment('')))
```
reuterbal, line 266: [no action required] This is not covered in tests, so we don't seem to be testing this.
(@mlange05 for info)

loki/transformations/single_column/annotate.py
```diff
@@ -194,11 +195,55 @@ def transform_subroutine(self, routine, **kwargs):
...
+            List of subroutines that are to be considered as part of
+            the transformation call tree.
+        """
+
+        acc_vars = defaultdict(list)
+
+        with pragma_regions_attached(routine):
+            with pragmas_attached(routine, ir.Loop):
+                for region in FindNodes(ir.PragmaRegion).visit(routine.body):
+                    if region.pragma.keyword.lower() == 'acc' and 'data' in region.pragma.content.lower():
```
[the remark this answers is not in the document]
awnawab: Of course! Nice catch, thanks!

loki/transformations/single_column/annotate.py
awnawab: Yes, it is much better indeed to only attach pragmas once, thanks! 🙏

reuterbal (commented):
Many thanks, changes look good to me!

mlange05 (approved):
Looks great, GTG. :shipit:

PR #392, opened 2024-10-09
SingleColumn: Fix vectorisation of nested else-if bodies

mlange05:
The recent special-casing for else-if branches in the SCC vectorisation only works for single-level nesting, but skips any further nested else-ifs and the final else-branch. TO fix this, i've introduced a `else_bodies` property on `Conditional` nodes that returns the nested bodies for multi-conditionals. Using this, this we can then apply vector-section extraction for every else body in a deeply nested multi-conditional.

### DIFF
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -704,6 +704,16 @@ class Conditional(InternalNode, _ConditionalBase):
             return f'Conditional:: {self.name}'
         return 'Conditional::'
 
+    @property
+    def else_bodies(self):
+        """
+        Return all nested node tuples in the ``ELSEIF``/``ELSE`` part
+        of the conditional chain.
+        """
+        if self.has_elseif:
+            return (self.else_body[0].body,) + self.else_body[0].else_bodies
+        return (self.else_body,)
+
 
 @dataclass_strict(frozen=True)
 class _PragmaRegionBase():
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -712,7 +712,7 @@ class Conditional(InternalNode, _ConditionalBase):
         """
         if self.has_elseif:
             return (self.else_body[0].body,) + self.else_body[0].else_bodies
-        return (self.else_body,)
+        return (self.else_body,) if self.else_body else ()
 
 
 @dataclass_strict(frozen=True)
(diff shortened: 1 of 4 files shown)


loki/ir/nodes.py
```diff
@@ -704,6 +704,16 @@ def __repr__(self):
...
         return 'Conditional::'
 
+    @property
+    def else_bodies(self):
+        """
+        Return all nested node tuples in the ``ELSEIF``/``ELSE`` part
+        of the conditional chain.
+        """
+        if self.has_elseif:
+            return (self.else_body[0].body,) + self.else_body[0].else_bodies
```
reuterbal, line 714: Now I have understood - very neat indeed 😄

loki/transformations/single_column/vector.py
```diff
@@ -141,7 +141,11 @@ def extract_vector_sections(cls, section, horizontal):
                 # as 'Conditional's rely on the fact that the first element of the 'else_body'
                 # (if 'has_elseif') is a Conditional itself
                 if separator.has_elseif and separator.else_body:
```
reuterbal, line 143: Since `separator.else_bodies` should, even in the `has_elseif==False` case, return the else_body correctly, we should be able to remove this check here and always use the new implementation, no?

reuterbal (commented):
Only a comment for now, but changes look very good. I think your solution actually allows to simplify the vector sections extraction again, removing the need to check for `has_elseif`.

I will let the nightly CI run against this branch to confirm whether it fixes the problem.

mlange05:
Thanks for the early review; you were right, the logic could be drastically simplified. Unfortunately, last nights version was still buggy, but I've just rebase-pushed a clean version that should bring the tco399 ecphys regression back (confirmed locally). Please have another look, but should be ready now.

loki/transformations/single_column/vector.py
```diff
@@ -140,12 +140,8 @@ def extract_vector_sections(cls, section, horizontal):
                 # we need to prevent that the whole 'else_body' is wrapped in a section,
                 # as 'Conditional's rely on the fact that the first element of the 'else_body'
                 # (if 'has_elseif') is a Conditional itself
```
MichaelSt98, line 142: Maybe rephrase this comment/explanation?

```suggestion
                # we need to prevent that all (possibly nested) 'else_bodies' are completely wrapped as a section,
                # as 'Conditional's rely on the fact that the first element of each 'else_body'
                # (if 'has_elseif') is a Conditional itself
```

MichaelSt98 (approved):
Very nice indeed!

reuterbal (approved):
The fixed fix seems to have resolved the problems in ifs-bundle. Many thanks for this!

I have accepted the comment suggestion from @MichaelSt98 as well.

PR #394, opened 2024-10-10
Logging: Small log-level sanitisation and CLI flags 

mlange05:
This PR brings a small clean-up and adjustments of our log-levels in preparation for some scheduler work. In particular it does:
* Introduce a new log-level DETAIL to capture per-file info in a batch processing context. This sits below PERF, so that PERF can be used for scheduler things without completely clobbering things with per-file output
* Docstrings for the specific log-levels to give some rough guidance on how to use them
* Adjustements to most of the frontend and scheduler timer logging; we now only output high-level timers for scheduler actions (no nested info!) and put the nested sub-components on PERF. Similarly, only the heavy IR parse stays at per-file level with PERF (as before with INFO), and all other frontend timers are put to DETAIL.
* The `--log-level` option is added to `loki-transform.py [convert | plan]` to expose log-levels explicitly. These then set the log_level, but can also be overridden with `LOKI_LOGGING` as before. The original central flag handling still works for debugging in the code base.

Last note: This is just a first attempt at improvements. There's a lot more we can do to improve logging in general, and I'm open for any pointers and/or opinions. Therefore posting as draft for initially for feedback.

### DIFF
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -16,11 +16,10 @@ from pathlib import Path
 import click
 
 from loki import (
-    Sourcefile, Frontend, as_tuple, set_excepthook,
-    auto_post_mortem_debugger, info
+    config as loki_config, Sourcefile, Frontend, as_tuple,
+    set_excepthook, auto_post_mortem_debugger, info
 )
 from loki.batch import Transformation, Pipeline, Scheduler, SchedulerConfig
-from loki.logging import set_log_level, log_levels
 
 # Get generalized transformations provided by Loki
 from loki.transformations.argument_shape import (
@@ -135,7 +134,7 @@ def convert(
     :any:`Transformation` objects to this call tree.
     """
 
-    set_log_level(log_levels[log_level])
+    loki_config['log-level'] = log_level
 
     info(f'[Loki] Batch-processing source files using config: {config} ')
 
@@ -447,7 +446,7 @@ def plan(
     given configuration.
     """
 
-    set_log_level(log_levels[log_level])
+    loki_config['log-level'] = log_level
 
     info(f'[Loki] Creating CMake plan file from config: {config}')
     config = SchedulerConfig.from_file(config)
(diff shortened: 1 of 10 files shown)


scripts/loki_transform.py
```diff
@@ -113,12 +114,15 @@ def cli(debug):
               help="Recursively derive explicit shape dimension for argument arrays")
 @click.option('--eliminate-dead-code/--no-eliminate-dead-code', default=True,
               help='Perform dead code elimination, where unreachable branches are trimmed from the code.')
+@click.option('--log-level', '-l', default='info', envvar='LOKI_LOGGING',
```
reuterbal, line 117: I just noticed that the log level could already be specified via this environment variable: https://github.com/ecmwf-ifs/loki/blob/1d95896f118ede6ada0e9bdd3e38921e81f3a8a9/loki/__init__.py#L49-L50
Not sure if having this twice might cause any problems?

reuterbal (approved):
Many thanks, this is great!
`detail` would now also be the log level where we should print much more information during Transformation processing, e.g., when skipping call nodes due to missing interprocedural type information etc.

scripts/loki_transform.py
mlange05: It doesn't cause problems _per se_, as the CLI version takes precedence (the naming alias is no coincidence). However, I still think exposing an explicit CLI option is a neat thing to have and having two ways of specifying env flags is iffy. The only remaining use case for this is now toggling log-level for the pytest test base invocation - I'm not sure if this is sufficiently useful to warrant having two ways of doing this?

reuterbal: In principle we could also provide an equivalent log level overwrite switch for pytest specifically:
https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_addoption

mlange05:
@reuterbal Nice! I merged your suggested PR and added a small fix on top to go through the config for setting log-level from the CLI entry points as well. Please have another look.

PR #395, opened 2024-10-11
Transformations: Remove routine pragmas when inlining functions

mlange05:
When inlining functions we need to remove any routine-level pragmas, such as `!$loki routine seq`, so as not to erroneously mark the caller body with them. This change adds a removal step and should hopefully fix ec-physics regression.

### DIFF
--- a/loki/transformations/inline.py
+++ b/loki/transformations/inline.py
@@ -601,7 +601,7 @@ def map_call_to_procedure_body(call, caller, callee=None):
         callee.body.body, scope=caller
     )
 
-    # Substitute 'loki routine' pragmas
+    # Remove 'loki routine' pragmas
     callee_body = Transformer(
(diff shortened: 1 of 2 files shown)


reuterbal (approved):
Many thanks, change looks good to me. I'll await confirmation before merging.

loki/transformations/inline.py
```diff
@@ -601,6 +601,12 @@ def _map_unbound_dims(var, val):
         callee.body.body, scope=caller
     )
 
+    # Substitute 'loki routine' pragmas
```
MichaelSt98, line 604:
```suggestion
    # Remove 'loki routine' pragmas
```

MichaelSt98 (approved):
Looks good to me. Thanks!

PR #396, opened 2024-10-14
Pytest CLI option for log-level

reuterbal:
As discussed offline: This provides a CLI option overwrite for the log-level when running pytest, and removes the duplicate env variable hook from the config.


mlange05 (approved):
Yes

PR #399, opened 2024-10-14
Improve representation of procedure pointers (fix #393)

### LINKED ISSUE #393: problem with procedure pointers
Procedure pointers are used in recent versions of the semi-implicit. However, the following leads to a bug in Loki :
spcmnew_pb.F90:
```fortran
SUBROUTINE SPCMNEW(YDGEOMETRY)

USE SPSI_MODNEW, ONLY : SPNSI
IMPLICIT NONE
PROCEDURE(SPNSI), POINTER :: SPNSIPTR
END SUBROUTINE SPCMNEW
```
launch.py:
```python
from loki import Sourcefile,fgen
source=Sourcefile.from_file('spcmnew_pb.F90',preprocess=True)
spcmnew=source['spcmnew']
print(fgen(spcmnew.spec))
```
(SPNSI is an abstract interface for a subroutine, included in spsi_modnew.F90).

reuterbal:
This was triggered by the fgen issue reported in #393. This looked first like the issue was simply that we did not account for the case that procedure information might be missing when generating the code for procedure declarations.

However, while trying to sort this out across frontends, I noticed that it should be perfectly possible to recognize the interface as a procedure in this situation, and that the handling of procedure types for symbols declared in procedure declarations was very inconsistent and partially wrong. This should fix this discrepancy.

### DIFF
--- a/loki/backend/tests/test_fgen.py
+++ b/loki/backend/tests/test_fgen.py
@@ -7,10 +7,11 @@
 
 import pytest
 
-from loki import Module, Subroutine
+from loki import Module, Subroutine, Sourcefile
 from loki.backend import fgen
 from loki.frontend import available_frontends, OMNI, OFP
 from loki.ir import Intrinsic, DataDeclaration
+from loki.types import ProcedureType, BasicType
 
 
 @pytest.mark.parametrize('frontend', available_frontends())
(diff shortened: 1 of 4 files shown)


mlange05 (approved):
Very neatly done. Looks good, GTG. :shipit:

Issue #400, opened 2024-10-14
Inlining of elemental functions with argument(s) being array (slices)


MichaelSt98:
### What happened?

Compile error, as a scalar variable can't be assigned to some array value ...

### What are the steps to reproduce the bug?

Problem being:

```fortran
elemental function elem(x, y)
    real :: x, y, elem, tmp
    temp = y * 2.0
    elem = x + tmp
end function elem

This function can be called

* with `x` and `y` being scalars
* with `x` and `y` being arrays of the same shape
* with `x` or `y` being an array and the non-array being a scalar

Thus, e.g.:

```fortran
real :: x_scalar, b_scalar, res_scalar
real :: x_arr(10). y_arr(10), res_arr(10)
real :: x_arr_2(12). y_arr_2(12), res_arr_2(12)

res_scalar = elem(x_scalar, y_scalar)
res_arr = elem(x_arr, y_arr)
res_arr = elem(x_arr, y_scalar)
res_arr = elem(y_scalar, x_arr)
res_arr_2 = elem(x_arr_2, y_arr_2)
...
```

A quick solution is not possible, since:

* adapting/adding the dimensions to the result var is not sufficient, since temporaries within the elemental function potentially need to be promoted as well
* as shown above the same routine can be called in various ways influencing the inlining

**The potential solution** probably should

* use a loop instead of vector notation to inline (to simplify the problem with promoting (some of) temporaries, ... in the end this could help us to solve it without need for (sophisticated) dataflow and dimension analysis)
* reuse variables/temporaries (maybe introduce `_scalar` and `_array` versions of temporaries (and result)

### Version

main

### Platform (OS and architecture)

Mac / HPC2020

PR #401, opened 2024-10-14
Inline elemental functions: skip calls with args being array (slices)

MichaelSt98:
We have a problem for things like:

```fortran
elemental function elem(x, y)
    real :: x, y, elem, tmp
    temp = y * 2.0
    elem = x + tmp
end function elem

real :: x_scalar, b_scalar, res_scalar
real :: x_arr(10). y_arr(10), res_arr(10)
real :: x_arr_2(12). y_arr_2(12), res_arr_2(12)

res_scalar = elem(x_scalar, y_scalar)
res_arr = elem(x_arr, y_arr)
res_arr = elem(x_arr, y_scalar)
res_arr = elem(y_scalar, x_arr)
res_arr_2 = elem(x_arr_2, y_arr_2)
...
```

As the fix is not straight-forward/trivial, this PR makes the inline to skip the problematic calls (for now). Thus, **this is not a fix to make those problematic calls being inlined correctly!**

A more detailed description of the problem: #400

### DIFF
--- a/loki/transformations/inline/functions.py
+++ b/loki/transformations/inline/functions.py
@@ -110,17 +110,7 @@ def _inline_functions(routine, inline_elementals_only=False, functions=None):
         def map_inline_call(self, expr, *args, **kwargs):
             if not self.visit(expr, *args, **kwargs):
                 return
-            if expr.procedure_type is BasicType.DEFERRED or isinstance(expr.routine, StatementFunction):
-                return
-            if not expr.procedure_type.is_function:
-                return
-            if self.inline_elementals_only:
-                if self.inline_elementals_only and not expr.procedure_type.is_elemental:
-                    return
-            if functions:
-                if expr.routine not in functions:
-                    return
-            if expr.procedure_type.is_elemental:
+            if not expr.procedure_type is BasicType.DEFERRED and expr.procedure_type.is_elemental:
                 if any(is_array(val) for val in expr.arg_map.values() if isinstance(val, sym.Array)):
                     warning(f"Call to elemental function '{expr.routine.name}' with array arguments."
                             f' There is currently no support to inline those calls!')
@@ -163,6 +153,15 @@ def _inline_functions(routine, inline_elementals_only=False, functions=None):
     # in the next call to this function.
     for node, calls in FindInlineCallsSkipInlineCallParameters(with_ir_node=True).visit(routine.body):
         for call in calls:
+            if call.procedure_type is BasicType.DEFERRED or isinstance(call.routine, StatementFunction):
+                continue
+            if not call.procedure_type.is_function:
+                continue
+            if inline_elementals_only and not call.procedure_type.is_elemental:
+                continue
+            if functions:
+                if call.routine not in functions:
+                    continue
             function_calls.setdefault(str(call.name).lower(),[]).append((call, node))
 
     if not function_calls:
--- a/loki/transformations/inline/functions.py
+++ b/loki/transformations/inline/functions.py
@@ -121,7 +121,7 @@ def _inline_functions(routine, inline_elementals_only=False, functions=None):
                 if expr.routine not in functions:
                     return
             if expr.procedure_type.is_elemental:
-                if any(is_array(val) for val in expr.arg_map.values()):
+                if any(is_array(val) for val in expr.arg_map.values() if isinstance(val, sym.Array)):
                     warning(f"Call to elemental function '{expr.routine.name}' with array arguments."
                             f' There is currently no support to inline those calls!')
                     return
--- a/loki/transformations/inline/functions.py
+++ b/loki/transformations/inline/functions.py
@@ -7,7 +7,8 @@
 
 from collections import ChainMap
 
-from loki.expression import symbols as sym, ExpressionRetriever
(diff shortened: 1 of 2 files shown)


loki/transformations/tests/test_inline.py
```diff
@@ -164,44 +164,89 @@ def test_transform_inline_elemental_functions_extended(tmp_path, builder, fronte
...
+  integer, intent(in) :: len  
+  real(kind=real64), intent(in) :: v1(len)
+  real(kind=real64), intent(inout) :: v2(len), v3(len)
+  real(kind=real64), parameter :: param1 = 100.
+  integer, parameter :: arr_index = 1
+
+  v2 = multiply(v1(:), 6._real64) + multiply_single_line(v1(:), 3._real64)
+  v3 = add(param1, 200._real64) + add(v1, 150._real64) + multiply(v1(arr_index), v2(1))
+end subroutine transform_inline_elemental_functions_extended_array
 """
```
reuterbal, line 182: Let's make the array case a separate test, please.

You can still avoid replicating the module code etc using the following pattern:

```py
@pytest.fixture(name='multiply_extended_mod', params=available_frontends())
def fixture_multiply_extended_mod(request, tmp_path):
    fcode_module = """
...
    """

    frontend = request.param
    module = Module.from_source(fcode_module, frontend=frontend, xmods=[tmp_path])
    return module, frontend

def test_transform_inline_elemental_functions_extended_scalar(multiply_extended_mod, tmp_path):
    module, frontend = multiply_extended_mod

    fcode = "..."

    routine = Subroutine.from_source(fcode, frontend=frontend, definitions=[module], xmods=[tmp_path])
    ...

def test_transform_inline_elemental_functions_extended_arr(multiply_extended_mod, tmp_path):
    module, frontend = multiply_extended_mod

    fcode_arr = "..."

    routine = Subroutine.from_source(fcode_arr, frontend=frontend, definitions=[module], xmods=[tmp_path])
    ...

```

loki/transformations/inline.py
```diff
@@ -389,6 +393,19 @@ def __init__(self, query, recurse_query=None, inline_elementals_only=False,
...
+                return
+            if self.inline_elementals_only:
+                if not (expr.procedure_type.is_function and expr.procedure_type.is_elemental):
+                    return
+            if functions:
+                if expr.routine not in functions:
+                    return
+            if (expr.procedure_type.is_function and expr.procedure_type.is_elemental):
+                if any(is_array(val) for val in expr.arg_map.values()):
+                    warning(f'Call to elemental function \'{expr.routine.name}\' with array arguments.'
```
reuterbal, line 406: You can avoid escaping if you mix quotation marks:
```suggestion
                    warning(f"Call to elemental function '{expr.routine.name}' with array arguments."
```

loki/transformations/inline.py
```diff
@@ -389,6 +393,19 @@ def __init__(self, query, recurse_query=None, inline_elementals_only=False,
         def map_inline_call(self, expr, *args, **kwargs):
             if not self.visit(expr, *args, **kwargs):
                 return
+            if expr.procedure_type is BasicType.DEFERRED or isinstance(expr.routine, StatementFunction):
+                return
```
reuterbal, line 397: We could add here a bailout for non-functions and save a few checks further down:
```suggestion
                return
            if not expr.procedure_type.is_function:
                return
```

loki/transformations/inline.py
```diff
@@ -389,6 +393,19 @@ def __init__(self, query, recurse_query=None, inline_elementals_only=False,
...
+            if self.inline_elementals_only:
+                if not (expr.procedure_type.is_function and expr.procedure_type.is_elemental):
```
reuterbal, line 399: This would then become
```suggestion
            if self.inline_elementals_only and not expr.procedure_type.is_elemental:
```

loki/transformations/inline.py
```diff
@@ -389,6 +393,19 @@ def __init__(self, query, recurse_query=None, inline_elementals_only=False,
...
+                    return
+            if functions:
+                if expr.routine not in functions:
+                    return
+            if (expr.procedure_type.is_function and expr.procedure_type.is_elemental):
```
reuterbal, line 404: And this would simplify to
```suggestion
            if expr.procedure_type.is_elemental:
```

reuterbal (changes requested):
Thanks for this! 

For readability, we could make the checks a bit shorter (suggestions in inline comments).

The test is blowing a bit out of proportion this way and I would prefer splitting the completely separate "array" case into a separate test. I left a suggestion how you can avoid code duplication by moving the module into a joint fixture.

Finally, this requires a rebase due to the refactoring of inline and extract packages.

MichaelSt98:
Thanks for the quick review!

loki/transformations/inline.py
MichaelSt98: Hmm this ends up with errors like:

```
AttributeError: 'BasicType' object has no attribute 'is_function'
```

reuterbal: You should add this below the checks in ll. 396-397

mlange05 (approved):
Yes, looks good to me. GTG

PR #403, opened 2024-10-14
Fix Scheduler instantiation without config (fix #373)

### LINKED ISSUE #373: Passing no config to Scheduler leads to runtime error.
The default argument for the config in the Scheduler leads to a runtime error. However, the documentation of the Scheduler class says that the config parameter is optional.

It seems to me that an easy fix could be to change the default argument from None to {}

reuterbal:
A rather trivial fix but it was correctly reported as a discrepancy between docs and observed behaviour.


mlange05 (approved):
GTG :shipit:

PR #405, opened 2024-10-14
Remove Maxeler transpilation module

reuterbal:
~_Note: Filing this as a draft PR for now to assess impact on test coverage._~

The Maxeler transpilation capability was developed during the EuroEXA project. However, since the completion of the project we do no longer have access to the commercially licenced simulation software nor any Maxeler Dataflow Engine hardware, therefore testing these capabilities is no longer possible. Hence, I propose removing the transpilation module from Loki.


reuterbal:
OK, coverage seems not to deteriorate as a result of this, meaning we don't seem to have had any indirect coverage purely due to the maxeler transpilation.

mlange05 (approved):
Yes! GTG :shipit:

PR #407, opened 2024-10-15
Update f90wrap to 0.2.15 as minimum to ensure compatibility with numpy 2.0+

reuterbal:
~_Note: Draft PR to trigger tests_~

What it says on the tin:
- Update f90wrap to 0.2.15 as minimum version, which restores compatibility with numpy 2.0 (https://github.com/jameskermode/f90wrap/releases/tag/v0.2.15)
- That allows to remove the restriction numpy<1.24
- Which allows to enable Python 12 as supported Python version
- ...and Python 12 added to the Github actions test matrix

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,6 @@ requires = [
   "setuptools >= 61",
   "wheel",
   "setuptools_scm[toml] >= 6.2",
-  "numpy<1.24"
 ]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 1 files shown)


reuterbal:
OK, the impact of the changes in f90wrap for support of Python 12 seems to be rather extensive: it mandates switching to meson (which somehow doesn't get installed reliably), requires ninja as build backend (which definitely doesn't get installed reliably) and somehow forgets to compile / use the source files before compiling the wrapper library. 

We don't have Python 3.12 available on the cluster, yet, but the problems can be replicated locally on MacOS. Hence, I limit the scope of this PR to supporting latest numpy via enforcing a recent f90wrap version, and we can look into Python 3.12 support separately.

mlange05 (approved):
Sounds sensible. GTG :shipit:

Issue #406, opened 2024-10-15
Future strategy regarding frontends


reuterbal:
### What maintenance does this project need?

We need to have a discussion on the future of Loki frontends.

Fparser is being actively worked on, is the most complete and our standard frontend.

Our use of OMNI compiler depends on the specific OMNI version packaged in the CLAW repository. That version is a few years old and doesn't build with newer compiler versions, e.g., GNU10+, anymore. But even the latest updates on the upstream OMNI compiler repository have been more than 2 years ago: https://github.com/omni-compiler/omni-compiler
Our Github CI currently runs with the ubuntu 20.04 runners, which will likely be EOL some time next year, when that Ubuntu version goes out of maintenance. Hence, we have a call to make regarding the future of OMNI in Loki.

The Open Fortran Parser relies on an outdated Python package that wraps an old ROSE compiler frontend version. We have already earmarked this for removal (#209).

The question is whether 
* we would be comfortable relying on a single frontend, or 
* if we should expend effort to either update one of the others to a more recent version
* explore other frontend options in lieu

Various options exist for the last:
* Use an up-to-date ROSE compiler version, which may provide other Python binding options
* Explore Fxtran as a frontend
* Explore Lfortran as a frontend
* ...?

reuterbal:
As explained in #408, removing CLAW allows to upgrade OMNI to the most recent release which currently remains compatible with the latest compiler versions and CI runners. If anything, this buys us a little time...

reuterbal:
PR #411 implements the deprecation of OFP for the next Loki release (v0.2.8?). The intention is to remove the frontend entirely in the release following after that (could justify a version increase as v0.3.0?).

As expected, this reduces the test coverage because the translation from the OFP parse tree to the Loki IR is no longer covered in the tests, and alongside the relevant constructors for `Sourcefile`, `Subroutine`, `Module`, etc. However, it incurred also a number of indirect coverage regressions due to code paths that only exist for OFP. These would have to be removed together with the frontend, or need to receive dedicated tests.

The following indirect coverage regressions are [capture by codecov](https://app.codecov.io/gh/ecmwf-ifs/loki/pull/411/indirect-changes):

- [ ] `loki/frontend/source.py`: `extract_source` and `extract_source_from_range`
  https://github.com/ecmwf-ifs/loki/blob/main/loki/frontend/source.py#L371-L438
[...]

PR #230, opened 2024-02-23
Regression test for CLOUDSC2

reuterbal:
This adds a regression test for CLOUDSC2 processing, and enables it as part of the Github actions CI tests.
Temporarily, it points to a branch for CLOUDSC2, which has a pending PR for integration: ecmwf-ifs/dwarf-p-cloudsc2-tl-ad#14

The PR is filed against the new Scheduler branch but doesn't have to be integrated now and can be left dangling for the moment.

Processing works fine and results are correct with GNU on host-side (GPU evaluation is tbd), however, in the Actions runner I'm seeing randomly massive deviations for steps 4 and 6 in the TL variant. No clue what is going on there.

PR #408, opened 2024-10-15
Add option to install "plain OMNI" to install script and upgrade Github actions runners

reuterbal:
This is partially in response to #406: It looks like we can keep OMNI alive for the time being (yay...?!) if we give up the specific version bundled with the CLAW, and instead update to the latest upstream. The way this is implemented does not yet remove CLAW entirely, instead a new `--with-omni` option is added to the `install` script and the existing `--with-claw` option is kept as is. The two cannot be combined but using `--with-omni` gives the newer F_Front version.

Notably, for our purpose it is even sufficient to use the [xcodeml-tools](https://github.com/omni-compiler/xcodeml-tools) component of the OMNI compiler, which provides the frontends.

With this we are able to use recent GNU versions to compile OMNI and thus can upgrade the Github runners to the latest Ubuntu image, which uses GNU 13.2.0 as default. This builds cleanly and all tests seem to work as expected. It should also resolve the spurious errors observed in #230, which I will rebase on top of this PR.

The one downside of these changes is that they means we are stopping the testing of the CLAW variants for CLOUDSC regression tests. But since they don't have any perspective anyway, I think this is acceptable - please shout if you disagree!

I used the opportunity to also update the module versions for HPC2020 to recent versions.
Piggy-backed are also a few fixes where no tmp_path was specified for xmods in tests.

### DIFF
--- a/install
+++ b/install
@@ -13,8 +13,8 @@ set -euo pipefail
 loki_ant_version=1.10.13
 
 hpc2020_java_version=11.0.6
-hpc2020_python_version=3.10.10-01
-hpc2020_cmake_version=3.25.2
+hpc2020_python_version=3.11.8-01
+hpc2020_cmake_version=3.28.3
 
 # Determine base path for loki
 # Either take the root of the current git tree or, if not inside a git repository, then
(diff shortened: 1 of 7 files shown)


install
```diff
@@ -61,6 +62,7 @@ print_usage_with_options() {
   echo "  --with[out]-jdk              Install JDK instead of using system version (default: use system version)"
   echo "  --with[out]-ant              Install ant instead of using system version (default: use system version)"
   echo "  --with[out]-claw             Install CLAW and OMNI Compiler (default: disabled)"
+  echo "  --with[out]-omni             Install OMNI Compiler; cannot be combined with --with-claw (default: disabled)"
```
mlange05, line 65: Maybe prefer that one in the `INSTALL.md` now in the example with `--hpc2020`?

mlange05 (approved):
I agree with the approach. And while I'm sad to see the original CLAW go, I think it's time to let go if this unlocks us going forward.

I left one tiny pointer, but otherwise, GTG from me. :shipit:

install
reuterbal: Thanks, I have update the instructions and also added the missing ability to install now OMNI instead of CLAW+OMNI via CMake. Please have another look

mlange05:
Very nice! CMake additions look good; still GTG for me.

mlange05:
Farewell, old fried!  🫡  😢 

https://media1.tenor.com/m/mPHvERT5iKgAAAAC/toy-story-buzz-lightyear.gif

PR #409, opened 2024-10-16
Fix SccAnnotate when existing acc pragmas declare a copy category more than once

reuterbal:
PR #389 introduced a regression in ecphys. There we have an ACC statement like this:

```
!$acc data copyin(var_a, var_b) &
!$acc &    copyin(var_c) create(var_d)
```

Now, the ACC statements are checked for existing offload directives. For that, `get_pragma_parameters` is used to extract offload "categories". However, for situations where a key exists more than once, the values are provided as a list, i.e., for the above we obtain
```
{'copyin': ['var_a, var_b', 'var_c'], 'create': 'var_d'}
```

Where previously a simple `str.split(',')` was used to build a list of variables of each category, we need to be careful to accomodate for the above situation where not a simple string constitutes the value. This PR expands the test to include the above case and fixes the transformation accordingly.

### DIFF
--- a/loki/transformations/single_column/tests/test_scc.py
+++ b/loki/transformations/single_column/tests/test_scc.py
@@ -247,16 +247,16 @@ def test_scc_annotate_openacc(frontend, horizontal, blocking, acc_data):
     INTEGER, INTENT(IN)   :: nlon, nz, nb  ! Size of the horizontal and vertical
     INTEGER, INTENT(IN)   :: nproma, nlev  ! Aliases of horizontal and vertical sizes
     REAL, INTENT(INOUT)   :: q(nlon,nz,nb)
-    REAL :: other_var(nlon)
+    REAL :: other_var(nlon), more_var(nlon)
     INTEGER :: b, start, end
 
     start = 1
     end = nlon
     {'!$acc data default(present)' if acc_data == 'default' else ''}
-    {'!$acc data copyin(other_var)' if acc_data == 'copyin' else ''}
+    {'!$acc data copyin(more_var) copyin(other_var)' if acc_data == 'copyin' else ''}
     !
     do b=1, nb
-      call compute_column(start, end, nlon, nproma, nz, q(:,:,b), other_var)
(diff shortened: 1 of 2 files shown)


reuterbal:
Build appears to be clean, so this is ready for a look. I haven't run ecphys, yet, though.

loki/transformations/single_column/tests/test_scc.py
```diff
@@ -338,7 +338,10 @@ def test_scc_annotate_openacc(frontend, horizontal, blocking, acc_data):
         if acc_data:
             assert driver_loops[0].pragma[0].content == 'parallel loop gang vector_length(nlon)'
         else:
-            assert driver_loops[0].pragma[0].content == 'parallel loop gang private(other_var) vector_length(nlon)'
+            assert driver_loops[0].pragma[0].content in (
+                'parallel loop gang private(other_var, more_var) vector_length(nlon)',
+                'parallel loop gang private(more_var, other_var) vector_length(nlon)'
```
mlange05, line 343: [no action] Curious, why would the ordering go off here? Should this not be deterministic?

mlange05 (approved):
Looks sensible. GTG from me.

loki/transformations/single_column/tests/test_scc.py
reuterbal: Well spotted! 🦅 - 👁️ 
I was wondering the same and I suspect it is here:
https://github.com/ecmwf-ifs/loki/blob/7f7337ef5515edb8b2ddeb9a92ebcfd36852c181/loki/transformations/single_column/annotate.py#L294
We could replace this by
```py
 private_arrays = ', '.join(dict.from_keys(v.name for v in arrays if not v.name_parts[0].lower() in acc_vars))
```
to have a stable sort but unique list. But didn't want to piggy-back this in this hotfix PR

mlange05: Ahhh yes. I think I've fixed this in one of my dev branches. Not urgent now, this will happen one way or another, I think.

PR #410, opened 2024-10-17
Prevent superfluous clone of loki in ecwam regression test

awnawab:
The ecwam-bundle project name in the regression test is outdated, which lead to the tagged loki release in the ewam bundle being cloned everytime the test was run. Since `--without-loki-install` is specified, that version is never built and the local copy of loki is indeed used. Nevertheless it would be good to prevent an unnecessary clone of loki during the regression test.


reuterbal (approved):
Nice catch, thanks!

Issue #209, opened 2024-01-16
Remove OFP frontend


mlange05:
The OFP frontend is getting increasingly hard to maintain. A recent brush with some of its inconsistencies (mostly due to skew in the generated AST node scoping due to upstream bugs) has highlighted that in particular better handling of functions/procedures will become very painful if we are to maintain this. 

I've checked and the most recent release (0.6.1) is not compatible with our frontend implementation, but it does still have many of the scoping quirks that we tend to bend over backwards to circumvent. As we have little use for it that we cannot deal with via other frontends, I propose we now fully deprecate and/or entirely remove the OFP frontend. This should improve test throughput and remove some of the special-casing, ideally before we continue to align procedures as proper IR nodes.

I'm happy to implement and test this, but input and feedback are very welcome (@reuterbal ).

reuterbal:
I agree. OFP is dead slow, which, among all the other issues, would make it undesirable for any productive use anyway. There is value in maintaining more than one frontend but arguably we should rather spend time on exploring other alternatives than fixing up OFP.

reuterbal:
Following the ideas tossed around in #406 and the thin thread we have thrown as a lifeline to OMNI in #408, the proposal is to proceed to deprecating OFP:

- [ ] Add a big warning in every outward API of OFP that makes clear that OFP will be removed in the next release and is no longer maintained and tested
- [ ] Remove OFP from the list returned by `available_frontends()`
- [ ] Clean up the test base to remove every trace of OFP from there
- [ ] After the next Loki release, remove OFP and its traces from the remaining Loki code base

PR #411, opened 2024-10-17
Frontend: Deprecate OFP and purge from test base

mlange05:
~_Note: Just testing for now..._~

As described in #209 , this PR deprecates the OFP frontend and adds a warning to the central OFP parse method. It also purges the `OFP` symbol from the outward-facing API and the test base (including the `available_frontends()` utiltiy), leaving internal mechanics to rely on the explicit mention of `Frontend.OFP`.

Please note that the frontend itself is still present and functional, but it's use is now discouraged.

### DIFF
--- a/loki/backend/tests/test_fgen.py
+++ b/loki/backend/tests/test_fgen.py
@@ -10,7 +10,7 @@ import pytest
 from loki import Module, Subroutine, Sourcefile
 from loki.backend import fgen
 from loki.expression import symbols as sym
-from loki.frontend import available_frontends, OMNI, OFP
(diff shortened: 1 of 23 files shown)


mlange05:
Tests all passing, but coverage is decreased (as expected). Farewell old friend. 🫡

reuterbal (approved):
Thank you very much, this is fantastic and sad and soothening and so many mixed feelings...

As discussed offline, I have captured the indirect coverage regressions that are incurred from this in https://github.com/ecmwf-ifs/loki/issues/406#issuecomment-2422215593

These should be taken care of when the factual removal of OFP is happening

PR #413, opened 2024-10-22
CMake/python_venv: Do not request COMPONENT Development

reuterbal:
A small hotfix PR that removes the "Development" component from the find_package for Python in a CMake installation.

This component is intended to allow applications to use Python headers - but Loki, as a pure Python package, doesn't need this. However, the find_package will fail if Python headers cannot be found, thus preventing installation.


mlange05 (approved):
Looks good to me. GTG :shipit:

PR #414, opened 2024-10-23
Dimension: Support stepping, implicit aliases and remove contrainsts

mlange05:
A small re-write of the convenience class `Dimension` that we use to map loop iteration spaces and variable data spaces to transformations. This is meant as a proposal for a more convenient usage, but it would eventually require external breaking changes to config files. I'm filing this here now for discussion, while trying to keep as much backward-compatibility as possible.

The changes themselves are in anticipation of declaring the co-mapped point-block dimensions for the custom IFS-style blocking pattern (`NPROMA`) in a single dimension. For this, we require an explicit stepping variable (the block size), and multiple lower/upper bound variables to capture the global point bounds and the block-specific bounds. Prior attempts at (ab)using the aliasing mechanics became very clunky, so I'm proposing this now.

To do this "properly", this PR changes the following details:
* Dimension objects now internally store `index`, `lower`, `upper`, `step` and `size` as core variable names
* Each core variable can now me a single variable string, or a tuple/list of strings to allow aliasing
* The `.bounds` and `.range` property are now derived from the first `lower` / `upper` pair, to stay backward compatible.
* `bounds_expression` and `index_expression` have been removed since they redundant now
* Previous aliiasing mechanics are retained in the constructor, but fed into the new internal layout for backward compatibility
* An explicit constructor test and an additional `SchedulerConfig` test have been added and existing tests have been retained

The idea is to now get rid of any explicit `alias` / `bounds_alias` / `size_alias` usage in externally used config files and simply overload the index | lower | upper fields in configs. Since this is a breaking change, we won't strip existing functionality just yet. And as mentioned above, all open for discussion.


mlange05:
Ok, I've made a few small tweaks to the design to keep better backward compatibility, but I think this is now ready for a first look and discussion. @reuterbal @awnawab

loki/dimension.py
```diff
@@ -69,68 +81,90 @@ def variables(self):
         return (self.index, self.size) + self.bounds
 
     @property
-    def size(self):
+    def sizes(self):
         """
-        String that matches the size expression of a data space (variable allocation).
+        Tuple of strings that matche the primary size and all secondary size expressions.
```
awnawab, line 86: Typo: `match` instead of `matche`.

awnawab (approved):
Many thanks for this rewrite! It's a big improvement over the existing `Dimension` utility, and very elegantly mops up my hackery around aliasing 😅

loki/dimension.py
awnawab, line 86: For this description to be true, don't we also need to add the output of `size_expressions`?

reuterbal: No, I don't think we do. `self._size` already contains all expressions that describe the size, including aliases. The only thing that `size_expressions` would add are expressions that are derived from iteration space bounds. Keeping the two explicitly separate is an intended outcome of this PR. It might, however, make sense to use `self.sizes` in `size_expressions` instead of only the primary size expression

loki/dimension.py
```diff
@@ -69,68 +81,90 @@ def variables(self):
...
+        Tuple of expression string that represent the bounds of an iteration space.
 
-        exprs = [(b,) for b in self.bounds]
-        if self._bounds_aliases:
-            exprs = [expr + (b,) for expr, b in zip(exprs, self._bounds_aliases)]
+        .. note:
 
-        return as_tuple(exprs)
+        If mutiple lower or upper bound string have been provided,
+        only the first pair will be used.
```
reuterbal, line 140: Sphinx markup correction required (and Github won't let me annotate this properly because of some deleted lines - sorry!): Double colon, indentation

```
        .. note::
            If mutiple lower or upper bound string have been provided,
            only the first pair will be used.
```

loki/dimension.py
```diff
@@ -69,68 +81,90 @@ def variables(self):
...
-    def index_expressions(self):
+    def range(self):
         """
-        A list of all expression strings representing the index expression of an iteration space (loop).
+        String that matches the range expression of an iteration space (loop).
+
+        .. note:
+
+        If mutiple lower or upper bound string have been provided,
+        only the first pair will be used.
```
reuterbal, line 155:
```suggestion
        .. note::
            If mutiple lower or upper bound string have been provided,
            only the first pair will be used.
```

loki/dimension.py
```diff
@@ -69,68 +81,90 @@ def variables(self):
...
-            exprs += list(self._index_aliases)
+    @property
+    def size_expressions(self):
+        """
+        A list of all expression strings representing the size of a data space.
 
-        return as_tuple(exprs)
+        This includes generic aliases, like ``end - start + 1`` or ``1:size`` ranges.
+        """
+        exprs = as_tuple(self.size)
```
reuterbal, line 166: As mentioned above: It might make sense to use `self.sizes` here to capture the aliases?

loki/transformations/block_index_transformations.py
```diff
@@ -216,7 +216,7 @@ def process_body(self, body, definitions, successors, targets, exclude_arrays):
             if call.name in targets:
                 _args = {a: d for d, a in call.arg_map.items() if isinstance(d, Array)}
                 _vars += [a for a, d in _args.items()
-                         if any(v in d.shape for v in self.horizontal.size_expressions) and a.parents]
+                         if any(v in d.shape for v in self.horizontal.sizes) and a.parents]
```
reuterbal, line 219:
```suggestion
                          if any(v in d.shape for v in self.horizontal.sizes) and a.parents]
```

loki/transformations/single_column/scc_cuf.py
```diff
@@ -292,10 +292,10 @@ def kernel_cuf(self, routine, horizontal, vertical, block_dim,
         loop_map = {}
         for loop in FindNodes(ir.Loop).visit(routine.body):
             if loop.variable == self.block_dim.index or loop.variable.name.lower()\
-                    in [_.lower() for _ in self.block_dim._aliases]:
+                    in [_.lower() for _ in self.block_dim.sizes]:
```
reuterbal, line 295: Looking at this pattern, there might be an argument to enforce lower case for all these epxressions in `Dimension` itself.

reuterbal (changes requested):
This looks great, many thanks! A few very nit-picky comments but nothing substantial.

Only idea that could be worthwhile to explore (but would also be fine to do in a separate PR if captured in an issue instead) is to enforce consistent casing on all these string expressions.

loki/dimension.py
mlange05: Yes, IIRC in some corner case the difference between "secondary alias" and "derived" is significant, so the two stay distinct. Once we mop up the aliases in the regression tests, we can mop this up though, I think (I'd need to check which one exactly though). In any case, I'll add a note to the docstring to clarify.

loki/dimension.py
```diff
@@ -39,22 +39,34 @@ class Dimension:
...
+    ):
         self.name = name
-        self._index = index
-        self._bounds = as_tuple(bounds)
-        self._size = size
-        self._aliases = as_tuple(aliases)
-        self._index_aliases = as_tuple(index_aliases)
 
+        if bounds:
+            # Backward compat for ``bounds`` contructor argument
```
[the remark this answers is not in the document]
mlange05: This one's tricky, as these are hit in a lot(!) of places, and in ec-phys on a per-file basis. So deprecation warnings would need to go into `detail`, which would then swallow them. So, while I'm all in favour of rigorous deprecation, I'm not quite seeing how to do this cleanly without overloading the output.

In any case, I'll make it explicit it in the docstring.

loki/transformations/single_column/scc_cuf.py
mlange05: No, I think this should really be mopped up by symbol-string-comparison. I'll adjust in the transformation as this makes it cleaner to read.

reuterbal (approved):
Looks great, nothing to add. Much appreciated!

PR #415, opened 2024-10-25
Transformations: "Parallel" sub-module with driver-level parallelilsation utilities

mlange05:
_Note: This sits on top of #414 , as it requires the new `Dimension` overloading mechanism for block-loop dimensions._

This change brings a set of utilities that form the basis of "driver"-level control flow manipulation for IFS-style block loops, OpenMP parallel regions and FIELD API boilerplate code. They are loose packaged in a new sub-module `loki.transformations.parallel` and are intended to be extended at a later stage. The come in remove / add pairs, and currently only pertain to host-side driver loops with FIELD-API view updates.

I've also piggy-backed two small IR-level fixes (logging string and a weird corner case for pragma region detection), the latter of which was needed for some of the testing of the new functionalities.

In a little more detail, the new sub-package provides:
* Utilities to remove `!$omp parallel` sections and optionally inject `!$loki parallel` regions in their place.
* A utility to re-inject `!$omp parallel` sections for `!$loki parallel` markers, where "private" and "firstprivate" clauses are derived using dataflow analysis
* Utilities to remove and create explicit copies for derived types that would trigger a known "firstprivate" bug. This common workarounds is inserted for explicitly named variables and is thus user-driven.
* Utilities to remove and create the common FIELD API `object%update_view(ibl)` for derived-type objects that have specifically named types. This can also insert the slightly deviating `dimension%update(ibl, icst, icend)` updates.

The transformations that require detection of IFS-style "driver" block loops depend on a specific overloading of the primary/secondary lower/upper/step properties of the re-written `Dimension` object - a very specific convention that I'm open to discussing further (still seems hacky!). 😞

### DIFF
--- a/loki/transformations/parallel/block_loop.py
+++ b/loki/transformations/parallel/block_loop.py
@@ -9,11 +9,17 @@
 Transformation utilities to remove and generate parallel block loops.
 """
 
-from loki.ir import nodes as ir, FindNodes, Transformer
+from loki.expression import symbols as sym
+from loki.ir import (
+    nodes as ir, FindNodes, Transformer, pragma_regions_attached,
+    is_loki_pragma
+)
+from loki.scope import SymbolAttributes
 from loki.tools import as_tuple
+from loki.types import BasicType
 
 
-__all__ = ['remove_block_loops']
+__all__ = ['remove_block_loops', 'add_block_loops']
 
 
 def remove_block_loops(routine, dimension):
@@ -52,3 +58,80 @@ def remove_block_loops(routine, dimension):
             return tuple(n for n in loop.body if n not in to_remove)
 
     routine.body = RemoveBlockLoopTransformer().visit(routine.body)
+
+
+def add_block_loops(routine, dimension, default_type=None):
+    """
+    Insert IFS-style (NPROMA) driver block-loops in ``!$loki
+    parallel`` regions.
+
+    The provided :any:`Dimension` object describes the variables to
+    used when generating the loop and default assignments. It
+    encapsulates IFS-specific convention, where a strided loop over
+    points, defined by ``dimension.index``, ``dimension.bounds`` and
+    ``dimension.step`` is created, alongside assignments that define
+    the corresponding block index and upper bound, defined by
+    ``dimension.indices[1]`` and ``dimension.upper[1]`` repsectively.
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The routine in which to add block loops.
+    dimension : :any:`Dimension`
+        The dimension object describing the block loop variables.
+    default_type : :any:`SymbolAttributes`, optional
+        Default type to use when creating variables; defaults to
+        ``integer(kind=JPIM)``.
+    """
+
+    _default = SymbolAttributes(BasicType.INTEGER, kind='JPIM')
+    dtype = default_type if default_type else _default
+
+    # TODO: Explain convention in docstring
+    lidx = routine.parse_expr(dimension.index)
+    bidx = routine.parse_expr(dimension.indices[1])
+    bupper = routine.parse_expr(dimension.upper[1])
+
+    # Ensure that local integer variables are declared
+    for v in (lidx, bupper, bidx):
+        if not v in routine.variable_map:
+            routine.variables += (v.clone(type=dtype),)
+
+    def _create_block_loop(body, scope):
+        """
+        Generate block loop object, including indexing preamble
+        """
+
+        bsize = scope.parse_expr(dimension.step)
+        lupper = scope.parse_expr(dimension.upper[0])
+        lrange = sym.LoopRange((sym.Literal(1), lupper, bsize))
+
+        expr_tail = scope.parse_expr(f'{lupper}-{lidx}+1')
+        expr_max = sym.InlineCall(
+            function=sym.ProcedureSymbol('MIN', scope=scope), parameters=(bsize, expr_tail)
+        )
+        preamble = (ir.Assignment(lhs=bupper, rhs=expr_max),)
+        preamble += (ir.Assignment(
+            lhs=bidx, rhs=scope.parse_expr(f'({lidx}-1)/{bsize}+1')
+        ),)
+
+        return ir.Loop(variable=lidx, bounds=lrange, body=preamble + body)
+
+    class InsertBlockLoopTransformer(Transformer):
+
+        def visit_PragmaRegion(self, region, **kwargs):
+            """
+            (Re-)insert driver-level block loops into marked parallel region.
+            """
+            if not is_loki_pragma(region.pragma, starts_with='parallel'):
+                return region
+
+            scope = kwargs.get('scope')
+
+            loop = _create_block_loop(body=region.body, scope=scope)
+
+            region._update(body=(ir.Comment(''), loop))
+            return region
+
+    with pragma_regions_attached(routine):
+        routine.body = InsertBlockLoopTransformer().visit(routine.body, scope=routine)
--- a/loki/transformations/parallel/block_loop.py
+++ b/loki/transformations/parallel/block_loop.py
@@ -1,0 +1,54 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""
+Transformation utilities to remove and generate parallel block loops.
+"""
+
+from loki.ir import nodes as ir, FindNodes, Transformer
+from loki.tools import as_tuple
+
+
+__all__ = ['remove_block_loops']
+
+
+def remove_block_loops(routine, dimension):
+    """
+    Remove any outer block :any:`Loop` from a given :any:`Subroutine.
+
+    The loops are identified accoerding to a given :any:`Dimension`
+    object, and will remove auxiliary assignments of index and bound
+    variables, as commongly used in IFS-style block loops.
+
+    Parameters
+    ----------
+    routine: :any:`Subroutine`
+        Subroutine from which to remove block loops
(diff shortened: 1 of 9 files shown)


loki/transformations/parallel/field_api.py
```diff
@@ -0,0 +1,154 @@
...
+from loki.logging import warning
+from loki.tools import as_tuple
+
+
+__all__ = [
+    'remove_field_api_view_updates', 'add_field_api_view_updates'
+]
+
+
+def remove_field_api_view_updates(routine, field_group_types, dim_object=None):
```
wertysas, line 25: This function is case-sensitive, but it may be good to make it case-insensitive.

mlange05: Nice catch! The transformation is actual case-insensitive, but the warning checks are not. I've added come camel-case testing and checking of the generated warnings now.

loki/transformations/parallel/tests/test_field_api.py
```diff
@@ -0,0 +1,129 @@
...
+    with caplog.at_level(WARNING):
+        field_group_types = ['state_type', 'aux_type', 'flux_type']
+        remove_field_api_view_updates(
+            routine, field_group_types=field_group_types, dim_object='DIMS'
+        )
+
+        assert len(caplog.records) == 2
+        assert '[Loki::ControlFlow] Removing STATE%update_VIEW call, but not in field group types!'\
+            in caplog.records[0].message
+        assert '[Loki::ControlFlow] Removing FLUXES(JF)%UPDATE_VIEW call, but not in field group types!'\
```
wertysas, line 68: I don't think there should be any warnings generated here, since the ``STATE`` and ``FLUXES`` objects are of the field group types ``STATE_TYPE`` and ``FLUX_type`` respectively, which are in the ``field_group_types``list if we don't do case-sensitive checks.

wertysas (commented):
I left some comments, mostly on typos in docstrings. It was useful for me to see some good examples of the visitor pattern in action in Loki! The caplog functionality of pytests seems very useful, I didn't know that it was this easy to check the warning messages.

loki/transformations/parallel/openmp_region.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def add_openmp_regions(routine, global_variables=None, field_group_types=None):
+    """
+    Add the OpenMP directives for a parallel driver region with an
+    outer block loop.
+    """
+    block_dim_size = 'YDGEOMETRY%YRDIM%NGPBLKS'
+
+    global_variables = global_variables or {}
```
awnawab, line 106: As I understand it, these are the variables that compilers struggle to firstprivatise, so perhaps it should be renamed to `skip_firstprivatise_vars` or something similar. `global_variables` should be reserved for module imports. And could the tests also be extended to cover this functionality?

loki/transformations/parallel/openmp_region.py
```diff
@@ -0,0 +1,271 @@
...
+                local_vars += tuple(dict.fromkeys(
+                    v for v in routine_arguments
+                    if v.name in symbols and str(v.type.dtype) in field_group_types
+                ))
+
+                # Filter out known global variables
+                local_vars = tuple(v for v in local_vars if v.name not in global_variables)
+
+                # Make field group types firstprivate
+                firstprivates = tuple(dict.fromkeys(
```
awnawab, line 153: [noaction] In principle it would be great to rely primarily on the dataflow to determine what should be privatised and unprivatised with a few known overrides. Unfortunately this quickly becomes impractical due to the missing intents on the `SELF` dummy argument in the type-bound procedures, not to mention the need to first parse all of these. So for now we can leave this as is, but it would be nice to generalise this in a future PR.

awnawab (changes requested):
Many many thanks for these great utilities 🙏  The lack of driver layer transformations has led to a not insignificant amount of code duplication in certain places, which of course causes conflicts when merging science changes. So these will really help cleanup loki integration with the IFS 🥳 

I've left a bunch of comments that I think should be addressed, mostly about extending test coverage and typos. There are also some more nitpicky ones which I've marked as optional.

Finally, these utilities in their current guise won't quite work out-of-the-box on ecWAM or the IFS dy-core, but I will address that in followup PRs.

loki/transformations/parallel/field_api.py
```diff
@@ -0,0 +1,154 @@
...
+        def visit_Assignment(self, assign, **kwargs):  # pylint: disable=unused-argument
+            if assign.lhs.type.dtype in field_group_types:
+                warning(f'[Loki::ControlFlow] Found LHS field group assign: {assign}')
+            return assign
+
+        def visit_Loop(self, loop, **kwargs):
+            loop = self.visit_Node(loop, **kwargs)
+            return loop if loop.body else None
+
+        def visit_Conditional(self, cond, **kwargs):
```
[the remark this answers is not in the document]
mlange05: Yes, there's a particular pattern where we sometime do 
```
IF (..) THEN
  DO JRF=1, <something>
     CALL TENDENCY(JRF)%UPDATE_VIEW(IBL)
  END
END IF
```
Admittedly, only the loop is used in the tests. I should add the conditional to the test too...

loki/transformations/parallel/field_api.py
```diff
@@ -0,0 +1,154 @@
...
+        call to ``<dim>%UPDATE(...)`` accordingly.
+    """
+    field_group_types = as_tuple(field_group_types)
+
+    class RemoveFieldAPITransformer(Transformer):
+
+        def visit_CallStatement(self, call, **kwargs):  # pylint: disable=unused-argument
+
+            if '%update_view' in str(call.name).lower():
+                if not call.name.parent:
```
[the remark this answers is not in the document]
mlange05: This warning is actually non-sensical, as the `%` in the line above ensures that this really cannot happen (and thus cannot be tested).

loki/transformations/parallel/openmp_region.py
```diff
@@ -0,0 +1,271 @@
...
+
+    with pragma_regions_attached(routine):
+        with dataflow_analysis_attached(routine):
+            for region in FindNodes(ir.PragmaRegion).visit(routine.body):
+                if not is_loki_pragma(region.pragma, starts_with='parallel'):
+                    return
+
+                # Accumulate the set of locally used symbols and chase parents
+                symbols = tuple(region.uses_symbols | region.defines_symbols)
+                symbols = tuple(dict.fromkeys(flatten(
```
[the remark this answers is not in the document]
awnawab: Never mind, my suggestion doesn't make sense, I had originally misunderstood what this line does. Sorry for the noise 😅

loki/transformations/parallel/openmp_region.py
mlange05: Agreed that "global" is ambiguous in this context, but `skip_firstprivate_vars` is also to long and unclear (they can also not be "private"). How about `shared_variables` to stick with OpenMP semantics?

loki/transformations/parallel/block_loop.py
```diff
@@ -0,0 +1,137 @@
...
+    variables += as_tuple(dimension.lower)
+    variables += as_tuple(dimension.upper)
+
+    class RemoveBlockLoopTransformer(Transformer):
+        """
+        :any:`Transformer` to remove driver-level block loops.
+        """
+
+        def visit_Loop(self, loop, **kwargs):  # pylint: disable=unused-argument
+            if not loop.variable == idx:
```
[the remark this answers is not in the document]
mlange05: I think this is actually tested, but due to missing recursion, we never his this line. I'll add recursion to be sure.

loki/transformations/parallel/openmp_region.py
mlange05: Agreed, but as I have a bunch of other modifications for this already lined up, I'd like to push this out to future PRs for now (when I have better real-life coverage for this).

mlange05:
Many thanks for the thorough review, @wertysas and @awnawab . I've pushed updates to the tests, comments and docstrings with small modifications to the methods themselves to address your comments. I've also renamed some of the "firstprivate-copies" routines to bring them in line with the naming convention for the other utilities. I've kept everything clean on top of the current branch to give you a clean diff - please let me know when if you're happy, so I can do another quick rebase over main.

loki/transformations/parallel/openmp_region.py
```diff
@@ -53,7 +52,7 @@ class RemoveOpenMPRegionTransformer(Transformer):
 
         def visit_PragmaRegion(self, region, **kwargs):
             """
-            Perform the fileterin and removeal of OpenMP pragma regions.
+            Perform the filetering and removal of OpenMP pragma regions.
```
wertysas, line 55:
```suggestion
            Perform the filtering and removal of OpenMP pragma regions.
```

awnawab (approved):
Many thanks for addressing all the comments, this now looks good to go 👌

loki/transformations/parallel/openmp_region.py
```diff
@@ -96,14 +95,26 @@ def visit_Pragma(self, pragma, **kwargs):
         routine.body = RemoveOpenMPRegionTransformer().visit(routine.body, active=False)
 
 
-def add_openmp_regions(routine, global_variables=None, field_group_types=None):
+def add_openmp_regions(
+        routine, dimension=None, shared_variables=None, field_group_types=None
```
wertysas, line 99:
```suggestion
        routine, dimension, shared_variables=None, field_group_types=None
```
I didn't realise this until the dimension was added to filter loops by loop variable, but attributes of the dimension parameter are accessed in this method, hence it shouldn't be set to ``None``.

loki/transformations/parallel/openmp_region.py
mlange05: Yes, thanks, will mop this up in the rebase.

loki/transformations/parallel/openmp_region.py
mlange05: Nice, very good catch! Again, I'll mop this up in the rebase.

wertysas (approved):

reuterbal (approved):
Thanks, this looks great!

Many thanks also to @awnawab and @wertysas for your diligent review and strict enforcement of testing - much appreciated!

PR #416, opened 2024-10-25
Expressions: Handle intrinsic function calls

mlange05:
This PR adds handling of intrinsic function calls by adding an `is_intrinsic` property to `ProcedureType` and adding the procedure type to the parsing scope in frontends before creating the respective `ProcedureSymbol`.

We also add the new `is_intrinsic` flag to the symbol attributes table for quick lookup in scheduler items, as we never want to chase intrinsics.

EDIT: This also needs the fix from #439 and needs a rebase once that is merged.

### DIFF
--- a/loki/batch/item.py
+++ b/loki/batch/item.py
@@ -685,7 +685,7 @@ class ProcedureItem(Item):
         inline_calls = tuple({
             call.function.name: call.function
             for call in FindInlineCalls().visit(self.ir.ir)
-            if isinstance(call.function, ProcedureSymbol)
+            if isinstance(call.function, ProcedureSymbol) and not call.function.type.dtype.is_intrinsic
         }.values())
         imports = tuple(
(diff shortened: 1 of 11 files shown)


reuterbal (changes requested):
Thanks, this looks great. One thing I'm missing here is the situation when intrinsic functions are hidden because of user-defined functions with the same name. This is unfortunately allowed by the standard and used in practice...

mlange05:
Hi @reuterbal , apologies, I rebased over main to sanity reasons. However, the response to your comment is in the second to last commit. With the top commit I piggy-backed a mitigation fix for ab upstream version incompatibility with last night's pydantic release. Please let me know if you'd like this to be a separate PR.

reuterbal (approved):
Many thanks, good to know about the limitation with shadowing in OMNI!
Thanks also for the pydantic hotfix.

PR #417, opened 2024-10-29
Handle Loki dimension pragmas for modules (and not only routines) for FP

MichaelSt98:
Infer shape via Loki dimension pragmas, e.g.,

```fortran
!$loki dimension(x/2, x**2, (x+y)/x)
real(kind=jprb), dimension(:, :, :), pointer :: v6 
```

not only for routines but also for modules (FP only).

### DIFF
--- a/loki/frontend/tests/test_frontends.py
+++ b/loki/frontend/tests/test_frontends.py
@@ -1963,7 +1963,7 @@ end module some_mod
 @pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'OMNI does not like Loki pragmas, yet!')]))
 def test_frontend_routine_variables_dimension_pragmas(frontend):
     """
-    Test that `!$loki dimension` pragmas can be used to verride the
+    Test that `!$loki dimension` pragmas can be used to override the
     conceptual `.shape` of local and argument variables.
     """
     fcode = """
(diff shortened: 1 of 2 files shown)


mlange05 (approved):
Nice and simple and well tested. LGTM. :shipit:

loki/frontend/tests/test_frontends.py
```diff
@@ -2020,6 +2020,69 @@ def to_str(expr):
...
+  real(kind=jprb), pointer, intent(inout) :: v4(:,:)
+  !$loki dimension(x+y,2*x)
+  real(kind=jprb), allocatable :: v5(:,:)
+  !$loki dimension(x/2, x**2, (x+y)/x)
+  real(kind=jprb), dimension(:, :, :), pointer :: v6 
+end module mod_variable_dimensions
+    """
+
+    def to_str(expr):
+        return str(expr).lower().replace(' ', '')
```
reuterbal, line 2051: This looks like we're missing a StrCompareMixin somewhere?

reuterbal (approved):
Fully agreed with @mlange05 's comment. Many thanks!

loki/frontend/tests/test_frontends.py
mlange05: Yes, but that's an (internally) know issue; we don't force `StrCompareMixin` on pure operations (`Sum`, `Product`, etc.). We could, probably, but beyond this PR I'd argue? Maybe we need a separate (formal) issue on this though?

reuterbal: Yes fully agreed - I didn't mean to imply this needs to be actioned here. Documenting this in an issue would be a good idea, though.

PR #418, opened 2024-10-30
Allow for optional case-sensitive 'recursive_expression_map_update'

MichaelSt98:
necessary for 'convert_to_lower_case' functionality.

@reuterbal it is unfortunately not possible without further changes to move the `with config_override` out of `recursive_expression_map_update` since the case-sensitivity produces problems within that function beyond the check for early termination.

### DIFF
--- a/loki/transformations/utilities.py
+++ b/loki/transformations/utilities.py
@@ -526,7 +526,7 @@ def recursive_expression_map_update(expr_map, max_iterations=10, mapper_cls=Subs
 
         # Check for early termination opportunities, either with case-sensitivity
         #  being the default (`{}`) or with the provided value (`case-sensitive`)
-        _case_sensitive = {'case-sensitive': case_sensitive} or {}
+        _case_sensitive = {'case-sensitive': case_sensitive} if case_sensitive is not None else {}
         with config_override(_case_sensitive):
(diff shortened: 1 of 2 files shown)


loki/transformations/utilities.py
```diff
@@ -549,7 +557,7 @@ def get_loop_bounds(routine, dimension):
     :any:`Subroutine`.
 
     Parameters
-    ----------
```
mlange05, line 552: Rogue character removal?

mlange05 (commented):
Yes, I can see how this creates a curious corner case - but I can't think of a better solution to this either. So, happy to go in like this.

I do think there's a rogue docstring change that snuck in there for no reason, so not approving yet. But otherwise GTG for me.

loki/transformations/utilities.py
MichaelSt98: Fair enough ... thanks for marking this! Should be resolved.

reuterbal (changes requested):
Thanks! We'll have to review the case-sensitivity approach more generally when the transpilation work progresses further, I think, but for now this seems like a good enough solution.

reuterbal (approved):
All addressed, many thanks!

PR #419, opened 2024-10-30
Expression: Expression cloning and mapper tests

### LINKED ISSUE #379: Literals not being rebuilt causing shallow copy issues
Loki seems to be shallow copying literals by not rebuilding the leaf nodes due to the assumption they are constant. @reuterbal identified the issue being due to the code at https://github.com/ecmwf-ifs/loki/blob/main/loki/expression/mappers.py#L552-L563

The reproducer for this issue is as follows:

```
from loki import Sourcefile, FindNodes, Loop, IntLiteral, SubstituteExpressions

source = Sourcefile.from_file('bar.F90', preprocess=True)
for routine in source.routines:
  o = FindNodes(Loop).visit(routine.body)[0]

  body_a = SubstituteExpressions().visit(o.body)
  body_b = SubstituteExpressions().visit(o.body)
  assert body_a[0].rhs.children[0] == body_b[0].rhs.children[0]     # Compare equal...
  assert body_a[0].rhs.children[0] is not body_b[0].rhs.children[0] # ...but not the same object
```

```
subroutine bar()
  integer :: c

  do i = 1, 5
    c = 1 + 3
  end do

end subroutine bar
```

mlange05:
~_Note: Only for testing so far ..._~

The primary intention behind this PR is to increase test coverage of the expression mappers and rigorously duplicate symbols when expression mappers (`LokiIdentityMapper` and `SubstituteExpressionMapper`). It also fixes a bunch of other issues along the way that this threw up and now fixes #379 .

In a little more detail:
* Add dedicated tests for expression walker (via `ExpressionRetriever`) and the mappers `LokitIdentityMapper` and `SubstituteExrpressionMapper`
* Lint-clean the expression tests directory
* Remove the now redundant `invalidate_source` argument from Loki mappers (we do not store source on expressions anymore)
  * Fix the resulting issue in `recursive_expression_map_update` by checking for ancestry of `SubstituteExpressionMapper` before handing the map to constructors
* Duplicate symbols in `LokiIdentityMapper` and `SubstotuteExpressionMapper`
  * Add a local `_rebuild()` method the base class that check for scope-carrying and scope-free symbols
  * Add a dedicated symbols test for correct re-instatiation of all our symbols types and fix the ones that do not behave
  * Always update the symbol after enrichment, to ensure that `DeferredTypeSymbols` are replaced with `ProcedureTypeSymbols` in `CallStatement.name` after the type updates.

### DIFF
--- a/loki/expression/tests/test_symbols.py
+++ b/loki/expression/tests/test_symbols.py
@@ -1,0 +1,77 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from loki.expression import symbols as sym
+from loki.scope import Scope
+from loki.types import BasicType, ProcedureType, SymbolAttributes
+
+
+def test_symbol_recreation():
+    """ 
+    Test the correct construction and re-construction of our symbol objects.
+    """
+    scope = Scope()
+    int_type = SymbolAttributes(BasicType.INTEGER, parameter=True)
+    real_type = SymbolAttributes(BasicType.REAL, kind='rick')
+    log_type = SymbolAttributes(BasicType.LOGICAL)
+    proc_type = SymbolAttributes(
+        ProcedureType(name='f', is_function=True, return_type=real_type)
+    )
+
+    i = sym.Scalar(name='i', type=int_type, scope=scope)
+    a = sym.Array(name='a', type=real_type, scope=scope)
+    b = sym.Variable(
+        name='b', dimensions=(i,), type=int_type, scope=scope
+    )
+    t = sym.Scalar(name='t', type=log_type, scope=scope)
+    f = sym.ProcedureSymbol(name='f', type=proc_type, scope=scope)
+
+    # Basic variables and symbols
+    exprs = [i, a, b, t, f]
+
+    # Literals
+    exprs.append( sym.FloatLiteral(66.6) )
+    exprs.append( sym.IntLiteral(42) )
+    exprs.append( sym.LogicLiteral(True) )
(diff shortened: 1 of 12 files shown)


mlange05:
FYI: Checking with ec-physics regression locall, there is no noticeable performance impact on loki processing speed from this.

loki/expression/tests/test_symbols.py
```diff
@@ -0,0 +1,77 @@
...
+        # TODO: Should we introduce a Mixin "Cloneable" to makes these sane?
+        cargs = dict(zip(expr.init_arg_names, expr.__getinitargs__()))
+        clone = type(expr)(**cargs)
+        assert clone == expr
+
+        if isinstance(expr, sym.TypedSymbol):
+            # Check that TypedSymbols replicate scope via .clone()
+            scoped_clone = expr.clone()
+            assert scoped_clone == expr
+            assert scoped_clone.scope == expr.scope
```
reuterbal, line 77: Do we want to strictly enforce duplication and identity?
```suggestion
        clone = type(expr)(**cargs)
        assert clone == expr
        assert clone is not expr

        if isinstance(expr, sym.TypedSymbol):
            # Check that TypedSymbols replicate scope via .clone()
            scoped_clone = expr.clone()
            assert scoped_clone == expr
            assert scoped_clone is not expr
            assert scoped_clone.scope is expr.scope
```

reuterbal (approved):
This is fantastic, many thanks. I also appreciate the number of clean-ups along the way, like the stray source handling for expression nodes.

Only comment is a suggestion for a stricter test, but otherwise nothing to add.

@Andrew-Beggs-ECMWF would you mind confirming if this resolves your issue?

Andrew-Beggs-ECMWF:
> This is fantastic, many thanks. I also appreciate the number of clean-ups along the way, like the stray source handling for expression nodes.
> 
> Only comment is a suggestion for a stricter test, but otherwise nothing to add.
> 
> @Andrew-Beggs-ECMWF would you mind confirming if this resolves your issue?

I'd like to reiterate Balthasar's comments, I'm really please to see this patch. Thank you for doing it! 😁

I can confirm all my tests now pass!

PR #421, opened 2024-10-30
extend 'remove_explicit_array_dimensions'

MichaelSt98:
to handle kwargs and inline calls (and not only args for calls to subroutines)

### DIFF
--- a/loki/transformations/array_indexing.py
+++ b/loki/transformations/array_indexing.py
@@ -19,7 +19,7 @@ from loki.analyse import dataflow_analysis_attached
 from loki.expression import symbols as sym, simplify, symbolic_op, is_constant
 from loki.ir import (
     nodes as ir, Assignment, Loop, VariableDeclaration, FindNodes,
-    Transformer, FindVariables, SubstituteExpressions
+    Transformer, FindVariables, SubstituteExpressions, FindInlineCalls
 )
 from loki.tools import as_tuple, CaseInsensitiveDict
 from loki.types import SymbolAttributes, BasicType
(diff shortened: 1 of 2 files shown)


loki/transformations/tests/test_array_indexing.py
```diff
@@ -1181,44 +1181,54 @@ def test_transform_resolve_vector_notation_common_loops(tmp_path, frontend):
...
+    fcode_driver = f"""
   SUBROUTINE driver_routine(nlon, nlev, a, b)
-    use kernel_mod, only: kernel_routine
+    use kernel_explicit_dimensions_mod, only: kernel_routine
     INTEGER, INTENT(IN)    :: nlon, nlev
     INTEGER, INTENT(INOUT) :: a(nlon,nlev)
     INTEGER, INTENT(INOUT)  :: b(nlon,nlev)
 
-    call kernel_routine(nlon, nlev, a, b)
+    call kernel_routine(nlon, nlev, {'a=' if as_kwargs else ''}a, {'b=' if as_kwargs else ''}b)
```
mlange05, line 1198: [no action] Just a minor nitpick: I don't think the additional parametrize and string-injection is needed. You could just leave `a` as argument and make `b` a keyword argument, but this way works too.

mlange05 (approved):
Yes, looks good to me and well tested. GTG :shipit:

loki/transformations/array_indexing.py
```diff
@@ -46,21 +46,40 @@ def remove_explicit_array_dimensions(routine, calls_only=False):
...
                 else:
                     arguments += (arg,)
-            call._update(arguments=arguments)
+            # handle kwargs
+            kwarguments = ()
+            for (kwarg_name, kwarg) in call.kwarguments:
+                if isinstance(kwarg, sym.Array) and all(dim==sym.RangeIndex((None, None)) for dim in kwarg.dimensions):
+                    kwarguments += ((kwarg_name, kwarg.clone(dimensions=None)),)
+                else:
+                    kwarguments += ((kwarg_name, kwarg),)
```
reuterbal, line 73: This line is currently untested. For full coverage, you could modify the `test_transform_excplicit_dimensions` to use a call like this (incorporating @mlange05's suggestion):
```
call kernel_routine(nlon, a, b=b, nlev=nlev)
```

reuterbal (commented):
Thanks, this is a useful extension and well-tested. A small suggestion to reach full test coverage but otherwise this is good to go!

PR #422, opened 2024-10-30
[F2C transpilation] (driver level) convert interface to import

MichaelSt98:
Convert interface to kernel calls (e.g., `interface kernel ...`) to the appropriate import (e.g., `kernel_fc_mod, only: kernel_fc`).

### DIFF
--- a/loki/transformations/transpile/fortran_c.py
+++ b/loki/transformations/transpile/fortran_c.py
@@ -182,6 +182,7 @@ class FortranCTransformation(Transformation):
             depth = depths[item]
 
         if role == 'driver':
+            self.interface_to_import(routine, targets)
             return
 
         for arg in routine.arguments:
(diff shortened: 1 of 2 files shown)


loki/transformations/transpile/fortran_c.py
```diff
@@ -231,6 +232,31 @@ def transform_subroutine(self, routine, **kwargs):
...
+            for b in i.body:
+                if isinstance(b, Subroutine):
+                    if targets and b.name.lower() in targets:
+                        # Create a new module import with explicitly qualified symbol
+                        modname = f'{b.name}_FC_MOD'
+                        new_symbol = Variable(name=f'{b.name}_FC', scope=routine)
+                        new_import = Import(module=modname, c_import=False, symbols=(new_symbol,))
+                        routine.spec.prepend(new_import)
+                        # Mark current import for removal
+                        removal_map[i] = None
```
reuterbal, line 255: We might want to use the API properties of `Interface` here:
```suggestion
            for s in i.symbols:
                if targets and s in targets:
                    # Create a new module import with explicitly qualified symbol
                    new_symbol = s.clone(name=f'{s.name}_FC', scope=routine)
                    modname = f'{new_symbol.name}_MOD'
                    new_import = Import(module=modname, c_import=False, symbols=(new_symbol,))
                    routine.spec.prepend(new_import)
                    # Mark current import for removal
                    removal_map[i] = None
```

loki/transformations/transpile/tests/test_transpile.py
```diff
@@ -1344,6 +1344,50 @@ def init_var(dtype, val=0):
...
+    assert not imports
+
+    f2c = FortranCTransformation()
+    f2c.apply(source=routine, path=tmp_path, targets=('kernel',), role='driver')
+
+    interfaces = FindNodes(ir.Interface).visit(routine.spec)
+    imports = FindNodes(ir.Import).visit(routine.spec)
+    assert len(interfaces) == 2
+    assert len(imports) == 1
+    assert imports[0].module.upper() == 'KERNEL_FC_MOD'
```
reuterbal, line 1389: You might want to verify here also the name of the imported symbol.

Suggested improvement that uses more API functions:
```suggestion
    assert len(routine.interfaces) == 2
    imports = routine.imports
    assert len(imports) == 1
    assert imports[0].module.upper() == 'KERNEL_FC_MOD'
    assert imports[0].symbols == ('KERNEL_FC',)
```

reuterbal (changes requested):
Thanks, this looks good to me. I've left two remarks how to improve the implementation a little bit but that's not mandatory. I'll leave it up to you whether you want to give this a try.

reuterbal (approved):
Many thanks, looks good now. I'll wait for tests to go green, then we can put this in before 0.2.9

PR #423, opened 2024-10-30
C-like-backends: skip/don't write Fortran interfaces

MichaelSt98:
For now: just don't generate code for Fortran interfaces when applying c-like-backends (`cgen`, `cppgen`, `cudagen`).


mlange05 (approved):
Looks good to me. GTG. :shipit:

reuterbal (approved):
Straightforward and with a nice test. GTG!

PR #425, opened 2024-11-01
Make builddir a runtime argument of `FileWriteTransformation`

awnawab:
A small PR that turns the builddir into a runtime argument of the `FileWriteTransformation`. This allows the builddir to be passed in via command line arguments even if the `FileWriteTransformation` is being overridden via a scheduler config.

### DIFF
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -120,7 +120,7 @@ class Scheduler:
         By default a full parse is executed, use this flag to suppress.
     frontend : :any:`Frontend`, optional
         Frontend to use for full parse of source files (default :any:`FP`).
-    builddir : str or path
+    output_dir : str or path
         Directory for the output to be written to
     """
 
@@ -129,7 +129,7 @@ class Scheduler:
 
     def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
                  includes=None, defines=None, definitions=None, xmods=None,
-                 omni_includes=None, full_parse=True, frontend=FP, builddir=None):
+                 omni_includes=None, full_parse=True, frontend=FP, output_dir=None):
         # Derive config from file or dict
         if isinstance(config, SchedulerConfig):
             self.config = config
@@ -156,7 +156,7 @@ class Scheduler:
             'xmods': xmods,
             'omni_includes': omni_includes,
             'frontend': frontend,
-            'builddir': builddir
+            'output_dir': output_dir
         }
 
         # Internal data structures to store the callgraph
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -120,6 +120,8 @@ class Scheduler:
         By default a full parse is executed, use this flag to suppress.
     frontend : :any:`Frontend`, optional
         Frontend to use for full parse of source files (default :any:`FP`).
+    builddir : str or path
+        Directory for the output to be written to
     """
 
     # TODO: Should be user-definable!
@@ -127,7 +129,7 @@ class Scheduler:
 
     def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
                  includes=None, defines=None, definitions=None, xmods=None,
-                 omni_includes=None, full_parse=True, frontend=FP):
+                 omni_includes=None, full_parse=True, frontend=FP, builddir=None):
         # Derive config from file or dict
         if isinstance(config, SchedulerConfig):
             self.config = config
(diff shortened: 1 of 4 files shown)


loki/batch/scheduler.py
```diff
@@ -495,7 +498,7 @@ def _get_definition_items(_item, sgraph_items):
                     _item.scope_ir, role=_item.role, mode=_item.mode,
                     item=_item, targets=_item.targets, items=_get_definition_items(_item, sgraph_items),
                     successors=graph.successors(_item, item_filter=item_filter),
-                    depths=graph.depths
+                    depths=graph.depths, build_args=self.build_args
```
mlange05, line 501: [no action] I would argue that we could also attach the `build_args` to the `Item`, so that we continue to use the `Item` as the universal meta-data object that gets handed from `Scheduler` to `Transformation`? Admittedly, this can be said for almost any of the arguments here - so this might also be part of a larger "scheduler-item" refactoring and need not go into this PR.

Any strong opinions on this, @reuterbal ?

mlange05 (commented):
All in all this looks good. I've left one conceptual question for @reuterbal , but once answered this is GTG for me.

loki/batch/scheduler.py
```diff
@@ -120,14 +120,16 @@ class Scheduler:
...
+        Directory for the output to be written to
     """
 
     # TODO: Should be user-definable!
     source_suffixes = ['.f90', '.F90', '.f', '.F']
 
     def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
                  includes=None, defines=None, definitions=None, xmods=None,
-                 omni_includes=None, full_parse=True, frontend=FP):
+                 omni_includes=None, full_parse=True, frontend=FP, builddir=None):
```
reuterbal, line 132: I would suggest we use a more generic term than `builddir` here, for example `output_path`. There is no reason that this needs to be an actual build directory, and for Loki processing that is not integrated with a build the name wouldn't make much sense.
It's just that in our use case we happen to specify the build directory as the output path

loki/batch/scheduler.py
reuterbal: Yes, good idea.

The only thing where we need to be a little careful is that the `definitions` part is updated dynamically when calling `_parse_items`. When re-parsing due to topology changes that introduce a new dependency that was previously not part of the graph, these definitions would then be missing for existing items. Hence, this is probably out of scope for this PR.

For the other arguments it depends. `role` and `targets` should be used from the Item and have only been kept for backwards compatibility. We should indeed remove them here. The other two are properties at a graph-level, not an item level, so there's an argument to keep them separate.

loki/transformations/build_system/file_write.py
```diff
@@ -72,6 +69,6 @@ def transform_file(self, sourcefile, **kwargs):
         path = Path(item.path)
         suffix = self.suffix if self.suffix else path.suffix
         sourcepath = Path(item.path).with_suffix(f'.{self.mode}{suffix}')
-        if self.builddir is not None:
-            sourcepath = self.builddir/sourcepath.name
+        if (builddir := kwargs.get('build_args', {}).get('builddir', None)) is not None:
```
reuterbal, line 72: I think we're be able to afford two lines to make this a little more readable :)
```suggestion
        build_args = kwargs.get('build_args')
        if build_args and (builddir := build_args.get('builddir')):
```

reuterbal (changes requested):
Thanks, looks great. A few nit-picks on nomenclature but otherwise good to go!

loki/batch/scheduler.py
mlange05: Ok, agreed. I'll look into this in a separate PR. 👍

loki/batch/scheduler.py
reuterbal: Yes, good point! The `paths` argument suffer from the same misnomer but that's a separate issue

reuterbal (approved):

loki/batch/tests/test_transformation.py
```diff
@@ -428,8 +428,8 @@ def test_transformation_file_write(tmp_path):
     ricks_path = tmp_path/'rick.roll.java'
     if ricks_path.exists():
         ricks_path.unlink()
-    FileWriteTransformation(mode='roll', suffix='.java').apply(source=source, item=item,
-                                                               build_args={'output_dir': tmp_path})
+    FileWriteTransformation(suffix='.java').apply(source=source, item=item,
```
reuterbal, line 431: [no action!] The suffix should probably become an item property as well in the future, particularly when thinking of transpilation use cases?

PR #429, opened 2024-11-05
Loki-transform: Add deprecation message about custom entry points

mlange05:
Pretty much what it says on the tin: adding a deprecation warning at the entry-point in `loki-transform.py convert` where we are planning to remove support for internally specified transformations. Once config-only mode is ready to be enforced, everything after this warning can.

_Integration note: Please announce this on the respective collaboration channels to make users aware._

Follow-on thought: Once enforced, I would like to add an alias `--pipeline`/`-p` for the `--mode`/`-m` CLI option, but that would require a lot of CMake-level renaming. Would this be useful, or overkill, @reuterbal ?

### DIFF
--- a/scripts/loki_transform.py
+++ b/scripts/loki_transform.py
@@ -17,7 +17,7 @@ import click
 
 from loki import (
     config as loki_config, Sourcefile, Frontend, as_tuple,
-    set_excepthook, auto_post_mortem_debugger, info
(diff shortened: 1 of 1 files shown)


reuterbal (approved):
Thanks, this looks good!

I agree, introducing a pipeline option sounds sensible! The CMake plumbing should be relatively minor on Loki side: we can directly use only the pipeline option there and  keep the mode option as an alias for the moment.

PR #430, opened 2024-11-06
Transformation to call loop transform utilities

awnawab:
A transformation to run the loop transform utilities in a specified order, with the aim to place this transformation in scheduler pipelines.

### DIFF
--- a/loki/transformations/tests/test_transform_loop.py
+++ b/loki/transformations/tests/test_transform_loop.py
@@ -16,11 +16,12 @@ from loki.expression import symbols as sym
 from loki.frontend import available_frontends
 from loki.ir import (
     is_loki_pragma, pragmas_attached, FindNodes, Loop, Conditional,
-    Assignment, FindVariables
+    Assignment, FindVariables, nodes as ir
 )
 
 from loki.transformations.transform_loop import (
-    do_loop_interchange, do_loop_fusion, do_loop_fission, do_loop_unroll
+    do_loop_interchange, do_loop_fusion, do_loop_fission, do_loop_unroll,
+    TransformLoopsTransformation
 )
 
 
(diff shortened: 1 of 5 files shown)


loki/transformations/transform_loop.py
```diff
@@ -752,3 +753,59 @@ def visit_Loop(self, o, *args, **kwargs):
...
+        loop_interchange(routine, project_bounds=self.project_bounds)
+
+        # Fuse loops
+        loop_fusion(routine)
+
+        # Split loops
+        loop_fission(routine, promote=self.promote, warn_loop_carries=self.warn_loop_carries)
+
+        # Unroll loops
+        loop_unroll(routine, warn_iterations_length=self.warn_iterations_length)
```
reuterbal, line 811: I would prefer having off switches for the individual loop transformations, with them disabled by default the same way we have this in the RemoveCodeTransformation:
https://github.com/ecmwf-ifs/loki/blob/c998cda4f465e34d17aeaf2e4803622ff4150522/loki/transformations/remove_code.py#L103-L110

Each of the utilities is on the more costly side of transformations, and if one of them is causing problems it's also handy to selectively switch them on or off. 

We should then also add a test that is parameterised such that each is switched on/off and that verifies that only the respective transformation is then active.

loki/transformations/transform_loop.py
```diff
@@ -752,3 +753,59 @@ def visit_Loop(self, o, *args, **kwargs):
...
+    warn_iterations_length : bool
+        This specifies if warnings should be generated when unrolling
+        loops with a large number of iterations (32). It's mainly to
+        disable warnings when loops are being unrolled for internal
+        transformations and analysis. Default: ``True``.
+    """
+
+    def __init__(
+            self, project_bounds=False, promote=True, warn_loop_carries=True,
+            warn_iterations_length=True
```
reuterbal, line 792: Without the loop transformation context these options are a little hard to map mentally. I suggest we prefix them according to the transformation they belong to:
```suggestion
            self, interchange_project_bounds=False,
            fission_auto_promote=True, fission_warn_loop_carries=True,
            unroll_warn_iterations_length=True
```

reuterbal (changes requested):
Thanks, conceptually this looks good but there's a few improvements that we should do right away to not require breaking the interface afterwards.

I've left inline comments, too, but to summarise:

- [ ] I would like to have explicit on-switches for each utility, with each off by default
- [ ] For the RemoveCodeTransformation, we renamed the utilities with `do_` prefix. We could use the same pattern here, e.g., call the relevant on-switch `loop_fusion` and the corresponding utility `do_loop_fusion`. This makes for a nicer config file appearance (@mlange05, any opinion on this?)
- [ ] We need a test that verifies the switches and their effectiveness by testing the full permutation matrix of all values

awnawab:
Many thanks @reuterbal for the detailed review and feedback 🙏 I've implemented all of your suggestions, please have a look again whenever you can.

loki/transformations/tests/test_transform_loop.py
```diff
@@ -2022,3 +2023,84 @@ def test_transform_loop_unroll_nested_neighbours(tmp_path, frontend):
 
     clean_test(filepath)
     clean_test(unrolled_filepath)
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+@pytest.mark.parametrize('loop_trafo', ['loop_interchange', 'loop_fusion', 'loop_fission',
+                                        'loop_unroll'])
+def test_transform_loop_transformation(frontend, loop_trafo):
```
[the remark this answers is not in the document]
awnawab: Ah yes, that is a much more comprehensive test, thanks!

PR #433, opened 2024-11-08
Sanitise: New transformation sub-package and some refactoring

mlange05:
This PR creates a new transformation sub-package for sanitisation and then refactors and improves several of the existing utilities for easier inclusion in external pipelines.

 In particular it separate associate-handling from sequence-association resolution, and adds a new utility for user-driven expression substitution. Each of those are now available as a single `Transformation` object and have been combined into a combined `SanitisePipeline`, each of which can now be used via external config constructors. The existing `SanitiseTransformation` is retained for backward compatibility.

In some more detail:
* Created a new sub-package `loki.transformations.sanitise` and moved according tests. In this sub-package, existing utilities for associates and sequence association have been separated
* Refactored sequence association resolver as an in-place `Transformer` to avoid awkward map handling. This requires some sub-classing in specific further use-case in `transformations.inline`.
* Provide `AssociateTransformation` to offer full or partial associate resolution together with the new associate-merging capabilities.
* Provide sequence association resolution as a standalone `Transformation`, which entails some method renaming
* Provide new `SubstituteExpressionTransformation` to expose string-based expression replacement
* Combine all of the above into a `SanitisePipeline`, which is intended as the long-term replacement for `SanitiseTransformation` - the latter, however, is retained for external backward compatibility
* Small tweaks to test (skip OMNI instead of xfail; it's faster and less messy in logs)

### DIFF
--- a/loki/transformations/sanitise/sequence_associations.py
+++ b/loki/transformations/sanitise/sequence_associations.py
@@ -5,6 +5,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+from loki.batch import Transformation
 from loki.expression import Array, RangeIndex
 from loki.ir import Transformer
 from loki.tools import as_tuple
@@ -12,8 +13,29 @@ from loki.types import BasicType
 
 
 __all__ = [
-    'transform_sequence_association', 'SequenceAssociationTransformer'
+    'SequenceAssociationTransformation',
+    'do_resolve_sequence_association',
+    'SequenceAssociationTransformer'
 ]
+
+
+class SequenceAssociationTransformation(Transformation):
+    """
+    :any:`Transformation` that resolves sequence association patterns
+    in :any:`CallStatement` nodes.
+
+    Parameters
+    ----------
+    resolve_sequence_associations : bool
+        Flag to trigger or suppress resolution of sequence associations
+    """
+
+    def __init__(self, resolve_sequence_associations=True):
+        self.resolve_sequence_associations = resolve_sequence_associations
+
+    def transform_subroutine(self, routine, **kwargs):  # pylint: disable=unused-argument
+        if self.resolve_sequence_associations:
+            do_resolve_sequence_association(routine)
 
 
 def check_if_scalar_syntax(arg, dummy):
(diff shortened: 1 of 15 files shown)


loki/transformations/sanitise/sequence_associations.py
```diff
@@ -0,0 +1,136 @@
...
+                n_dims = len(dummy.shape)
+                new_dims = []
+
+                if not arg.shape:
+                    # Hack: If we don't have a shape, short-circuit here
+                    new_dims = tuple(RangeIndex((None, None)) for _ in dummy.shape)
+                else:
+                    for s, lower in zip(arg.shape[:n_dims], arg.dimensions[:n_dims]):
+                        if isinstance(s, RangeIndex):
+                            new_dims += [RangeIndex((lower, s.stop))]
```
awnawab, line 123: This is probably quite a common use case, and so could the tests please be extended to cover this?

loki/transformations/sanitise/sequence_associations.py
```diff
@@ -0,0 +1,136 @@
...
+        for dummy, arg in call.arg_map.items():
+            if check_if_scalar_syntax(arg, dummy):
+                found_scalar = True
+
+                n_dims = len(dummy.shape)
+                new_dims = []
+
+                if not arg.shape:
+                    # Hack: If we don't have a shape, short-circuit here
+                    new_dims = tuple(RangeIndex((None, None)) for _ in dummy.shape)
```
awnawab, line 119: [no action] this line is also not covered by the testing, but I suspect we would only arrive here from incomplete enrichment of imported symbols, so probably no pressing need to extend the tests to cover this.

awnawab (changes requested):
Many thanks for this cleanup! Having access to some of these utilities directly via scheduler configs will be very helpful indeed👌 I just have two small comments on extending the test coverage of the `SequenceAssociationTransformer`, one of which is optional but I feel the other is an important use case to cover.

loki/transformations/sanitise/sequence_associations.py
mlange05: Yes, it is and it was not covered in Santeri's original testing. I've slightly refactored his test and added specific cases for this. Please check if this now covers it.

loki/transformations/sanitise/sequence_associations.py
mlange05: Apologies, this one snuck in by accident (I needed it in one of the dev branches). I've now also added a test for this, so please see if this is ok.

awnawab (approved):
Thanks a lot for covering the missing lines, this is now good to go! 🙏

PR #434, opened 2024-11-11
Loki: Pin pymbolic version to 2022.2 due to upstream incompatibility

mlange05:
Upstream update to Pymbolic `v2024.1` is causing issues in ecWAM regression suite that need separate investigation. Pinning to known-good version with this.


reuterbal (approved):
Many thanks! Confirmed that this is indeed needed right now and pinning the version restores functionality.

PR #436, opened 2024-11-11
extending dependency trafo to access specs

MichaelSt98:
rename/transform private/public access specifiers for modules

@reuterbal and @mlange05: asking for feedback on this

problem for OMNI:

```
        # This is a call to a subroutine declared via header-included interface
        item_name = f'#{proc_name}'.lower()
        if self._is_ignored(item_name, config, ignore):
            return None
        if config and config.is_disabled(item_name):
            return None
        if item_name not in self.item_cache:
            if not config or config.default.get('strict', True):
>               raise RuntimeError(f'Procedure {item_name} not found in self.item_cache.')
E               RuntimeError: Procedure #kernel_3 not found in self.item_cache.
```

### DIFF
--- a/loki/transformations/build_system/dependency.py
+++ b/loki/transformations/build_system/dependency.py
@@ -115,6 +115,8 @@ class DependencyTransformation(Transformation):
         """
         role = kwargs.get('role')
 
+        # remember/keep track of the module subroutines (even if some of those are removed)
+        routines = tuple(routine.name.lower() for routine in module.subroutines)
         if role == 'kernel':
             # Change the name of kernel modules
             module.name = self.derive_module_name(module.name)
@@ -137,16 +139,15 @@ class DependencyTransformation(Transformation):
         if self.remove_inactive_items and not kwargs.get('items') is None:
             active_nodes = [item.scope_ir.name.lower() for item in kwargs['items']]
         # rename target names in an access spec for both public and private access specs 
-        variables = module.variables
         if module.public_access_spec:
             module.public_access_spec = self.rename_access_spec_names(
                 module.public_access_spec, targets=targets, active_nodes=active_nodes,
-                variables=variables
+                routines=routines
             )
         if module.private_access_spec:
             module.private_access_spec = self.rename_access_spec_names(
                 module.private_access_spec, targets=targets, active_nodes=active_nodes,
-                variables=variables
+                routines=routines
             )
 
     def transform_subroutine(self, routine, **kwargs):
@@ -344,7 +345,7 @@ class DependencyTransformation(Transformation):
         if import_map:
             source.spec = Transformer(import_map).visit(source.spec)
 
-    def rename_access_spec_names(self, access_spec, targets=None, active_nodes=None, variables=None):
+    def rename_access_spec_names(self, access_spec, targets=None, active_nodes=None, routines=None):
         """
         Rename target names in an access spec
         
@@ -359,20 +360,18 @@ class DependencyTransformation(Transformation):
             Optional list of subroutine names for which to modify access specs
         active_nodes : list of str
             Optional list of active nodes
-        variables : list of :any:`Expression`
-            Optional list of module variables
+        routines : list of :any:`Subroutine`
+            Optional list of subroutines
         """
-        module_variables = variables or ()
+        module_routines = routines or ()
         if active_nodes is not None:
-            access_spec_routines = tuple(elem for elem in access_spec if elem in active_nodes)
-        else:
-            access_spec_routines = tuple(elem for elem in access_spec if elem not in module_variables)
-        access_spec_vars = tuple(elem for elem in access_spec if elem in module_variables)
+            access_spec = tuple(elem for elem in access_spec if elem in active_nodes
+                                or elem.lower() not in module_routines)
         return tuple(
-            f'{elem}{self.suffix}' if not targets or elem in targets
+            f'{elem}{self.suffix}' if elem.lower() in module_routines and (not targets or elem in targets)
             else elem
-            for elem in access_spec_routines
-        ) + access_spec_vars
+            for elem in access_spec
+        )
 
     def rename_interfaces(self, intfs, targets=None):
         """
--- a/loki/transformations/build_system/dependency.py
+++ b/loki/transformations/build_system/dependency.py
@@ -137,13 +137,16 @@ class DependencyTransformation(Transformation):
         if self.remove_inactive_items and not kwargs.get('items') is None:
             active_nodes = [item.scope_ir.name.lower() for item in kwargs['items']]
         # rename target names in an access spec for both public and private access specs 
+        variables = module.variables
         if module.public_access_spec:
             module.public_access_spec = self.rename_access_spec_names(
-                module.public_access_spec, targets=targets, active_nodes=active_nodes
+                module.public_access_spec, targets=targets, active_nodes=active_nodes,
+                variables=variables
             )
         if module.private_access_spec:
             module.private_access_spec = self.rename_access_spec_names(
-                module.private_access_spec, targets=targets, active_nodes=active_nodes
+                module.private_access_spec, targets=targets, active_nodes=active_nodes,
+                variables=variables
             )
 
     def transform_subroutine(self, routine, **kwargs):
@@ -341,7 +344,7 @@ class DependencyTransformation(Transformation):
         if import_map:
             source.spec = Transformer(import_map).visit(source.spec)
 
-    def rename_access_spec_names(self, access_spec, targets=None, active_nodes=None):
+    def rename_access_spec_names(self, access_spec, targets=None, active_nodes=None, variables=None):
         """
         Rename target names in an access spec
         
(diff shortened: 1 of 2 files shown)


loki/transformations/build_system/dependency.py
```diff
@@ -329,6 +333,34 @@ def rename_imports(self, source, imports, targets=None):
...
+                new_access_spec = module.public_access_spec
+            module.public_access_spec = tuple(f'{elem}{self.suffix}' if not targets or elem in targets else
+                    elem for elem in new_access_spec)
+        if module.private_access_spec:
+            if active_nodes is not None:
+                new_access_spec = tuple(elem for elem in module.private_access_spec if elem in active_nodes)
+            else:
+                new_access_spec = module.public_access_spec
+            module.private_access_spec = tuple(f'{elem}{self.suffix}' if not targets or elem in targets
+                    else elem for elem in new_access_spec)
```
reuterbal, line 362: The docstring doesn't tell much... But I would make this utility even more basic, something like
```suggestion
    def rename_access_spec_names(self, access_spec, targets=None, active_nodes=None):
        """
        Rename target names in an access spec
        
        For all names in the access spec that are contained in :data:`targets`, rename them as 
        ``{name}{self.suffix}``. If :data:`active_nodes` are given, then all names
        that are not in the list of active nodes, are being removed from the list.

        Parameters
        ----------
        access_spec : list of str
            List of names from an access spec
        targets : list of str
            Optional list of subroutine names for which to modify access specs
        active_nodes : list of str
            Optional list of active nodes
        """
        if active_nodes:
            access_spec = tuple(elem for elem in access_spec if elem in active_nodes)
        return tuple(
            f'{elem}{self.suffix}' if not targets or elem in targets
            else elem
            for elem in access_spec
        )
```

And then use that utility simply as:
```python
if module.public_access_spec:
    module.public_access_spec = self.rename_access_spec(
        module.public_access_spec, targets=targets, active_nodes=active_nodes
    )

if module.private_access_spec:
    module.private_access_spec = self.rename_access_spec(
        module.private_access_spec, targets=targets, active_nodes=active_nodes
    )
```

loki/transformations/build_system/dependency.py
```diff
@@ -329,6 +333,34 @@ def rename_imports(self, source, imports, targets=None):
...
+                new_access_spec = tuple(elem for elem in module.public_access_spec if elem in active_nodes)
+            else:
+                new_access_spec = module.public_access_spec
+            module.public_access_spec = tuple(f'{elem}{self.suffix}' if not targets or elem in targets else
+                    elem for elem in new_access_spec)
+        if module.private_access_spec:
+            if active_nodes is not None:
+                new_access_spec = tuple(elem for elem in module.private_access_spec if elem in active_nodes)
+            else:
+                new_access_spec = module.public_access_spec
```
reuterbal, line 360: This should likely be `module.private_access_spec`?

loki/transformations/build_system/tests/test_dependency.py
```diff
@@ -121,6 +121,118 @@ def test_dependency_transformation_globalvar_imports(frontend, use_scheduler, tm
...
+            frontend=frontend, xmods=[tmp_path]
+        )
+        assert (
+            Sourcefile.from_source(kernel_fcode, frontend=frontend, xmods=[tmp_path]).to_fortran() ==
+            scheduler.item_factory.item_cache[str(tmp_path/'kernel_mod.F90')].source.to_fortran()
+        )
+
+    else:
+        kernel = Sourcefile.from_source(kernel_fcode, frontend=frontend, xmods=[tmp_path])
+        driver = Sourcefile.from_source(driver_fcode, frontend=frontend, xmods=[tmp_path])
```
reuterbal, line 207: Do we not want the enrichment, here?
```suggestion
        driver = Sourcefile.from_source(driver_fcode, frontend=frontend, xmods=[tmp_path], definitions=kernel.definitions)
```

reuterbal (changes requested):
Looks conceptually great. Just some pointers for the implementation

reuterbal (commented):
Thanks, looks great! While looking through this I noticed a potential problem when global variables (that are not represented as items in the scheduler graph) are part of the accessibility statement. I'm wondering if we might accidentally remove them and thus cause problems, see comments in code.

A simple "solution" might be, when building a list of active nodes to make sure we also keep any module variables that are marked as public in there

MichaelSt98:
Agreed ... @reuterbal: Don't we have a problem/bug even without this PR?

Executing:

```py
from loki import fgen, DependencyTransformation, Sourcefile

if __name__ == "__main__":
    kernel_fcode = """
MODULE kernel_access_spec_mod

  INTEGER, PUBLIC :: some_const
  INTEGER :: another_const

PRIVATE
PUBLIC kernel, kernel_2, unused_kernel, another_const
CONTAINS
    SUBROUTINE kernel(a, b, c)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a, b, c

    call kernel_2(a, b)
    call kernel_3(c)
  END SUBROUTINE kernel
  SUBROUTINE kernel_2(a, b)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a, b

    a = 1
    b = 2
  END SUBROUTINE kernel_2
  SUBROUTINE kernel_3(a)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a

    a = 3
  END SUBROUTINE kernel_3
  SUBROUTINE unused_kernel(a)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a

    a = 3
  END SUBROUTINE unused_kernel
END MODULE kernel_access_spec_mod
    """.strip()

    driver_fcode = """
SUBROUTINE driver(a, b, c)
    USE kernel_access_spec_mod, only: kernel, another_const
    USE kernel_access_spec_mod, only: some_const
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a, b, c

    CALL kernel(a + some_const, b + another_const ,c)
END SUBROUTINE driver
    """.strip()

    transformation = DependencyTransformation(suffix='_test', module_suffix='_mod')

    kernel = Sourcefile.from_source(kernel_fcode)
    driver = Sourcefile.from_source(driver_fcode, definitions=kernel.definitions)

    kernel.apply(transformation, role='kernel')
    driver['driver'].apply(transformation, role='driver', targets=('kernel', 'kernel_access_spec_mod'))

    print(f"{fgen(driver)}")
    print(f"------------------------")
    print(f"{fgen(kernel)}")
```

yields:

```fortran
SUBROUTINE driver (a, b, c)
  USE kernel_access_spec_test_mod, ONLY: kernel_test
  USE kernel_access_spec_mod, ONLY: another_const
  USE kernel_access_spec_mod, ONLY: some_const
  IMPLICIT NONE
  INTEGER, INTENT(INOUT) :: a, b, c
  
  CALL kernel_test(a + some_const, b + another_const, c)
END SUBROUTINE driver
------------------------
MODULE kernel_access_spec_test_mod
  
  private
  PUBLIC :: kernel, kernel_2, unused_kernel, another_const
  INTEGER, PUBLIC :: some_const
  INTEGER :: another_const
  

  CONTAINS
  SUBROUTINE kernel_test (a, b, c)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a, b, c
    
    CALL kernel_2(a, b)
    CALL kernel_3(c)
  END SUBROUTINE kernel_test
  SUBROUTINE kernel_2_test (a, b)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a, b
    
    a = 1
    b = 2
  END SUBROUTINE kernel_2_test
  SUBROUTINE kernel_3_test (a)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a
    
    a = 3
  END SUBROUTINE kernel_3_test
  SUBROUTINE unused_kernel_test (a)
    IMPLICIT NONE
    INTEGER, INTENT(INOUT) :: a
    
    a = 3
  END SUBROUTINE unused_kernel_test
END MODULE kernel_access_spec_test_mod
```

Is 

```fortran
  USE kernel_access_spec_test_mod, ONLY: kernel_test
  USE kernel_access_spec_mod, ONLY: another_const
  USE kernel_access_spec_mod, ONLY: some_const
```

what we want to have?

reuterbal:
Excellent question, I think that depends on the transformed code. Afaik this situation occurs in ecWAM, where you want to duplicate the module routines to supply offload code, but you _don't_ want to duplicate the module variables because they are also used outside in host code. Instead, data offload is applied to the original module, without duplicating. Therefore, the outcome you describe looks correct to me. This may, of course, incur problems if the module variables are used by module procedures.

For the scope of this PR, I would say as long as it retains current behaviour that's fine.

For the future, we should support two modes for these situations.
Consider an example like this:
```
module a_mod
    implicit none
    private
    public :: var, sub
    integer :: var
contains
    subroutine sub(arg)
        integer, intent(inout) :: arg
        arg = arg + var
    end subroutine sub
end module a_mod
```

First mode: global module variables are replicated, i.e., we keep them in the public access spec.

```
module a_loki_mod
    implicit none
    private
    public :: var, sub_loki
    integer :: var
contains
    subroutine sub_loki(arg)
        integer, intent(inout) :: arg
        arg = arg + var
    end subroutine sub_loki
end module a_loki_mod
```

Second mode: global variables are kept in the original module and used from there, i.e., the declaration is replaced by an import:

```
module a_loki_mod
    use a_mod, only: var
    implicit none
    private
    public :: sub_loki
contains
    subroutine sub_loki(arg)
        integer, intent(inout) :: arg
        arg = arg + var
    end subroutine sub
end module a_loki_mod
```
This declaration-to-import change should probably be a separate transformation. Something similar would also be useful for the derived type definitions in ecrad.

MichaelSt98:
Is it necessary to take special care of `module.typedefs` and `module.interfaces` regarding access specifiers?

reuterbal:
Oh, good point. Typedefs presumably in the same way as variables, interfaces should be treated like procedures (so no special handling required).

mlange05 (approved):
Looks great! Many thanks to both for the thorough review and implementation. GTG :shipit:

PR #437, opened 2024-11-11
DataOffload: Add new FieldOffloadTransformation for FIELD API boilerplate injection

wertysas:
This adds a new transformation, ``FieldOffloadTransformation``,   for offloading F-API fields to GPU. Specifically this transformation is meant to operate on CPU driver code that passes view pointers to the kernels. The transformation

1. Removes the view update calls
2. Adds new device pointers for the arrays to offload
3. Adds F-API data offload and sync calls
4. Replaces the old *view-based* arguments in kernel calls with device pointer slices

The corresponding ``dwarf-p-cloudsc`` implementation is on the branch [dwarf-p-cloudsc/je-field-api-offload-v2](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/tree/je-field-api-offload-v2). This adds a new transformation pipeline for converting view based CPU driver code to GPU offloaded F-API code,  ``[pipelines.scc-field]``, and a new Loki target ``dwarf-cloudsc-loki-scc-field``.

### DIFF
--- a/loki/transformations/data_offload.py
+++ b/loki/transformations/data_offload.py
@@ -953,13 +953,13 @@ def get_field_type(a: sym.Array) -> sym.DerivedType:
 
 
 def field_new(field_ptr, data, scope):
-    return ir.CallStatement(sym.ProcedureSymbol('FIELD_NEW', scope=scope),
-                            (field_ptr,), (('DATA', data),))
+    return ir.CallStatement(name=sym.ProcedureSymbol('FIELD_NEW', scope=scope),
+                            arguments=(field_ptr,), kwarguments=(('DATA', data),))
 
 
 def field_delete(field_ptr, scope):
-    return ir.CallStatement(sym.ProcedureSymbol('FIELD_DELETE', scope=scope),
-                            (field_ptr,))
+    return ir.CallStatement(name=sym.ProcedureSymbol('FIELD_DELETE', scope=scope),
+                            arguments=(field_ptr,))
 
 
 class FieldAPITransferType(Enum):
@@ -977,13 +977,13 @@ def field_get_device_data(field_ptr, dev_ptr, transfer_type: FieldAPITransferTyp
     if transfer_type == FieldAPITransferType.WRITE_ONLY:
         suffix = 'WRONLY'
     procedure_name = 'GET_DEVICE_DATA_' + suffix
-    return ir.CallStatement(sym.ProcedureSymbol(procedure_name, parent=field_ptr, scope=scope),
-                            (dev_ptr.clone(dimensions=None),))
+    return ir.CallStatement(name=sym.ProcedureSymbol(procedure_name, parent=field_ptr, scope=scope),
+                            arguments=(dev_ptr.clone(dimensions=None),), )
 
 
 def field_sync_host(field_ptr, scope):
     procedure_name = 'SYNC_HOST_RDWR'
-    return ir.CallStatement(sym.ProcedureSymbol(procedure_name, parent=field_ptr, scope=scope), ())
+    return ir.CallStatement(name=sym.ProcedureSymbol(procedure_name, parent=field_ptr, scope=scope), arguments=())
 
 
 def find_array_arguments(routine, calls):
--- a/loki/transformations/data_offload.py
+++ b/loki/transformations/data_offload.py
@@ -1074,10 +1074,10 @@ class FieldOffloadTransformation(Transformation):
                 yield outarg, self.devptrs[i+start]
 
 
-    def __init__(self, devptr_prefix='LOKI_DEVPTR_', **kwargs):
+    def __init__(self, **kwargs):
         self.devptr_prefix = kwargs.get('devptr_prefix', 'loki_devptr')
-        field_group_types = kwargs.get('field_group_types', ['CLOUDSC_STATE_TYPE', 'CLOUDSC_AUX_TYPE', 'CLOUDSC_FLUX_TYPE'])
         self.assume_deviceptr = kwargs.get('assume_deviceptr', False)
+        field_group_types = kwargs.get('field_group_types', ['CLOUDSC_STATE_TYPE', 'CLOUDSC_AUX_TYPE', 'CLOUDSC_FLUX_TYPE'])
         self.field_group_types = tuple(typename.lower() for typename in field_group_types)
 
     def transform_subroutine(self, routine, **kwargs):
--- a/loki/transformations/data_offload.py
+++ b/loki/transformations/data_offload.py
@@ -7,23 +7,19 @@
 
 from collections import defaultdict
 from itertools import chain
-from enum import Enum
 
 from loki.analyse import dataflow_analysis_attached
 from loki.batch import Transformation, ProcedureItem, ModuleItem
 from loki.expression import Scalar, Array, symbols as sym
 from loki.ir import (
     FindNodes, PragmaRegion, CallStatement, Pragma, Import, Comment,
-    Transformer, pragma_regions_attached, get_pragma_parameters, pragmas_attached,
-    FindInlineCalls, FindVariables, SubstituteExpressions
+    Transformer, pragma_regions_attached, get_pragma_parameters,
+    FindInlineCalls, SubstituteExpressions
 )
-import loki.ir as ir
-from loki.scope import Scope
-from loki.transformations.utilities import find_driver_loops
 from loki.logging import warning
 from loki.tools import as_tuple, flatten, CaseInsensitiveDict, CaseInsensitiveDefaultDict
 from loki.types import BasicType, DerivedType
-
+from loki.transformations.parallel import FieldAPITransferType, field_get_device_data, field_sync_host
 
 __all__ = [
     'DataOffloadTransformation', 'GlobalVariableAnalysis',
@@ -925,66 +921,6 @@ class GlobalVarHoistTransformation(Transformation):
 ################################################################################
 # Field API helper routines
 ################################################################################
-def get_field_type(a: sym.Array) -> sym.DerivedType:
-    """
-    Returns the corresponding FIELD API type for an array.
-
-    This transformation is IFS specific and assumes that the
-    type is an array declared with one of the IFS type specifiers, e.g. KIND=JPRB
-    """
-    type_map = ["jprb",
-                "jpit",
-                "jpis",
-                "jpim",
-                "jpib",
-                "jpia",
-                "jprt",
-                "jprs",
-                "jprm",
-                "jprd",
-                "jplm"]
-
-    type_name = a.type.kind.name
-    assert type_name.lower() in type_map, ('Error array type kind is: '
-                                           f'"{type_name}" which is not a valid IFS type specifier')
-    rank = len(a.shape)
-    field_type = sym.DerivedType(name="field_" + str(rank) + type_name[2:4])
-    return field_type
-
-
-def field_new(field_ptr, data, scope):
-    return ir.CallStatement(name=sym.ProcedureSymbol('FIELD_NEW', scope=scope),
-                            arguments=(field_ptr,), kwarguments=(('DATA', data),))
-
-
-def field_delete(field_ptr, scope):
-    return ir.CallStatement(name=sym.ProcedureSymbol('FIELD_DELETE', scope=scope),
-                            arguments=(field_ptr,))
-
-
-class FieldAPITransferType(Enum):
-    READ_ONLY = 1
-    READ_WRITE = 2
-    WRITE_ONLY = 3
-
-
-def field_get_device_data(field_ptr, dev_ptr, transfer_type: FieldAPITransferType, scope: Scope):
-    assert isinstance(transfer_type, FieldAPITransferType)
-    if transfer_type == FieldAPITransferType.READ_ONLY:
-        suffix = 'RDONLY'
-    if transfer_type == FieldAPITransferType.READ_WRITE:
-        suffix = 'RDWR'
-    if transfer_type == FieldAPITransferType.WRITE_ONLY:
-        suffix = 'WRONLY'
-    procedure_name = 'GET_DEVICE_DATA_' + suffix
-    return ir.CallStatement(name=sym.ProcedureSymbol(procedure_name, parent=field_ptr, scope=scope),
-                            arguments=(dev_ptr.clone(dimensions=None),), )
-
-
-def field_sync_host(field_ptr, scope):
-    procedure_name = 'SYNC_HOST_RDWR'
-    return ir.CallStatement(name=sym.ProcedureSymbol(procedure_name, parent=field_ptr, scope=scope), arguments=())
-
 
 def find_array_arguments(routine, calls):
     """
--- a/loki/transformations/data_offload.py
+++ b/loki/transformations/data_offload.py
@@ -1011,8 +1011,7 @@ class FieldOffloadTransformation(Transformation):
 
 
     def __init__(self, **kwargs):
-        self.devptr_prefix = kwargs.get('devptr_prefix', 'loki_devptr')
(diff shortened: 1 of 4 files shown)


loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+    :targets: collection of :any:`Subroutine`
+        Iterable object of subroutines or functions called
+    :returns: list of :any:`CallStatement`
+    """
+    calls = FindNodes(CallStatement).visit(region)
+    calls = [c for c in calls if str(c.name).lower() in targets]
+    return calls
+
+
+class FieldOffloadTransformation(Transformation):
```
mlange05, line 944: External-facing routines really should have a docstring.

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+    class FieldPointerMap:
```
mlange05, line 945: Similarly, utilities like this need to be explained in a docstring.

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+        @property
+        def out_pairs(self):
+            start = len(self.inargs)+len(self.inoutargs)
+            for i, outarg in enumerate(self.outargs):
+                yield outarg, self.devptrs[i+start]
+
+
+    def __init__(self, **kwargs):
+        self.deviceptr_prefix = kwargs.get('devptr_prefix', 'loki_devptr_')
+        field_group_types = kwargs.get('field_group_types', ['CLOUDSC_STATE_TYPE',
```
mlange05, line 972: I don't think we want CLOUDS-specific defaults here. The natural default should probably be None, and and then converted to an empty tuple.

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+                yield inoutarg, self.devptrs[i+start]
+
+        @property
+        def out_pairs(self):
+            start = len(self.inargs)+len(self.inoutargs)
+            for i, outarg in enumerate(self.outargs):
+                yield outarg, self.devptrs[i+start]
+
+
+    def __init__(self, **kwargs):
```
mlange05, line 970: To make docstrings and interfaces easier to read, it might be better to list the required and optional arguments explicitly here, rather than pulling them out of `**kwargs`

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+        self.offload_index = kwargs.get('offload_index', 'IBL')
+
+    def transform_subroutine(self, routine, **kwargs):
+        role = kwargs['role']
+        targets = as_tuple(kwargs.get('targets'), (None))
+        if role == 'driver':
+            self.process_driver(routine, targets)
+
+    def process_driver(self, driver, targets):
+        remove_field_api_view_updates(driver, self.field_group_types + tuple(s.upper() for s in self.field_group_types))
```
mlange05, line 985: The lower/upper-case conversion here should not be necessary. If `remove_field_api_view_updates` is not sufficiently case-agnostic, it should be fixed there.

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+        with pragma_regions_attached(driver):
+            for region in FindNodes(PragmaRegion).visit(driver.body):
+                # Only work on active `!$loki data` regions
+                if not DataOffloadTransformation._is_active_loki_data_region(region, targets):
+                    continue
+                kernel_calls = find_target_calls(region, targets)
```
mlange05, line 991: [no action] Most of these need not live on the object itself, but could be standalone, re-usable utility methods. But we have bunch of these, so I think we can leave this to a consolidation effort in a follow-on.

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,160 @@ def _append_routine_arguments(self, routine, item):
...
+                offload_map = self.FieldPointerMap(device_ptrs, *offload_variables)
+                self._add_field_offload_calls(driver, region, offload_map)
+                self._replace_kernel_args(driver, kernel_calls, offload_map)
+
+    def find_offload_variables(self, driver, calls):
+        inargs = ()
+        inoutargs = ()
+        outargs = ()
+
+        for call in calls:
```
mlange05, line 1003: [no action] This routine only works for list of calls, and does not work on code regions yet. In a follow-on we might want to do this per-call and add the ability to do regions as well (just a note to self, really).

loki/transformations/parallel/field_api.py
```diff
@@ -150,3 +152,58 @@ def visit_Loop(self, loop, **kwargs):  # pylint: disable=unused-argument
...
+        raise TypeError(f"transfer_type must be of type FieldAPITransferType, but is of type {type(transfer_type)}")
+    if transfer_type == FieldAPITransferType.READ_ONLY:
+        suffix = 'RDONLY'
+    elif transfer_type == FieldAPITransferType.READ_WRITE:
+        suffix = 'RDWR'
+    elif transfer_type == FieldAPITransferType.WRITE_ONLY:
+        suffix = 'WRONLY'
+    else:
+        suffix = ''
+    procedure_name = 'GET_DEVICE_DATA_' + suffix
```
mlange05, line 202: [no action] Will need a toggle for GET_HOST_DATA_<suffix> soon....

loki/transformations/parallel/field_api.py
```diff
@@ -150,3 +152,58 @@ def visit_Loop(self, loop, **kwargs):  # pylint: disable=unused-argument
...
+
+
+
+class FieldAPITransferType(Enum):
+    READ_ONLY = 1
+    READ_WRITE = 2
+    WRITE_ONLY = 3
+
+
+def field_get_device_data(field_ptr, dev_ptr, transfer_type: FieldAPITransferType, scope: Scope):
```
mlange05, line 191: Docstring would be helpful here.

loki/transformations/tests/test_data_offload.py
```diff
@@ -42,10 +43,45 @@ def fixture_config():
...
+def fixture_parkind_mod(tmp_path, frontend):
+    fcode = """
+    module parkind1
+      integer, parameter :: jprb=4
+    end module
+    """
+    return Module.from_source(fcode, frontend=frontend, xmods=[tmp_path])
+
+@pytest.fixture(name="field_module")
+def fixture_field_module(tmp_path, frontend):
```
mlange05, line 56: Neat! 😏

loki/transformations/tests/test_data_offload.py
```diff
@@ -81,11 +117,25 @@ def test_data_offload_region_openacc(frontend, assume_deviceptr):
     end do
   END SUBROUTINE kernel_routine
 """
-    driver = Sourcefile.from_source(fcode_driver, frontend=frontend)['driver_routine']
-    kernel = Sourcefile.from_source(fcode_kernel, frontend=frontend)['kernel_routine']
-    driver.enrich(kernel)
-
-    driver.apply(DataOffloadTransformation(assume_deviceptr=assume_deviceptr), role='driver',
+    with caplog.at_level(log_levels['ERROR']):
```
mlange05, line 120: I'd keep the caplog context tightly within the "expected failure" conditional below.

loki/transformations/tests/test_data_offload.py
```diff
@@ -805,3 +859,457 @@ def test_transformation_global_var_derived_type_hoist(here, config, frontend, ho
...
+                                            offload_index='i',
+                                            field_group_types=['state_type']),
+                 role='driver',
+                 targets=['kernel_routine'])
+
+    calls = FindNodes(CallStatement).visit(driver.body)
+    kernel_call = next(c for c in calls if c.name=='kernel_routine')
+
+    # verify that field offloads are generated properly
+    in_calls = [c for c in calls if 'get_device_data_rdonly' in c.name.name.lower()]
```
mlange05, line 929: Is case-sensitive comparison really needed here? Same goes for a bunch of lines below.

loki/transformations/tests/test_data_offload.py
```diff
@@ -805,3 +859,457 @@ def test_transformation_global_var_derived_type_hoist(here, config, frontend, ho
...
+        do i=1,nlev
+            call state%update_view(i)
+            call kernel_routine(nlon, nlev, a, state%b, state2%c)
+        end do
+        !$loki end data
+
+      end subroutine driver_routine
+    end module driver_mod
+    """
+    with caplog.at_level(log_levels['WARNING']):
```
mlange05, line 1296: Again, ideally keep the caplog tightly around the call you're expecting to trigger this. Otherwise unrelated constructor changes can make this test fail very easily.

mlange05 (changes requested):
Excellent contribution and very nicely structured and test; many thanks! This is a great start towards a very comprehensive set of tools for dealing with FIELD API boilerplate.

I've left a bunch of cosmetic comments (mostly docstrings, and minor style things), and a few notes to myself. The reason for the latter is that this one does not yet cover more general cases with compute code in the driver loop, and it will also likely require some restructuring for the utilities to be used outside the `DataOffloadTransformation` - but all of these will happen in a follow-on consolidation effort and should not stop this.

loki/transformations/tests/test_data_offload.py
wertysas: Yes, actually the tests fails without case-sensitive comparison, because the generated ``ProcedureSymbol``s actually have uppercase names.

mlange05: Ahhh, I see, and they have parents, so `==` comparison doesn't work. Yeah, ok, makes sense then. Please leave as is.

loki/transformations/tests/test_data_offload.py
mlange05: Ahhh, ignore this! I had misunderstood the caplog fixture and the log-level management in `.at_level`. Please ignore and leave this as is.

loki/transformations/tests/test_data_offload.py
mlange05: Ignore, I had misunderstood how caplog works. This is fine.

mlange05 (approved):
Very nice! And many thanks for addressing this so promptly. GTG from me. :shipit:

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,224 @@ def _append_routine_arguments(self, routine, item):
...
+            if call.routine is BasicType.DEFERRED:
+                error(f'[Loki] Data offload: Routine {driver.name} has not been enriched ' +
+                        f'in {str(call.name).lower()}')
+                raise RuntimeError
+            for param, arg in call.arg_iter():
+                if not isinstance(param, Array):
+                    continue
+                try:
+                    parent = arg.parent
```
awnawab, line 1076: Couldn't we do this with a `parent = arg.getattr('parent', None)` and print the warning if `not parent`?

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,224 @@ def _append_routine_arguments(self, routine, item):
...
+        outargs = tuple(set(outargs))
+        return inargs, inoutargs, outargs
+
+
+    def _declare_device_ptrs(self, driver, offload_variables):
+        device_ptrs = tuple(self._devptr_from_array(driver, a) for a in chain(*offload_variables))
+        driver.variables += device_ptrs
+        return device_ptrs
+
+    def _devptr_from_array(self, driver, a: sym.Array):
```
awnawab, line 1108: Really love the type hints ❤️

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,224 @@ def _append_routine_arguments(self, routine, item):
...
+        host_to_device += tuple(field_get_device_data(self._get_field_ptr_from_view(inarg), devptr,
+                                FieldAPITransferType.READ_WRITE, driver) for inarg, devptr in offload_map.out_pairs)
+        device_to_host = tuple(field_sync_host(self._get_field_ptr_from_view(inarg), driver)
+                               for inarg, _ in chain(offload_map.inout_pairs, offload_map.out_pairs))
+        update_map = {region: host_to_device + (region,) + device_to_host}
+        Transformer(update_map, inplace=True).visit(driver.body)
+
+    def _get_field_ptr_from_view(self, field_view):
+        type_chain = field_view.name.split('%')
+        field_type_name = 'F_' + type_chain[-1]
```
awnawab, line 1136: [noaction] this is a sensible default, which we might want to make overridable per derived-type.

loki/transformations/data_offload.py
```diff
@@ -908,3 +924,224 @@ def _append_routine_arguments(self, routine, item):
...
+        return field_view.parent.get_derived_type_member(field_type_name)
+
+    def _replace_kernel_args(self, driver, kernel_calls, offload_map):
+        change_map = {}
+        offload_idx_expr = driver.variable_map[self.offload_index]
+        for arg, devptr in chain(offload_map.in_pairs, offload_map.inout_pairs, offload_map.out_pairs):
+            dims = (sym.RangeIndex((None, None)),) * (len(devptr.shape)-1) + (offload_idx_expr,)
```
awnawab, line 1143: Should we not clone `arg.dims` here and add the `offload_idx_expr` to it? We don't always want to pass the full array, we might for example pass `FIELD_PTR(:,1,IBL)` if the field is 3D  but the dummy argument is 2D. These are probably mistakes that we should fix in source, but the transformation should nevertheless support this edge-case.

loki/transformations/parallel/field_api.py
```diff
@@ -150,3 +152,85 @@ def visit_Loop(self, loop, **kwargs):  # pylint: disable=unused-argument
...
+    
+    Parameters
+    ----------
+    field_ptr: pointer to field object
+        Pointer to the field to call ``SYNC_HOST`` from.
+    scope: :any:`Scope`
+        Scope of the created :any:`CallStatement`
+    """
+
+    procedure_name = 'SYNC_HOST_RDWR'
```
awnawab, line 235: [noaction] we might want to mirror the suffix logic of `field_get_device_data` here.

awnawab (changes requested):
Many thanks for this superb piece of work 👌 The implementation is really neat and well tested 🙏 I especially love your dedication to type hints (even though I am too lazy to adhere to it myself 😇 ).

I've left a few comments, mostly reminders for us/myself for the future but there are a couple which I would like you to please address.

loki/transformations/tests/test_data_offload.py
wertysas: Yeah! It was a good suggestion to use a fixture for this to reduce boilerplate in the tests.

loki/transformations/data_offload.py
```diff
@@ -49,6 +52,12 @@ def __init__(self, **kwargs):
         self.has_data_regions = False
         self.remove_openmp = kwargs.get('remove_openmp', False)
         self.assume_deviceptr = kwargs.get('assume_deviceptr', False)
+        self.assume_acc_mapped = kwargs.get('assume_acc_mapped', False)
```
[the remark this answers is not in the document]
wertysas: I agree, updated it according to your suggestion now.

loki/transformations/data_offload.py
wertysas: Yes, but I left it as is now, since ``getattr`` does something similar under the hood.

loki/transformations/data_offload.py
wertysas: Nice catch! I had missed that. Fixed it now.

awnawab (approved):
Thanks a lot for addressing the changes so promptly 🙏  looks great 👌

Issue #438, opened 2024-11-15
bug with statement functions 


ecossevin:
### What happened?

I have a bug in routines with statement functions : FOLH in this case. 

The code has a strange behavior, in my test case there are two fortran routines, one with comment and one without. The bug is occuring in one case and not in another.
The bug seems to be due to FOLH type that is lost at some point... 

I mentioned already this problem in the loki chat a few weeks ago.

I'm using the lastest version of loki with Python 3.10.12. 

[bug_statement.tar.gz](https://github.com/user-attachments/files/17781756/bug_statement.tar.gz)

I have other test cases, that are bigger. And i still have strange behavior : bug when i remove fortran comments or not, bug if i do a print inside python loki or not.

### What are the steps to reproduce the bug?

untar the attached file
load loki venv
run python3 test.py   :: change the `file_name` variable in test.py. 

### Version

latest, from today

### Platform (OS and architecture)

Linux pxdesrtls020 6.8.0-47-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct  2 16:16:55 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell
dbg1 FOLH
Traceback (most recent call last):
  File "/home/cossevine/bug_loki/substitute/test.py", line 19, in <module>
    routine_t1 = routine.clone()
  File "/home/cossevine/tmp/loki_env_test/lib/python3.10/site-packages/loki/subroutine.py", line 293, in clone
    return super().clone(**kwargs)
  File "/home/cossevine/tmp/loki_env_test/lib/python3.10/site-packages/loki/program_unit.py", line 450, in clone
    obj = super().clone(**kwargs)
  File "/home/cossevine/tmp/loki_env_test/lib/python3.10/site-packages/loki/scope.py", line 343, in clone
    return type(self)(**kwargs)
  File "/home/cossevine/tmp/loki_env_test/lib/python3.10/site-packages/loki/subroutine.py", line 89, in __init__
    self.__initialize__(
  File "/home/cossevine/tmp/loki_env_test/lib/python3.10/site-packages/loki/subroutine.py", line 117, in __initialize__
    super().__initialize__(
  File "/home/cossevine/tmp/loki_env_test/lib/python3.10/site-packages/loki/program_unit.py", line 96, in __initialize__
    self.rescope_symbols()
[...]

PR #439, opened 2024-11-17
Sanitise: Update scope on unchanged expressions in remove_associates

mlange05:
The default case is the one, where the symbol is associates by the closest associate, but that associates is about to be removed. So we simply update the scope to ensure nothing becomes unscoped.

Note, I believe this should fix issue #438 . The PR includes a much simplified test case.


reuterbal (approved):
Many thanks, this looks great and has been confirmed to fix the reported issue!

PR #440, opened 2024-11-19
JoinableStringList: Do not break lines within quoted strings

reuterbal:
The line break heuristic in JoinableStringList did not account for quoted strings, sometimes putting line breaks into the middle of a string.
This caused issues with an initialiser list for a character array, where the list entries contained white spaces.

The fix is to search for quoted strings in the remainder string and treat them as unbreakable entities, while retaining the same splitting heuristic as before outside of quoted strings. This allowed to remove a four year old "TODO" comment that points out that the above problem might occur... 🤦 

Two tests are included to validate the behaviour directly on the class as well as in the original Fortran context where the problem first occured.

Piggy-backed is also a small hotfix to the CMake layer to fix an issue that occured if no frontend was chosen explicitly.

### DIFF
--- a/cmake/loki_transform_helpers.cmake
+++ b/cmake/loki_transform_helpers.cmake
@@ -132,7 +132,7 @@ macro( _loki_transform_env_setup )
     set( _LOKI_TRANSFORM_ENV )
     set( _LOKI_TRANSFORM_PATH )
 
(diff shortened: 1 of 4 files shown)


mlange05 (approved):
Looks great and seems to work as intended. GTG from me.

PR #441, opened 2024-11-20
CMake plan and enrichment bugfix

reuterbal:
This fixes two errors spotted with complex Fortran module dependencies:

1. When including module files in transformations, we had inconsistent behaviour between the files in a dependency graph that are going to be written, and those that are listed in the CMake plan. 74409002d6bd05e83504804806a92ae393e17186 applies the methodology that is used to derive the file graph for the `FileWriteTransformation` also when deriving the file list in the CMake plan writing, thus ensuring consistent behaviour.
2. For indirect imports, e.g., when importing a symbol from a module, where the symbol itself is only available because of an import from a third module, enrichment failed non-gracefully because we had the presumed origin module available but that module didn't have the symbol information available. e9a77f425bbc884fff26f0e3b3533c4c5edd60ed fixes the failure by not enriching the symbol, which is the correct approach if the information is not available. A corresponding test is supplied.

### DIFF
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -635,8 +635,15 @@ class Scheduler:
         sources_to_transform = []
 
         # Filter the SGraph to get a pure call-tree
-        item_filter = None if self.config.enable_imports else ProcedureItem
-        for item in SFilter(self.sgraph, item_filter=item_filter):
+        item_filter = ProcedureItem
+        if self.config.enable_imports:
+            item_filter = as_tuple(item_filter) + (ModuleItem,)
+        graph = self.sgraph.as_filegraph(
+            self.item_factory, self.config, item_filter=item_filter,
+            exclude_ignored=True
+        )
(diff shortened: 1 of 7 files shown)


reuterbal:
I added a test for the FileWriteTransformation/CMake planning - I had this but forgot to push it previously. This verifies the aligned behaviour between the two.

reuterbal:
Thanks @awnawab for the pointer regarding the `replicate` property on items. This turned out to be indeed the issue that broke the ecwam regression test. I have now added a test that verifies this behaviour in Loki's testbase itself.

Similarly to the `is_ignored` property, we had to propagate this property upward to the `FileItem` to make sure this property is honoured when iterating over the file graph.

mlange05 (approved):
Yes, looks great. GTG from me. :shipit:

awnawab (approved):
Many thanks for this, removing this inconsistency will really help to clean up some of my config files 🙏 G2G 👌

PR #442, opened 2024-11-21
DataOffload: fix generation of offload pragmas

awnawab:
A small fix to how the global variable offload pragmas are generated which allows them to be split into multiple lines by `fgen`.

### DIFF
--- a/loki/transformations/data_offload.py
+++ b/loki/transformations/data_offload.py
@@ -484,7 +484,7 @@ class GlobalVarOffloadTransformation(Transformation):
         offload_variables = offload_variables - declared_variables
         if offload_variables:
             module.spec.append(
-                Pragma(keyword='acc', content=f'declare create({",".join(v.name for v in offload_variables)})')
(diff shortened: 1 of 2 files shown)


mlange05 (approved):
Yes, makes sense and looks good to me. Approving now, but needs rebase once Pydantic version compatibility is fixed.

reuterbal (commented):
Thanks! The fix looks good and also personally I'm of the opinion that white space behind commas is not a luxury that should lay beyond us :)

The question is whether we wouldn't want to make commas itself a delimiter for chunking in `JoinableStringList`:
https://github.com/ecmwf-ifs/loki/blob/72d35fbb91b783f50d2df66fce4553bcea1c33e0/loki/tools/strings.py#L57

This pattern governs the split points in strings when long lines need to be wrapped.

I have pushed a rebase to include the pydantic compatibility change.

reuterbal (approved):

PR #444, opened 2024-11-21
Started on: making Loki plan trafo (pipeline) aware

MichaelSt98:
Corresponding CLOUDSC branch: **nams-cloudsc-loki-plan** (necessary since CLOUDSC wasn't using `loki_transform_target` yet).

* more or less just copied/utilised the infrastructure and mechanism for (dispatching) transformations and duplicated for planning
* the only trafo that uses this for now is the `FileWriteTransformation`
    * consequently `write_cmake_plan` is not renaming the modules/routines anymore

Next step: 

* create new item (and/or delete existing item)
    * I've added `cloudsc_init.F90`  in the above CLOUDSC branch as well as a `DuplicateKernel` and `RemoveKernel` trafo to test that (obviously not working yet)

### DIFF
--- a/loki/batch/item.py
+++ b/loki/batch/item.py
@@ -517,6 +517,12 @@ class Item(ItemConfig):
         """
         return self.source.path
 
+    @property
+    def orig_path(self):
+        """
+        The filepath of the associated source file
+        """
+        return self.source.orig_path
 
 class FileItem(Item):
     """
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -508,6 +508,120 @@ class Scheduler:
             self._discover()
             self._parse_items()
 
+    def process_plan(self, transformation):
+        """
+        """
+        if isinstance(transformation, Transformation):
+            self.process_plan_transformation(transformation=transformation)
+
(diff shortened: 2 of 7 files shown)


reuterbal (commented):
Many thanks for this @MichaelSt98 !

I looked through this yesterday and noticed the need for significant code duplication if we make the planning a separate traversal mode in the Scheduler. So, I've played around with an approach, where planning becomes an option in the Scheduler's `process` and Transformation's `apply` entry points, with the latter taking care of dispatching to `plan_<...>` instead of `transform_<...>`. For me, that has the advantage that the planning mode really becomes closely linked to the processing of pipelines/transformations. 

However, I have not gone as far as testing your `DuplicateKernel`/`RemoveKernel` transformations or CLOUDSC in this setup.

You can take a look at the changes here:
https://github.com/ecmwf-ifs/loki/compare/main...nabr-pipeline-plan

~~Note that this sits on top of my changes from #441, because that allowed me to utilize the additional tests added there.~~ I have rebased the branch following the merge, the diff is now clean.

FYI: @mlange05

MichaelSt98:
Out-dated and superseded due to/by #453.

PR #445, opened 2024-11-22
Inline: Fix rescoping of intrinsic procedure symbols in elementals

mlange05:
This caused problems by a race condition, where the elemental scope that was the original scope of intrisincs could be re-build, making the scope weakref invalid before the final catch-all rescoping.

To fix this, I'm explicitly foxing the function body to be rescoped, before it gets inserted. This was problematic, as intrisic procedure symbols were not updated correctly, so I enforce indiscriminant rescoping for intrisic procedure symbols to the given, closest scope in `AttachScopesMapper`.


reuterbal (approved):
Thanks, these are notoriously hard to find and test. Much appreciate the quick fix!

PR #446, opened 2024-11-22
Offload: Refactor `loki.transformations.data_offload` into separate sub-package

mlange05:
~_Currently sits on top of #442 , so I'll only activate it once this is merged and rebased._~

Small housekeeping PR that touches a lot of lines, but changes nothing but imports (I promise, but please double-check).

The idea is to split the 5 different transformation classes in `loki.transformation.data_offload` into `offload.py`, `global_var.py`, `field_api.py`. The tests have been split accordingly and once source set needed moving.

I've also piggybacked a small move of the `InlineTransformation` in `loki.transformations.inline` into a separate `inline.transformation` file. The reason for this is that the `__all__ =[...]` statement was overriding the `*` imports in `__init__.py` and thus certain symbols were not propagated to the outside correctly (that'll teach me! 😉 )

EDIT: Due to aliasing test names between `transformations.parallel` and `transformations.data_offload`, I started wrapping test sub-ackages via `__init__.py`, as per https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html . As a result, the linter now picks up the tests too, so a small linter pass was needed. I've only done this for the `transformation` sub-package, and leave the rest for a follow-on.


reuterbal (approved):
Many thanks, changes look good to me and the sub-package organisation is very neat.

awnawab (approved):
Thanks a lot for this much needed reorganisation 👌 data_offload looks way nicer now, and ready for us to put a lot more things in it 🤓

loki/transformations/inline/transformation.py
wertysas, line 1: I am wondering why this file is named transformation.py? To me it seems to break the naming pattern with the the DataOffloadTrafo being in offload.py

wertysas (commented):
I like these design changes! 😃 

The only comment is that the name ``transformation.py`` for the file containing ``InlineTransformation`` seems inconsistent to me.

loki/transformations/inline/transformation.py
mlange05: That's because the inline sub-package has a different layout; it's largely a set of different methods, with a single `Transformation` as the single entry entry point with a lot of flags. I don't think the naming pattern is set in stone, and might change across sub-packages, I just wanted to avoid `inline.inline.py` 😉

PR #447, opened 2024-11-22
Loki: Turn test sub-directories into sub-pacakges

mlange05:
This make test integration cleaner and avoids eventual namespace clashes in the future. It also now runs the linter over tests, without exporting test symbols as part of `from loki import *` (please double-check this). 

Note, the other half of this is hidden in PR #446 , where this was necessary to avoid namespace clashes in tests.

EDIT: Looks like I have to leave `loki.lint.tests` untouched, as the introduction of the test sub-packages makes the dynamic `rules` imports go haywire. I leave that for a follow-on PR.


reuterbal (approved):
Many thanks for this, changes look good and the linter-compliance pass on the tests is very much appreciated!

Issue #428, opened 2024-11-05
Enrichment of module fails to update the TypeDef information of imported types


wertysas:
### What happened?

The `ProgramUnit.enrich` method does not update the type information in the enriched module with the type information of other modules properly.

### What are the steps to reproduce the bug?

Consider the example below where the module `state_type_mod` defines a derived type `state_type` and is imported in the module `driver_mod`.

```python
from loki.sourcefile import Sourcefile

fcode_state_mod = """
module state_type_mod
  implicit none

  type state_type
    real, pointer, dimension(:,:) :: a
  end type state_type

end module state_type_mod
"""
fcode_driver_mod = """
module driver_mod
  use state_type_mod, only: state_type
  implicit none

contains
  subroutine driver_routine(state)
    type(state_type), intent(inout) :: state

    state%a = 1

  end subroutine driver_routine
end module driver_mod
"""
state_mod = Sourcefile.from_source(fcode_state_mod)['state_type_mod']
driver_mod = Sourcefile.from_source(fcode_driver_mod)['driver_mod']
# # enrichment of driver_mod with state_type_mod
driver_mod.enrich([state_mod], recurse=True)
driver = driver_mod['driver_routine']
driver.enrich([state_mod])

state = driver.variable_map['state']
print(state.type.dtype.typedef)     # BasicType.DEFERRED even after driver being enriched

```

Executing the above example code shows that the driver has not been enrichened and that the type of the argument `state` is still not known in the subroutine `driver_routine`:

```bash
(loki-venv):  python minimal_reproducible_enrich_failure.py
BasicType.DEFERRED
```

The expected output would be `TypeDef:: state_type` and not `BasicType.DEFERRED`. 

If the `state_type` and subroutine are defined in the same module, the output will instead be `TypeDef::state_type` as expected (see example below)

```python
from loki.sourcefile import Sourcefile

fcode_driver_mod = """
module driver_mod
  use state_type_mod, only: state_type
  implicit none

  type state_type
    real, pointer, dimension(:,:) :: a
  end type state_type

contains
  subroutine driver_routine(state)
    type(state_type), intent(inout) :: state

    state%a = 1

  end subroutine driver_routine
end module driver_mod
"""
driver_mod = Sourcefile.from_source(fcode_driver_mod)['driver_mod']
driver = driver_mod['driver_routine']

state = driver.variable_map['state']
print(state.type.dtype.typedef)     # TypeDef::state_type
```

### Version

Loki main branch

[...]

PR #448, opened 2024-11-24
Module: Fix enrichment of type info via `Module` imports

mlange05:
This addresses issue #428 . Many thanks to @wertysas for the excellent reproducer and test

The problem was essentially that we would drop type info for subroutines if the info enrichment came though type imports on the parent `Module`. This is easily fixed by including parent imports in the enrichment loop over Imports.

I took the opportunity to also clean up the import on `test_module.py` in the process.


reuterbal (approved):
Fantastic, many thanks for the excellently described issue @wertysas and neat fix @mlange05!

PR #449, opened 2024-11-24
SGraph (re-)build optimisations

reuterbal:
While profiling the building of the SGraph, I noticed two things:

1. A non-negligible amount of time was spent on checking if items are part of any of the exclusion lists. The reason that this is costly is the pattern support, which incurs some overhead for repeatedly compiling search patterns. This can be partially reduced by using `fnmatch.filter` instead of a comprehension with `any`. On top of that, I tried reducing the amount of pattern compilations overall via some reordering or caching, however, since `fnmatch` already comes with a built-in pattern cache, this did worsen performance in all cases.
2. The `Module` class includes module variables in the list of definitions, which is correct from a Fortran point-of-view but not required for the Scheduler, where we don't capture explicitly the dependency on individual variables but only the module itself. Hence, searching directly only for procedures, interfaces and typedefs brings the search effort down considerably.

Depending on the structure of the graph, this dropped graph building time by 10-20% for me. Much bigger gains could be achieved by caching definitions or dependencies, but this seems a little trickier to do.

### DIFF
--- a/loki/batch/configure.py
+++ b/loki/batch/configure.py
@@ -5,7 +5,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-from fnmatch import fnmatch
+import fnmatch
 from itertools import accumulate
 from pathlib import Path
 import re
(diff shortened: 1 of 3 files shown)


mlange05 (commented):
Changes look good to me and local ecphys regression also seemed fine. Small, but noticeable improvements during Loki processing - so very much appreciate this. 

Will formally approve once test case green, but otherwise GTG from me.

reuterbal:
Thanks! I fixed the failing test - turns out the new logic for yielding definitions for Module items removes some redundant reported items.

mlange05 (approved):
Fantastic! Confirmed still functional and faster than before. All GTG. :shipit:

PR #453, opened 2024-11-29
CMake Plan: Pipeline dry-run and transformation-based plan creation

reuterbal:
This is an alternative implementation suggestion to #444 

The objective here would be to retain current functionality and implement support for transformations that change scheduler graph topology in subsequent PRs.

Please have a look and point out anything that looks odd to you. In particular, please consider the interface - e.g., is a `plan` argument acceptable and the naming ok?

With this, the `loki-transform.py plan` command is redirected to the `loki-transform.py convert` command, clearing the way for subsequent removal of the custom transformation entry point and further homogenisation in the CLI and CMake layer.

I will also add another commit to add some more documentation (and I expect the linter pass to fail due to the local import).

### DIFF
--- a/loki/batch/transformation.py
+++ b/loki/batch/transformation.py
@@ -128,6 +128,11 @@ class Transformation:
             Keyword arguments for the transformation.
         """
 
+    def plan_subroutine(self, routine, **kwargs):
+        """
+        ...
+        """
+
     def transform_module(self, module, **kwargs):
         """
         Defines the transformation to apply to :any:`Module` items.
@@ -142,6 +147,11 @@ class Transformation:
             The module to be transformed.
         **kwargs : optional
             Keyword arguments for the transformation.
+        """
+
+    def plan_module(self, module, **kwargs):
+        """
+        ...
         """
 
     def transform_file(self, sourcefile, **kwargs):
@@ -160,7 +170,12 @@ class Transformation:
             Keyword arguments for the transformation.
         """
 
-    def apply(self, source, post_apply_rescope_symbols=False, **kwargs):
+    def plan_file(self, sourcefile, **kwargs):
+        """
+        ...
+        """
+
+    def apply(self, source, post_apply_rescope_symbols=False, plan=False, **kwargs):
         """
         Dispatch method to apply transformation to :data:`source`.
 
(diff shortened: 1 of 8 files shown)


MichaelSt98:
I skimmed it briefly and I really like the implementation!

Would it make sense to have instead of `process(self, transformation, plan=False)` something like `process(self, transformation, mode=<batch|normal|plan|...>)`?

reuterbal:
> Would it make sense to have instead of `process(self, transformation, plan=False)` something like `process(self, transformation, mode=<batch|normal|plan|...>)`?

That's not a bad idea, however, `mode` as a name is also used to specify the "transformation mode", i.e., `scc`, `scc-stack` etc. Not sure if we would want to avoid that conflict?

Other than that, what would the various options be? I can think of:
- `plan`: The dry-run mode from this PR
- `atomic` (not sure that's a good name): The current mode of applying each transformation to every item before applying the next transformation
- `batch`: Could be a future mode that applies to an item all transformations that share the same traversal mode before going to the next item, thus "batching" the processing. Out of scope for this PR, just to highlight a potential future mode.
- `default`: Should then alias to whatever we think is the best default, e.g. `atomic` right now?

MichaelSt98:
> > Would it make sense to have instead of `process(self, transformation, plan=False)` something like `process(self, transformation, mode=<batch|normal|plan|...>)`?
> 
> That's not a bad idea, however, `mode` as a name is also used to specify the "transformation mode", i.e., `scc`, `scc-stack` etc. Not sure if we would want to avoid that conflict?
> 
> Other than that, what would the various options be? I can think of:
> 
> * `plan`: The dry-run mode from this PR
> * `atomic` (not sure that's a good name): The current mode of applying each transformation to every item before applying the next transformation
> * `batch`: Could be a future mode that applies to an item all transformations that share the same traversal mode before going to the next item, thus "batching" the processing. Out of scope for this PR, just to highlight a potential future mode.
> * `default`: Should then alias to whatever we think is the best default, e.g. `atomic` right now?

Naming things is always the hardest part ...
What about `process_mode`?
Yes, those options are what I was thinking about and I think `atomic` is a good name, however, we could also go with `serial`?
Moreover, we could think about calling `transform_file`/`transform_routine` or `plan_file`/`plan_routine` in dependence of `mode`/`process_mode` like calling `<process_mode>_file`/`<process_mode>_routine`, but this would probably be over-engineering right now...

reuterbal:
Thanks, I went for `proc_strategy` and list available options in an enum `ProcessingStrategy`. That seemed most appropriate to me while minimising accidental aliasing with other meanings.

Inside transformations I kept the boolean option because there it really is just the switch between "transformation" and "planning".

I fixed the linter warnings and added some documentation, so I think this should be ready for another look.

loki/batch/transformation.py
```diff
@@ -206,22 +258,27 @@ def apply_file(self, sourcefile, **kwargs):
...
         items = kwargs.pop('items', None)
         role = kwargs.pop('role', None)
         targets = kwargs.pop('targets', None)
 
         # Apply file-level transformations
-        self.transform_file(sourcefile, item=item, role=role, targets=targets, items=items, **kwargs)
+        if plan_mode:
+            self.plan_file(sourcefile, item=item, role=role, targets=targets, items=items, **kwargs)
+        else:
+            if not plan_mode and sourcefile._incomplete:
```
MichaelSt98, line 278:
```suggestion
            if sourcefile._incomplete:
```

double-checking for `plan_mode` to be `False`?!

MichaelSt98 (approved):
Really nice!

mlange05 (approved):
Agreed, very neat and a great step towards making this all a bit more accessible. GTG from me. :shipit:

PR #455, opened 2024-12-02
Sanitise: Resolve free range indices when resolving associates

mlange05:
The current `Associate` resolution does not account for array index ranges if they are specific in the associate clauses (index remapping). This PR adds a partial resolution step to `ResolveAssociates`, which resolves free index ranges (`:`) in association clause against explicit indices from the original expression, should should both exist. The respective changes in the expression mapper require some care with respect to early termination and recursion into `.dimension` and `.type` properties of the given array expression.

Note: It does not resolve index shifts, where offsets are introduced by using bounded ranges (eg. `3:8`). Instead it warns about such (ab)use.

### DIFF
--- a/loki/transformations/sanitise/associates.py
+++ b/loki/transformations/sanitise/associates.py
@@ -158,7 +158,9 @@ class ResolveAssociateMapper(LokiIdentityMapper):
 
     def map_array(self, expr, *args, **kwargs):
         """ Partially resolve dimension indices and handle shape """
-        new = self.map_scalar(expr, *args, **kwargs)
+
+        # Recurse over existing array dimensions
+        expr_dims = self.rec(expr.dimensions, *args, **kwargs)
 
         # Recurse over the type's shape
         _type = expr.type
@@ -166,14 +168,19 @@ class ResolveAssociateMapper(LokiIdentityMapper):
             new_shape = self.rec(expr.type.shape, *args, **kwargs)
             _type = expr.type.clone(shape=new_shape)
 
+        # Stop if scope is not an associate
+        if not isinstance(expr.scope, ir.Associate):
+            return expr.clone(dimensions=expr_dims, type=_type)
+
+        new = self.map_scalar(expr, *args, **kwargs)
+
         # Recurse over array dimensions
         if isinstance(new, sym.Array):
             # Resolve unbound range symbols form existing indices
             new_dims = self.rec(new.dimensions, *args, **kwargs)
-            new_dims = self._match_range_indices(new_dims, expr.dimensions)
+            new_dims = self._match_range_indices(new_dims, expr_dims)
         else:
-            # Recurse over existing array dimensions
-            new_dims = self.rec(expr.dimensions, *args, **kwargs)
+            new_dims = expr_dims
 
         return new.clone(dimensions=new_dims, type=_type)
 
--- a/loki/transformations/sanitise/associates.py
+++ b/loki/transformations/sanitise/associates.py
@@ -175,7 +175,7 @@ class ResolveAssociateMapper(LokiIdentityMapper):
         new = self.map_scalar(expr, *args, **kwargs)
 
         # Recurse over array dimensions
-        if isinstance(new, sym.Array):
+        if isinstance(new, sym.Array) and new.dimensions:
             # Resolve unbound range symbols form existing indices
             new_dims = self.rec(new.dimensions, *args, **kwargs)
             new_dims = self._match_range_indices(new_dims, expr_dims)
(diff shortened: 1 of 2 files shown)


loki/transformations/sanitise/associates.py
```diff
@@ -143,17 +157,31 @@ def map_scalar(self, expr, *args, **kwargs):
...
+
         # Recurse over array dimensions
-        new_dims = self.rec(expr.dimensions, *args, **kwargs)
+        if isinstance(new, sym.Array) and new.dimensions:
+            # Resolve unbound range symbols form existing indices
+            new_dims = self.rec(new.dimensions, *args, **kwargs)
+            new_dims = self._match_range_indices(new_dims, expr_dims)
+        else:
+            new_dims = expr_dims
+
```
reuterbal, line 184: I'm not sure I'm following the logic here. The first recursion to expr_dims seems to only be used if the array's scope isn't an associate - could we not move it into the conditional just before the return statement?

Or alternatively, in which situation do we expect the return value of the recursion on `new.dimensions` to differ from the first recursion? Could we maybe use the previous recursion value here?

I suspect the second wouldn't work but the first should allow us to save some recursion maybe?

loki/transformations/sanitise/tests/test_associates.py
```diff
@@ -148,6 +148,57 @@ def test_transform_associates_array_call(frontend):
...
+  integer :: i
+  integer, parameter :: idx_a = 2
+
+  associate (a => arr2d(:, 1), b=>arr2d(:, idx_a) )
+    b(:) = 42.0
+    do i=1, 5
+      a(i) = b(i+2)
+      call another_routine(i, a(2:4), b)
+    end do
+  end associate
```
reuterbal, line 172: Do we also have a test that covers the reverse situation, where a dimension expression contains an associated symbol but the array is defined outside?

Something like:
```
real, intent(inout) :: arr2d(:,:)

associate(var_idx=>some%type%idx)
  do i=1,5
    arr2d(i, var_idx) = 1.0
  end do
end associate
```

reuterbal (commented):
Thanks for this! I have some questions about the recursion and potentially a missing test but not sure on either, so marking this as a comment only for now

loki/transformations/sanitise/tests/test_associates.py
mlange05: Yes, agreed, good point! I've extended the array slicing test to cover this example.

loki/transformations/sanitise/associates.py
mlange05: Ah, yes, this was quite fiddly. The initial recursion is required for resolving associations in the dimensions of arrays that are not actually scope to the `Associate`; that's also why it has to happen before the short-circuiting. For example, consider 
```
associate(a => b)
  some%obj(i, :, a) = ...
end associate
```

The second recursion is then done on the potentially replaced symbol (`new`), which now might have entirely different `.dimensions`. For example consider
```
associate(a => b(:, : i))
  a = 42.0
  ! or even
  a(j, k) = 66.6
end associate
```
In the first line of this example `expr == 'a'` and `expr.dimensions == None` and `new.dimensions == (:, :, i)`, the latter of which we have not recursed into yet. In the second example line `new.dimensions` then also needs to be matched against `(j, k)` to yield `b(j, k, i)`. Does that make sense?

mlange05:
Hi @reuterbal, thanks for the review and apologies for the delay. In addition to the things you flagged, I've independently found another bug in this, which I've now fixed and extended the slice array test for. This pertains to the matching of ':' when matching the free range symbols in the utility method of the mapper class. Please have another look. 🙏

loki/transformations/sanitise/associates.py
reuterbal: Thanks, I suspected something along these lines.

My question was mostly on the grounds of avoiding redundant recursion, purely judged from the dependencies within this method. Something like this:

```python
        # Recurse over the type's shape
        _type = expr.type
        if _type.shape:
            new_shape = self.rec(_type.shape, *args, **kwargs)
            _type = expr.type.clone(shape=new_shape)

        # Stop if scope is not an associate, but still recurse on the dimensions
        # in case an associated symbol is in there
        if not isinstance(expr.scope, ir.Associate):
            expr_dims = self.rec(expr.dimensions, *args, **kwargs)
            return expr.clone(dimensions=expr_dims, type=_type)

        # For arrays that represent an associate symbol, re-use the
        # scalar implementation to resolve the symbol
        new = self.map_scalar(expr, *args, **kwargs)

        # Recurse over array dimensions
        if isinstance(new, sym.Array) and new.dimensions:
            # Resolve unbound range symbols form existing indices
            new_dims = self.rec(new.dimensions, *args, **kwargs)
            new_dims = self._match_range_indices(new_dims, expr_dims)
        else:
            new_dims = expr_dims
        return new.clone(dimensions=new_dims, type=_type)
```

This should save the need to do the recursion on the dimensions for symbols from the associate since you're recursing on the `new_dims` anyway

PR #461, opened 2024-12-18
Inline: Skip explicit intrinsics when inlining stmt functions

mlange05:
The stmt function inliner does not actually test the case where the stmt function contains calls to intrinsics. This PR remedies this and fixes the corner case by including explicitly marked intrinsic calls in the skip condition.

This caused issue when the following lines attempt to touch the non-existent procedure definition.


reuterbal (approved):
Looks great, many thanks!

PR #462, opened 2024-12-18
Pipeline-plan duplicate/remove transformation changing dependencies

MichaelSt98:
Proof-of-concept transformations `DuplicateKernel` and `RemoveKernel` with both changing the dependencies for the scheduler.

This builds on top of #453 

`DuplicateKernel`  takes a kernel and creates a duplicated kernel with corresponding duplicated items and adds a call to this duplicated kernel.

`RemoveKernel` removes a kernel from the scheduler and also removes the call.


MichaelSt98:
@mlange05 and @reuterbal I would wait for some short feedback before I add more tests to improve code coverage

reuterbal (changes requested):
_Many_ thanks for this, this is a very useful demonstration of this functionality!

Conceptually, there's a few gotchas and things that I would prefer to do differently, and I'll try to explain here:

1. Item creation should always go through the item_factory, and the introduced `clone` method shortcuts this. Instead, we should have a `get_or_create_from_item` method or similar on the ItemFactory, to make sure any new items automatically land in the cache.

2. You inject the dependency changes incurred from plan_data in the `dependencies` property. However, that property returns (and I would like to keep it that way) IR nodes that constitute the dependency, e.g., `CallStatement`, `Import`, etc. So, we have two options: (a) inject (and remove) relevant IR nodes during planning to fake these dependencies or (b) inject the plan dependencies at a point where we are already working with items. I think (b) is a little easier and I have explored doing this via `create_dependency_items` instead, which seems to work (more details below).

3. The need for a recursive clone can be inferred from not specifying an IR to the clone method. For contained subroutines in ProgramUnit we are already doing this, and I would prefer to do the same for source files instead of introducing a new rebuild-method on scoped nodes.

4. We need to separate plan and trafo in the tests - both need to be able to apply the relevant sgraph topology changes independently, because we run either plan or trafo mode in practice. Getting both to work, i.e., trafo after plan, might be rather tricky but I also don't see a need for that at the moment.

Admittedly, half the observations above I only figured out when starting to play with this directly and trying to apply the relevant changes. So, I've decided not to bother you with too many inline comments that I kept editing and changing and I will instead file a PR on top of yours against your PR branch. I hope this doesn't come across as dismissive, but I've underestimated how fiddly this gets...

reuterbal:
Ah, I forgot to delete the inline comments. Disregard them or at least don't action on them, yet :)

MichaelSt98:
> _Many_ thanks for this, this is a very useful demonstration of this functionality!
> 
> Conceptually, there's a few gotchas and things that I would prefer to do differently, and I'll try to explain here:
> 
> 1. Item creation should always go through the item_factory, and the introduced `clone` method shortcuts this. Instead, we should have a `get_or_create_from_item` method or similar on the ItemFactory, to make sure any new items automatically land in the cache.
> 2. You inject the dependency changes incurred from plan_data in the `dependencies` property. However, that property returns (and I would like to keep it that way) IR nodes that constitute the dependency, e.g., `CallStatement`, `Import`, etc. So, we have two options: (a) inject (and remove) relevant IR nodes during planning to fake these dependencies or (b) inject the plan dependencies at a point where we are already working with items. I think (b) is a little easier and I have explored doing this via `create_dependency_items` instead, which seems to work (more details below).
> 3. The need for a recursive clone can be inferred from not specifying an IR to the clone method. For contained subroutines in ProgramUnit we are already doing this, and I would prefer to do the same for source files instead of introducing a new rebuild-method on scoped nodes.
> 4. We need to separate plan and trafo in the tests - both need to be able to apply the relevant sgraph topology changes independently, because we run either plan or trafo mode in practice. Getting both to work, i.e., trafo after plan, might be rather tricky but I also don't see a need for that at the moment.
> 
> Admittedly, half the observations above I only figured out when starting to play with this directly and trying to apply the relevant changes. So, I've decided not to bother you with too many inline comments that I kept editing and changing and I will instead file a PR on top of yours against your PR branch. I hope this doesn't come across as dismissive, but I've underestimated how fiddly this gets...

That is perfectly fine for me and I wasn't expecting this PR to go in as it is but to serve as a first step/proof of concept.

MichaelSt98:
Multi-level call tree duplication/removal for a follow-up PR.

reuterbal (approved):
Thanks for mopping up the outstanding housekeeping tasks, @MichaelSt98!

I think this is ready for integration, more testing and expansion to multi-level graph changes etc. should happen in follow-up PRs.

mlange05 (approved):
Agreed! Looks great now and next steps go into follow-ons. Many thanks to both! GTG :shipit:

PR #463, opened 2024-12-20
Suggested changes for Pipeline planning of Duplicate/Remove

reuterbal:
These are my suggested changes for #462, as mentioned. This looks like everything's different, but it is mostly moving conceptually things to different places and streamlining the implementation where necessary.

Importantly, I have moved the ItemFactory into a separate file as it was growing a bit. This is the first commit on this branch over #462: c1e9fc55d3d7de9989dce66ccf550515833e0e49

To properly assess the differences to @MichaelSt98's implementation, you may want to take a look at the diff against this commit: https://github.com/ecmwf-ifs/loki/compare/c1e9fc55d3d7de9989dce66ccf550515833e0e49...nabr-nams-pipeline-plan-duplicate-remove

Conceptually, this implements the following changes that I have also tried to keep in separate commits:

1. Split the tests into planning and processing
2. Fix proper deep-cloning for ProgramUnit and Sourcefile, and remove the recursive_clone method
3. Move the injection of the dependency changes from the `Item.dependency` property into `Item.create_dependency_items`
4. Move `Item.clone` into `ItemFactory.get_or_create_item_from_item`
5. Apply relevant changes to `DuplicateTrafo` and `RemoveTrafo`
6. Add a test that combines duplication and removal


loki/transformations/dependency.py
```diff
@@ -16,96 +16,107 @@ class DuplicateKernel(Transformation):
...
+    def __init__(self, duplicate_kernels=None, duplicate_suffix='duplicated',
                  duplicate_module_suffix=None):
         self.suffix = duplicate_suffix
         self.module_suffix = duplicate_module_suffix or duplicate_suffix
-        print(f"suffix: {self.suffix}")
-        print(f"module_suffix: {self.module_suffix}")
-        self.kernels = tuple(kernel.lower() for kernel in as_tuple(kernels))
+        self.duplicate_kernels = tuple(kernel.lower() for kernel in as_tuple(duplicate_kernels))
+
+    def _create_duplicate_items(self, successors, item_factory, config):
```
MichaelSt98, line 27: I feel like this should be a (more generic) utility that lives outside of this transformation (at least in the future).

MichaelSt98 (approved):
Really nice! And much better compared to my original implementation 👍

loki/transformations/dependency.py
mlange05: I'd agree here in principle, although this feels like it needs an `Item.clone()` method first and could/should possibly be done in a follow-up.

mlange05 (approved):
Looks great and adds some much needed structure. GTG from me.

loki/transformations/dependency.py
reuterbal: An Item.clone() is something we don't want, though - because you want that to always go through the factory.

PR #464, opened 2024-12-22
Transpile: Split C-kernel generation from ISO-C wrapper generation

mlange05:
_Note: Apologies for the size of this PR; it's mainly housekeeping. If this is too much, please let me know, as it can be split into more digestible chunks. I promise I've not changed any meaningful logic beyond what's mentioned below._

This PR primarily aims at separating the two stages of C-style language transpilation:
* Generation of the respective C[CPP/CUDA/HIP] kernel of a given `Subroutine`
* Generation of ISO-C compatible wrappers for the given `Subroutine` and used `Module`, including the generation of associated C-style header files.

The key change to achieve this is the creation of a new sub-package `loki.transforamtion.transpile.fortran_iso_c_wrapper` with an additional transformation `FortranISOCWrapperTransformation`. This can now be used separately in scheduler pipelines and may have a different iteration space, as it may needs to include `Module` objects for type information.

In addition, both Transformations can now use the `build_args` handed over by the scheduler, if no explicit `path` argument is given. I've kept the explicit override, as it is used extensively throughout the test base. This will eventually allow us to properly express C-transpilation as config-based pipelines. In future changes, we can hopefully further unpick the C-transpilation to build specific C/CUDA/HIP pipelines that can re-use the ISO-C wrapper generation.

In some more detail:
* Get output path from `build_args` for C-transpile transformations, when used with the scheduler.
* Create new `fortran_iso_c_wrapper` sub-package for transpile, including the respective utilities for dering with ISO-C types, derived-types and structs
* Add a `FortranISOCWrapperTransformation` to execute wrapper and header generation for subroutines and modules
* Derive the output files from utilities in tests and stop storing them on the transformation; this made a transformation instance non-reusable, as repeated use would override the output file names
* More SCC-CUDA test into separate test sub-package
* Add better docstrings to the Fortran-ISO-C wrapper methods


loki/transformations/transpile/fortran_c.py
```diff
@@ -96,8 +94,7 @@ def visit_CallStatement(self, o, **kwargs):
 
 class FortranCTransformation(Transformation):
```
MichaelSt98, line 95: both `use_c_ptr` and `path` should be removed from the doctstring.

scripts/loki_transform.py
```diff
@@ -417,14 +419,18 @@ def convert(
     mode = mode.replace('-', '_')  # Sanitize mode string
     if mode in ['c', 'cuda_parametrise', 'cuda_hoist']:
         if mode == 'c':
-            f2c_transformation = FortranCTransformation(path=build)
+            f2c_transformation = FortranCTransformation()
+            f2c_wrapper = FortranISOCWrapperTransformation()
         elif mode in ['cuda_parametrise', 'cuda_hoist']:
-            f2c_transformation = FortranCTransformation(path=build, language='cuda', use_c_ptr=True)
+            f2c_transformation = FortranCTransformation(language='cuda', use_c_ptr=True)
```
MichaelSt98, line 425:
```suggestion
            f2c_transformation = FortranCTransformation(language='cuda')
```

There is no `use_c_ptr` argument for `FortranCTransformation` anymore.

MichaelSt98 (approved):
Nice! Thank you.
Two small changes requested otherwise happy with this PR.

Also tested with/for CLOUDSC using the branch `nams-integrate-loki-cuda` and the merged this branch/PR into the Loki branch `nams-integrate-cloudsc-loki-cuda` to test the C and CUDA transpilation with CLOUDSC.

loki/transformations/transpile/fortran_iso_c_wrapper.py
```diff
@@ -0,0 +1,518 @@
...
+                # Pass by reference for array types
+                value = isinstance(arg, sym.Scalar) and arg.type.intent and arg.type.intent.lower() == 'in'
+                kind = iso_c_intrinsic_kind(arg.type, intf_fct, use_c_ptr=use_c_ptr)
+                ctype = SymbolAttributes(arg.type.dtype, value=value, kind=kind)
+                dimensions = arg.dimensions if isinstance(arg, sym.Array) else None
+                var = sym.Variable(name=arg.name, dimensions=dimensions, type=ctype, scope=intf_fct)
+                intf_args += (var,)
+            intf_fct.arguments = intf_args
+            sanitise_imports(intf_fct)
+            intfs.append(intf_fct)
```
reuterbal, line 448: [no action required] Codecov points out that this is unused. We should probably add a test for module routine transpilation in the future?

reuterbal (approved):
Thanks, this looks great and is an important stepping stone towards following through with some of the deprecations.

I left a few observation comments but none of these are strictly required.

loki/transformations/transpile/fortran_iso_c_wrapper.py
```diff
@@ -0,0 +1,518 @@
...
+            build_args = kwargs.get('build_args')
+            path = Path(build_args.get('output_dir'))
+
+        role = kwargs.get('role', 'kernel')
+
+        c_structs = {}
+        for name, td in module.typedef_map.items():
+            c_structs[name.lower()] = c_struct_typedef(td, use_c_ptr=self.use_c_ptr)
+
+        if role == 'header':
```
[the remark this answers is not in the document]
mlange05: Nice catch, these are even redundant, as they don't get propagated. I've fix this here and in the module entry point below now.

loki/transformations/transpile/fortran_iso_c_wrapper.py
```diff
@@ -0,0 +1,518 @@
...
+    if _type.dtype == BasicType.REAL:
+        kind = str(_type.kind)
+        if kind.lower() in ('real32', 'c_float'):
+            return sym.Variable(name='c_float', scope=scope)
+        if kind.lower() in ('real64', 'jprb', 'selected_real_kind(13, 300)', 'c_double', 'c_ptr'):
+            if use_c_ptr and is_array:
+                return sym.Variable(name='c_ptr', scope=scope)
+            return sym.Variable(name='c_double', scope=scope)
+
+    return None
```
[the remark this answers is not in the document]
mlange05: I tried to retro-fit these, please have a look.

loki/transformations/transpile/fortran_iso_c_wrapper.py
mlange05: Yeah, this one is puzzling me a bit. I've tried triggering this, but I think this is some leftover from the previous way we dealt with stmt funcs. Not quite sure what to do about it, but I think we want to carefully rethink some of these wrappers anyway, so we can mop this all up.

loki/transformations/transpile/fortran_iso_c_wrapper.py
```diff
@@ -0,0 +1,518 @@
...
+        elif self.language == 'cuda':
+            self.codegen = cudagen
+        else:
+            raise ValueError(f'language "{self.language}" is not supported!'
+                             f' (supported languages: "{self._supported_languages}")')
+
+    def file_suffix(self):
+        if self.language == 'cpp':
+            return '.cpp'
+        return '.c'
```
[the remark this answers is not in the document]
reuterbal: No, I think the wrapper transformation doesn't write any c/cpp files at all, so I meant this entire method is redundant here. The same exists in the actual FortranCTransformation, where this is indeed used to write the transpiled code.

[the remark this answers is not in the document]
mlange05: Ohh, right, yes that's an oversight then. I've removed the redundant one now. Good catch!

loki/transformations/transpile/fortran_iso_c_wrapper.py
reuterbal: I've had to try this myself. The relevant code path is hit with a test like this:

```python
@pytest.mark.parametrize('frontend', available_frontends())
@pytest.mark.parametrize('language', ['c', 'cpp'])
def test_transpile_module_function(tmp_path, builder, frontend, language):
    fcode_mod = """
module my_mod
    use iso_fortran_env, only: real64
    implicit none
    contains

    function scalar_func(a, b)
        real(kind=real64), intent(in) :: a, b
        real(kind=real64) :: scalar_func
        scalar_func = a + b
    end function scalar_func

    function array_func(a, b, n)
        real(kind=real64), intent(in) :: a(n), b(n)
        integer, intent(in) :: n
        real(kind=real64) :: array_func(n)
        array_func(:) = a(:) + b(:)
    end function array_func
end module my_mod
    """.strip()

    fcode = """
subroutine driver(a, b, c, d, n)
    use my_mod, only: scalar_func, array_func
    use iso_fortran_env, only: real64
    implicit none
    real(kind=real64), intent(in) :: a(n), b(n)
    real(kind=real64), intent(inout) :: c(n), d(n)
    integer, intent(in) :: n
    integer :: j

    do j=1,n
        c(j) = scalar_func(a(j), b(j))
    enddo

    d = array_func(a, b, n)
end subroutine driver
    """.strip()

    module = Module.from_source(fcode_mod, xmods=[tmp_path], frontend=frontend)
    routine = Subroutine.from_source(fcode, xmods=[tmp_path], frontend=frontend, definitions=[module])
    refname = f'ref_{routine.name}_{frontend}'
    reference = jit_compile_lib([module, routine], path=tmp_path, name=refname, builder=builder)

    n = 10
    a = np.ones(shape=(n,), order='F', dtype=np.float64)
    b = np.ones(shape=(n,), order='F', dtype=np.float64)
    c = np.zeros(shape=(n,), order='F', dtype=np.float64)
    d = np.zeros(shape=(n,), order='F', dtype=np.float64)
    cd_ref = a + b

    reference.driver(a=a, b=b, c=c, d=d, n=n)
    assert np.all(c == cd_ref)
    assert np.all(d == cd_ref)

    f2cwrap = FortranISOCWrapperTransformation(language=language)
    f2c = FortranCTransformation()

    for trafo in [f2c, f2cwrap]:
        for source in [routine, module, *module.subroutines]:
            trafo.apply(source=source, path=tmp_path, role='kernel')
    f2cwrap.apply(source=module, path=tmp_path, role='header')
```

But I don't quite know how to complete the test here (i.e., what to check for). Should we put this in an issue for now?

mlange05: Ok, as discussed offline, this can then probably be removed. I'm fairly confident that the use case for this has been removed from CLOUDSC, and the approach is really quite hacky. If CLOUDSC regression is fine after the removal I think this is safe.

reuterbal:
Very nice, many thanks!

PR #465, opened 2024-12-24
Loki-transform: Remove custom entry point options

mlange05:
~_Note: This sits on top of PR #464 and requires CLOUDSC  CLOUDSC PR 106 to be merged first._~

~_Note: For testing only at this stage._~

Following the conversion to config-file based pipelines, this PR now removes all the custom entry point options under `loki-transform.py convert` and their associated use in CMake macros. It also removes the now unused `loki_transform_convert` entry point.

### DIFF
--- a/cmake/loki_transform.cmake
+++ b/cmake/loki_transform.cmake
@@ -386,62 +386,6 @@ endfunction()
 ##############################################################################
 # .rst:
 #
-# loki_transform_convert
-# ======================
-#
-# Deprecated interface to loki-transform.py. Use loki_transform( COMMAND convert ) instead.::
-#
-##############################################################################
-
-function( loki_transform_convert )
-
-    ecbuild_warn( "\
-loki_transform_convert() is deprecated and will be removed in a future version!
-Please use
-    loki_transform( COMMAND convert [...] )
-or
-    loki_transform_target( COMMAND convert [...] ).
-"
-    )
-
-    set( options
-         CPP DATA_OFFLOAD REMOVE_OPENMP ASSUME_DEVICEPTR GLOBAL_VAR_OFFLOAD
-         TRIM_VECTOR_SECTIONS REMOVE_DERIVED_ARGS INLINE_MEMBERS
-	 RESOLVE_SEQUENCE_ASSOCIATION DERIVE_ARGUMENT_ARRAY_SHAPE
-    )
-    set( oneValueArgs
-         MODE DIRECTIVE FRONTEND CONFIG PATH OUTPATH
-    )
-    set( multiValueArgs
-         OUTPUT DEPENDS INCLUDES HEADERS DEFINITIONS OMNI_INCLUDE XMOD
-    )
-
-    cmake_parse_arguments( _PAR "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN} )
-
-    if( _PAR_UNPARSED_ARGUMENTS )
-        ecbuild_critical( "Unknown keywords given to loki_transform_convert(): \"${_PAR_UNPARSED_ARGUMENTS}\"")
-    endif()
-
-    #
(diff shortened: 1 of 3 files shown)


cmake/loki_transform.cmake
```diff
@@ -291,26 +291,6 @@ function( loki_transform_target )
             list( APPEND _TRANSFORM_OPTIONS CPP )
         endif()
 
-        if( _PAR_T_INLINE_MEMBERS )
```
reuterbal, line 294: With these removed, we should now also remove the corresponding arguments from the CMake function - that will make sure that any options that have been specified are not silently ignored but do actually raise an error in CMake directly. For that, ll. 226-231 need to look like this:

```cmake
    set( options NO_PLAN_SOURCEDIR COPY_UNMODIFIED CPP CPP_PLAN )
    set( single_value_args TARGET COMMAND MODE FRONTEND CONFIG PLAN )
    set( multi_value_args SOURCES HEADERS DEFINITIONS INCLUDES )
```

Please also update the docstring to match the reduced interface.

cmake/loki_transform_helpers.cmake
```diff
@@ -84,41 +84,6 @@ macro( _loki_transform_parse_options )
         list( APPEND _ARGS --cpp )
     endif()
 
-    if( _PAR_DATA_OFFLOAD )
```
reuterbal, line 87: With these removed, the argument definition to `loki_transform` (ll. 48-57 in `loki_transform.cmake`) should be updated accordingly:
```cmake
    set( options CPP )
    set( oneValueArgs COMMAND MODE FRONTEND CONFIG BUILDDIR )
    set( multiValueArgs OUTPUT DEPENDS SOURCES HEADERS INCLUDES DEFINITIONS OMNI_INCLUDE XMOD )
```
Please also update the docstring of `loki_transform` accordingly.

Also, since we only have the `CPP` option left now, it might be worthwhile to just remove this macro and instead inline the remaining 3 lines of code into `loki_transform` and `loki_transform_plan` directly

scripts/loki_transform.py
```diff
@@ -193,260 +133,29 @@ def convert(
...
-
-        if callgraph:
-            scheduler.callgraph(callgraph)
-
-        return
-
-    if plan_file is not None:
-        msg = '[Loki] ERROR: Plan mode requires a pipeline definition in the config file.\n'
+    if not mode in config.pipelines:
+        msg = '[Loki] ERROR: Pipeline or transformation mode not found in config file.\n'
```
reuterbal, line 137: Might be helpful to mention the specified mode here, to spot typos quickly:
```suggestion
        msg = f'[Loki] ERROR: Pipeline or transformation mode "{mode}" not found in config file.\n'
```

scripts/loki_transform.py
```diff
@@ -193,260 +133,29 @@ def convert(
...
-            scheduler.write_cmake_plan(plan_file, rootpath=root)
-
-        if callgraph:
-            scheduler.callgraph(callgraph)
-
-        return
-
-    if plan_file is not None:
-        msg = '[Loki] ERROR: Plan mode requires a pipeline definition in the config file.\n'
+    if not mode in config.pipelines:
```
reuterbal, line 136: Pet peeve, I find this much more readable:
```suggestion
    if mode not in config.pipelines:
```

reuterbal (changes requested):
This is _very_ pleasing to see!

In addition to removing the processing of the CMake options we should directly remove them from the corresponding function interfaces - this will make sure to hard fail with an error instead of silently ignoring deprecated options, which I think is better at this stage.

mlange05:
@reuterbal Indeed, very satisfying! 😉  Thanks for the pointers, I think it's ready for another look.

reuterbal (approved):
Many thanks!

PR #467, opened 2025-01-08
Expression: Remove logical literals in and/or operations during simplify

mlange05:
The current implementation would only resolve .and./.or. operations if all arguments are logical literals. Instead, we can evaluate `.true. and a` => `.true.` and `.false. and a` => `a`, and vice versa for `or` operations. This allows us to further trim code paths in dead code elimination where this is used to remove conditionals.


reuterbal (approved):
Neat! Looks good to me! :shipit:

PR #468, opened 2025-01-08
Inline: Avoid Import duplication from multiple callees

mlange05:
The previous logic does not account for the same import modules being used by multiple callees, as it attempts to propagate imported symbols in one go. Instead, we now call a utility method for each callee, so that the previous redundancy mechanics avoid duplication across multiple callees, not just between caller and callee.


reuterbal (approved):
Looks great and neatly tested. Many thanks!

PR #469, opened 2025-01-08
Frontend: Final removal of OFP frontend

mlange05:
o7


reuterbal (changes requested):
Thanks, looks good.

However, as mentioned in a comment during the deprecation (https://github.com/ecmwf-ifs/loki/issues/406#issuecomment-2422215593) we need to now also remove the routines that were only kept around for OFP. Because they have not had code coverage after the deprecation, this doesn't show up at this point anymore.

Could you take care of the checklist in the above linked comment, please?

mlange05:
> Could you take care of the checklist in the above linked comment, please?

Yes, thanks for the pointer, I had forgotten about that. I've removed the main utilities and the test now - the false positives I think we want to address once we expand more low-level unit testing of the interior IR nodes and scoping behaviour.

reuterbal:
Great, thanks. Agreed on the false positives, they should stay in.
Any particular reason for keeping the disc-cached utility? https://github.com/ecmwf-ifs/loki/blob/main/loki/tools/files.py#L186-L227
It is currently not being tested: https://app.codecov.io/gh/ecmwf-ifs/loki/pull/469/blob/loki/tools/files.py#L186

mlange05:
> Great, thanks. Agreed on the false positives, they should stay in. Any particular reason for keeping the disc-cached utility? 

Ah, good catch, I had removed the option, but not the utility. I've done that now (top commit), and rebased over main.

reuterbal (approved):
Many thanks, looks good to me now!

Now comes also with a nice uplift of the code coverage to >96%!
(The Github check for codecov is stuck somehow but will override upon merge)

Issue #459, opened 2024-12-12
bug with statement functions again


ecossevin:
### What happened?

I have a bug in routines with statement functions : FMINJ in this case.

The code has a strange behavior, the bug is difficult to reproduce, this is why my reproducer is so big, sorry for that. Let me know if you can't reproduce the bug. 

Using the version from https://github.com/ecmwf-ifs/loki/pull/439 I have no bug. Using the lastest version of loki from main, the bug is back (end the commit fixing the bug is in loki latest main (git log | grep 97d0943).

### What are the steps to reproduce the bug?

download loki `pip install "loki @ git+https://github.com/ecmwf-ifs/loki.git"` 
run the test `python3 -m pdb test.py` and then continue through the breakpoint. 

[bug_statement.tar.gz](https://github.com/user-attachments/files/18110456/bug_statement.tar.gz)

### Version

latest, from today

### Platform (OS and architecture)

Linux taranislogin1.taranishpc.meteo.fr 3.10.0-1160.102.1.el7.x86_64 #1 SMP Mon Sep 25 05:00:52 EDT 2023 x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell
(Pdb) c
Traceback (most recent call last):
  File "/opt/softs/anaconda3/envs/Python310/lib/python3.10/pdb.py", line 1723, in main
    pdb._runscript(mainpyfile)
  File "/opt/softs/anaconda3/envs/Python310/lib/python3.10/pdb.py", line 1583, in _runscript
    self.run(statement)
  File "/opt/softs/anaconda3/envs/Python310/lib/python3.10/bdb.py", line 598, in run
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "/home/gmap/mrpm/cossevine/tmp/dbg/test.py", line 92, in <module>
    horizontal_idx=get_horizontal_idx(routine, lst_horizontal_idx)
  File "/home/gmap/mrpm/cossevine/tmp/dbg/test.py", line 62, in get_horizontal_idx
    routine.body=SubstituteExpressions(rename_map).visit(routine.body)
  File "/home/gmap/mrpm/cossevine/tmp/test/venv3/lib/python3.10/site-packages/loki/ir/transformer.py", line 248, in visit
    obj = super().visit(o, *args, **kwargs)
  File "/home/gmap/mrpm/cossevine/tmp/test/venv3/lib/python3.10/site-packages/loki/ir/visitor.py", line 124, in visit
[...]

PR #470, opened 2025-01-10
Sanitise: Fix rescoping of symbols for nested associates

mlange05:
When nested associates are about to be removed, we need to account for that when updating the symbol-scope (a priori). This new logic does this by counting out the nested associates and picking the appropriate one, according to the septh of the symbol in the nest.

I've extended the stat-func test, where this most often triggers.

Note, this looks like a recurrence of the problem in #439 and hopefully should address issue #459 (@ecossevin to confirm).


reuterbal (approved):
What a mean little issue, many thanks for figuring that one out! Looks good to me

PR #471, opened 2025-01-10
Fix dependency on header files in loki_transform_target

reuterbal:
There was a typo in the variable name, possibly causing race conditions in parallel build when a header module is also generated.


mlange05 (approved):
Yes, looks good. GTG.

PR #472, opened 2025-01-15
Transformations: Fix downcasing of nested `InlineCall` symbols

mlange05:
Nested calls to functions is not captured in `convert_to_lower_case`, which can break the Fortran-to-Python converter. This simple fix remedies this.


reuterbal (approved):
Concise and uncontroversial! GTG

PR #474, opened 2025-01-15
POC: allow multi-level/subgraph duplication of kernels

MichaelSt98:
POC duplicating a kernel and optionally duplicating/separating the underlying subgraph.

E.g., starting with:

callgraph_orig.pdf

the result of the `DuplicateKernel` trafo with `duplicate_kernels=('kernel',)` is:

callgraph.pdf

However, with this PR one can specify `duplicate_subgraph=True`, which yields for the example above:

callgraph.pdf

Further, the actual transformation part works as well, producing e.g.,:

```fortran
MODULE kernel_mod_duplicated
  IMPLICIT NONE
  CONTAINS
  SUBROUTINE kernel_duplicated (klon, field1)
    USE kernel_nested_mod_duplicated, ONLY: kernel_nested_vector_duplicated, kernel_nested_seq_duplicated
    USE compute_2_mod_duplicated, ONLY: compute_2_duplicated
    IMPLICIT NONE
    INTEGER, INTENT(IN) :: klon
    INTEGER, INTENT(INOUT) :: field1(klon)
    INTEGER :: tmp1(klon)
    INTEGER :: jl
    
    CALL kernel_nested_vector_duplicated(klon, field1)
    
    DO jl=1,klon
      CALL kernel_nested_seq_duplicated(field1(jl))
      CALL compute_2_duplicated(field1(jl))
      tmp1(jl) = 0
      field1(jl) = tmp1(jl)
    END DO
    
  END SUBROUTINE kernel_duplicated
END MODULE kernel_mod_duplicated
```

### DIFF
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -545,6 +545,8 @@ class Scheduler:
             if transformation.renames_items or transformation.creates_items:
                 kwargs['item_factory'] = self.item_factory
                 kwargs['scheduler_config'] = self.config
+                kwargs['graph'] = graph
+                kwargs['item_filter'] = item_filter
 
             for _item in traversal:
                 if isinstance(_item, ExternalItem):
--- a/loki/transformations/dependency.py
+++ b/loki/transformations/dependency.py
@@ -11,7 +11,6 @@ from loki.tools.util import as_tuple, CaseInsensitiveDict
 
 __all__ = ['DuplicateKernel', 'RemoveKernel']
 
-
 class DuplicateKernel(Transformation):
     """
     Duplicate subroutines which includes the creation of new :any:`Item`s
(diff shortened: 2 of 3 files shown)


loki/batch/scheduler.py
```diff
@@ -545,6 +545,8 @@ def _get_definition_items(_item, sgraph_items):
             if transformation.renames_items or transformation.creates_items:
                 kwargs['item_factory'] = self.item_factory
                 kwargs['scheduler_config'] = self.config
+                kwargs['graph'] = graph
```
reuterbal, line 548: We need to consider whether we want to directly provide the SGraph to transformations or build an API to semantically capture the changes and have the scheduler apply them afterwards.
The latter seems more elegant (not hacking with core data structures) but it might be tricky to build something equally flexible.
Also note that tthe `graph` may be a temporary object here when traversing a file graph. In that case, this would not work at all.

reuterbal:
Has this been superseded by #489 and #500?

MichaelSt98:
Superseded by #489 and #500.

Issue #420, opened 2024-10-30
ir.CallStatement constructor fails if the arguments are not passed as keyword arguments


wertysas:
### What happened?

After the addition of a pre_init method to the CallStatment node class it is not possible to construct a CallStatement without passing a keyword argument.

### What are the steps to reproduce the bug?

```python
import loki.ir as ir
import loki.expression.symbols as sym

myfun = sym.ProcedureSymbol(name='myfun')
call = ir.CallStatement(myfun, (sym.IntLiteral(1),))   # this fails
# call = ir.CallStatement(name=myfun, arguments=(sym.IntLiteral(1),))   # this doesnt fail
```

### Version

main branch

### Platform (OS and architecture)

Mac OS

### Relevant log output

```shell
File "callstmt_ctor_minimal_reproducible.py", line 5, in <module>
    call = ir.CallStatement(myfun, (sym.IntLiteral(1),))   # this fails
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/loki-venv/lib/python3.11/site-packages/pydantic/_internal/_dataclasses.py", line 141, in __init__
    s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
  File "loki/loki/ir/nodes.py", line 1005, in pre_init
    if 'arguments' in values.kwargs:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

mlange05:
With PR #476 merged this can be closed now. Confirmed offline with @wertysas

PR #476, opened 2025-01-16
IR: Better tuple autocasting via pydantic field_validators

mlange05:
Instead of using `model_validator` objects, we can use inidividual `field_validator` utilities to do per-argument auto-casting. The effect of this is that we can then use positional constructor arguments without using the funcationality and dealing the full model `ArgsKwargs` objects.

This should address issue #420, @wertysas to confirm.

### DIFF
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -21,7 +21,7 @@ from typing import Any, Tuple, Union, Optional
 from pymbolic.primitives import Expression
 
 from pydantic.dataclasses import dataclass as dataclass_validated
-from pydantic import model_validator
+from pydantic import field_validator
 
 from loki.expression import (
     symbols as sym, Variable, parse_expr, AttachScopesMapper,
@@ -257,17 +257,10 @@ class InternalNode(Node, _InternalNode):
 
     _traversable = ['body']
 
-    @model_validator(mode='before')
+    @field_validator('body', mode='before')
     @classmethod
-    def pre_init(cls, values):
-        """ Ensure non-nested tuples for body. """
-        if values.kwargs and 'body' in values.kwargs:
-            values.kwargs['body'] = _sanitize_tuple(values.kwargs['body'])
-        if values.args:
-            # ArgsKwargs are immutable, so we need to force it a little
-            new_args = (_sanitize_tuple(values.args[0]),) + values.args[1:]
-            values = type(values)(args=new_args, kwargs=values.kwargs)
-        return values
+    def ensure_tuple(cls, value):
+        return _sanitize_tuple(value)
 
     def __repr__(self):
         raise NotImplementedError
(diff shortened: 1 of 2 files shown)


loki/ir/nodes.py
```diff
@@ -257,17 +257,10 @@ class InternalNode(Node, _InternalNode):
 
     _traversable = ['body']
 
-    @model_validator(mode='before')
+    @field_validator('body', mode='before')
```
reuterbal, line 260: Likely my ignorance of pydantic showing here: The `pre_init` method did also stuff to `args` and `kwargs` before. Is this implicitly handled now or not required anymore?

loki/ir/nodes.py
```diff
@@ -960,8 +949,8 @@ class _CallStatementBase():
     """ Type definitions for :any:`CallStatement` node type. """
 
     name: Expression
-    arguments: Optional[Tuple[Expression, ...]] = None
-    kwarguments: Optional[Tuple[Tuple[str, Expression], ...]] = None
+    arguments: Optional[Tuple[Expression, ...]] = ()
+    kwarguments: Optional[Tuple[Tuple[str, Expression], ...]] = ()
```
reuterbal, line 953: Is this a change in behaviour, i.e., `arguments` and `kwarguments` always a tuple now instead of `None`? (I'm not opposed, just trying to understand the impact)

loki/ir/tests/test_ir_nodes.py
```diff
@@ -148,6 +154,14 @@ def test_conditional(scope, n, a_i):
     )
     assert not cond.body and cond.else_body == (assign, assign, assign)
 
+    # Test auto-casting with unnamed constructor args
+    cond = ir.Conditional(condition)
+    assert cond.body == () and not cond.else_body
```
reuterbal, line 159: Pylint doesn't like `cond.body == ()`, maybe 
```suggestion
    assert isinstance(cond.body, tuple) and not (cond.body or cond.else_body)
```

loki/ir/tests/test_ir_nodes.py
```diff
@@ -235,6 +249,12 @@ def test_section(n, a_n, a_i):
     sec = ir.Section((assign, (func,), assign, None))
     assert sec.body == (assign, func, assign)
 
+    # Test auto-casting with unnamed constructor args
+    sec = ir.Section()
+    assert sec.body == ()
```
reuterbal, line 254:
```suggestion
    assert isinstance(sec.body, tuple) and not sec.body
```

reuterbal (changes requested):
Thanks, looks cleaner indeed and appears to be fine like this. Pylint's unhappy, so requesting changes, but these are trivial. More interested in better understanding the questions I put in inline comments, which are only for my own education.

loki/ir/tests/test_ir_nodes.py
mlange05: Or even just `sec.body is ()`...

loki/ir/tests/test_ir_nodes.py
mlange05: Same here, `cond.body is ()` should do to appease pylint...

loki/ir/nodes.py
mlange05: Yes, but this one worked on the entire "Model" in pydantic speak, so it got a single `ArgsKwargs` argument that held all the constructor arguments as `ArgsKwarg.args` and `ArgsKwargs.kwargs`. This became an issue, as the `ArgsKwargs` object is immutable, but the `ArgsKwarg.kwargs` dict could be changed. This means, if any kw-args were given, I could change the dict, but I cannot replace `None` with `()` as `ArgsKwargs.args = ()` is not allowed. 

Long story short, this new validator is applied to each arg individual when pydantic creates the `ArgsKwargs` object, so all our troubles fade away and pydantic takes care of identifying unnamed args for free.

loki/ir/nodes.py
mlange05: Internally, we would always force this into `()` in the pre-validator autocast mechanism. The reason I need to change it here is that the before-validator does not get invoked if the arg is omitted so the default is used directly. This should be agnostic to the previous behaviour. 🤞

PR #477, opened 2025-01-17
Loki: Remove the CLAW and its associated transformations

mlange05:
o7


mlange05:
Ok, I've rebased this and removed the temporary integration commit. Tests will fail until the corresponding PR has been merged, but otherwise it's good to review.

reuterbal (approved):
Looks like the table has turned. The Claw no longer decides who stays and who goes.

PR #478, opened 2025-01-17
FieldOffload: Allow and skip non-enriched calls in offload region

mlange05:
Turns out in our CLOUDSC regression test we were not yet testing the new FIELD-API generation transformation, and it was broken. 🙀 The issue was that there are several calls to timer routines that are harmless, but break the rigorous sanity checks. So, instead of breaking we now warn ans kip, and enable the test in the regression suite.

### DIFF
--- a/loki/transformations/data_offload/field_offload.py
+++ b/loki/transformations/data_offload/field_offload.py
@@ -12,7 +12,7 @@ from loki.ir import (
     nodes as ir, FindNodes, FindVariables, Transformer,
     SubstituteExpressions, pragma_regions_attached, is_loki_pragma
 )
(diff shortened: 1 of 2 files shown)


mlange05:
Integration note: I tried enabling this transformation in the CLOUDSC regression test, but it turns out that this is only switched on if OpenACC is enabled in Field-API, which in practice requires an Nvidia compiler toolchain. So, I've left this off for now. We might want to consider letting the regression workflow to cover multiple compiler toolchains at some point.

reuterbal (approved):
Looks great, many thanks for sorting this. Agreed also on running field-api variants as part of regression tests soon

PR #480, opened 2025-01-24
SCC re-vector ad driver level and OpenMP offload

MichaelSt98:
extend SCC family pipelines to allow revectorisation at driver level and OpenMP offload

* you can test this via e.g., CLOUDSC branch [`nams-loki-scc-omp`](https://github.com/ecmwf-ifs/dwarf-p-cloudsc/tree/nams-loki-scc-omp)
    * new config file `src/cloudsc_loki/cloudsc_loki_omp.config` (which is currently set in `src/cloudsc_loki/CMakeLists.txt` to build `SCC`, `SCC-HOIST` (some minor problem do debug) and `SCC-STACK`

### DIFF
--- a/loki/transformations/data_offload/offload.py
+++ b/loki/transformations/data_offload/offload.py
@@ -28,6 +28,9 @@ class DataOffloadTransformation(Transformation):
 
     Parameters
     ----------
+    directive : str, optional
+        Pragma/Directive language to be used (OpenACC via 'openacc'
+        or OpenMP offload via 'omp-gpu')
     remove_openmp : bool
         Remove any existing OpenMP pragmas inside the marked region.
     present_on_device : bool
@@ -37,9 +40,11 @@ class DataOffloadTransformation(Transformation):
         is being managed outside of structured OpenACC data regions.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, directive='openacc', **kwargs):
         # We need to record if we actually added any, so
         # that down-stream processing can use that info
+        self.directive = directive
+        assert self.directive in ['openacc', 'omp-gpu']
         self.has_data_regions = False
         self.remove_openmp = kwargs.get('remove_openmp', False)
         self.assume_deviceptr = kwargs.get('assume_deviceptr', False)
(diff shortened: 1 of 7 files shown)


loki/transformations/data_offload/offload.py
```diff
@@ -148,28 +153,40 @@ def insert_data_offload_pragmas(self, routine, targets):
...
+                    pragma_post = Pragma(keyword='acc', content='end data')
+                elif self.directive == 'omp-gpu':
+                    if self.present_on_device:
+                        ... # TODO: OpenMP offload if self.present_on_device
+                    else:
+                        copyin = f'map(to: {", ".join(inargs)})' if inargs else ''
+                        copy = f'map(tofrom:{", ".join(inoutargs)})' if inoutargs else ''
+                        copyout = f'map(from: {", ".join(outargs)})' if outargs else ''
+                        pragma = Pragma(keyword='omp', content=f'target data {copyin} {copy} {copyout}')
+                    pragma_post = Pragma(keyword='omp', content='end target data')
```
reuterbal, line 189: I think something that would be great here would be an abstraction of the encoded concept: We inject statements to perform data allocation and movement, by 
1) specifying a list of variables that are in/out/inout/create.
2) specifying the programming model.
3) specifying the IR (PragmaRegion) to apply this to

I would create an abstract interface that takes this information and applies it to a region, with implementations for OpenACC, OpenMP, FIELD_API etc. That makes this easier to test standalone and then to re-use everywhere (offload trafo, pool allocator, global var offload, ...)

loki/transformations/single_column/annotate.py
```diff
@@ -69,19 +69,40 @@ def annotate_vector_loops(self, routine):
...
+                            pragma_post = ir.Pragma(keyword='omp', content='end parallel do')
+                            # pragma_new = ir.Pragma(keyword='omp', content='loop bind(parallel)')
+                            # pragma_post = ir.Pragma(keyword='omp', content='end loop')
+
+                            # Replace existing loki pragma and add post-pragma
+                            loop_pragmas = tuple(p for p in as_tuple(loop.pragma) if p is not pragma)
+                            loop._update(
+                                pragma=loop_pragmas + (pragma_new,),
+                                pragma_post=(pragma_post,) + as_tuple(loop.pragma_post)
+                            )
```
reuterbal, line 105: Same comment, essentially: Encode the parallelisation concepts used here as a generic interface that applies them, and then overload as programming model-specific implementations for OpenMP and OpenACC.

I'm undecided whether this should be captured together with the data offload interfaces in common programming-model specific classes or have them separate by function (i.e., parallelization, data movement).

loki/transformations/data_offload/offload.py
mlange05: Yes, I think we have a rudimentary version of this for FIELD-API only already, although it would probably need extending/generalising:
https://github.com/ecmwf-ifs/loki/blob/main/loki/transformations/field_api.py#L33

 I fully agree though, that having a common abstraction for this would be great.

reuterbal:
I think this can be closed now @MichaelSt98 ?

MichaelSt98:
> I think this can be closed now @MichaelSt98 ?

Yes!

MichaelSt98:
Superseded by #483 and #485.

PR #481, opened 2025-01-28
PoolAllocator improvement

MichaelSt98:
improvement so that stack size calculation and incrementing guarantees 8 byte alignment.

With this, 

```fortran
    ISTSZ = (C_SIZEOF(REAL(1, kind=JPRB))*NPROMA + 13*C_SIZEOF(REAL(1, kind=JPRB))*NLEV*NPROMA + 5*C_SIZEOF(REAL(1, kind=JPRB)) &
    & *NLEV*NPROMA*NCLV + C_SIZEOF(REAL(1, kind=JPRB))*NPROMA*NCLV) / C_SIZEOF(REAL(1, kind=JPRB))
    IF (.not.(MOD(C_SIZEOF(REAL(1, kind=JPRB))*NPROMA + 13*C_SIZEOF(REAL(1, kind=JPRB))*NLEV*NPROMA + 5*C_SIZEOF(REAL(1,  &
    & kind=JPRB))*NLEV*NPROMA*NCLV + C_SIZEOF(REAL(1, kind=JPRB))*NPROMA*NCLV, C_SIZEOF(REAL(1, kind=JPRB))) == 0)) ISTSZ = ISTSZ &
    &  + 1
```

becomes:

```fortran
    ISTSZ = ISHFT(7 + C_SIZEOF(REAL(1, kind=JPRB))*NPROMA + C_SIZEOF(REAL(1, kind=JPRB))*NLEV*NPROMA, -3) + 12*ISHFT(7 +  &
    & C_SIZEOF(REAL(1, kind=JPRB))*NLEV*NPROMA, -3) + 4*ISHFT(7 + C_SIZEOF(REAL(1, kind=JPRB))*NLEV*NPROMA*NCLV, -3) + ISHFT(7 +  &
    & C_SIZEOF(REAL(1, kind=JPRB))*NPROMA*NCLV + C_SIZEOF(REAL(1, kind=JPRB))*NLEV*NPROMA*NCLV, -3)
```

Further,

```fortran
    YLSTACK_L = YLSTACK_L + KLON*(KLEV + 1)*C_SIZEOF(REAL(1, kind=JPRB))
    IP_ZTP1 = YLSTACK_L
    YLSTACK_L = YLSTACK_L + KLON*KLEV*C_SIZEOF(REAL(1, kind=JPRB))
    IP_ZLI = YLSTACK_L
    YLSTACK_L = YLSTACK_L + KLON*KLEV*C_SIZEOF(REAL(1, kind=JPRB))
```

[...]

PR #482, opened 2025-01-28
PoolAllocator: fix handling of first shape/dimension being a range index

MichaelSt98:
We already handled range indices like e.g., `0:10` properly as long as those aren't the first dimension/shape ...

Thanks @piotrows for pointing this out!


reuterbal (approved):
Looks great, many thanks! This may create conflicts with #481 - in which order would you like to merge them?

PR #483, opened 2025-01-30
SCC revector transformation that gives sequential kernels

MichaelSt98:
New/additional SCC revector transformation that hoists the horizontal loop to the driver and passes the horizontal/loop variable to the kernel(s).

### DIFF
--- a/loki/transformations/single_column/vector.py
+++ b/loki/transformations/single_column/vector.py
@@ -547,11 +547,11 @@ class SCCSeqRevectorTransformation(Transformation):
                     if (reduction_clause := re.search(r'reduction\([\w:0-9 \t]+\)', region.pragma.content)):
 
                         loops = FindNodes(ir.Loop).visit(region)
-                        assert len(loops) == 1
-                        pragma = ir.Pragma(keyword='loki', content=f'loop vector {reduction_clause[0]}')
-                        # Update loop and region in place to remove marker pragmas
-                        loops[0]._update(pragma=(pragma,))
-                        region._update(pragma=None, pragma_post=None)
+                        if loops:
+                            pragma = ir.Pragma(keyword='loki', content=f'loop vector {reduction_clause[0]}')
+                            # Update loop and region in place to remove marker pragmas
+                            loops[0]._update(pragma=(pragma,))
+                            region._update(pragma=None, pragma_post=None)
 
 
     def mark_seq_loops(self, section):
@@ -626,10 +626,10 @@ class SCCSeqRevectorTransformation(Transformation):
             Tuple of target routine names for determining "driver" loops
         """
         role = kwargs['role']
-        targets = kwargs.get('targets', ())
+        targets = tuple(str(t).lower() for t in as_tuple(kwargs.get('targets', None)))
         # ignore = kwargs.get('ignore', ())
         item = kwargs.get('item', None)
-        ignore = item.ignore if item else ()
+        ignore = item.ignore if item else () + tuple(str(t).lower() for t in as_tuple(kwargs.get('ignore', None)))
 
         if role == 'kernel':
             # Skip if kernel is marked as `!$loki routine seq`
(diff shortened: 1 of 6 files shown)


reuterbal (changes requested):
Very nice work and overall looks very good to me.

I have flagged some docstring typos (might be copy-pasted in which case feel free to fix the source, too ;-) ).

Two other remarks:
1. I haven't done a line-by-line comparison to see if everything is duplicated or actually has some subtle differences between Seq and Vec implementations - but if it's indeed the same it might be worthwhile to insert a base class that provides these.
2. Codecov shows a few lines missing coverage. They've been likely already missing before but if it's not too much effort it might be useful to add corresponding tests.

For the fine details I'll rely on @mlange05's feedback :)

loki/transformations/single_column/vector.py
```diff
@@ -450,6 +480,196 @@ def transform_subroutine(self, routine, **kwargs):
...
+            for region in FindNodes(ir.PragmaRegion).visit(section):
+                if is_loki_pragma(region.pragma, starts_with='vector-reduction'):
+                    if (reduction_clause := re.search(r'reduction\([\w:0-9 \t]+\)', region.pragma.content)):
+
+                        loops = FindNodes(ir.Loop).visit(region)
+                        if loops:
+                            pragma = ir.Pragma(keyword='loki', content=f'loop vector {reduction_clause[0]}')
+                            # Update loop and region in place to remove marker pragmas
+                            loops[0]._update(pragma=(pragma,))
+                            region._update(pragma=None, pragma_post=None)
```
reuterbal, line 542: Thanks for expanding the tests a bit. These 4 lines are still not hit in the test base - just pointing this out in case that is easy to fix, not making this a requirement here.

loki/transformations/single_column/vector.py
```diff
@@ -450,6 +480,196 @@ def transform_subroutine(self, routine, **kwargs):
...
+        ignore = item.ignore if item else () + tuple(str(t).lower() for t in as_tuple(kwargs.get('ignore', None)))
+
+        if role == 'kernel':
+            # Skip if kernel is marked as `!$loki routine seq`
+            if check_routine_sequential(routine):
+                return
+
+            if self.horizontal.index not in routine.variables:
+                jl = get_integer_variable(routine, self.horizontal.index)
+                routine.arguments += (jl.clone(type=jl.type.clone(intent='in')),)
```
reuterbal, line 585: This one is also not covered. I think that becomes relevant when you have a routine that only dispatches to other kernels but doesn't have any loops itself?

loki/transformations/single_column/vector.py
```diff
@@ -450,6 +480,196 @@ def transform_subroutine(self, routine, **kwargs):
...
+                    if check_routine_sequential(call.routine):
+                        continue
+                    if self.horizontal.index not in call.arg_map:
+                        new_kwarg = (self.horizontal.index, get_integer_variable(routine, self.horizontal.index))
+                        updated_call = call.clone(kwarguments=call.kwarguments + (new_kwarg,))
+                        call_map[call] = updated_call
+                    if call.routine.name.lower() in ignore:
+                        if self.horizontal.index not in call.routine.variables:
+                            jl = get_integer_variable(call.routine, self.horizontal.index)
+                            call.routine.arguments += (jl.clone(type=jl.type.clone(intent='in')),)
```
reuterbal, line 606: Same here - would probably be resolved with the same case as the previous one.

reuterbal (approved):
Many thanks, from my point of view this looks really great!

I've left again two pointers where we currently don't have coverage in the tests in case this is something that's easy to add, but not mandating this here.

loki/transformations/single_column/vector.py
```diff
@@ -450,6 +480,196 @@ def transform_subroutine(self, routine, **kwargs):
...
+                                call_map[call] = wrap_vector_section((updated_call,), routine, bounds=loop_bounds,
+                                        insert_pragma=True, index=self.horizontal.index)
+
+                    loop._update(body=Transformer(call_map).visit(loop.body))
+
+                    # Revector all marked sections within the driver loop body
+                    loop._update(body=self.revector_section(routine, loop.body))
+
+                    # Check for explicitly labelled vector-reduction regions
+                    super().mark_vector_reductions(routine, loop.body)
```
mlange05, line 665: [no action] Ahh, so in driver compute regions we fall back to SCCVec behaviour, yes? Very neat!

mlange05 (approved):
This is absolutely fantastic! Very neatly implemented and well-tested. Thank you very much for this!

I only left discussion comments, but nothing that needs action. I also played with it a little in CLOUDSC and it behaves exactly as intended. Very well done! GTG from me :shipit:

loki/transformations/single_column/vector.py
MichaelSt98: Exactly, yes.

PR #484, opened 2025-02-05
Enable inlining of a subroutine with an array subrange in an argument

awnawab:
A small fix to the inlining trafo that enables inlining of calls where we pass a subrange of an array argument, i.e.:
```
real :: a(8)

!$loki inline
call kernel(a(5:8)

....

subroutine kernel
real, intent(in) :: a(4)

```

### DIFF
--- a/loki/transformations/inline/procedures.py
+++ b/loki/transformations/inline/procedures.py
@@ -95,12 +95,24 @@ def map_call_to_procedure_body(call, caller, callee=None):
 
         indices = [index for index, dim in enumerate(val.dimensions) if isinstance(dim, sym.Range)]
 
-        for index, dim in enumerate(var.dimensions):
+        lbounds_diff = [sym.IntLiteral(0) for _ in var.shape]
+        var_ubounds = [getattr(dim, 'upper', dim) for dim in var.shape]
+        if var.shape and val.shape:
+            decl_lbounds = [(getattr(val.shape[i], 'lower', sym.IntLiteral(1)),
+                             getattr(dim, 'lower', sym.IntLiteral(1))) for i, dim in enumerate(var.shape)]
+
+            for i, (lb_val, lb_var) in enumerate(decl_lbounds):
+                # we can't simply check if lb_val here as that would return a false negative if lb_val == 0
+                if lb_val is not None and lb_var is not None:
+                    lbounds_diff[i] = simplify(sym.Sum((lb_val, sym.Product((lb_var, sym.IntLiteral(-1))))))
+
+        for (index, dim), lbdiff in zip(enumerate(var.dimensions), lbounds_diff):
             # if the argument contains an array range, we must map the bounds accordingly
             if isinstance(val.dimensions[index], sym.Range) and (lower := val.dimensions[index].lower):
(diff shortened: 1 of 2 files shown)


loki/transformations/inline/tests/test_inline_transformation.py
```diff
@@ -338,18 +339,24 @@ def test_inline_transformation_adjust_imports(frontend, tmp_path):
...
+  real(kind=8), intent(out) :: c(4), d(4)
   real(kind=8) :: tmp(m)
   integer :: i
 
   tmp(1:m) = x
   do i=1, n
     a(i) = b(i) + sum(tmp)
   end do
+  do i=1,4
+    c(i) = 0.
```
mlange05, line 356: Could we add an assignment `d(i) = 1.` here too, to check that this index is shifted too?

loki/transformations/inline/tests/test_inline_transformation.py
```diff
@@ -338,18 +339,24 @@ def test_inline_transformation_adjust_imports(frontend, tmp_path):
   implicit none
   contains
 
-subroutine test_inline_inner(a, b)
+subroutine test_inline_inner(a, b, c, d)
   use BNDS_module, only: n, m
   use another_module, only: x
 
   real(kind=8), intent(inout) :: a(n), b(n)
+  real(kind=8), intent(out) :: c(4), d(4)
```
mlange05, line 347: What happens to `real, intent(inout) :: e(0:3)` in the two cases below? Could we add a test for this too, just to cover our bases?

loki/transformations/inline/procedures.py
```diff
@@ -87,12 +87,29 @@ def _map_unbound_dims(var, val):
...
+
         new_dimensions = list(val.dimensions)
 
         indices = [index for index, dim in enumerate(val.dimensions) if isinstance(dim, sym.Range)]
 
         for index, dim in enumerate(var.dimensions):
-            new_dimensions[indices[index]] = dim
+            # if the argument contains an array range, we must map the bounds accordingly
+            if isinstance(val.dimensions[index], sym.Range) and (lower := val.dimensions[index].lower):
+                if isinstance(dim, sym.Range):
```
mlange05, line 101: [no action] Note to self, there's a similar index-shifting problem in the associate resolver. We could possibly re-use a shared utility for this if we were to externalise this? This is beyond this PR tho.

mlange05 (changes requested):
Looks great and very useful. I've flagged a few corner cases, that I think we want to cover in the tests to avoid future debug pain. Otherwise great, many thanks! 🙏 (I might just steal that for the associate resolve some time. 😉 )

loki/transformations/inline/tests/test_inline_transformation.py
awnawab: That was indeed a corner case I'd missed, thanks!

awnawab:
Please wait to merge this, I think I've missed the edge case where we are passing a subrange of a dimension and there are different lbounds for the argument and dummy. I'll confirm today and push a fix + test if necessary.

mlange05 (approved):
Looks great! Many thanks for extending to the corner cases. GTG :shipit:

PR #485, opened 2025-02-06
Introduce 'PragmaModelTransformation' and preliminary OpenMP offload

MichaelSt98:
Introduce 'PragmaModelTransformation' and preliminary OpenMP offload

With this all transformation insert "generic" loki pragmas which are then transformed/mapped as (one of the last steps) using the 'PragmaModelTransformation'  to a specific Pragma model (e.g., OpenACC, OpenMP offload).

### DIFF
--- a/loki/ir/pragma_utils.py
+++ b/loki/ir/pragma_utils.py
@@ -23,7 +23,7 @@ __all__ = [
     'attach_pragmas', 'detach_pragmas',
     'pragmas_attached', 'attach_pragma_regions', 'detach_pragma_regions',
     'pragma_regions_attached', 'PragmaAttacher', 'PragmaDetacher',
-    'get_pragma_start_and_parameters'
+    'get_pragma_command_and_parameters'
 ]
 
 
@@ -124,7 +124,7 @@ class PragmaParameters:
         parameters = {k: v if len(v) > 1 else v[0] for k, v in parameters.items()}
         return parameters
 
-def get_pragma_start_and_parameters(pragma):
+def get_pragma_command_and_parameters(pragma, only_loki_pragmas=True):
     """
     Parse a pragma in the form
 
@@ -140,26 +140,20 @@ def get_pragma_start_and_parameters(pragma):
     Returns
     -------
     tuple(str, dict) :
-        ...
+        tuple being (command, parameters as dictionary)
     """
-    pragma_parameters = PragmaParameters()
-    parameters = defaultdict(list)
-    if pragma.keyword.lower() != 'loki':
+    pragma_parameters = list(get_pragma_parameters(pragma, only_loki_pragmas=only_loki_pragmas).items())
+    if not pragma_parameters:
         return None, None
-    content = pragma.content or ''
-    # Remove any line-continuation markers
-    content = content.replace('&', '')
-    starts_with = content.split(' ')[0]
-    if starts_with == 'end':
-        starts_with = f"{starts_with}-{content.split(' ')[1]}"
-    if not starts_with:
-        return None, None
-    content = content[len(starts_with):]
-    parameter = pragma_parameters.find(content)
-    for key in parameter:
-        parameters[key].append(parameter[key])
-    parameters = {k: v if len(v) > 1 else v[0] for k, v in parameters.items()}
-    return starts_with, parameters
+    if pragma_parameters[0][0] == 'end':
+        if len(pragma_parameters) < 2 or pragma_parameters[1][1] is not None:
+            debug('get_pragma_command_and_parameters: Failed to match end-command in pragma {pragma}')
+            return None, None
+        pragma_parameters = [
+            (f'{pragma_parameters[0][0]}-{pragma_parameters[1][0]}', None),
+            *pragma_parameters[2:]
+        ]
+    return pragma_parameters[0][0], dict(pragma_parameters[1:])
 
 def get_pragma_parameters(pragma, starts_with=None, only_loki_pragmas=True):
     """
--- a/loki/ir/pragma_utils.py
+++ b/loki/ir/pragma_utils.py
@@ -146,7 +146,7 @@ def get_pragma_command_and_parameters(pragma, only_loki_pragmas=True):
     if not pragma_parameters:
         return None, None
     if pragma_parameters[0][0] == 'end':
-        if len(pragma_parameters) < 2 or pragma_parameters[1][1] is not None:
+        if len(pragma_parameters) < 2: #  or pragma_parameters[1][1] is not None:
             debug('get_pragma_command_and_parameters: Failed to match end-command in pragma {pragma}')
             return None, None
         pragma_parameters = [
--- a/loki/ir/pragma_utils.py
+++ b/loki/ir/pragma_utils.py
@@ -22,7 +22,8 @@ __all__ = [
     'is_loki_pragma', 'get_pragma_parameters', 'process_dimension_pragmas',
     'attach_pragmas', 'detach_pragmas',
     'pragmas_attached', 'attach_pragma_regions', 'detach_pragma_regions',
-    'pragma_regions_attached', 'PragmaAttacher', 'PragmaDetacher'
+    'pragma_regions_attached', 'PragmaAttacher', 'PragmaDetacher',
+    'get_pragma_start_and_parameters'
 ]
 
 
@@ -123,6 +124,42 @@ class PragmaParameters:
         parameters = {k: v if len(v) > 1 else v[0] for k, v in parameters.items()}
         return parameters
 
+def get_pragma_start_and_parameters(pragma):
+    """
+    Parse a pragma in the form
+
+    ``!$loki [end] command [param] [param_with_val(val)]``
+
+    and return ``command, {param: None, param_with_val: val}``
+
+    Parameters
+    ----------
+    pragma : :any:`Pragma`
+        the pragma to parse and return parameters
+
+    Returns
+    -------
+    tuple(str, dict) :
+        ...
+    """
+    pragma_parameters = PragmaParameters()
+    parameters = defaultdict(list)
+    if pragma.keyword.lower() != 'loki':
+        return None, None
+    content = pragma.content or ''
+    # Remove any line-continuation markers
+    content = content.replace('&', '')
+    starts_with = content.split(' ')[0]
+    if starts_with == 'end':
+        starts_with = f"{starts_with}-{content.split(' ')[1]}"
+    if not starts_with:
+        return None, None
+    content = content[len(starts_with):]
+    parameter = pragma_parameters.find(content)
+    for key in parameter:
+        parameters[key].append(parameter[key])
+    parameters = {k: v if len(v) > 1 else v[0] for k, v in parameters.items()}
+    return starts_with, parameters
 
 def get_pragma_parameters(pragma, starts_with=None, only_loki_pragmas=True):
     """
@@ -167,7 +204,6 @@ def get_pragma_parameters(pragma, starts_with=None, only_loki_pragmas=True):
             parameters[key].append(parameter[key])
     parameters = {k: v if len(v) > 1 else v[0] for k, v in parameters.items()}
     return parameters
-
 
 def process_dimension_pragmas(ir, scope=None):
     """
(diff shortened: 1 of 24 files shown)


MichaelSt98:
OpenMP offload incomplete ...

| Loki | OpenACC | OMP-GPU |
| ---- | ------- | ------- |
| `create device(...)`| `declare create(...)` | |
| `update device(...) host(...)` | `update device(...) self(...)` | |
| `unscoped-data in(...) create(...)` | `enter data copyin(...) create(...)` | `target enter data map(to: ...) map(alloc: ...)` |
| `unscoped-end-data out(...) delete(...)` | `exit data copyout(...) delete(...)` | `target exit data map(from: ...) map(delete: ...) map(release: ... ???)` |
| `scoped-data inout(...) in(...) out(...) create(...)` | `data copy(...) copyin(...) copyout(...) create(...)` | `target data map(tofrom: ...) map(to: ...) map(from: ...)` |
| `scoped-end-data inout(...) in(...) out(...)` | `end data` | `end target data` |
| `loop gang private(...) vlength(...)` | `parallel loop gang private(...) vector_length(...)` | `target teams distribute thread_limit(...) ???` |
| `end-loop gang` | `end parallel loop`  | `end target teams distribute` |
| `loop vector private(...) reduction(...)` | `loop vector private(...) reduction(...)` | `parallel do` |
| `end-loop vector` | - | `end parallel do` |
| `loop seq` | `loop seq` | - | |
| `end-loop seq` | - | - | |
| `routine vector` | `routine vector` | **not allowed/supported?** |
| `routine seq` | `routine seq` | `declare target` |
| `data device-present vars(...)` | `data present()` | ? |
| `device-present vars (...)` | `data present(...)` | ? |
| `end-device-present vars(...)` | `end data` | ? |
| `device-ptr vars (...)` | `data deviceptr(...)` | ? |
| `end-device-ptr vars(...)` | `end data` |

MichaelSt98:
Problems with modules being incomplete ... That's why the regression tests are failing.

loki/transformations/pragma_model.py
```diff
@@ -0,0 +1,324 @@
...
+        The directive(s) to be used, used to determine which
+        child class of :any:`GenericPragmaMapper` is used.
+    keep_loki_pragmas: bool
+        Keep or remove generic Loki pragmas that are not
+        mapped.
+    """
+
+    item_filter = (ProcedureItem, ModuleItem)
+
+    def __init__(self, directive=None, keep_loki_pragmas=True, process_module_items=False):
```
reuterbal, line 281: In what situation would I not want to process module items?

loki/ir/pragma_utils.py
```diff
@@ -123,6 +124,42 @@ def _match_spans(open_, close_):
...
+    if starts_with == 'end':
+        starts_with = f"{starts_with}-{content.split(' ')[1]}"
+    if not starts_with:
+        return None, None
+    content = content[len(starts_with):]
+    parameter = pragma_parameters.find(content)
+    for key in parameter:
+        parameters[key].append(parameter[key])
+    parameters = {k: v if len(v) > 1 else v[0] for k, v in parameters.items()}
+    return starts_with, parameters
```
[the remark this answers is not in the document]
reuterbal: And while we're at it, can we please have a test for this utility?

loki/transformations/single_column/annotate.py
```diff
@@ -263,8 +270,8 @@ def device_alloc_column_locals(cls, routine, column_locals):
 
         if column_locals:
             vnames = ', '.join(v.name for v in column_locals)
-            pragma = ir.Pragma(keyword='acc', content=f'enter data create({vnames})')
-            pragma_post = ir.Pragma(keyword='acc', content=f'exit data delete({vnames})')
+            pragma = ir.Pragma(keyword='loki', content=f'unscoped-data create({vnames})')
+            pragma_post = ir.Pragma(keyword='loki', content=f'unscoped-enddata delete({vnames})')
```
reuterbal, line 274: This entire section is not covered by tests. What situation is this for, what would be required to capture this?
(Cc @mlange05)

loki/transformations/pragma_model.py
```diff
@@ -0,0 +1,324 @@
...
+        meth = self.lookup_method(starts_with.lower().replace('-', '_'))
+        print(f"map to meth: {meth} | starts_with: {starts_with}")
+        if meth is not None:
+            return meth(pragma, parameters, **kwargs)
+        return self.default_retval() # pragma # or return None
+
+
+class OpenACCPragmaMapper(GenericPragmaMapper):
+    """
+    Loki generic pragmas to OpenACC mapper.
```
reuterbal, line 77: Can we please add here the table with the directive translation between loki, openacc and openmp?
And a documentation of the constructor arguments, please.

loki/transformations/pragma_model.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+
+    def pmap_create(self, pragma, parameters, **kwargs): # pylint: disable=unused-argument
+        if param_device := parameters.get('device'):
+            return Pragma(keyword='acc', content=f'declare create({param_device})')
+        return self.default_retval()
```
reuterbal, line 83: The default_retval case is never hit in tests, and I'm wondering if we shouldn't better raise an exception than returning something if we have a Loki directive that doesn't match the "spec".

reuterbal (changes requested):
Many thanks, this is a really cool development!
Conceptually this is fantastic and I don't have anything to change there. But a few other things could be improved, which I've mostly marked with inline comments. 

A few more general remarks here:

I tried to do this during the review but it would be good if you could also double-check where the switch to loki directives has implications on the transformation beyond updating the keyword and naming of directives. For example:
- The SCCAnnotate transformation shouldn't care about the directive type anymore, and anything related to this should be possible to remove
- In a few places there used to be a direct translation from `!$loki ...` to `!$acc ...` - which seems sometimes redundant now. I've tried to flag this but it's likely I overlooked this in other situtaitons.

I know that even after this PR the Loki directives are not set in stone and we can update them as required. But it's likely a good idea to sanity check them on this occasion. I've already flagged some things in inline comments, but particularly for structured directives that consist of `!$loki command` and `!$loki end command` pairs, I don't see any reason why the end statement should include any additional arguments, as it currently does for `end scoped-data`.

Other than that, the usual nagging about some small things, docs and tests ;-)

loki/transformations/pragma_model.py
MichaelSt98: I think there is a currently a problem with mode `strict=True` and modules being incomplete. That is why I introduced that.

MichaelSt98:
> I don't see any reason why the end statement should include any additional arguments, as it currently does for end scoped-data

I think we should (and I wasn't too strict about that yet) have all the additional arguments for both "start" and `end` statement. So that there is a clear and easy to find one-to-one mapping of start and end directives. Further with this we could for example convert structured directives to unstructured directives within the pragma model transformation...

loki/transformations/single_column/annotate.py
```diff
@@ -124,21 +124,24 @@ def annotate_kernel_routine(self, routine):
...
                 # falsely attributed to the body!
-                routine.spec.append(pragma.clone(keyword='acc'))
                 pragma_map[pragma] = None
-        pragma_transformer = Transformer(pragma_map)
-        routine.spec = pragma_transformer.visit(routine.spec)
-        routine.body = pragma_transformer.visit(routine.body)
+                pragma_transformer = Transformer(pragma_map)
+                routine.spec = pragma_transformer.visit(routine.spec)
+                routine.body = pragma_transformer.visit(routine.body)
+                routine.spec.append(pragma.clone(keyword='loki'))
```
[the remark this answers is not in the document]
reuterbal: Sure, but what about the suggestion above? An explicit translation from loki routine to loki routine doesn't seem particularly useful to me?

The translation of the Loki directive to OpenACC should then be taken care of by the PragmaModelTrafo, no?

reuterbal (commented):
> >  I don't see any reason why the end statement should include any additional arguments, as it currently does for end scoped-data
>
> I think we should (and I wasn't too strict about that yet) have all the additional arguments for both "start" and end statement. So that there is a clear and easy to find one-to-one mapping of start and end directives. Further with this we could for example convert structured directives to unstructured directives within the pragma model transformation...

Interesting point and something I hadn't thought about before. But the entire point of the structured directives (compared to unstructured ones) is that they can be considered together, i.e., we can always view the pair of start and end-pragma together and have the contained code as a PragmaRegion. 

So, I'd still argue we should only have the specs on the opening pragma and use `pragma_regions_attached` before invoking the transformer.

MichaelSt98:
As discussed offline: The `PragmaModelTransformation` is the first transformation to really process modules for CLOUDSC2, ecphys, ... which is why I am convinced that we need to PR to the respective branches and not "fix" it within this PR.

loki/transformations/pragma_model.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def pmap_end_loop(self, pragma, parameters, **kwargs):
+        if 'vector' in parameters:
+            return Pragma(keyword='omp', content='end parallel do')
+        if 'gang' in parameters:
+            return Pragma(keyword='omp', content='end target teams distribute')
+        return self.default_retval()
+
+
+class OpenMPPragmaMapper(GenericPragmaMapper):
```
reuterbal, line 270: [optional] Should we call this OpenMPThreadingPragmaMapper to explicitly differentiate from offloading?

loki/transformations/pragma_model.py
```diff
@@ -0,0 +1,346 @@
...
+        return self.default_retval()
+
+
+class PragmaModelTransformation(Transformation):
+    """
+    Transformation to map Loki generic pragmas to a specific
+    pragma model using a child class of :any:`GenericPragmaMapper`.
+
+    Parameters
+    __________
```
reuterbal, line 296:
```suggestion
    ----------
```

loki/transformations/raw_stack_allocator.py
```diff
@@ -334,29 +329,29 @@ def create_stacks_kernel(self, routine, stack_dict, successors):
...
-                    if present_pragma:
-                        pragma_map = {present_pragma: None}
-                        routine.body = Transformer(pragma_map).visit(routine.body)
-                        content = re.sub(r'\bpresent\(', f'present({pragma_string}, ', present_pragma.content.lower())
-                        present_pragma = present_pragma.clone(content = content)
-                        pragma_data_end = None
-                    else:
-                        present_pragma = Pragma(keyword='acc', content=f'data present({pragma_string})')
-                        pragma_data_end = Pragma(keyword='acc', content='end data')
+        # if self.directive:
```
reuterbal, line 332: Leftover? Also, remove/update the comment in the line before that

loki/transformations/raw_stack_allocator.py
```diff
@@ -334,29 +329,29 @@ def create_stacks_kernel(self, routine, stack_dict, successors):
...
+        if pragma_string:
+            pragma_string = pragma_string[:-2].lower()
+
+            # if self.directive == 'openacc':
```
reuterbal, line 336:
```suggestion
```

loki/transformations/single_column/annotate.py
```diff
@@ -192,7 +177,7 @@ def transform_subroutine(self, routine, **kwargs):
             self.annotate_vector_loops(routine)
 
             # Mark all non-parallel loops as `!$acc loop seq`
```
reuterbal, line 179: Comment needs an update

reuterbal (commented):
Many thanks, this looks very good now.
I agree that the fix for the regression tests should be to update the config in cloudsc2. Do you want to take care of raising a PR for this? Since that change on CLOUDSC2 shouldn't break functionality even before this PR is merged, we can get that in and merge the Loki PR after.

Also, do I remember correctly that you were working on putting the directive table into the docs? Do you want to do this as part of the current PR or separately?

Finally, something to consider going forward: For structured directives such as `!$loki structured-data ... !$loki end structured-data`, we might consider a use such as:
```
with pragma_regions_attached(routine):
    <apply pragma model trafo>
```
This might be required when we might add in the future an implementation that uses explicit API calls (think HIP/CUDA/F-API) for data movement instead of equivalent directives, e.g.,

```
!$loki structured-data in(a) out(b)
...
!$loki end structured-data
```
then translated to something conceptually like
```
call copy_to_device(a)
...
call copy_to_host(b)
```

With attached pragma regions we can process the pair of entry/exit pragmas at the same time, with all parameters available to both. 

In any case, it's not required for the current state but should also not be too difficult to accommodate in the current structure.

reuterbal (approved):

PR #486, opened 2025-02-12
Add support for assumed size declarations to Fparser

awnawab:
This PR adds support for assumed size declarations (😑 ) for Fparser, and also adds the mechanics to the `ArgumentShapeAnalysis` to resolve the corresponding explicit shape.

### DIFF
--- a/loki/frontend/tests/test_frontends.py
+++ b/loki/frontend/tests/test_frontends.py
@@ -18,7 +18,7 @@ from loki import (
 )
 from loki.build import jit_compile, clean_test
 from loki.expression import symbols as sym
-from loki.frontend import available_frontends, OMNI, FP
+from loki.frontend import available_frontends, OMNI, FP, HAVE_FP
 from loki.ir import nodes as ir, FindNodes, FindVariables
 
 
@@ -782,7 +782,7 @@ module mod_variable_dimensions
   !$loki dimension(x+y,2*x)
   real(kind=jprb), allocatable :: v5(:,:)
   !$loki dimension(x/2, x**2, (x+y)/x)
-  real(kind=jprb), dimension(:, :, :), pointer :: v6 
+  real(kind=jprb), dimension(:, :, :), pointer :: v6
 end module mod_variable_dimensions
     """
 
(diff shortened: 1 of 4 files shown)


loki/frontend/tests/test_frontends.py
```diff
@@ -1015,3 +1015,38 @@ def test_frontend_derived_type_imports(tmp_path, frontend):
...
+    """
+    Test if assumed size declarations are correctly parsed.
+    """
+
+    fcode = """
+subroutine kernel(a, b, c)
+  implicit none
+  real, intent(in) :: a(*)
+  real, intent(in) :: b(8,*)
+  real, intent(in) :: c(8,0:*)
```
reuterbal, line 1031: Would you mind adding a `d(2,4,3:*)` to demonstrate that non-zero lower bounds and multiple preceeding explicit-shape-specs are working as well? It looks to me like the implementation would handle this correctly but worth double-checking.

loki/transformations/argument_shape.py
```diff
@@ -28,6 +29,7 @@
 )
 from loki.tools import as_tuple, CaseInsensitiveDict
 from loki.types import BasicType
+from loki.expression import symbols as sym, simplify
```
reuterbal, line 32: Please add this to the existing import from `loki.expression` above

loki/transformations/argument_shape.py
```diff
@@ -80,6 +82,35 @@ def transform_subroutine(self, routine, **kwargs):  # pylint: disable=arguments-
                                          if d == ':']
                             vmap[arg] = arg.clone(type=arg.type.clone(shape=new_shape))
 
+                    elif re.search(r'\*$', str(arg.shape[-1])):
```
reuterbal, line 85: Do we really need a regex here? Wouldn't the following work?
```suggestion
                    elif str(arg.shape[-1])[-1] == '*':
```

loki/transformations/argument_shape.py
```diff
@@ -80,6 +82,35 @@ def transform_subroutine(self, routine, **kwargs):  # pylint: disable=arguments-
...
+                        # determine extent of explicit shape part of assumed size declaration
+                        expl_shape_len = [loc for loc, dim in enumerate(arg.shape) if '*' in dim][0]
```
reuterbal, line 87: The standard explicitly mandates `expl_shape_len = len(arg.shape)-1`, can't we use that knowledge here?

> The rank is equal to one plus the number of explicit-shape-specs.
https://j3-fortran.org/doc/year/10/10-007.pdf#page=114

loki/transformations/argument_shape.py
```diff
@@ -119,13 +150,15 @@ class ExplicitArgumentArrayShapeTransformation(Transformation):
 
     def transform_subroutine(self, routine, **kwargs):  # pylint: disable=arguments-differ
 
+        def assumed(dims):
+            return all(d == ':' for d in dims) or re.search(r'\*$', str(dims[-1]))
```
reuterbal, line 154: Same comment about the need for a regex

reuterbal (approved):
Many thanks for making the changes. Looks good to me now!

mlange05 (approved):
Looks fantastic, great addition! Many thanks. GTG :shipit:

PR #488, opened 2025-02-19
PoolAllocator: revise stack creation for 'read-only' temporaries

MichaelSt98:
Filter out unused vars used to rely on dataflow_analysis only putting temporaries on the stack that are defined/written to.
However, there exist some cases where temporaries are only read and it is not the responsibility of this transformation to decide whether that is reasonable.
With this only unused temporaries are not put on the stack, but all other are put on the stack disregarding whether they are written to or only read.

### DIFF
--- a/loki/transformations/tests/test_pool_allocator.py
+++ b/loki/transformations/tests/test_pool_allocator.py
@@ -17,7 +17,8 @@ from loki.expression import (
 from loki.frontend import available_frontends, OMNI, FP
 from loki.ir import (
     FindNodes, CallStatement, Assignment, Allocation, Deallocation,
-    Loop, Pragma, get_pragma_parameters, FindVariables, FindInlineCalls
+    Loop, Pragma, get_pragma_parameters, FindVariables, FindInlineCalls,
+    Intrinsic
 )
 
 from loki.transformations.pool_allocator import TemporariesPoolAllocatorTransformation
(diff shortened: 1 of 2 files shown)


loki/transformations/tests/test_pool_allocator.py
```diff
@@ -348,6 +349,112 @@ def test_pool_allocator_temporaries(tmp_path, frontend, generate_driver_stack, b
...
+        ! shouldn't end up on stack since not used at all
+        real(kind=jprb) :: tmp3(klon, klev)
+        ! shouldn't end up on stack since all dimension are constant
+        real(kind=jprb) :: tmp4(nclv)
+        ! shouldn't end up on stack since all dimension are constant
+        real(kind=jprb) :: tmp5(2)
+        ! should be on the stack although only read and not written to
+        real(kind=jprb) :: tmp6(klon, nclv)
+        ! should be on the stack although only used as an argument in a call
+        real(kind=jprb) :: tmp7(klon, nclv)
```
reuterbal, line 390: This is very nice, covering the matrix of cases with comments explaining what's going on and to expect! ❤️

reuterbal (approved):
Very nice, thanks!

loki/transformations/tests/test_pool_allocator.py
mlange05: ⬆️ What he said! Very neat! 🙏

mlange05 (approved):
Looks great! GTG from me. :shipit:

PR #489, opened 2025-02-20
transformations getting subgraph instead of successors from scheduler

MichaelSt98:
instead of successors, pass the subgraph/subtree to transformations, the subgraph can then be used to derive successors

### DIFF
--- a/loki/batch/sgraph.py
+++ b/loki/batch/sgraph.py
@@ -332,7 +332,8 @@ class SGraph:
             dependencies will also be included until they eventually resolve to a
             :any:`ProcedureItem`.
         """
-        if ProcedureItem in as_tuple(item_filter):
+        item_filter = as_tuple(item_filter)
+        if ProcedureItem in item_filter:
             # ProcedureBindingItem and InterfaceItem are intermediate nodes that take
             # essentially the role of an edge to ProcedureItems. Therefore
             # we need to make sure these are included if ProcedureItems are included
--- a/loki/batch/sgraph.py
+++ b/loki/batch/sgraph.py
@@ -332,7 +332,7 @@ class SGraph:
             dependencies will also be included until they eventually resolve to a
             :any:`ProcedureItem`.
         """
-        if ProcedureItem in item_filter:
+        if ProcedureItem in as_tuple(item_filter):
             # ProcedureBindingItem and InterfaceItem are intermediate nodes that take
             # essentially the role of an edge to ProcedureItems. Therefore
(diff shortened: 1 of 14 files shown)


loki/batch/sgraph.py
```diff
@@ -333,6 +375,33 @@ def successors(self, item, item_filter=None):
                     successors += (child,)
         return successors
 
+    def subsgraph(self, item, item_filter=None):
```
reuterbal, line 378: Im a little undecided about the naming, because the word looks a little weird and I have more than once read "_subs-graph_". Python `snake_case` would suggest to use `sub_sgraph` - what's your thoughts on this @MichaelSt98 @mlange05 ? Open to other suggestions or being overruled here.

reuterbal (changes requested):
Thanks, this is an important first step towards the next topology-modifying transformation additions. 
Changes to transformations that can use this topology information instead of the previous successors data will be introduced in future PRs.

The implementation looks really good! Only comments are a small stylistic change and a question about the naming.

loki/batch/sgraph.py
mlange05: Hmm, naming is hard. Mh intuition would be to stick with `subgraph` to avoid weirdness, but snake case would be fine too for me.

loki/batch/sgraph.py
```diff
@@ -292,6 +312,36 @@ def dependencies(self):
...
+        ----------
+        item_filter : list of :any:`Item` subclasses, optional
+            Filter successor items to only include items of the provided type. By default,
+            all items are considered. Note that including :any:`ProcedureItem` in the
+            ``item_filter`` automatically adds :any:`ProcedureBindingItem` and
+            :any:`InterfaceItem` as well, since these are intermediate nodes. Their
+            dependencies will also be included until they eventually resolve to a
+            :any:`ProcedureItem`.
+        """
+        if ProcedureItem in as_tuple(item_filter):
```
reuterbal, line 335:
```suggestion
        item_filter = as_tuple(item_filter)
        if ProcedureItem in item_filter:
```
Otherwise the conditional in the next line fails if `item_filter` is not a tuple.

reuterbal (changes requested):
Thanks. One more detail to fix, unfortunately, otherwise all good from me!

reuterbal (approved):
Many thanks, everything clear from me!

mlange05 (approved):
Fantastic progress, neatly implemented. GTG from me. :shipit:

PR #491, opened 2025-02-20
Scheduler: Allow explicit file paths to be added as sources

mlange05:
This change allows individual source file locations to be added without resorting to directory globs. This is particularly useful for auto-inlining individual external utility routines, without having to search their entire parent location during discover.

This extends one of the existing tests instead of adding a new one.


reuterbal (approved):
Many thanks, this is a great addition towards more fine-grained control of source trees. GTG!

PR #492, opened 2025-02-24
Loki: Restrict numpy version to avoid JIT incompatibilities

mlange05:
Hoping to get CI tests green again...


reuterbal:
I think this can be closed in favour of #493

PR #493, opened 2025-02-25
Python 3.12/3.13 compatibility

reuterbal:
The distutils build system that underpinned the f90wrap-f2py JIT compilation used in Loki's test base was deprecated a while ago - and Python 3.12+ now ship with a version that no longer supports this. For that reason, Numpy 2+ (f2py) and f90wrap have switched to a Meson build backend. We put off adapting to this for a while and it has now become a problem.

This PR remedies this via the following changes:

- `loki.build` has been modified to use the Meson build backend. To get this to work, I had to change the previous invocation of `f2py-f90wrap` to providing the .F90 instead of a .o file.
- Numpy version requirement has been updated to version >= 2.0 to ensure the availability of the meson build backend in f2py. 
  - However, Numpy 2 only has support for Python 3.9+, so I had to disable Python 3.8 in the test base. Given that this version is EOL, I deem this acceptable and I am not aware of any system we are currently using that wouldn't provide a newer version than this via the module system. In any case, I have not increased the minimum version requirement in pyproject.toml because we are not yet incompatible.
  - Secondly, the latest DaCe release (1.0) doesn't support NumPy 2, yet (but it's been fixed on their main-branch already), and I had to disable dace tests for the moment.
  - On the plus side, this change allowed to enable Python 3.12 and 3.13 support, which means we now fully support the very narrow band of all Python versions that have official support. Python 3.14 might also work already but I haven't tested this, yet.
  - On the downside (again), this breaks MacOS tests, see #496, which remain deactivated for now
- For some reason unclear to me, f90wrap now no longer supports derived type arguments. To unblock this PR, tests have been marked as xfail (which can be disabled by setting environment variable `XFAIL_DERIVED_TYPE_JIT_TESTS=0`. Issue #494 tracks this problem to resolve this subsequently in a follow-on PR.
- A few instances of wrong test implementations have been fixed, which were wrongly implemented before but happened to work due to the underlying implicit behaviour of the jit backend.
- A particularly weird test failure occurs only in Python 3.12 (but is fine in all other versions, including 3.13) when using multiple parallel workers for linting, see #495
- Along the way I also found and removed a few remnants of OFP installation.
- A bug was also found in the chasing of import dependencies in Loki's `Obj` implementation when the import statement includes a module nature specification (e.g., `NON_INTRINSIC`). This has been fixed and a corresponding test has been added but for some reason, `loki/build/tests/test_build.py` does not get executed??? (Would appreciate any insight here!)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,7 +100,7 @@ if( NOT HAVE_NO_INSTALL )
     # Setup Python virtual environment
     include( python_venv )
     set( loki_VENV_PATH ${CMAKE_CURRENT_BINARY_DIR}/loki_env )
-    set( PYTHON_VERSION 3.8...<3.12 )
+    set( PYTHON_VERSION 3.8 )
     setup_python_venv( ${loki_VENV_PATH} ${PYTHON_VERSION} )
     install( DIRECTORY ${loki_VENV_PATH} DESTINATION . USE_SOURCE_PERMISSIONS )
 
--- a/.github/workflows/regression_tests.yml
+++ b/.github/workflows/regression_tests.yml
@@ -17,11 +17,23 @@ jobs:
   regression_tests:
     name: Python ${{ matrix.python-version }}
 
-    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false  # false: try to complete all jobs
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        name:
+          - linux gnu-13
+
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+
+        include:
+          - name: linux gnu-13
+            os: ubuntu-24.04
+            install-options: --with-omni --with-tests --without-dace
+            env: JAVA_HOME=${JAVA_HOME_11_X64}
+            pkg-dependencies: graphviz gfortran byacc flex openjdk-11-jdk ant cmake meson ninja-build libhdf5-dev libopenmpi-dev
+            pip-dependencies: pyyaml fypp
+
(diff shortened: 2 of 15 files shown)


CMakeLists.txt
```diff
@@ -100,7 +100,7 @@ if( NOT HAVE_NO_INSTALL )
     # Setup Python virtual environment
     include( python_venv )
     set( loki_VENV_PATH ${CMAKE_CURRENT_BINARY_DIR}/loki_env )
-    set( PYTHON_VERSION 3.8...<3.12 )
+    set( PYTHON_VERSION 3.8 )
```
mlange05, line 103: Shouldn't this be 3.9 now? No need to fix here, just looking for confirmation to fix in a follow-up.

mlange05 (approved):
Ok, this looks fantastic (even though the follow-on work list is gruesome). The general direction is perfectly right though, and the JIT engine needs some TLC anyway. Many thanks though for this gigantic effort in record time, and apologies for all the leftover mess that was my fault. GTG from me. :shipit:

CMakeLists.txt
reuterbal: I left it intentionally at 3.8 because Loki technically still works with that - so no need to break the integration there. It's just that we can't test it anymore (or at least not with JIT). But I realize I've actually increased the version requirement in pyproject.toml, so it wouldn't work as is. My proposal would be to lower pyproject.toml to 3.8 again until we hit some blocking issue that renders Loki itself (not just the JIT tests) incompatible

reuterbal:
See #499 for the pytest issue with loki.build

PR #500, opened 2025-02-26
optionally duplicating/separating the underlying subgraph when duplicating item/call 

MichaelSt98:
Duplicating a kernel and optionally duplicating/separating the underlying subgraph.

E.g., starting with:

callgraph_0.pdf

the result of the DuplicateKernel trafo with duplicate_kernels=('kernel',) is:

callgraph_1.pdf

However, with this PR one can specify duplicate_subgraph=True, which yields for the example above:

callgraph_2.pdf

### DIFF
--- a/loki/transformations/dependency.py
+++ b/loki/transformations/dependency.py
@@ -29,18 +29,135 @@ class DuplicateKernel(Transformation):
     duplicate_module_suffix : str, optional
         Suffix to be used to append the original module name(s),
         if defined, otherwise `duplicate_suffix`
+    duplicate_subgraph : bool, optional
+        Whether or not duplicate the subgraph beneath the kernel(s)
+        that are duplicated.
     """
 
     creates_items = True
     reverse_traversal = True
 
     def __init__(self, duplicate_kernels=None, duplicate_suffix='duplicated',
-                 duplicate_module_suffix=None):
+                 duplicate_module_suffix=None, duplicate_subgraph=False):
         self.suffix = duplicate_suffix
         self.module_suffix = duplicate_module_suffix or duplicate_suffix
         self.duplicate_kernels = tuple(kernel.lower() for kernel in as_tuple(duplicate_kernels))
+        self.duplicate_subgraph = duplicate_subgraph
 
-    def _create_duplicate_items(self, successors, item_factory, config):
+    def _get_new_item_name(self, item):
+        """
+        Get new/duplicated item name, more specifically ``local_name``,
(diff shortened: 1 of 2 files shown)


loki/transformations/dependency.py
```diff
@@ -29,18 +29,135 @@ class DuplicateKernel(Transformation):
...
+            new_call_name = f'{call_name}{self.suffix}'.lower()
+            call_new_item = new_dependencies[new_call_name]
+            proc_symbol = call_new_item.ir.procedure_symbol.rescope(scope=new_item.ir)
+            call_map[call] = call.clone(name=proc_symbol)
+        # TODO: imports at module level ...
+        imp_map = {}
+        for imp in FindNodes(ir.Import).visit(new_item.ir.spec):
+            new_symbols = [symbol.clone(name=f'{symbol.name}{self.suffix}') for symbol in imp.symbols]
+            imp_map[imp] = imp.clone(module=f'{imp.module.lower()}{self.module_suffix}',
+                                     symbols=as_tuple(new_symbols))
```
reuterbal, line 153: Does this really do the right thing? This looks like it blindly appends the suffix to every import and symbol, regardless of whether these are actually part of the replicated call tree (i.e., I think this would also rename a `use iso_fortran_env, only: real64` to `use iso_fortran_env{suffix}, only: real64{suffix}`). Maybe you can add a test for this?

loki/transformations/dependency.py
```diff
@@ -29,18 +29,135 @@ class DuplicateKernel(Transformation):
...
+            to be renamed.
+        new_dependencies : dict
+            Dictionary used to get information about how
+            to rename calls and imports.
+        """
+        call_map = {}
+        for call in FindNodes(ir.CallStatement).visit(new_item.ir.body):
+            call_name = str(call.name).lower()
+            new_call_name = f'{call_name}{self.suffix}'.lower()
+            call_new_item = new_dependencies[new_call_name]
```
reuterbal, line 145: What happens if we have here a call to something that is disabled/ignored? I'm thinking of, for example, calls to `DR_HOOK` or procedures from an external library? Don't we need to somehow check if the call is part of the subgraph?

reuterbal (changes requested):
Thanks, this looks great. I think there are a few omissions in the handling of the renaming which would be good to take care off, otherwise very happy with this.

reuterbal (approved):
Fantastic, many thanks. Agreed that we may discover more edge cases but this way at least those that came to mind are already covered :)

Issue #497, opened 2025-02-26
Discussion: external module item


MichaelSt98:
Executing:

```py
import pathlib
from loki import (
        Transformation, ProcedureItem, ModuleItem,
        Scheduler, SchedulerConfig
)

class Trafo(Transformation):

    item_filter = (ProcedureItem, ModuleItem)

    def transform_module(self, module, **kwargs):
        pass

    def transform_subroutine(self, routine, **kwargs):
        pass

if __name__ == "__main__":

    fcode_parkind_mod = """
module parkind1
  implicit none
  integer, parameter :: jprb = selected_real_kind(13,300)
  integer, parameter :: jpim = selected_int_kind(9)
  integer, parameter :: jplm = jpim
end module parkind1
    """.strip()
    fcode_driver = """
module driver_mod
  contains
  subroutine driver(nlon, klev, nb, ydml_phy_mf)
    use parkind1, only: jpim, jprb
    use kernel1_mod, only: kernel1
    implicit none
    type(model_physics_mf_type), intent(in) :: ydml_phy_mf
    integer(kind=jpim), intent(in) :: nlon
    integer(kind=jpim), intent(in) :: klev
    integer(kind=jpim), intent(in) :: nb
    integer(kind=jpim) :: jstart
    integer(kind=jpim) :: jend
    integer(kind=jpim) :: b
    jstart = 1
    jend = nlon
    do b = 1, nb
        call kernel1()
    enddo
  end subroutine driver
end module driver_mod
    """.strip()
    fcode_kernel1 = """
module kernel1_mod
  contains
  subroutine kernel1()
    use parkind1, only: jpim, jprb
  end subroutine kernel1
end module kernel1_mod
    """.strip()
    
    here = pathlib.Path().resolve()

    (here/'driver.F90').write_text(fcode_driver)
    (here/'kernel1_mod.F90').write_text(fcode_kernel1)

    config = {
        'default': {
            'mode': 'idem',
            'role': 'kernel',
            'expand': True,
            'strict': True,
            'ignore': ['parkind1'],
        },
        'routines': {
            'driver': {'role': 'driver'}
        }
    }

    definitions = ()
    scheduler = Scheduler(paths=[here], config=SchedulerConfig.from_dict(config),
                          definitions=definitions, xmods=[here])
    scheduler.process(transformation=Trafo())
```

gives:

```
Module parkind1 not found in self.item_cache. Marking parkind1 as an external dependency
Traceback (most recent call last):
[...]

PR #501, opened 2025-03-03
SFilter: fix for external and ignored items

MichaelSt98:
Fix for #497.

Thanks @reuterbal!

### DIFF
--- a/loki/batch/sfilter.py
+++ b/loki/batch/sfilter.py
@@ -8,7 +8,6 @@
 import networkx as nx
 
 from loki.batch.item import Item, ExternalItem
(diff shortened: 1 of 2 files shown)


reuterbal (changes requested):
Thanks. The underlying issue was that SFilter skips
- external items, if `include_external == False`
- ignored items, if `exclude_ignored == True`

But it did not consider the case that an external item might be marked as ignored and should therefore be skipped.

Would you mind adding the reproducer from #497 as a test for this situation?

reuterbal (approved):
Many thanks, GTG!

PR #503, opened 2025-03-11
Fix numpy deprecation warnings

reuterbal:
We have recently acquired a substantial list of warnings when executing tests, of the form

> DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)

These are related to our various ways of initializing 1-sized arrays as `out` or `inout` arguments to compiled subroutines. While this is in principal the recommended approach, it seems that Numpy wants the user to only use `np.array(5)`-style syntax to create these objects. See also the [documentation](https://numpy.org/doc/stable/f2py/python-usage.html#scalar-arguments).

Piggy-backed is the re-enabling of MacOS in the test matrix.


mlange05 (approved):
Very nice find and neat fix! Much appreciated. GTG from me. :shipit:

Issue #504, opened 2025-03-12
Invalid parsing of empty PRINT statements with fparser


quepas:
### What happened?

When parsing and regenerating back an empty PRINT statement with Loki:
```fortran
PRINT *
```
what we get is an invalid Fortran code:
```fortran
PRINT *, None
```

#### Expected result

We should get back the original code:

```fortran
PRINT *
```

### What are the steps to reproduce the bug?

To reproduce, just run the code below:

```python
from loki import Sourcefile

fcode = """
SUBROUTINE test()
    IMPLICIT NONE
    print *
    print*, "test1"
END SUBROUTINE test
""".strip()
source = Sourcefile.from_source(fcode)
print(source['test'].to_fortran())
```

### Version

v0.3.1

### Platform (OS and architecture)

Ubuntu 24.04 (x64); Python 3.12.3

### Organisation

École Polytechnique/IPSL

Acknowledgments: this bug has been found and documented by  @remi-kazeroni (IPSL/CEA)

PR #505, opened 2025-03-12
Fix: Invalid parsing of empty PRINT statements with fparser

quepas:
This simple PR tries to resolve the bug mentioned in #504.

Tell me if some unit tests for this case are necessary in your opinion.

### DIFF
--- a/loki/frontend/tests/test_frontends.py
+++ b/loki/frontend/tests/test_frontends.py
@@ -1102,3 +1102,27 @@ end subroutine kernel
     assert c.shape[1].lower == 0
 
     assert all('*' in str(shape) for shape in [a.shape, b.shape, c.shape])
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_empty_print_statement(frontend):
+    """
+    Test if an empty print statement (PRINT *) is parsed correctly.
+    """
+    fcode = """
+SUBROUTINE test_routine()
+    IMPLICIT NONE
+    print *
+    ! Using single quotes to simplify the test comparison (see below)
+    print *, 'test_text'
+END SUBROUTINE test_routine
+    """.strip()
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+    print_stmts = [
+        intr for intr in FindNodes(ir.Intrinsic).visit(routine.ir)
+        if 'print' in intr.text.lower()
+    ]
+    assert print_stmts[0].text.lower() == "print *"
+    # NOTE: OMNI always uses single quotes ('') to represent string data in PRINT statements
+    #       while fparser will mimic the quotes used in the parsed source code
+    assert print_stmts[1].text.lower() == "print *, 'test_text'"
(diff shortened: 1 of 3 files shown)


reuterbal (commented):
Many thanks for finding this and filing the fix. This looks very good!

It would be great indeed if you could add a simple test that verifies the behaviour. Something like this (untested) in loki/frontend/tests/test_frontends.py should work:

```py
@pytest.mark.parametrize('frontend', available_frontends())
def test_empty_print_statement(frontend):
    fcode = """
    SUBROUTINE test()
        IMPLICIT NONE
        print *
        print*, "test1"
    END SUBROUTINE test
    """.strip()
    print_stmts = [
        intr for intr in FindNodes(ir.Intrinsic).visit(routine.ir)
        if 'print' in intr.text.lower()
    ]
    assert all('none' not in intr.text.lower() for intr in print_stmts)
```

reuterbal:
NB: Please bring this PR up-to-date with main - yesterday, a breaking change was introduced in the regression tests that makes this necessary. Apologies about that.

quepas:
> It would be great indeed if you could add a simple test that verifies the behaviour. Something like this (untested) in loki/frontend/tests/test_frontends.py should work:

Of course! I am on it :)

loki/frontend/tests/test_frontends.py
```diff
@@ -1102,3 +1102,27 @@ def test_assumed_size_declarations():
...
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+    print_stmts = [
+        intr for intr in FindNodes(ir.Intrinsic).visit(routine.ir)
+        if 'print' in intr.text.lower()
+    ]
+    assert print_stmts[0].text.lower() == "print *"
+    # NOTE: OMNI uses single quotes ('') while fparser uses double quotes ("")
+    #       to represent strings with PRINT statement
+    print_with_str = "print *, 'test_text'" if frontend == OMNI else 'print *, "test_text"'
+    assert print_stmts[1].text.lower() == (print_with_str)
```
quepas, line 1128: This part of the test shows that PRINT statements are differently stored as text by OMNI and fparser frontends. The first one uses single quotes and the second one uses double quotes. @reuterbal : I don't know if it makes sense but it is possible to unify the two, just by storing the strings in PRINT statements with the same type of quotes.

reuterbal: Sorry about that, these "harmonisations" by OMNI are annoying. You could try using single quotes in your original fcode - I suspect fparser might retain them like that, and then you would end up with the same representation for both.
Alternatively, you could explicitly substitute single by double quotes before the comparision:
```py
# OMNI replaces the double quotes by single quotes
assert print_stmts[1].text.replace("'", '"').lower() == 'print *, "test_text"'
```

quepas: Ahh yes, I didn't think about the fact that fparser might just mimics the type of quotes it sees at the beginning! Thank you. I will try with using single quotes only and leaving some appropriate comment then!

reuterbal (approved):
Thank you very much, looks great!

Issue #506, opened 2025-03-12
Missing PROTECTED access specifier in the fgen backend


quepas:
### What happened?

The `PROTECTED` variable attribute is correctly parsed by fparser, however it is not taken into account by the fgen backend.

Parsing and regenerating this:

```fortran
INTEGER, SAVE, PROTECTED :: var0
```
results in:
```fortran
INTEGER, SAVE :: var0
```

But it should be as before!

### What are the steps to reproduce the bug?

Just run the code below:

```python
from loki import Sourcefile
fcode = """
module mod1
    implicit none
    integer, save, protected :: var0
    integer, save            :: var1
end module mod1
    """.strip()
source = Sourcefile.from_source(fcode)
mod1 = source['mod1']
print(mod1.symbols[0].type) # Output is: <SymbolAttributes BasicType.INTEGER, save, protected>
print(mod1.symbols[1].type) # Output is: <SymbolAttributes BasicType.INTEGER, save>
print(source.to_fortran())
```

### Version

v0.3.1

### Platform (OS and architecture)

Ubuntu 24.04 (x64); Python 3.12.3

### Organisation

École Polytechnique/IPSL

Acknowledgments: this bug has been found and documented by @remi-kazeroni (IPSL/CEA)

PR #507, opened 2025-03-12
Missing PROTECTED access specifier in the fgen backend

quepas:
Fixes missing PROTECTED attribute in the fgen backend #506

### DIFF
--- a/loki/backend/fgen.py
+++ b/loki/backend/fgen.py
@@ -861,6 +861,8 @@ class FortranCodegen(Stringifier):
             attributes += ['PRIVATE']
         if o.public:
             attributes += ['PUBLIC']
+        if o.protected:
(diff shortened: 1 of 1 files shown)


reuterbal (changes requested):
Many thanks - this looks great! Could you please add a corresponding test?

Unfortunately, you also have to update your branch with latest main due to a breaking change in one of the regression tests.

quepas:
> Many thanks - this looks great! Could you please add a corresponding test?
> 
> Unfortunately, you also have to update your branch with latest main due to a breaking change in one of the regression tests.

Done! There was also one small thing missing from the OMNI parser, buit I have added it.

reuterbal (approved):
Very nice work, many thanks!

PR #508, opened 2025-03-12
#499: Rename loki.build to loki.jit_build

### LINKED ISSUE #499: Rename `loki.build`
### What maintenance does this project need?

Pytest skips subdirectories called `build` when collecting tests, which means the tests defined in `loki/build/tests` are never run. See the following links for more details:
https://github.com/pytest-dev/pytest/discussions/12621#discussioncomment-10066085
https://github.com/pytest-dev/pytest/issues/12625

We should rename `loki.build` to something else to enable testing this component.

### Organisation

_No response_

reuterbal:
This ensures the tests inside the build package are being run, which fixes #499 . Draft-PR for now because one of the tests in there doesn't actually pass at the moment...


reuterbal:
@mlange05 Many thanks for figuring this out - the fix is fine as the specific pattern causing problems here isn't something we encounter in reality. 

I can't approve the PR as I'm the original author but from my side this is ready for integrating.

mlange05 (approved):
Thanks for finding this odd idiosyncrasy. Renaming is fine, and I agree with the fix being ok for now. GTG now.

I'll file a reproducer to upstream f90wrap, for completeness.

mlange05:
For completeness, I've filed a reproducer here:
https://github.com/jameskermode/f90wrap/issues/253

PR #509, opened 2025-03-12
Update test_cloudsc

reuterbal:
This makes the regression test compatible with the changes introduced in ecmwf-ifs/dwarf-p-cloudsc#113

~The top commit in this branch points to a temporary cloudsc branch and should be removed once the above PR is merged.~


mlange05 (approved):
Yes, looks good. Many thanks. :shipit:

Issue #510, opened 2025-03-12
Access-spec statements ignored when importing symbols from other modules


remi-kazeroni:
### What happened?

Currently, access control when importing symbols from another module does not check symbols listed in `private_access_spec` and `public_access_spec` which can result in a wrong type inference. It seems that access-spec statements are not used at parsing: https://github.com/ecmwf-ifs/loki/blob/main/loki/frontend/fparser.py#L480-L483

```fortran
module private_var
  integer :: var
  private :: var
end module private_var

module public_var
  integer :: var
  public :: var
end module public_var

module mod_main
  use public_var
  use public_var

  contains

    subroutine test()
      integer :: result
      result = var
      ! `var` is imported from private_var module, but it should be accessed from public_var
    end subroutine test

end mod_main
```

This also applies to procedures whose access is defined in access-spec statements. However `private` attributes at variable declaration and default private access are correctly interpreted. For example:
```fortran
module private_var
  integer, private :: var
end module private_var
```
or
```fortran
module private_var
  private
  integer :: var
end module private_var
```

### What are the steps to reproduce the bug?

_adapted from test https://github.com/ecmwf-ifs/loki/blob/main/loki/frontend/tests/test_frontends.py#L830-L872_

```python
code_var_private = """
module var_private
    integer :: var
    private :: var
end module var_private
    """
code_var_public = """
module var_public
    integer :: var
    public :: var
end module var_public
    """
code_mod_main = """
module mod_main
    use var_public
    use var_private
contains

    subroutine test()
        integer :: result
        result = var
    end subroutine test

end module mod_main
    """

var_private = Module.from_source(code_var_private)
var_public = Module.from_source(code_var_public)
mod_main = Module.from_source(code_mod_main, definitions=[var_private, var_public])
var = mod_main.subroutines[0].body.body[0].rhs
# Check if this is really our symbol
assert var.name == "var"
assert var.scope is mod_main
# Check if the symbol is imported
assert var.type.imported is True
# Check if the symbol comes from the var_public module
print(var.type) # Output is <SymbolAttributes BasicType.INTEGER, imported, module=Module:: var_private>
assert var.type.module is var_public # Failing check
```
```python
AssertionError                            Traceback (most recent call last)
Cell In[90], line 37
     35 assert var.type.imported is True
     36 # Check if the symbol comes from the var_public module
---> 37 assert var.type.module is var_public

AssertionError: 
```

### Version

0.3.1

### Platform (OS and architecture)

[...]

PR #511, opened 2025-03-12
Check access-spec lists when importing symbols from modules

remi-kazeroni:
As mentioned in #510 , symbols listed in `private_access_spec` and `public_access_spec` of modules are not analyzed when importing symbols from other modules. This PR proposes a fix for the issue. I can add a specific unit test if needed

### DIFF
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -479,8 +479,15 @@ class FParser2IR(GenericVisitor):
                 # Import symbol attributes from module, if available
                 for k, v in module.symbol_attrs.items():
                     # Don't import private module symbols
-                    if v.private or (module.default_access_spec == "private" and not v.public):
+                    if v.private:
                         continue
+                    else:
+                        if module.default_access_spec == "private":
+                            if k not in module.public_access_spec and not v.public:
+                                continue
+                        else:
+                            if k in module.private_access_spec:
+                                continue
                     if k in rename_list:
                         local_name = rename_list[k].name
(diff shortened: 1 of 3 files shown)


loki/frontend/fparser.py
```diff
@@ -479,8 +479,15 @@ def visit_Use_Stmt(self, o, **kwargs):
...
-                    if v.private or (module.default_access_spec == "private" and not v.public):
+                    if v.private:
                         continue
+                    else:
+                        if module.default_access_spec == "private":
+                            if k not in module.public_access_spec and not v.public:
+                                continue
+                        else:
+                            if k in module.private_access_spec:
+                                continue
```
reuterbal, line 490: The else isn't required after a continue.
```suggestion
                    if module.default_access_spec == "private":
                        if k not in module.public_access_spec and not v.public:
                            continue
                    else:
                        if k in module.private_access_spec:
                            continue
```

reuterbal (commented):
Many thanks for this. We are not very diligent with handling access specs, yet, unfortunately.

The fix looks great. It would be much appreciated if you could add a test for this, too!

reuterbal:
NB: Please bring this PR up-to-date with main - yesterday, a breaking change was introduced in the regression tests that makes this necessary. Apologies!

remi-kazeroni:
@reuterbal - Thanks for your feedback on this issue. I think I've addressed all your points. Please let me know if there is anything else to do or to improve here. And overall, thanks to the team for providing this great tool with very nice capabilities for source-to-source transformation 👍

reuterbal (approved):
Fantastic, many thanks! 🙏

PR #512, opened 2025-03-14
CMakePlan: ensure file list matches what the build system provided to the Scheduler

reuterbal:
When using a system where overlay file systems are in place, or if symlinks are used in a repository, the premature resolution of file paths to absolute paths can cause aliasing problems in the CMake macros. There, we rely on the implicit assumption that the relative paths of the source files in a CMake target match the lists we write into the CMake plan.

This adds a test for this behaviour using an artificial symlink setup and fixes the CMake transformation accordingly.


mlange05 (approved):
Looks fantastic and, as always, expertly and diligently tested! Much obliged! GTG from me :shipit:

Issue #494, opened 2025-02-26
f90wrap no longer works for subroutines with user-defined derived type arguments


reuterbal:
### What happened?

This is supposed to be the key selling point of f90wrap, so it's unclear why this fails after the switch to Meson in #493. There, tests are now marked as xfail-ing but can be re-enabled by setting the environment variable `XFAIL_DERIVED_TYPE_JIT_TESTS=0`.

### What are the steps to reproduce the bug?

Run pytest. Primarily affected are tests in `test_frontend.py` and `test_derived_types.py`.

### Version

0.3.0-dev

### Platform (OS and architecture)

Any

### Relevant log output

```shell

```

PR #514, opened 2025-03-18
Fix JIT tests with derived types

mlange05:
As described in issue #494 the f90wrap step of our built-in JIT compiler does has started having problems with the recent switch to meson. This PR fixes this by adding a dedicated compilation step of the pure Fortran sources into a small, static utility lib that then gets linked into the dynamic lib alongside the wrappers by `f2py-f90wrap`. This follows the internal examples of 90wrap.


reuterbal (approved):
Very nice, thanks so much for figuring this out!

PR #517, opened 2025-03-24
Backend: Style dataclasses for fgen code generation

mlange05:
The main focus of this PR is to make the "look-and-feel" of the generated Fortran code in the `fgen` backend more configurable through the addition of a set of `Style` dataclasses. The new classes contain various options that define mostly the use of spaces in the Fortran backend, and allow a customised `IFSFortranStyle` class to be added that make the generated code look a little bit more like the code that was input into the Loki toolchains. This is intended to braing the do-nothing parse-unparse cycle closer to neutral for regular IFS source files, without forcing IFS-specific layout choices on the Loki default behaviour.

Of course, this initial draft is mainly concerned with the mechanics of exposing "style" classes and leaves many additional "style-choices" configurable at a later stage. For now, I've mainly focussed at making node-specific indentation and spaces around `END DO/ENDDO` and `END IF/ENDIF` configurable. As part of the enabling mechanical changes, I've also moved the `Stringifier` and it's associated test into `loki.backend`.

In some more detail:
* Moved `Stringifier` and tests to `loki.backend`
* Added a `DefaultStyle` dataclass for all stringifiers that defines linewidth and default indentation (character and width).
* Added whitespace trimming at the end of each line (including empties).
* Switched from one double-space default indentation to two single-space default indents
* Added a dedicated `FortranStyle` for the `fgen` stringifier that allows configuring default indentation for modules procedures and certain node types.
* Added a dedicated `IFSFortranStyle` that indents less and uses compact `ENDIF`/`ENDDO` notation and exposed this to the `fgen` utilities and the `FileWriterTransformation`. 

P.S. Further additions at a later stage could aim to make the use of "SHOUTCASE" configurable and the way we handle interleaved-comments in call and procedure signatures. But I've left those for a later PR.


loki/backend/pygen.py
```diff
@@ -233,9 +236,9 @@ def visit_Loop(self, o, **kwargs):
...
             cntrl = f'range({start}, {end} + 1)'
         header = self.format_line('for ', var, ' in ', cntrl, ':')
-        self.depth += 1
+        self.depth += self.style.indent_default
         body = self.visit(o.body, **kwargs)
-        self.depth -= 1
+        self.depth -= self.style.indent_default
         return self.join_lines(header, body)
 
     def visit_WhileLoop(self, o, **kwargs):
```
reuterbal, line 244: Because codecov flags them, I'll point out a few coverage omissions in case you'd like to do some testing TLC.
This visitor method isn't tested.

loki/backend/pygen.py
```diff
@@ -267,14 +270,14 @@ def visit_Conditional(self, o, **kwargs):
         is_elseif = kwargs.pop('is_elseif', False)
         keyword = 'elif' if is_elseif else 'if'
         header = self.format_line(keyword, ' ', self.visit(o.condition, **kwargs), ':')
-        self.depth += 1
+        self.depth += self.style.indent_default
         body = self.visit(o.body, **kwargs)
         if o.has_elseif:
-            self.depth -= 1
+            self.depth -= self.style.indent_default
```
reuterbal, line 276: The `has_elseif` case isn't covered.

loki/backend/pygen.py
```diff
@@ -267,14 +270,14 @@ def visit_Conditional(self, o, **kwargs):
...
             else_body = [self.visit(o.else_body, is_elseif=True, **kwargs)]
         else:
             else_body = [self.visit(o.else_body, **kwargs)]
-            self.depth -= 1
+            self.depth -= self.style.indent_default
             if o.else_body:
                 else_body = [self.format_line('else:')] + else_body
```
reuterbal, line 282: The `else_body` case isn't covered.

loki/sourcefile.py
```diff
@@ -424,10 +424,10 @@ def make_complete(self, **frontend_args):
     def source(self):
         return self._source
 
-    def to_fortran(self, conservative=False, cuf=False):
+    def to_fortran(self, conservative=False, cuf=False, style=None):
         if cuf:
-            return cufgen(self, conservative=conservative)
-        return fgen(self, conservative=conservative)
+            return cufgen(self, conservative=conservative, style=style)
```
reuterbal, line 429: `cufgen` via this entry point is currently not covered.

reuterbal (approved):
Many thanks, this looks fantastic. The structure is nice and flexible and easy to read.

Due to the injection of the style argument, codecov lists a few omissions in our test coverage. I've pointed them out in case you'd like to mop them up here, but it doesn't look like anything serious.

loki/sourcefile.py
```diff
@@ -424,10 +424,10 @@ def make_complete(self, **frontend_args):
     def source(self):
         return self._source
 
-    def to_fortran(self, conservative=False, cuf=False):
+    def to_fortran(self, conservative=False, cuf=False, style=None):
```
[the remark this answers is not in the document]
mlange05: Hmmm, good point. We could probably group the specialisations of the backends (F/C/Py) and then dispatch particular flavours based on the style-class inheritance. In particular the dispatch from `fgen/cufgen`, etc. could be unified, but I think this is beyond this PR. In fact, there's a bunch of "easy" refactoring steps that could be build on that, and we should probably create a wishlist issue to track them.

loki/sourcefile.py
mlange05: This is actually quite awkward to test, as it's merely a convenience entry point. I think it's safe to leave this until we generalised the backend dispatch a bit more (see above comment).

mlange05:
Hi @reuterbal , thanks for the review. I've fixed the pygen coverage as you suggested, but left the other pieces you flagged for later (explanation inlined). I've also created issue #519 to track further work and ideas on this topic, as well as facilitate questions and discussions. Let's see what codecov says once the runners are back.

loki/sourcefile.py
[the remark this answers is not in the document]
mlange05: Done, issue #519

reuterbal (approved):

PR #518, opened 2025-03-26
Frontend: Make storing `Source` optional

mlange05:
The primary objective of this PR is to allow toggling the storage of `Source` objects in frontends (in particular FParser), and add explicit test coverage for source extraction. In particular this adds a central loki config option "frontend-store-source" (on by default), which allows us to switch off source-object generation during FParser parses. 

This paves the way for:
* Clean integration of a dedicated UID to make IR nodes unique (we previously implicitly relied on the `Source` to do so); hopefully this will allow us to finally land PR #153 .
* Enable closer control over `Source` objects for developing a clean "conservative" Fortran backend for clean diff generation.

Integration note: Along the way, I've moved some tests around again to reduce the per-test file length and group them more cleanly. For this reason I've also had to piggy-back the first commit of PR #517 . The PR then also separates the IR tests for visitors (mostly finders) and transformers. I've also managed to remove two unused utility methods from the FParser frontend.


reuterbal (approved):
Thanks, the changes look very good to me. Have you by any chance checked how the rest of the test base behaves if you disable storing the source via the env variable?

I'd expect a few tests to fail where we capture the specific behaviour for which uuids will be required - which is fine but useful to know whether we capture this already and have something to verify a uuid implementation subsequently.

PR #522, opened 2025-04-02
Transform multiple targets/libs with a single invocation of `loki_transform_target`

MichaelSt98:
Allow multiple targets for 'loki_transform_target' invocations, items/routines can be assigned using the new attribute `lib`.

Imagine having:

```
    projA: driverB -> kernelB -> compute_l1<replicated> -> compute_l2
                         |
    projB:          ext_driver -> ext_kernel
```

and defining a config like:

```
        'default': {
            'role': 'kernel',
            'expand': True,
            'strict': True,
            'ignore': ('header_mod',),
            'mode': 'foobar',
            'lib': 'projAlib'
        },
        'routines': {
            'driverB': {'role': 'driver'},
            'ext_driver': {'lib': 'projBlib'}
        }
```

it is possible to transform files for both libs `projAlib` and `projBlib` with a single invocation and without the necessity of defining `ignore` and calling `loki_transform_target` twice.


loki/transformations/build_system/plan.py
```diff
@@ -96,36 +96,72 @@ def plan_file(self, sourcefile, **kwargs):
...
         """
+        sources_to_transform = []
+        sources_to_append = []
+        sources_to_remove = []
+        for _, vals in self.sources_to_transform.items():
+            sources_to_transform.extend(vals)
+        for _, vals in self.sources_to_append.items():
+            sources_to_append.extend(vals)
+        for _, vals in self.sources_to_remove.items():
+            sources_to_remove.extend(vals)
```
reuterbal, line 132: Instead of `items`, the canonical pattern would be to use `values`, which you can then shorten as:

```suggestion
        sources_to_transform = [s for sources in self.sources_to_transform.values() for s in sources]
        sources_to_append = [s for sources in self.sources_to_append.values() for s in sources]
        sources_to_remove = [s for sources in self.sources_to_remove.values() for s in sources]
```

cmake/loki_transform.cmake
```diff
@@ -295,59 +298,75 @@ function( loki_transform_target )
...
+        endforeach()
 
-        # Add the Loki-generated sources to our target
-        target_sources( ${_PAR_T_TARGET} PRIVATE ${LOKI_SOURCES_TO_APPEND} )
-    endif()
+        if( NOT _PAR_T_COPY_UNMODIFIED )
+            # Update the target source list
+            set_property( TARGET ${_target} PROPERTY SOURCES ${_target_sources} )
+        else()
+            # Copy the unmodified source files to the build dir
```
reuterbal, line 330: Are we actually using this feature (copying unmodified files) anywhere? Otherwise we could consider removing this in a future PR.

cmake/loki_transform.cmake
```diff
@@ -225,21 +225,23 @@ function( loki_transform_target )
     endif()
 
     if( NOT _PAR_T_TARGET )
-        ecbuild_critical( "The call to loki_transform_target() doesn't specify the TARGET." )
+        ecbuild_critical( "The call to loki_transform_target() doesn't specify the TARGET(S)." )
```
reuterbal, line 228: I would suggest to stick to the exact name of the option.
```suggestion
        ecbuild_critical( "The call to loki_transform_target() doesn't specify the TARGET." )
```

cmake/loki_transform.cmake
```diff
@@ -225,21 +225,23 @@ function( loki_transform_target )
...
     endif()
 
     if( NOT _PAR_T_COMMAND )
         set( _PAR_T_COMMAND "convert" )
     endif()
 
     if( NOT _PAR_T_PLAN )
-        ecbuild_critical( "No PLAN specified for loki_transform_target()" )
+        ecbuild_critical( "No PLAN(S) specified for loki_transform_target()" )
```
reuterbal, line 236:
```suggestion
        ecbuild_critical( "No PLAN specified for loki_transform_target()" )
```

cmake/loki_transform.cmake
```diff
@@ -225,21 +225,23 @@ function( loki_transform_target )
...
     endif()
 
-    ecbuild_info( "[Loki] Loki scheduler:: target=${_PAR_T_TARGET} mode=${_PAR_T_MODE} frontend=${_PAR_T_FRONTEND}")
+    ecbuild_info( "[Loki] Loki scheduler:: target(s)=${_PAR_T_TARGET} mode=${_PAR_T_MODE} frontend=${_PAR_T_FRONTEND}")
```
reuterbal, line 239:
```suggestion
    ecbuild_info( "[Loki] Loki scheduler:: target=${_PAR_T_TARGET} mode=${_PAR_T_MODE} frontend=${_PAR_T_FRONTEND}")
```

reuterbal (approved):
Very nice! Many thanks!

Only some minor stylistic remarks but nothing I deem essential. Very happy to bring this in!

cmake/loki_transform.cmake
MichaelSt98: Good question! I don't think so, but I am not sure.

mlange05 (approved):
Fantastic addition that will resolve a bunch of integration issues! Very cool, and very much obliged! GTG from me! :shipit:

PR #523, opened 2025-04-04
Offline Loki installation support for CMake

reuterbal:
This updates the Loki CMake scripts and enhances them with the support for the upcoming `populate` step in ecbundle.
The underlying idea is that the download of Python wheels can be done before the actual build, thus enabling the use on air-gapped systems without internet access.

The method is described in `INSTALL.md`. The CMake tests in pytest have been amended to validate the functionality.

Along the way I have applied some TLC to installation description and CMake scripts, removing leftovers from the CLAW and similar.

Cc @wdeconinck for info.


reuterbal:
I have pushed a cloudsc-bundle branch that adds support for the populate step: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/tree/nabr-populate

When using the correct Python version for the populate step already, everything works as expected:
```
./cloudsc-bundle create
module load python/3.11
./cloudsc-bundle populate
./cloudsc-bundle build --arch arch/ecmwf/hpc2020/intel/2021.4.0 --with-loki=ON
```

But I seem unable to get it to work to set the python version in the populate step, because somehow the relevant option gets wrapped into single and double quotes and thus is no longer recognizable as an integer to pip:
```
--python-version error: invalid --python-version value: '"311"': each
version part must be an integer
```
Any ideas @awnawab ?

awnawab:
I couldn't manage to consistently preserve the type of the variable as an integer all the through the entire chain of bundle-> populate script -> cmake script to download wheels. The simplest (yet inelegant) workaround I found was to simply strip the double quotes from the argument of the cmake macro:
```
string(REPLACE "\"" "" _PYTHON_VERSION ${_PAR_WHEEL_PYTHON_VERSION})
list( APPEND PIP_OPTIONS --python-version=${_PYTHON_VERSION} )
```

reuterbal:
Thanks for this, very good point. I've added a corresponding fix, which seems to get PIP to accept the options. I'm not 100% sure it always does the right thing with regards to honouring the platform spec but that could be a separate problem.

awnawab (approved):
Thanks for this and for highlighting the problem with the pip options, I had missed this previously. Now looks G2G 👌

mlange05 (approved):
Looks great! And many thanks for the clean-up! 🙏 :shipit:

PR #524, opened 2025-04-08
IR: Source status, invalidation and Subroutine testing

mlange05:
This is a preparatory set of changes that is intended to pave the way for a string-conservative Fortran backend that renders changes in a minimally invasive way. For this purpose it introduces an explicit concept of "status" for the `Source` object that IR nodes can be equipped with and extends the invalidation of the source object to track the status of a `Source` object as being in of these three states:
* `SourceStatus.VALID` - A node has not been touched since the `Source` object was created by the frontend.
* `SourceStatus.INVALID_NODE` - The node has been directly altered or changed in in such a way that internal properties are directly contradicting the recorded source string.
* `SourceStatus.INVALID_CHILDREN` - The interior properties (eg. condition expression in conditionals) have not been modified, but some child nodes have been altered, thus rendering the recorded string as partially incorrect.

In addition to the new source status and invalidation in `Transformer` objects, we also now record source objects on `docstring`, `spec` and `body` properties on `Subroutine` objects. I've also made the transformer test non-recursive and test source invalidation explicitly in there now.

Importantly, I've also piggy-bagged a clean-up of some of the old `Subroutine` tests, which needlessly use JIT compilation (slow!) and often have redundancies between them. This makes things a little neater and speeds up the test base a little.

### DIFF
--- a/loki/ir/expr_visitors.py
+++ b/loki/ir/expr_visitors.py
@@ -256,7 +256,7 @@ class ExpressionTransformer(Transformer):
         else:
             new = self.expr_mapper(o)
         # Invalidate `Source` object if we've changed the expression
-        if kwargs.get('source') and not o == new:
+        if kwargs.get('source') and o != new:
             kwargs['source'].status = SourceStatus.INVALID_NODE
         return new
 
--- a/loki/ir/transformer.py
+++ b/loki/ir/transformer.py
@@ -99,8 +99,8 @@ class Transformer(Visitor):
         args_frozen.update(args)
         if self.invalidate_source and 'source' in args_frozen:
             # If any child node has been invalidated, mark this node as invalid too
-            if any(isinstance(c, Node) and not is_source_valid(c) for c in flatten(children)):
-                if is_source_valid(args_frozen.get('source')):
+            if is_source_valid(args_frozen.get('source')):
+                if any(isinstance(c, Node) and not is_source_valid(c) for c in flatten(children)):
                     args_frozen['source'] = args_frozen['source'].clone(
                         status=SourceStatus.INVALID_CHILDREN
                     )


loki/frontend/source.py
```diff
@@ -35,13 +59,16 @@ class Source:
...
+        Flag indicating if the associated node has been altered
     """
 
-    def __init__(self, lines, string=None, file=None):
+    def __init__(self, lines, string=None, file=None, status=None):
         assert lines and len(lines) == 2 and (lines[1] is None or lines[1] >= lines[0])
         self.lines = lines
         self.string = string
         self.file = file
+        self.status = status if status else SourceStatus.VALID
```
reuterbal, line 71:
```suggestion
        self.status = status if status is not None else SourceStatus.VALID
```

You might end up having `auto()` assign `0`, which would be falsey and reordering the enum could then change behaviour.

loki/frontend/source.py
```diff
@@ -127,6 +156,12 @@ def clone_lines(self, span=None):
             for idx, line in enumerate(self.string.splitlines())
         ]
 
+    def invalidate(self):
+        """
+        Return an invalid copy of this source with ``status=SourceStatus=INVALID_NODE``.
+        """
+        return self.clone(status=SourceStatus.INVALID_NODE)
+
```
reuterbal, line 164: This seems only to be used from tests, not in the actual transformer. Is that intentional?

loki/ir/expr_visitors.py
```diff
@@ -237,13 +238,27 @@ class ExpressionTransformer(Transformer):
...
         Call the associated mapper for the given expression node
         """
         if kwargs.get('recurse_to_declaration_attributes'):
-            return self.expr_mapper(o, recurse_to_declaration_attributes=True)
-        return self.expr_mapper(o)
+            new = self.expr_mapper(o, recurse_to_declaration_attributes=True)
+        else:
+            new = self.expr_mapper(o)
+        # Invalidate `Source` object if we've changed the expression
+        if kwargs.get('source') and not o == new:
```
reuterbal, line 259:
```suggestion
        if kwargs.get('source') and o != new:
```
?

loki/ir/transformer.py
```diff
@@ -107,10 +98,12 @@ def _rebuild(self, o, children, **args):
         args_frozen = o.args_frozen
         args_frozen.update(args)
         if self.invalidate_source and 'source' in args_frozen:
-            if any(isinstance(child, Node) for child in flatten(children)):
-                child_has_no_source = [getattr(i, 'source', None) is None for i in flatten(children)]
-                if any(child_has_no_source) or len(child_has_no_source) != len(flatten(o.children)):
-                    return self._rebuild_without_source(o, children, **args_frozen)
+            # If any child node has been invalidated, mark this node as invalid too
+            if any(isinstance(c, Node) and not is_source_valid(c) for c in flatten(children)):
+                if is_source_valid(args_frozen.get('source')):
```
reuterbal, line 103: Might want to reverse the nesting of the conditionals here to save the loop through children

reuterbal (changes requested):
Many thanks, this looks great! I've left a few minor comments

loki/frontend/fparser.py
```diff
@@ -1799,7 +1799,7 @@ def visit_Subroutine_Subprogram(self, o, **kwargs):
             decls = FindNodes((ir.VariableDeclaration, ir.ProcedureDeclaration)).visit(spec)
             if not decls:
                 # No other declarations: add it to the end
-                spec.append(return_var_decl)
+                spec.prepend(return_var_decl)
```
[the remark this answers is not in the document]
mlange05: Argh, yes, you're right! The whole thing is not safe, as we always add a `Declaration` node to the spec for prefix-defined return types. This, of course, cannot be string neutral, as it upsets the line-count of the stored source, so we first have to stop adding "ghost declarations'. 😬

loki/frontend/source.py
mlange05: Urgh, good point! Should we not explicitly avoid `0` then and simply prescribe enum values explicitly?

reuterbal: Meh, I should just have read the documentation:
> [auto](https://docs.python.org/3/library/enum.html#enum.auto)
>    Instances are replaced with an appropriate value for Enum members. [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) defaults to the lower-cased version of the member name, while other Enums default to 1 and increase from there.

loki/frontend/fparser.py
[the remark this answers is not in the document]
mlange05: This was addressed in PR #552. I've now rebased this.

loki/frontend/source.py
mlange05: Sort of. It's a neat utility that can freely be used by user-level transformations, when for example a changed clone of a node is used in a dict-based `Transformer` replacement (as in the tests). However, the only core-level automatic invalidation that happens in this PR is based by in-place changes to the status, so it does not use the utility.

mlange05:
Ok, @reuterbal , sorry for letting this hang for so long. Following the enabling PR-merge, I've now rebased this and addressed the other comments. Please have another look if you have a sec (assuming tests pass 😉 ).

reuterbal (approved):
Many thanks, this looks great and #552 is a welcome addition sparked by this :)

PR #526, opened 2025-04-08
Utility to parametrise those scalars which are only written to once

MichaelSt98:
and assigned a constant value.

**Not really sophisticated, however, this will eventually be superseded by some kind of constant propagation anyway**

E.g.

```fortran
  integer, intent(in) :: invar
  integer, intent(out) :: ret
  integer, parameter :: param = 1
  integer :: a, b, c, d
  real :: x, y, z
  logical :: l1, l2, l3

  ret = 0
  x = 3.5
  y = invar * 2.0
  z = 4.5 + 3.5
  a = ret
  b = 10
  c = invar*invar
  l1 = .false.
  l3 = .true.
  if (c .gt. 0) then
    l2 = .true.
  else
    l2 = .false.
  endif
```

to 

```fortran
  INTEGER, INTENT(IN) :: invar
  INTEGER, INTENT(OUT) :: ret
  INTEGER, PARAMETER :: param = 1
  INTEGER :: a, c, d
  INTEGER, PARAMETER :: b = 10
  REAL :: y
  REAL, PARAMETER :: x = 3.5
  REAL, PARAMETER :: z = 4.5 + 3.5
  LOGICAL :: l2
  LOGICAL, PARAMETER :: l1 = .false.
  LOGICAL, PARAMETER :: l3 = .true.

  ret = 0
  y = invar*2.0
  a = ret
  c = invar*invar
  IF (c > 0) THEN
    l2 = .true.
  ELSE
    l2 = .false.
  END IF
```

### DIFF
--- a/loki/transformations/parametrise.py
+++ b/loki/transformations/parametrise.py
@@ -84,13 +84,70 @@ using the transformation
 
 from loki.batch import Transformation
 from loki.expression import symbols as sym
-from loki.ir import nodes as ir, Transformer, FindNodes
+from loki.ir import nodes as ir, Transformer, FindNodes, FindInlineCalls
 from loki.tools.util import as_tuple, CaseInsensitiveDict
 
+from loki.transformations.utilities import single_variable_declaration
 from loki.transformations.inline import inline_constant_parameters
 
 
-__all__ = ['ParametriseTransformation']
+__all__ = ['ParametriseTransformation', 'declare_fixed_value_scalars_as_constants']
+
+
+def declare_fixed_value_scalars_as_constants(routine):
+    """
+    Mark local scalars that are assigned a fixed value as parameters.
+
+    This is not really sophisticated and will eventually be superseded by
+    a constant propagation transformation.
+    """
+    def is_constant_rhs(expr):
+        # expr is a literal e.g., a IntLiteral
+        if issubclass(type(expr), sym._Literal):
+            return True
+        # expr is a Product/Sum and all children are literals
+        if isinstance(expr, (sym.Product, sym.Sum)):
+            return all(issubclass(type(_expr), sym._Literal) for _expr in expr.children)
+        return False
+
+    assignments = FindNodes(ir.Assignment).visit(routine.body)
+    # filter for local variables and scalars
+    variables = [var for var in routine.variables if var not in routine.arguments and not isinstance(var, sym.Array)]
+    # don't bother with those being used in (inline) calls (although intent 'in' would be fine)
+    calls = as_tuple(FindNodes(ir.CallStatement).visit(routine.body)) + as_tuple(FindInlineCalls().visit(routine.body))
+    args = set()
+    for call in calls:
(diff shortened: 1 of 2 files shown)


loki/transformations/parametrise.py
```diff
@@ -87,10 +87,60 @@
...
+            return all(issubclass(type(_expr), sym._Literal) for _expr in expr.children)
+        return False
+
+    assignments = FindNodes(ir.Assignment).visit(routine.body)
+    # local variables
+    variables = [var for var in routine.variables if var not in routine.arguments]
+    assignments_dic = {}
+    for assignment in assignments:
+        if assignment.lhs in variables:
+            assignments_dic.setdefault(assignment.lhs, []).append(assignment)
```
reuterbal, line 119: I'm not sure but you might be running into case-sensitivity issues here. Could be better to build `assignments_dic` as a `CaseInsensitiveDict` with `assignment.lhs.name` as keys

loki/transformations/parametrise.py
```diff
@@ -87,10 +87,60 @@
...
 
+from loki.transformations.utilities import single_variable_declaration
 from loki.transformations.inline import inline_constant_parameters
 
 
-__all__ = ['ParametriseTransformation']
+__all__ = ['ParametriseTransformation', 'parametrise_single_val_scalars']
+
+
+def parametrise_single_val_scalars(routine):
```
reuterbal, line 97: The namings not ideal (as discussed offline). In particular, `parametrise` in the transformation refers to replacing values by constant parameters, whereas here we essentially mark local variables as compile-time constants. How about `declare_fixed_value_scalars_as_constants`?

loki/transformations/tests/test_parametrise.py
```diff
@@ -473,3 +473,57 @@ def test_parametrise_non_driver_entry_points(tmp_path, testdir, frontend, config
...
+  z = 4.5 + 3.5
+  a = ret
+  b = 10
+  c = invar*invar
+  l1 = .false.
+  l3 = .true.
+  if (c .gt. 0) then
+    l2 = .true.
+  else
+    l2 = .false.
```
reuterbal, line 502: Try upper-casing the lhs here to make sure my comment above is redundant ;-)

loki/transformations/parametrise.py
```diff
@@ -87,10 +87,60 @@
...
+    """
+    Parametrise those scalars which are only written to once and assigned a constant value.
```
reuterbal, line 99:
```suggestion
    Mark local scalars that are assigned a fixed value as parameters
```

reuterbal (changes requested):
A few minor comments to make sure this handles a few edge cases better, otherwise I think this is nice and simple!

reuterbal (approved):
Many thanks for the expanded test case and addressing the comments.

mlange05 (approved):
Looks great! GTG from me. :shipit:

PR #527, opened 2025-04-08
Fix indentation in fgen after Associate blocks and add TypeConditional node

reuterbal:
This is a bugfix PR for a line continuation problem I was facing. In the end it turned out to be the 1-character fix in 0ff12f700de0f5a08a85bf9d00cf4dc45c10387c that was missing, but to get to that realization I implemented a TypeConditional node and a safeguarding against excessive indentation in JoinableStringList. So, we have those now, too.

### DIFF
--- a/loki/backend/fgen.py
+++ b/loki/backend/fgen.py
@@ -810,7 +810,7 @@ class FortranCodegen(Stringifier):
         footer = self.format_line('END ASSOCIATE')
         self.depth += self.style.associate_indent
         body = self.visit(o.body, **kwargs)
-        self.depth += self.style.associate_indent
+        self.depth -= self.style.associate_indent
         return self.join_lines(header, body, footer)
 
     def visit_CallStatement(self, o, **kwargs):
--- a/loki/backend/fgen.py
+++ b/loki/backend/fgen.py
@@ -682,6 +682,35 @@ class FortranCodegen(Stringifier):
         branches = [item for branch in zip(cases, bodies) for item in branch]
         return self.join_lines(header, *branches, footer)
 
+    def visit_TypeConditional(self, o, **kwargs):
+        """
+        Format as
+          [name:] SELECT TYPE (<expr>)
+          [CLASS IS (<value>) [name]]
+            [...body...]
+          [TYPE IS (<value>) [name]]
+            [...body...]
+          [CLASS DEFAULT [name]]
+            [...body...]
+          END SELECT [name]
+        """
+        header_name = f'{o.name}: ' if o.name else ''
+        header = self.format_line(header_name, 'SELECT TYPE (', self.visit(o.expr, **kwargs), ')')
+        cases = []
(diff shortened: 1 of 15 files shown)


loki/backend/fgen.py
```diff
@@ -781,7 +810,7 @@ def visit_Associate(self, o, **kwargs):
         footer = self.format_line('END ASSOCIATE')
         self.depth += self.style.associate_indent
         body = self.visit(o.body, **kwargs)
-        self.depth += self.style.associate_indent
+        self.depth -= self.style.associate_indent
```
mlange05, line 813: Ooopsie! 🙀

loki/frontend/fparser.py
```diff
@@ -2354,6 +2354,138 @@ def visit_Case_Value_Range(self, o, **kwargs):
...
+        elif o.children[0] == 'TYPE IS':
+            is_polymorphic = False
+        elif o.children[0] == 'CLASS DEFAULT':
+            is_polymorphic = None
+        else:
+            raise ValueError(f'Unsupported first child of Type_Guard_Stmt: {o.children[0]}')
+
+        return self.visit(o.children[1], **kwargs), is_polymorphic
+
+        # # The banter before the construct...
```
mlange05, line 2471: Intentionally left in, or dev leftover?

loki/tools/tests/test_tools.py
```diff
@@ -174,6 +174,16 @@ def test_joinable_string_list_long():
     assert str(obj) == ref
 
 
+def test_joinable_string_list_exceeding_indentation():
+    items = ['a', 'b', 'c']
+    for depth in range(100):
+        # This should not throw an assertion
+        obj = JoinableStringList(
+            items, sep=' ', width=132,
+            cont=f' &\n{"  " * depth}&')
```
mlange05, line 183: Not sure I understand this - aren't you just setting a large continuation indentation, without ever needing it because your `sep` is still just `' '`? For 3 single-char elements that should generate (`a b c`) and not need the continuation at all - or am I Misreading this (quite likely)?

mlange05 (approved):
Looks great and is very neatly implemented. I've marked a potential debug leftover and a potential test inconsistency (which might well be me misunderstanding things), but these are minor, so GTG once explained.

And apologies fro the mistake (and apparent lack of test coverage) in the first place. I still like having this made explicit, so thanks!

loki/frontend/fparser.py
reuterbal: Ah, no, leftover - thanks!

loki/tools/tests/test_tools.py
reuterbal: Oh, you're absolutely right. The test did trigger the original assertion, which simply looked at the length of the continuation string, but it did not assess the underlying formatting problem due to the lack of causing line length overflow. Top commit includes a fix.

mlange05:
Looking good, going in.

PR #528, opened 2025-04-09
Fix symlink to loki scripts after CMake installation when a relative prefix path was used

reuterbal:
The changes by the recently merged #523 cause a problem if Loki is installed with a relative prefix path (i.e., `cmake --install <builddir> --prefix <relative/path>`), leaving the symlinks from `<prefix>/bin` into the installed venv dangling.
This fixes the issue and adds testing for this use case.


mlange05 (approved):
Looks sane to me. GTG :shipit:

PR #529, opened 2025-04-09
new utility: 'substitute_variables_for_definitions'

MichaelSt98:
necessary to enable SCC-HOIST for ecphys

### DIFF
--- a/loki/transformations/single_column/hoist.py
+++ b/loki/transformations/single_column/hoist.py
@@ -9,7 +9,7 @@ from loki.expression import symbols as sym
 from loki.ir import nodes as ir
 
 from loki.transformations.hoist_variables import HoistVariablesTransformation
-from loki.transformations.utilities import get_integer_variable
+from loki.transformations.utilities import get_integer_variable, substitute_variables_for_definitions
 
 
 __all__ = ['SCCHoistTemporaryArraysTransformation']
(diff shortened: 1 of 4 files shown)


loki/transformations/utilities.py
```diff
@@ -680,3 +680,25 @@ def check_routine_sequential(routine):
             return True
 
     return False
+
+def remap_variables(routine, variables):
+    """
+    'Remap' variables if variable was initialized with another var/value.
```
reuterbal, line 686: Can we rename this to something like `substitute_variable_definitions` or so?

loki/transformations/single_column/hoist.py
```diff
@@ -56,6 +56,7 @@ def driver_variable_declaration(self, routine, variables):
             )
 
         block_var = get_integer_variable(routine, self.block_dim.size)
+        block_var = remap_variables(routine, variables=block_var)[0]
```
reuterbal, line 59: Would it be easy to capture the effect of this in one of the tests for hoist? This seems important enough to warrant capturing this pattern in the test base.

reuterbal (changes requested):
Nice, clean and simple. I had no idea what `remap_variables` should mean, so maybe we can find a somewhat more suggestive naming?

Also, given that this seems crucial for ecphys to have hoisting work, it would be good if the hoist tests capture the effect somehow, to make sure we don't break this.

PR #530, opened 2025-04-10
Hotfix: Set SETUPTOOLS_SCM_PRETEND_VERSION during installation if required

reuterbal:
Otherwise, `cmake --install` will not work if used from a tarball instead of a git worktree. This breaks Github Actions for packages that use Loki at the moment.


mlange05 (approved):
Looks good to me. :shipit:

PR #531, opened 2025-04-10
Backend: New generalised "string-conservative" FGen backend

mlange05:
~_Integration note: This is based on top of PR #524 and needs to be reviewed/merged after._~

This PR brings in a rudimentary rethink of the "conservative" `fgen` option that is based on a dedicated backend visitor that aims to re-create as much of the original source string as possible. The purpose of this is to then create string output that only captures the changes to the IR/expression tree in comparison to the originally parsed source, which would allow us to generate reasonably a diff/patch when comparing to the original file. **Please note that this is not yet far enough for diff-generation(!) in all cases**, but provides a first set of changes that can recapture a significant amount the original source (eg. non-conformant indentation).

In particular, up to now, we hard-invalidated the source as soon as a node, or any of its children was touched. Following PR #524 we now retain the source object and track its status, allowing us to even recover partial matches when a node or its children have changed. For example, for an `Assignment` or `VariableDeclaration` we let `fgen` create the respective string to the left and right of the delimiter (`=` and `::` respectively), and compare the expression to the original source - allowing us to prefer if we detect no change in, eg. the LHS expression. Similarly, if only the children of a `Loop` or `Conditional` have changed, we use the original source to re-capture the loop or condition or conditional expression. And finally we attempt to re-capture the header/footer of subroutines if possible to capture non-standard indentation and whitespaces.

Final remark: I'm sure there's a lot more that can be done to break this, but I had to start somewhere. 😉 

Edit: Post-rebase I also slightly changed the behaviour of the source invalidation API, so that we can now tell each node to `node.source.invalidate(children=True/False)`, and then query it with `node.source.is_valid()`. This avoids having to import the enum type that we use to track the status.

### DIFF
--- a/loki/backend/fgen.py
+++ b/loki/backend/fgen.py
@@ -190,29 +190,42 @@ class FortranCodegen(Stringifier):
 
         return self.join_lines(header, docstring, spec, contains, footer)
 
-    def visit_Subroutine(self, o, **kwargs):
-        """
-        Format as
-          [<prefix>] SUBROUTINE <name> ([<args>]) [BIND(c, name=<name>)]
-            ...docstring...
-            ...spec...
-            ...body...
-          [CONTAINS]
-            [...member...]
-          END SUBROUTINE <name>
-        """
+    def _construct_procedure_footer(self, o, **kwargs):
+        ftype = 'FUNCTION' if o.is_function else 'SUBROUTINE'
+        return self.format_line('END ', ftype, ' ', o.name if self.style.procedure_end_named else '')
+
+    def _construct_function_header(self, o, **kwargs):
         prefix = self.join_items(o.prefix, sep=' ')
         if o.prefix:
             prefix += ' '
+        if not o.result_name in o.variable_map:
+            prefix += f'{self.visit(o.return_type)} '
         arguments = self.join_items(o.argnames)
+        result = f' RESULT({o.result_name})' if o.result_name\
+                and o.result_name.lower() != o.name.lower() else ''
         if isinstance(o.bind, str):
             bind_c = f' BIND(c, name="{o.bind}")'
         elif isinstance(o.bind, StringLiteral):
             bind_c = f' BIND(c, name={o.bind})'
         else:
             bind_c = ''
-        header = self.format_line(prefix, 'SUBROUTINE ', o.name, ' (', arguments, ')', bind_c)
-        footer = self.format_line('END SUBROUTINE ', o.name if self.style.procedure_end_named else '')
+
+        return self.format_line(prefix, 'FUNCTION ', o.name, ' (', arguments, ')', result, bind_c)
+
+    def visit_Function(self, o, **kwargs):
+        """
+        Format as
+          [<prefix>] FUNCTION <name> ([<args>]) [RESULT(<name>)] [BIND(c, name=<name>)]
+            ...docstring...
+            ...spec...
+            ...body...
+          [CONTAINS]
+            [...member...]
+          END FUNCTION <name>
+        """
+
+        header = self._construct_function_header(o, **kwargs)
+        footer = self._construct_procedure_footer(o, **kwargs)
 
         self.depth += self.style.procedure_spec_indent
         docstring = self.visit(o.docstring, **kwargs)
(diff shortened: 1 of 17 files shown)


MichaelSt98 (commented):
Looks awesome!
I agree, there are probably many things to do/improve, but looks like it is already quite powerful and useful.

reuterbal (commented):
Really pleased how straightforward and clean this looks now! Nothing that needs major changes.

The test coverage is still patchy. One thing to potentially include in the testing is a module and possibly some mixed upper/lower case. That would probably also confirm @MichaelSt98's suspicion about the case-sensitivity of attributes.

loki/backend/fgen.py
```diff
@@ -1038,10 +1053,14 @@ def fgen(ir, style=None, depth=0, conservative=False):
     """
     Generate standardized Fortran code from one or many IR objects/trees.
     """
+    from loki.backend.fgencon import FortranCodegenConservative  # pylint: disable=import-outside-toplevel,cyclic-import
```
MichaelSt98, line 1056: If having this import as an outside top-level import, wouldn't it make more sense to put it within `if conservative`?

loki/backend/fgencon.py
```diff
@@ -0,0 +1,207 @@
...
+class FortranCodegenConservative(FortranCodegen):
+    """
+    Strictly conservative version of :any:`FortranCodegen` visitor
+    that will attempt to use existing :any:`Source` information from
+    the frontends where possible.
+    """
+
+    def visit_Node(self, o, *args, **kwargs):
+        if o.source and o.source.status == SourceStatus.VALID:
+            return o.source.string
```
MichaelSt98, line 24:
```py
if o.source and o.source.status == SourceStatus.VALID:
    return o.source.string
```

is a common pattern in `fgencode`.
Wouldn't it make sense to wrap that in e.g.,

```py
def return_valid_source(self, o):
     if o.source and o.source.status == SourceStatus.VALID:
        return o.source.string
```

MichaelSt98 (approved):
Really nice and already quite powerful!

loki/backend/fgencon.py
mlange05: Unfortunately no, since the early return is what prevents the rest of the method to be executed. If you wrap this in a utility you lose the early return and have to add another conditional around the utility, which defeats the point for a two-line snippet.

loki/backend/fgen.py
```diff
@@ -241,33 +244,34 @@ def visit_Subroutine(self, o, **kwargs):
...
-          [<prefix>] FUNCTION <name> ([<args>]) [RESULT(<name>)] [BIND(c, name=<name>)]
-            ...docstring...
-            ...spec...
-            ...body...
-          [CONTAINS]
-            [...member...]
-          END FUNCTION <name>
-        """
+    def _construct_subroutine_header(self, o, **kwargs):
+        ftype = 'FUNCTION' if o.is_function else 'SUBROUTINE'
```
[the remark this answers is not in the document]
mlange05: This is no longer the case (the comment was raised before the rebase over the subroutine/function separation).

mlange05:
Ok, apologies for the wait. I've now rebased this and added two more tests: one explicitly aimed at partial re-generation of modules and one that aims at dispatching to the full fgen backend after IR-level changes (to achieve better coverage). It's still quite hard to fully get the new transformer covered though, but hopefully more usage will fix that. 

As part of the new module tests, I've found and fixed a few more issues (including storing the source for modules in the frontend). I've also tried to use native string comparison of select expressions (assign.lhs/rhs and variable symbols in declarations) to better capture string-level differences between stored source and internal representation (this was triggere by @MichaelSt98 comment about case-sensitive comparisons). 

I've also rebase over latest main, and I think this is now ready for another look. @MichaelSt98 @reuterbal

loki/backend/fgen.py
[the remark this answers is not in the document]
reuterbal: Hmm, I think it is still the case? l. 273 calls `_construct_subroutine_header` for subroutines, and l. 227 calls `_construct_function_header` for functions - hence there's no need to have an ftype distinction inside the method anymore. The function_header method directly uses `FUNCTION` and I think the subroutine_header method could use `SUBROUTINE` directly, too.

Only the footer is constructed using a joint `procedure_footer` method.

reuterbal (approved):
Many thanks and the additional test coverage is much appreciated (and has clearly been useful - so I feel only half bad that I asked for it ;-) ).

Only one nitpick follow-up on the earlier `ftype` comment, which I think is still applicable. Otherwise this looks arrrr-rite 🏴‍☠️

PR #533, opened 2025-04-12
SingleColumn: Use in-place `Transformer` in devector and revector steps

mlange05:
A small change in the SCC components that replaces a "find-and-transform" traversal pattern to a single traversal using custom `Transformer` objects for:
* The removal of horizontal loops in `SCCDevectorTransformation`, where we replace `SCCDevectorTransformation.kernel_remove_vector_loops` with a standalone `RemoveLoopTransformer`
* The re-insertion of vector loops in `SCCRevectorTransformation` (both vec and seq). Here we replace `wrap_vector_section_from_dimension` and `BaseRevectorTransformation.revector_section` with `RevectorSectionTransformer`.

The primary intention here is to make the code more compact and improve processing time. The speedups, using ec-physics and surface components regression on local fast SSDs, are non-negligible:

**Surface:**
```
[Loki::Scheduler] Applied transformation <SCCDevectorTransformation> in 12.93s
[Loki::Scheduler] Applied transformation <SCCVecRevectorTransformation> in 5.12s
   => 
[Loki::Scheduler] Applied transformation <SCCDevectorTransformation> in 8.79s
[Loki::Scheduler] Applied transformation <SCCVecRevectorTransformation> in 1.61s
```

**Physics:**
```
[Loki::Scheduler] Applied transformation <SCCDevectorTransformation> in 54.35s
[Loki::Scheduler] Applied transformation <SCCVecRevectorTransformation> in 25.64s
 => 
[Loki::Scheduler] Applied transformation <SCCDevectorTransformation> in 39.45s
[Loki::Scheduler] Applied transformation <SCCVecRevectorTransformation> in 8.98s
```

One shortcoming that this highlighted was that we did not have the correct auto-fixer constructors for `MultiConditional` nodes, which I now implemented. These are needed to catch nested tuples that get created when replacing a loop by its body (a tuple), within an existing tuple (eg. `Section`).

And one final housekeeping step I then added was to move the `SCCDemoteTransformation` into a separate sub-package, and separate devector and revector utilities into sub-packages. Apologies if this muddles the diff a little, but looking at individual commits should show the respective changes in isolation.


loki/ir/tests/test_ir_nodes.py
```diff
@@ -350,3 +350,68 @@ def test_associate(scope, a_i):
     assoc.rescope_symbols()
     assert assign.lhs.scope is assoc
     assert assign2.lhs.scope is scope
+
+
+def test_multiconditional(scope, a_i, i):
+    """
+    Test constructors and scoping bahviour of :any:`MultiConditional`.
```
MichaelSt98, line 357:
```suggestion
    Test constructors and scoping behaviour of :any:`MultiConditional`.
```

loki/transformations/single_column/demote.py
```diff
@@ -0,0 +1,152 @@
...
+        role = kwargs['role']
+        item = kwargs.get('item', None)
+
+        if role == 'kernel':
+            demote_locals = self.demote_local_arrays
+            preserve_arrays = []
+            if item:
+                demote_locals = item.config.get('demote_locals', self.demote_local_arrays)
+                preserve_arrays = item.config.get('preserve_arrays', [])
+            self.process_kernel(routine, demote_locals=demote_locals, preserve_arrays=preserve_arrays)
```
MichaelSt98, line 118: Shouldn't we early exit whenever `demote_locals = False`?

MichaelSt98 (approved):
Nice! Looks good to me.

awnawab (approved):
Many thanks, looks much neater and the performance improvements are very welcome! G2G 👌

loki/transformations/single_column/demote.py
alissarfenric: In theory, yes, but this is just a plain copy-paste to move things around, so I'd like to address this in a separate PR for testing. (FWIW, I think there's a bit more that can be cleaned up here).

mlange05: Yes, in theory we should, but this was a straight copy-paste to move things around. So, I'd prefer to address this in a separate PR for testing (there's more that can be done here, I think).

loki/ir/nodes.py
```diff
@@ -1785,6 +1785,16 @@ class MultiConditional(LeafNode, _MultiConditionalBase):
 
     _traversable = ['expr', 'values', 'bodies', 'else_body']
 
+    @field_validator('else_body', mode='before')
```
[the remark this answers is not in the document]
mlange05: Yes, correct. Importantly, it will also flatten nested tuples and convert `None` to `()`. 

The nested tuples is something we can stumble over if we're using in-place transformers, as it will convert `(Comment, Loop)` into `(Comment, (Assignment))` when it replaces the loop with its body. With dict-based transformers, the `visit_tuple` method will rectify the tuple nesting before invoking the constructor, but that's the method we're explicitly trying to avoid here, as the lookups are costly.

loki/transformations/single_column/devector.py
```diff
@@ -0,0 +1,248 @@
...
+    # pylint: disable=unused-argument
+
+    def __init__(self, dimension, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.dimension = dimension
+
+    def visit_Loop(self, loop, **kwargs):
+        if loop.variable == self.dimension.index:
+            # Recurse and return body as replacement
+            return self.visit(loop.body)
```
reuterbal, line 48:
```suggestion
            return self.visit(loop.body, **kwargs)
```
Otherwise any derived implementations that may want to utilitze additional kwargs won't be able to pass them through here.

loki/transformations/single_column/devector.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Rebuild loop after recursing to children
+        return self._rebuild(loop, self.visit(loop.children))
```
reuterbal, line 51:
```suggestion
        return self._rebuild(loop, self.visit(loop.children, **kwargs))
```

reuterbal (approved):
Many thanks, this looks great. Just two minor issues which I will directly commit before merging

loki/ir/nodes.py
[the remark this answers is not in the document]
reuterbal: See #537

PR #539, opened 2025-04-28
Remove unused arguments in a call-tree

awnawab:
This PR adds the ability to optionally remove unused arguments from a call-tree. A single reverse pass is used, which means the `Subroutine` signature is updated before the `CallStatement`. That's why `call.arg_map` isn't used but rather the order of the arguments is relied upon to map unused dummy arguments to arguments.

### DIFF
--- a/loki/transformations/remove_code.py
+++ b/loki/transformations/remove_code.py
@@ -11,10 +11,11 @@ section and to perform Dead Code Elimination.
 """
 
 from loki.batch import Transformation
-from loki.expression.symbolic import simplify
+from loki.expression import simplify, symbols as sym
 from loki.tools import flatten, as_tuple
-from loki.ir import Conditional, Transformer, Comment
+from loki.ir import Conditional, Transformer, Comment, CallStatement, FindNodes, FindVariables
 from loki.ir.pragma_utils import is_loki_pragma, pragma_regions_attached
+from loki.analyse import dataflow_analysis_attached
 
 
 __all__ = [
@@ -65,14 +66,18 @@ class RemoveCodeTransformation(Transformation):
         "kernel"; default: ``False``
     """
 
+    _key = 'RemoveCodeTransformation'
+
     # Recurse to subroutines in ``contains`` clause
     recurse_to_internal_procedures = True
+    reverse_traversal = True
 
     def __init__(
             self, remove_marked_regions=True, mark_with_comment=True,
             remove_dead_code=False, use_simplify=True,
             call_names=None, intrinsic_names=None,
-            remove_imports=True, kernel_only=False
+            remove_imports=True, kernel_only=False,
+            remove_unused_args=False
     ):
         self.remove_marked_regions = remove_marked_regions
         self.mark_with_comment = mark_with_comment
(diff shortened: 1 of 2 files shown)


loki/transformations/remove_code.py
```diff
@@ -85,30 +90,86 @@ def __init__(
...
+            unused_args = self._find_unused_args(routine)
+
+            # store unused args
+            item.trafo_data[self._key] = {'unused_args': unused_args}
+
+            # remove unused args
+            routine.variables = [a for a in routine.variables
+                                 if not a.name.lower() in unused_args]
+
+    @staticmethod
```
mlange05, line 137: I know we used this pattern a lot before, but it's actually not great tying this to a class for no reason. Could we instead simply make this a public utility method of this module? This makes re-use in other modules easier, I think.

And making it "public" (removing the `_`) then also warrants a docstring, I'd suggest. 😉

loki/transformations/remove_code.py
```diff
@@ -85,30 +90,86 @@ def __init__(
...
+            # Apply Dead Code Elimination
+            if self.remove_dead_code:
+                do_remove_dead_code(routine, use_simplify=self.use_simplify)
+
+        if self.remove_unused_args and (item := kwargs['item']):
+            if item.config.get('remove_unused_args', True):
+                successors = kwargs['sub_sgraph'].successors(item=item)
+                self.do_remove_unused_args(routine, item, successors, kwargs['role'])
+
+    def do_remove_unused_args(self, routine, item, successors, role):
```
mlange05, line 121: If you're following the `do_something` calling convention, maybe consider making `do_remove_unused_args` a freely reusable utility method and let the `Transformation` update its `trafo_data` with the returned `unused_args`? 

What I mean here is that we could actually rename the slightly non-descript `_update_callstatements` as `do_remove_unused_arguments(routine, ..., unused_args)` and let it do the callstatement and `routine.variable`. The `item.trafo_data` and `if role =="kernel"` logic would then just be a concise chunk in `transform_subroutine`, but outside of the scheduler, one could easily do
```
unused_args = find_unused_args(routine_a)
do_remove_unused_args(routine_b)
```
Is this possible, or am I missing something here?

loki/transformations/remove_code.py
```diff
@@ -85,30 +90,86 @@ def __init__(
...
+            used_or_defined_symbols |= set(FindVariables().visit(used_or_defined_array_shapes))
+
+            used_or_defined_symbols |= set(variable_map.get(v.name_parts[0], v) for v in used_or_defined_symbols)
+
+            unused_args = {a.clone(dimensions=None): c for c, a in enumerate(routine.arguments)
+                           if not a.name.lower() in used_or_defined_symbols}
+
+        return unused_args
+
+    def _update_callstatements(self, routine, successors):
```
mlange05, line 155: See above comments about name and missing docstring).

loki/transformations/tests/test_remove_code.py
```diff
@@ -439,3 +523,58 @@ def test_remove_code_transformation(frontend, source, include_intrinsics, kernel
...
+@pytest.mark.parametrize('kernel_override', [True, False])
+def test_remove_code_unused_args(frontend, source_with_args, kernel_override, tmp_path):
+    """
+    Test the removal of unused arguments in a call tree.
+    """
+
+    config = {
+        'default': {
+            'role': 'kernel', 'expand': True, 'strict': False,
+            'disable': ['dr_hook', 'abor1'], 'enable_imports': True,
```
mlange05, line 538: `disable` entry is not needed here, right?

mlange05 (changes requested):
This is a great general utility that is very useful for code sanitisation and a great addition to this utility transformation! 

However, I think it might be great if we could slightly refactor out the two core pieces - finding unused args in the callee routine and removing them from all calls in the caller - and make them available in a standalone fashion. This way we can then only use the "mode" and `trafo_data` logic in the `Transformation` to orchestrate this, while leaving the bare-bone utilities open for reuse elsewhere. Unless that is, I'm missing something that makes the mechanics closely tied t the trafo itself?

Otherwise this looks great and has neat test coverage! Many thanks

loki/transformations/tests/test_remove_code.py
awnawab: Yes you're right, not needed here. I'll remove it 👍

loki/transformations/remove_code.py
awnawab: Yes that's a good idea. There would be limited value in running this without a scheduler, since you would have to make sure to run this in reverse order and collect all the unused args from children before you update the `CallStatement`s, but supporting that possibility will make the code nicer anyhow 👍

loki/transformations/remove_code.py
```diff
@@ -11,10 +11,12 @@
...
 
 from loki.batch import Transformation
-from loki.expression.symbolic import simplify
+from loki.expression import simplify, symbols as sym
 from loki.tools import flatten, as_tuple
-from loki.ir import Conditional, Transformer, Comment
+from loki.ir import Conditional, Transformer, Comment, CallStatement, FindNodes, FindVariables
 from loki.ir.pragma_utils import is_loki_pragma, pragma_regions_attached
+from loki.analyse import dataflow_analysis_attached
+from loki.types import BasicType
```
reuterbal, line 19: We're trying to stick to alphabetical import orders:
```py
from loki.analyse import dataflow_analysis_attached
from loki.batch import Transformation
from loki.expression import simplify, symbols as sym
from loki.ir import Conditional, Transformer, Comment, CallStatement, FindNodes, FindVariables
from loki.ir.pragma_utils import is_loki_pragma, pragma_regions_attached
from loki.tools import flatten, as_tuple
from loki.types import BasicType
```

loki/transformations/tests/test_remove_code.py
```diff
@@ -110,6 +110,90 @@ def fixture_source(srcdir):
     (srcdir/'never_gonna_give.F90').unlink()
 
 
+@pytest.fixture(scope='module', name='source_with_args')
+def fixture_source_with_args(srcdir):
```
reuterbal, line 114: As usual, I would suggest littering the test's fcode with some random capitalization of variables and routine names in declarations/uses to ensure case-insensitivity is appropriately implemented. 
It looks fine in the implementation afaict, but that prevents problems from sneaking in in the future.

awnawab: Good call, I've added random capitals to the unused arguments to see if these are still picked up 👍

mlange05 (approved):
Yes, very happy now, and thanks for refactoring as suggested. 🙏  GTG from me :shipit:

PR #540, opened 2025-04-28
MODULE_WRAP: fix bug in `replace_ignore_items` option

awnawab:
A small PR to fix the `replace_ignore_items` option in the `ModuleWrapTransformation`.

### DIFF
--- a/loki/transformations/build_system/module_wrap.py
+++ b/loki/transformations/build_system/module_wrap.py
@@ -110,7 +110,7 @@ class ModuleWrapTransformation(Transformation):
         """
 
         targets = tuple(str(t).lower() for t in as_tuple(kwargs.get('targets')))
-        if self.replace_ignore_items and (item := kwargs.get('item')):
+        if (item := kwargs.get('item')) and self.replace_ignore_items:
             targets += tuple(str(i).lower() for i in item.ignore)
 
         def _update_item(proc_name, module_name):
--- a/loki/transformations/build_system/module_wrap.py
+++ b/loki/transformations/build_system/module_wrap.py
@@ -110,7 +110,8 @@ class ModuleWrapTransformation(Transformation):
         """
 
         targets = tuple(str(t).lower() for t in as_tuple(kwargs.get('targets')))
-        if (item := kwargs.get('item')) and self.replace_ignore_items:
+        item = kwargs.get('item')
+        if item and self.replace_ignore_items:
             targets += tuple(str(i).lower() for i in item.ignore)
 
         def _update_item(proc_name, module_name):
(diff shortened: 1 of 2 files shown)


loki/transformations/build_system/tests/test_dependency.py
```diff
@@ -415,40 +419,80 @@ def test_dependency_transformation_module_wrap(frontend, use_scheduler, tmp_path
...
             scheduler.process(transformation)
 
-        kernel = scheduler['kernel_test_mod#kernel_test'].source
+        if replace_ignore_items:
+            kernel = scheduler['kernel_test_mod#kernel_test'].source
+        else:
+            with pytest.raises(AttributeError):
+                kernel = scheduler['#kernel'].source
+                kernel = scheduler['kernel_mod#kernel'].source
+                kernel = scheduler['kernel_test_mod#kernel_test'].source
```
reuterbal, line 458: Don't you need to separate these to test that the exception is raised for each of them? I expect this to bail out on the first exception occurence
```suggestion
            for item_name in ['#kernel', 'kernel_mod#kernel', 'kernel_test_mod#kernel_test']:
                with pytest.raises(AttributeError):
                    _ = scheduler[item_name].source
```

loki/transformations/build_system/module_wrap.py
```diff
@@ -110,7 +110,7 @@ def update_imports(self, source, imports, **kwargs):
         """
 
         targets = tuple(str(t).lower() for t in as_tuple(kwargs.get('targets')))
-        if self.replace_ignore_items and (item := kwargs.get('item')):
+        if (item := kwargs.get('item')) and self.replace_ignore_items:
```
reuterbal, line 113: Is that change significant?
(Not opposed, just curious)

reuterbal (approved):
Thanks, not sure in what way this was broken before but the additional test coverage is very welcome!

loki/transformations/build_system/tests/test_dependency.py
awnawab: Yes good catch! I'll update the test.

loki/transformations/build_system/module_wrap.py
awnawab: If `self.replace_ignore_items` was false then the second part of the expression wasn't evaluated and we could end up with an undefined item.

reuterbal: Right - I was misled by the formatting in the PR which suggested this conditional was the last statement in this method...
Arguably, `item` should then be assigned outside the conditional here. The walrus shorthand doesn't make sense if the value is to be used regardless of the boolean value of the conditional expression:
```suggestion
        item = kwargs.get('item')
        if item and self.replace_ignore_items:
```

reuterbal (approved):

PR #541, opened 2025-04-30
Add `force_inject_arrays` option and also enrich the driver in `BlockViewToFieldView` transformation

awnawab:
A small PR that:
- Adds an option in the `InjectBlockIndexTransformation` to specify arrays that should be forcibly injected via `force_inject_array`. This is primarily aimed at sequence associated arguments that we can't fix in source
- The driver layer is also also enriched with the dummy `FIELD_RANKSUFF_ARRAY` type definitions in the `BlockViewToFieldView` transformation.

### DIFF
--- a/loki/transformations/block_index_transformations.py
+++ b/loki/transformations/block_index_transformations.py
@@ -353,11 +353,13 @@ class InjectBlockIndexTransformation(Transformation):
         targets = tuple(str(t).lower() for t in as_tuple(kwargs.get('targets', None)))
 
         exclude_arrays = []
+        force_inject_arrays = []
         if (item := kwargs.get('item', None)):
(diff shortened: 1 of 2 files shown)


loki/transformations/block_index_transformations.py
```diff
@@ -188,7 +193,8 @@ def process_driver(self, routine, successors):
         if field_array_mod_imports:
             definitions += self._create_dummy_field_api_defs(field_array_mod_imports)
 
-        # propagate dummy field_api wrapper definitions to children
+        # propagate dummy field_api wrapper definitions to self and children
+        routine.enrich(definitions)
```
reuterbal, line 197: Ooh, that's a neat trick!

reuterbal (approved):
Thanks, looks good to me and well-tested

PR #542, opened 2025-05-06
Introduce subdir/subpackage 'temporaries'

MichaelSt98:
for transformations that deal with temporaries ...
 
 Preparatory work for upcoming stack/pool transformations.


reuterbal (approved):
Many thanks - looks good to me!

PR #543, opened 2025-05-06
Subroutine: Clean up subroutine tests and enable OMNI

mlange05:
A small housekeeping PR that aims to tidy up `test_subroutine.py` one of the older parts of the code base. This paves the way for some TLC further down the road.

For this we do:
* Remove JIT compilation from some of the simpler tests and explicitly test internal properties of the generated object. Some JIT tests are left that often trigger when frontend-backend pairs disagree, but overall throughput is improved by this
* Merge several tests that deal with variables and argument handling
* Enabled OMNI for all tests by providing header info and fixing tests that are preempted by OMNI's parameter resolution.

And finally, the partial removal of JIT testing speeds-up the respective test quite a bit. 😉

### DIFF
--- a/loki/tests/test_subroutine.py
+++ b/loki/tests/test_subroutine.py
@@ -8,7 +8,6 @@
 # pylint: disable=too-many-lines
 from pathlib import Path
 import pytest
-import numpy as np
 
 from loki import Sourcefile, Module, Subroutine, fgen, fexprgen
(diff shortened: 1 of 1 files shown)


reuterbal (approved):
Thanks, looks good to me. Very much appreciate the reduction of JIT compilation in the tests!

Issue #195, opened 2023-11-27
Shape inference from allocations should be a utility/transformation


mlange05:
The `Scheduler` constructor call `_infer_allocatable_shapes` (https://github.com/ecmwf-ifs/loki/blob/main/loki/subroutine.py#L131) routinely for every object. This is a remnant of the early days of Loki development and should be moved into the shape-handling transformation utility.

PR #544, opened 2025-05-07
Subroutine: Remove automatic derivation of shape from ALLOCATE

mlange05:
~_Note: Sits on top of #543 out of sheer laziness, but should address #195._~

Another housekeeping PR that removes a remnant from the past. This little hack lived in the constructor of subroutines and is no longer needed, I think. I've run this with ecphys regression, and everything is fine, so I would like to remove it instead of re-implementing, as suggested in #195.


reuterbal (approved):
Thanks, this looks good to me and I'm not aware of any cases where we still depend on this. However, since this can potentially break someone else's transformations, we need to make sure to point this out explicitly in the next release notes.

mlange05:
Yes, absolutely agree about sign-posting this in the release notes. 👍

PR #546, opened 2025-05-07
make the regex frontend capable of parsing pragmas

MichaelSt98:
This adds a `PragmaPattern` class for the regex frontend.


loki/frontend/source.py
```diff
@@ -173,8 +173,12 @@ def _sanitize_raw_source(self, raw_source):
         """
         if FortranStringReader is None:
             raise RuntimeError('FortranReader needs fparser2')
-        reader = FortranStringReader(raw_source)
+        # do not ignore comments as this would also ignore pragmas
+        reader = FortranStringReader(raw_source, ignore_comments=False)
         self.sanitized_lines = tuple(item for item in reader)
+        # remove comments but keep pragmas
+        self.sanitized_lines = tuple(l for l in self.sanitized_lines if
+                len(l.line) > 1 and not (l.line[0] == '!' and not l.line[1] == '$'))
```
reuterbal, line 181: I have managed to break it after all but can fix this with a slightly expanded logic:
```suggestion
        # do not ignore comments during reading as this would also ignore pragmas ...
        reader = FortranStringReader(raw_source, ignore_comments=False)
        lines = tuple(l for l in reader)
        # ...but remove all comments that do not look like pragmas and are not inline comments
        def is_not_comment(line, prev=None):
            prev_end = prev.span[1] if prev else 0
            return not isinstance(line, ReadComment) or (
                line.span[0] > prev_end and line.line[:2] == '!$'
            )
        self.sanitized_lines = tuple(takewhile(is_not_comment, lines[:1])) + tuple(
            l for l, prev in zip(lines[1:], lines) if is_not_comment(l, prev)
        )
```

reuterbal: The stress test for this is the following:
```py
def test_regex_comments():
    """
    Make sure the REGEX frontend doesn't match any comments
    """
    fcode = """
SUBROUTINE my_routine
! use my_mod
use other_mod, only: foo
use third_mod ! use fourth_mod
use fifth_mod! , only: bar
implicit none

! type my_type
type other_type
end type

integer :: var !, val

! $acc not an acc pragma
!$ acc also not an acc pragma
var = 1 !$acc definitely not a pragma
!!$acc not a pragma either
!$$acc no pragma

var = 1 &
    &+1!$acc again no pragma

call some_routine(var)
var = var ! + function(val)
var = var + 1 ! call other_routine(val)
!call third routine(val)
END SUBROUTINE my_routine
    """.strip()

    source = Sourcefile.from_source(fcode, frontend=REGEX)
    routine = source['my_routine']
    assert len(routine.imports) == 3
    assert [imprt.module for imprt in routine.imports] == ['other_mod', 'third_mod', 'fifth_mod']
    assert len(calls := FindNodes(ir.CallStatement).visit(routine.ir)) == 1 and calls[0].name == 'some_routine'
    assert not FindNodes(ir.Pragma).visit(routine.ir)

```

reuterbal (changes requested):
Thanks, that this seems to straightforward is exciting. I've found some small edge cases where comments are not correctly stripped or mis-interpreted as directives, but have also provided a potential fix.

MichaelSt98:
@reuterbal thanks for breaking it and thanks for providing the corresponding fixes! :smile:

reuterbal (approved):
Very nice, many thanks!

PR #548, opened 2025-05-08
hack for OpenMP updating global variables on device

MichaelSt98:
Having:

```fortran
module some_mod
  integer, allocatable :: foo(:, :)
!$omp declare target(foo)
end module some_mod
```

Unfortunately, this doesn't work:

```fortran
use some_mod, only: foo

!$omp target update to(foo)
```

However, the following works (as workaround):

```fortran
use some_mod, only: foo

!$omp target update to(foo)
!$omp target enter data map(to: foo)
```

### DIFF
--- a/loki/transformations/data_offload/tests/test_global_var.py
+++ b/loki/transformations/data_offload/tests/test_global_var.py
@@ -336,7 +336,8 @@ def test_global_variable_offload(frontend, key, config, global_variable_analysis
 
 
 @pytest.mark.parametrize('frontend', available_frontends())
-def test_transformation_global_var_import(here, config, frontend, tmp_path):
+@pytest.mark.parametrize('directive', ['openacc', 'omp-gpu'])
+def test_transformation_global_var_import(here, config, frontend, directive, tmp_path):
     """
     Test the generation of offload instructions of global variable imports.
     """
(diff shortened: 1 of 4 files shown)


loki/transformations/data_offload/tests/test_global_var.py
```diff
@@ -366,38 +367,82 @@ def test_transformation_global_var_import(here, config, frontend, tmp_path):
...
-    assert 'var3' in pragmas[0].content
+    assert len(pragmas) == 5
 
-    pragmas = FindNodes(Pragma).visit(moduleC.spec)
-    assert len(pragmas) == 1
-    assert pragmas[0].keyword == 'acc'
-    assert 'declare create' in pragmas[0].content
-    assert 'var4' in pragmas[0].content
-    assert 'var5' in pragmas[0].content
+    if directive == 'acc':
```
reuterbal, line 372:
```suggestion
    if directive == 'openacc':
```

reuterbal (changes requested):
Great... neat fix, though!
A little naming inconsistency in the test, which renders the OpenACC checks ineffective. Also, if you don't mind please add the directive to the table: https://github.com/ecmwf-ifs/loki/blob/main/docs/source/loki_pragma_model.csv

loki/transformations/data_offload/tests/test_global_var.py
```diff
@@ -366,38 +367,84 @@ def test_transformation_global_var_import(here, config, frontend, tmp_path):
...
+        assert 'update device' in pragmas[0].content
+        assert 'var2' in pragmas[0].content
+        assert 'var3' in pragmas[0].content
+
+        assert pragmas[1].keyword == 'loki'
+        assert 'omp-update-global-vars in(' in pragmas[1].content
+        assert 'var2' in pragmas[1].content
+        assert 'var3' in pragmas[1].content
+
+        assert pragmas[2].keyword == 'acc'
```
reuterbal, line 384: OMNI seems to be converting these to upper case

reuterbal (changes requested):

reuterbal (approved):
Looks great now!

PR #550, opened 2025-05-09
FParser: Better `Subroutine` and `Module` construction

mlange05:
Another small housekeeping PR that aims to tidy up the construction of `Subroutine` and `Module` objects in the FParser frontend. The key changes are a slight generalisation of utilities to move comments between spec and body, and a common utility to pre-instantiate sub-procedure scopes.

_As a bonus, this gets rid of the last use of a naked dict-based `Transformer` in the AST traversal._ 

To make this safer, I also extended one of the `Subroutine` tests to capture the corner case that forces us to pre-instantiate internal and module procedures. We do this, so that all subroutines are created up front, so that `InlineCall`s between two internal routines are handled cleanly and the cross-calling type info is generated safely.

I also piggy-backed a small change that makes sure we use the `get_source()` utility throughout, even for node types with distinct `end_xxx` AST nodes. This all is in preparation to separate Function and Subroutine constructors, which have a lot of natural overlap. Hope this makes sense.

### DIFF
--- a/loki/frontend/fparser.py
+++ b/loki/frontend/fparser.py
@@ -2077,30 +2077,23 @@ class FParser2IR(GenericVisitor):
             ]
 
         # Build the spec
-        docs = ()
-        spec = None
-        spec_ast = get_child(o, Fortran2003.Specification_Part)
-        if spec_ast:
-            spec = self.visit(spec_ast, **kwargs)
-            spec = sanitize_ir(spec, FP, pp_registry=sanitize_registry[FP], pp_info=self.pp_info)
+        spec = self.visit(get_child(o, Fortran2003.Specification_Part), **kwargs)
+        spec = sanitize_ir(spec, FP, pp_registry=sanitize_registry[FP], pp_info=self.pp_info)
 
-            # Infer any additional shape information from `!$loki dimension` pragmas
-            spec = attach_pragmas(spec, ir.VariableDeclaration)
-            spec = process_dimension_pragmas(spec)
-            spec = detach_pragmas(spec, ir.VariableDeclaration)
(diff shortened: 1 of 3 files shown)


loki/frontend/fparser.py
```diff
@@ -1825,24 +1863,11 @@ def visit_Subroutine_Subprogram(self, o, **kwargs):
...
 
-        # Now we move comments from the end to the beginning of the body as those
-        # can potentially be pragmas.
-        for node in reversed(spec.body):
-            if not isinstance(node, (ir.Pragma, ir.Comment, ir.CommentBlock)):
-                break
-            body.prepend(node)
-            comment_map[node] = None
-        spec = Transformer(comment_map, invalidate_source=False).visit(spec)
+        # Move trailing comments from spec to the body as those can be pragmas.
```
reuterbal, line 1869: [no action only and observation] The comment's not 100% accurate anymore. We're moving comments and directives, but the latter are now identified as different node types (which wasn't the case at the time this was first built, iirc). The conceptual step is still required, though, and we can't bail out early either, because you might have interleaved comments and preprocesser directives and you wouldn't want to miss any of the latter

loki/frontend/fparser.py
```diff
@@ -229,6 +229,41 @@ def rget_child(node, node_type):
     return None
 
 
+def _get_comments_from_section(sec, include_pragmas=False, reverse=False):
```
reuterbal, line 232: You could consider making this a common frontend utility. We're doing the same hack for OMNI: https://github.com/ecmwf-ifs/loki/blob/main/loki/frontend/omni.py#L495

reuterbal (approved):
Very nice, no objections! Small stylistic comments/questions, otherwise all good from my side

PR #551, opened 2025-05-10
SCC: privatise derived-type scalars

awnawab:
Derived-types are classified as "aggregate variables" in the OpenACC and OpenMP offload standards and have the same implicit data attributes as arrays. This PR therefore extends `SCCAnnotate` to also privatise derived-type scalars (derived-type arrays would have already been privatised according to the existing implementation).

Other changes included are:
- Skip driver loop labeling for loops with existing parallel pragmas
- The ability to use a derived-type component to set the vector_length in a gang loop

### DIFF
--- a/loki/transformations/single_column/revector.py
+++ b/loki/transformations/single_column/revector.py
@@ -169,6 +169,13 @@ class BaseRevectorTransformation(Transformation):
         This method assumes that pragmas have been attached via
         :any:`pragmas_attached`.
         """
+
+        # Skip loops with existing parallel annotations
+        if loop.pragma:
+            if any(pragma.keyword.lower() in ['omp', 'acc'] and 'parallel' in pragma.content.lower()
+                   for pragma in loop.pragma):
+                return
+
         # Find a horizontal size variable to mark vector_length
         symbol_map = routine.symbol_map
         sizes = tuple(
--- a/loki/transformations/single_column/revector.py
+++ b/loki/transformations/single_column/revector.py
@@ -172,8 +172,8 @@ class BaseRevectorTransformation(Transformation):
         # Find a horizontal size variable to mark vector_length
         symbol_map = routine.symbol_map
         sizes = tuple(
-            symbol_map.get(size) for size in self.horizontal.size_expressions
-            if size in symbol_map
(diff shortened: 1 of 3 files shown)


MichaelSt98 (approved):
Looks great!

The only thing that could be improved in my opinion is to add an explanatory comment for this change in code. Something like:

> Derived-types are classified as "aggregate variables" in the OpenACC and OpenMP offload standards and have the same implicit data attributes as arrays. Therefore, not only derived-type arrays need to be privatised but derived-type scalars too.

loki/transformations/single_column/revector.py
```diff
@@ -169,11 +169,18 @@ def mark_driver_loop(self, routine, loop):
         This method assumes that pragmas have been attached via
         :any:`pragmas_attached`.
         """
+
+        # Skip loops with existing parallel annotations
+        if loop.pragma:
+            if any(pragma.keyword.lower() in ['omp', 'acc'] and 'parallel' in pragma.content.lower()
+                   for pragma in loop.pragma):
+                return
```
reuterbal, line 177: What's the idea behind this? We've converted all transformations to be programming-model agnostic and make them work with loki directives instead. If this is meant as an injection mechanism for handcrafted directives to take precedence over auto-generated ones, then I would suggest to insert loki directives in the code instead.

reuterbal (approved):
Looks great as far as I can tell. One small question in-line.

loki/transformations/single_column/revector.py
awnawab: The driver layers for the dycore look something like this:
```
IF (ON_GPU) THEN

!$loki data offload
DO IBL=1,...
CALL KERNEL
ENDDO
!$loki end data

ELSE

!$loki data set_pointers
!$OMP PARALLEL
DO IBL=1,...
CALL KERNEL
ENDDO
!$OMP END PARALLEL
!$loki end data

ENDIF
```

This bail-out is to prevent annotation of the second loop, which is the CPU optimised Loki variant. I've now realised that the `pragma_model` mapper also has an openmp threading backend. The only thing missing is the thread scheduling logic, so I can add that in a follow-up PR and replace the hard-coded openmp loop with a loki pragma.

PR #553, opened 2025-05-11
`SubstitutePragmaStrings` transformer

awnawab:
A simple in-place transformer that updates the content of pragmas based on a given string map.


reuterbal (changes requested):
Small fix to the test, otherwise looks great and very useful!

reuterbal (approved):

mlange05 (approved):
Looks great and very neatly tested!  🙏 :shipit:

PR #554, opened 2025-05-12
New stack variants: raw/direct indexing and based on Fortran pointers

MichaelSt98:
Refactored `TemporariesRawStackTransformation` and implemented:

* another variant of the "raw stack"/direct indexing
* a stack based on Fortran pointers

Tested with ecwam.

### DIFF
--- a/loki/transformations/temporaries/stack_allocator.py
+++ b/loki/transformations/temporaries/stack_allocator.py
@@ -28,14 +28,31 @@ from loki.types import BasicType, SymbolAttributes
 from loki.transformations.utilities import recursive_expression_map_update, single_variable_declaration
 
 
-__all__ = ['PoolAllocatorFtrPtrTransformation', 'PoolAllocatorRawTransformation']
+__all__ = ['FtrPtrStackTransformation', 'DirectIdxStackTransformation']
 
+class BaseStackTransformation(Transformation):
+    """
+    Base Transformation to inject a stack that allocates large scratch spaces per block
+    and per datatype on the driver and maps temporary arrays in kernels to this scratch space.
 
-class PoolAllocatorBaseTransformation(Transformation):
+    Parameters
+    ----------
+    block_dim : :any:`Dimension`
+        :any:`Dimension` object to define the blocking dimension.
+    horizontal : :any:`Dimension`
+        :any:`Dimension` object to define the horizontal dimension.
+    stack_name : str, optional
+        Name of the stack (default: 'STACK')
+    local_int_var_name_pattern : str, optional
+        Local integer variable names pattern
+        (default: 'JD_{name}')
+    int_kind : str, optional
+        Integer kind (default: 'JWIM')
+    """
 
     _key = 'PoolAllocatorBaseTransformation'
+
     reverse_traversal = True
-    process_ignored_items = True # TODO: remove?
 
     type_name_dict = {
         BasicType.REAL: {'kernel': 'P', 'driver': 'Z'},
@@ -75,7 +92,7 @@ class PoolAllocatorBaseTransformation(Transformation):
         sub_sgraph = kwargs.get('sub_sgraph', None)
         successors = as_tuple(sub_sgraph.successors(item)) if sub_sgraph is not None else ()
 
-        # TODO: shouldn't happen here ...
+        # TODO: probably shouldn't happen here ...
         for call in FindNodes(CallStatement).visit(routine.body):
             if call.routine is not BasicType.DEFERRED:
                 call.convert_kwargs_to_args()
@@ -199,14 +216,6 @@ class PoolAllocatorBaseTransformation(Transformation):
         for a in arrays:
             type_dict.setdefault(a.type.dtype, {})
             type_dict[a.type.dtype].setdefault(a.type.kind, []).append(a)
-        # for a in arrays:
-        #     if a.type.dtype in type_dict:
-        #         if a.type.kind in type_dict[a.type.dtype]:
-        #             type_dict[a.type.dtype][a.type.kind] += [a]
-        #         else:
-        #             type_dict[a.type.dtype][a.type.kind] = [a]
-        #     else:
-        #         type_dict[a.type.dtype] = {a.type.kind: [a]}
 
         return type_dict
 
@@ -230,27 +239,27 @@ class PoolAllocatorBaseTransformation(Transformation):
         }
         call_map = {}
 
-        #Loop over calls and check if they call a successor routine and if the
-        #transformation data is available
+        # loop over calls and check if they call a successor routine and if the
+        # transformation data is available
         for call in FindNodes(CallStatement).visit(routine.body):
             if call.name in successor_map and self._key in successor_map[call.name].trafo_data:
                 successor_stack_dict = successor_map[call.name].trafo_data[self._key]['stack_dict']
 
                 call_stack_args = []
 
-                #Loop over dtypes and kinds in successor arguments stacks
-                #and construct list of stack arguments
+                # loop over dtypes and kinds in successor arguments stacks
+                # and construct list of stack arguments
                 for dtype in successor_stack_dict:
                     for kind in successor_stack_dict[dtype]:
                         call_stack_args += list(stack_arg_dict[dtype][kind])
 
-                #Get position of optional arguments so we can place the stacks in front
+                # get position of optional arguments so we can place the stacks in front
                 arg_pos = [call.routine.arguments.index(arg) for arg in call.routine.arguments if arg.type.optional]
 
                 arguments = call.arguments
                 if arg_pos:
-                    #Stack arguments have already been added to the routine call signature
-                    #so we have to subtract the number of stack arguments from the optional position
+                    # stack arguments have already been added to the routine call signature
+                    # so we have to subtract the number of stack arguments from the optional position
                     arg_pos = min(arg_pos) - len(call_stack_args)
                     arguments = arguments[:arg_pos] + as_tuple(call_stack_args) + arguments[arg_pos:]
                 else:
@@ -278,7 +287,7 @@ class PoolAllocatorBaseTransformation(Transformation):
             The items corresponding to successor routines called from :data:`routine`
         """
 
-        #Block variables
+        # block variables
         kgpblock = self._get_int_var(name=self.block_dim.size, scope=routine)
         jgpblock = self._get_int_var(name=self.block_dim.index, scope=routine)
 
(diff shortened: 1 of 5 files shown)


loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+        # block variables
+        kgpblock = self._get_int_var(name=self.block_dim.size, scope=routine)
+        jgpblock = self._get_int_var(name=self.block_dim.index, scope=routine)
+
+        stack_vars = []
+        stack_arg_dict = {}
+        assignments = []
+        deallocs = []
+        pragma_string = ''
+        pragma_data_start = None
```
reuterbal, line 299: This is used only locally further down in the conditional, so shouldn't need to be initialized here
```suggestion
```

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+        if pragma_string:
+            # remove ', '
+            pragma_string = pragma_string[:-2].lower()
+
+            present_pragma = None
+            acc_pragmas = [p for p in FindNodes(Pragma).visit(routine.body) if p.keyword.lower() == 'loki'] # acc
+            for pragma in acc_pragmas:
+                if pragma.content.lower().startswith('device-present'):
+                    present_pragma = pragma
+                    break
```
reuterbal, line 422: Seems like this search could actually be combined:

```suggestion
            present_pragmas = [
                p for p in FindNodes(Pragma).visit(routine.body)
                if p.keyword.lower() == 'loki' and p.content.lower().startswith('device-present')
            ]
```

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+            if present_pragma:
+                pragma_map = {present_pragma: None}
```
reuterbal, line 424: ...and then a small change here:
```suggestion
            if present_pragmas:
                present_pragma = present_pragmas[0]
                pragma_map = {present_pragma: None}
```

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+                                           rhs=stack_dict[dtype][kind]), stack_alloc, stack_used_var_init]
+                deallocs += [stack_dealloc]
+                pragma_string += f'{stack_var.name}, '
+
+        # add to routine
+        routine.variables = routine.variables + as_tuple(stack_vars)
+        nodes_to_add = assignments
+
+        if pragma_string:
+            pragma_string = pragma_string[:-2].lower()
```
reuterbal, line 339: [optional] You could avoid this if you collected variable names in a list and then just did `pragma_string = ', '.join(stack_data_vars)`

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+
+            present_pragma = None
+            acc_pragmas = [p for p in FindNodes(Pragma).visit(routine.body) if p.keyword.lower() == 'loki'] # acc
+            for pragma in acc_pragmas:
+                if pragma.content.lower().startswith('device-present'):
+                    present_pragma = pragma
+                    break
+            if present_pragma:
+                pragma_map = {present_pragma: None}
+                routine.body = Transformer(pragma_map).visit(routine.body)
```
reuterbal, line 425: Not entirely sure I'm following what's happening here. My understanding is, we're
1. Looking for existing `!$loki device-present` pragmas
2. Remove the first such pragma and re-insert it at the beginning of the body with the var list amended
3. We don't add a corresponding `end` pragma
4. If no such pragma exists, we insert a `device-present` pragma pair around the body

Apologies if these are stupid questions:
1. are we certain there's always just one such pragma? 
2. If there can only be one, is there any reason to expect this pragma to be in the wrong place? Because then we could simply do an in-place `_update` with the amended string here, thus speeding up processing and removing some control flow complexity

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+    def apply_pool_allocator_to_temporaries(self, routine, item=None): # pylint: disable=unused-argument
+        """
+        Apply raw stack allocator to local temporary arrays
+
+        This appends the relevant argument to the routine's dummy argument list and
+        creates the assignment for the local copy of the stack type.
+        For all local arrays, a Cray pointer is instantiated and the temporaries
+        are mapped via Cray pointers to the pool-allocated memory region.
+
+        The cumulative size of all temporary arrays is determined and returned.
```
reuterbal, line 458: This docstring needs updating to reflect the base-class nature of the method here.

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+            var for var in temporary_arrays
+            if not all(is_dimension_constant(d) for d in var.shape)
+        ]
+
+        # Filter out variables whose first dimension is not horizontal
+        # temporary_arrays = [
+        #     var for var in temporary_arrays if (
+        #     isinstance(var.shape[0], Scalar) and
+        #     var.shape[0].name.lower() == self.horizontal.size.lower())
+        # ]
```
reuterbal, line 509: debug leftover?

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+                    else:
+                        stack_dict[dtype] = {kind: [local_stack_dict[dtype][kind]]}
+
+        # if several expressions, return MAX, else just add the expression
+        for (dtype, kind_dict) in stack_dict.items():
+            for (kind, stacks) in kind_dict.items():
+                if len(stacks) == 1:
+                    kind_dict[kind] = stacks[0]
+                else:
+                    kind_dict[kind] = InlineCall(function = Variable(name = 'MAX'), parameters = as_tuple(stacks))
```
reuterbal, line 615:
```suggestion
                    kind_dict[kind] = InlineCall(function=Variable(name='MAX'), parameters=as_tuple(stacks))
```

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+            routine.symbol_attrs[var.name] = var.type.clone(pointer=True, contiguous=True)
+        declarations = FindNodes(VariableDeclaration).visit(routine.spec)
+        # adapt declarations
+        decl_map = {}
+        for decl in declarations:
+            if decl.symbols[0] in temporary_arrays:
+                new_dimensions = as_tuple(RangeIndex((None, None)) for _ in decl.symbols[0].dimensions)
+                new_symbol = decl.symbols[0].clone(dimensions=new_dimensions)
+                if decl.dimensions is not None:
+                    decl_map[decl] = decl.clone(dimensions=as_tuple(RangeIndex((None, None)) for _ in decl.dimensions),
```
reuterbal, line 829: [optional] Shorthand:
```suggestion
                    decl_map[decl] = decl.clone(dimensions=(RangeIndex((None, None),) * len(decl.dimensions)),
```

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+        return stack_dict
+
+class FtrPtrStackTransformation(BaseStackTransformation):
+    """         
+    Transformation to inject a stack that allocates large scratch spaces per block
+    and per datatype on the driver and maps temporary arrays in kernels to this scratch space.
+
+    Starting from:
+
+    .. code-block:: fortran
```
reuterbal, line 626: This doesn't seem to render correctly, maybe missing empty lines or something (same in the other docstring): https://sites.ecmwf.int/docs/loki/554/loki.transformations.temporaries.stack_allocator.html#module-loki.transformations.temporaries.stack_allocator

I suggest to check the output of the documentation builder CI run for any warnings.

loki/transformations/temporaries/tests/test_stack_allocator.py
```diff
@@ -0,0 +1,381 @@
...
+    if frontend == OMNI:
+        (tmp_path/'parkind_mod.F90').write_text(fcode_parkind_mod)
+        parkind_mod = Sourcefile.from_file(tmp_path/'parkind_mod.F90', frontend=frontend, xmods=[tmp_path])
+        (tmp_path/'yomphy_mod.F90').write_text(fcode_yomphy_mod)
+        yomphy_mod = Sourcefile.from_file(tmp_path/'yomphy_mod.F90', frontend=frontend, xmods=[tmp_path])
+        (tmp_path/'mf_phys_mod.F90').write_text(fcode_mf_phys_mod)
+        mf_phys_mod = Sourcefile.from_file(tmp_path/'mf_phys_mod.F90', frontend=frontend, xmods=[tmp_path])
+        definitions = parkind_mod.definitions + yomphy_mod.definitions + mf_phys_mod.definitions
+    else:
+        definitions = ()
```
reuterbal, line 261: Since OMNI is disabled for the test anyway, this can be skipped.

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+                    stack_dict[dtype][kind] = Literal(0)
+
+                # store type information of temporary allocation
+                if item:
+                    if kind in routine.imported_symbols:
+                        item.trafo_data[self._key]['kind_imports'][kind] = routine.import_map[kind.name].module.lower()
+                    for array in arrays:
+                        dims = [d for d in array.shape if d in routine.imported_symbols]
+                        for d in dims:
+                            item.trafo_data[self._key]['kind_imports'][d] = routine.import_map[d.name].module.lower()
```
reuterbal, line 1210: This seems not to be covered by tests.

loki/transformations/temporaries/stack_allocator.py
```diff
@@ -0,0 +1,1372 @@
...
+
+    @staticmethod
+    def _insert_stack_at_loki_pragma(routine, insert):
+        pragma_map = {}
+        for pragma in FindNodes(Pragma).visit(routine.body):
+            if pragma.keyword == 'loki' and 'stack-insert' in pragma.content:
+                pragma_map[pragma] = insert
+            if pragma_map:
+                routine.body = Transformer(pragma_map).visit(routine.body)
+                return True
```
reuterbal, line 139: This part seems not to be covered in the tests.

reuterbal (changes requested):
This is a hugely complex and impressive piece of work! The fact that you are able to get these two variants to work and develop shared infrastructure to support them is a next level achievement!

I can't claim to have fully thought-through all details of the transformations, but the overall structure and approach seems sensible and in-line with offline discussions. The comments mostly cover some implementation details or suggestions, but no major show-stoppers.

loki/transformations/temporaries/stack_allocator.py
MichaelSt98: Not stupid at all. Thinking about it we can probably just add the pragmas around the body either way, since having multiple of those device-present pragmas shouldn't be a problem at all and it shouldn't happen that one of the stack variables ends up twice in the pragmas ... What do you think?

MichaelSt98:
Thanks for the review and the feedback @reuterbal!

PR #555, opened 2025-05-13
Hotfix for GH Actions: Replace package distribution version of JDK with setup-java action

reuterbal:
The download of OpenJDK in the dependency installation for OMNI failed spuriously but increasingly often in the last two days. This replaces the installation of the JDK by the official setup-java action, which is hopefully more robust and independent from the runner image.


mlange05 (approved):
Looks great! Many thanks! :shipit:

PR #556, opened 2025-05-14
Small bug fixes in the `DataflowAnalysis`

awnawab:
This PR fixes the following bugs in the dataflow analysis:
- Stripping dimensions of nested derived-type accesses
- Extending the exclusion of memory status checking calls to loop bounds

The PR also includes a behaviour change. Previously, all parents of a derived-type expression were marked as used if a single type member was used. Now, only the full derived-type expression is included in the used symbols set. For example, the `used_symbols` for the assignment `D = A%B%C` will be limited to `A%B%C`, whereas previously it would have also included `A%B, A`.

### DIFF
--- a/loki/analyse/tests/test_analyse_dataflow.py
+++ b/loki/analyse/tests/test_analyse_dataflow.py
@@ -436,10 +436,14 @@ implicit none
 
   real,intent(out) :: a(:,:)
   real             :: b(10)
-  integer                     :: bsize
+  integer          :: bsize, i
 
   if(size(a) > 0) a(:,:) = 0.
   bsize = size(b)
+
+  do i=1,size(b)
+    print *, i
+  enddo
 
 end subroutine test
     """.strip()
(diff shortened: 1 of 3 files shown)


loki/analyse/tests/test_analyse_dataflow.py
```diff
@@ -676,3 +680,45 @@ def test_analyse_associate(frontend):
...
+
+end module
+"""
+
+    source = Sourcefile.from_source(fcode, frontend=frontend, xmods=[tmp_path])
+    routine = source['kernel']
+
+    with dataflow_analysis_attached(routine):
+        assert routine.body.defines_symbols == {r'a%b%c'}
+        assert routine.body.uses_symbols == {r'd%b%c'}
```
reuterbal, line 724: This isn't a regular expression, so I'm confused why the string has the `r` prefix?
```suggestion
        assert routine.body.defines_symbols == {'a%b%c'}
        assert routine.body.uses_symbols == {'d%b%c'}
```

reuterbal (approved):
Thanks, looks good to me and the behavioural change makes sense, as discussed offline. This will need a careful regression CI run with IFS to make sure the behaviour wasn't relied upon implicitly.

loki/analyse/tests/test_analyse_dataflow.py
reuterbal: It's used to substitute variables into strings when printing formatted strings, pretty much in the same way as printf. But it's harmless in a string like this.

PR #557, opened 2025-05-15
Inline Trafo CMake planing when inlining marked routines

MichaelSt98:
Inline Trafo CMake planing when inlining marked routines.

In the planning step:

* look for calls with and without `!$loki inline` pragmas
* adapt `removed_dependencies` and `additional_dependencies` according to which calls are inlined


reuterbal (approved):
Very nice! Many thanks for sorting out the renaming implications along the way!

mlange05 (approved):
Fantastic work, very much appreciated as this has caught me a number of times now! Implementation looks neat and well tested, and I can confirm that it indeed behaves as expected in local ecphys regression tests. GTG from me! 🙏 :shipit:

PR #558, opened 2025-05-19
SCC: Make privatising derived-type variables configurable

mlange05:
This cause trouble in EC-physics, as we rely on them being treated as globals in various places. So, by default we leave this option off for now, until this is tested further.

As a bonus, this also replaces a use of `set` for unique variable lists with `dict.fromkeys()` to ensure reproducible string output in the list of private variable names.

This PR should bring EC-physics regression back to green.


mlange05:
@awnawab for context, PR #551 seems to have broken EC-physics regression. This is just a hotfix for now - more discussion offline.

reuterbal (approved):
Looks good, thanks!

PR #559, opened 2025-05-22
`Scheduler`: skip external items in plan_mode

awnawab:
We frequently have to process sourcefiles with Loki that are absent at configure time. This PR enables the scheduler to skip the processing of these items if we are in planning mode, and a `RuntimeError` is only raised if the items are still marked as external in the build step.


reuterbal (changes requested):
I'm not convinced we want to make it generally acceptable like this to skip external items because it renders the strict mode ineffective until much later during the build. This becomes a problem specifically when trying to build up the exclude/ignore lists, which then might require a change to the config once the external item has been encountered, which then triggers a new round of configure and build. Having it fail during configure already shortens the turnaround time considerably.
Maybe a list of items that are marked as "build-time generated" and thus accepted as external during planning might be a safer approach here?

awnawab:
After the discussion offline, I've adapted the implementation to only skip those `ExternalItem`s marked as `generated` during the planning step. Please have another look when you can.

mlange05 (approved):
Looks good to me. 🙏 :shipit:

PR #564, opened 2025-06-04
Add offload utils

awnawab:
This small PR contributes the following offload utilities:
- `attach/detach` clauses for OpenACC unstructured data pragmas
- Optional `finalize` clause for OpenACC unstructured data pragmas
- Addition of `GET_HOST_DATA_*` and `DELETE_DEVICE_DATA`

### DIFF
--- a/docs/source/loki_pragma_model.csv
+++ b/docs/source/loki_pragma_model.csv
@@ -1,8 +1,8 @@
 Loki,OpenACC,OMP-GPU
-``create device(...)``,``declare create(...)``,
+``create device(...)``,``declare create(...)``,``declare target(...)``
 ``update device(...) host(...)``,``update device(...) self(...)``,
-``unstructured-data in(...) create(...)``,``enter data copyin(...) create(...)``,``target enter data map(to: ...) map(alloc: ...)``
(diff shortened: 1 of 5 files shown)


reuterbal (commented):
Thanks, implementation looks good to me. Can you please update the directives table with the additional keywords?
https://github.com/ecmwf-ifs/loki/blob/main/docs/source/loki_pragma_model.csv

This will ensure the documentation is up-to-date at https://sites.ecmwf.int/docs/loki/564/programming_models.html#loki-directives

Am I correct in assuming there's no equivalent in OpenMP for attach/detach?

awnawab:
Ah yes, forgot about that, thanks for the reminder. Will do shortly.

And yes, OpenMP has no equivalent for attach. For F-API I have been mapping `attach` to `map(to:`, for both nvhpc and rocm-afar this doesn't trigger a recopy (at least on unit-test sized cases) and does effectively the same job as attach. The reason why I didn't add the OpenMP directives in this PR is because the OpenMP offload backend is missing many pieces, at least for the deepcopy, so I'll prepare a PR for OpenMP offload once I try using that for the deepcopy.

PR #565, opened 2025-06-12
Basic exception handling with some origin information for Transformation.apply

reuterbal:
This should give a slightly more informative message when hitting an exception in a transformation, giving an indication on what program unit node the problem occurs.


mlange05 (approved):
Great little utility that can be very helpful. I've confirmed that nested tracebacks still work fine, so we get the file, transformation, and where in the transformation things go wrong, which is super nice. GTG from me. 🙏 :shipit:

PR #567, opened 2025-06-14
Restore editable pip installation in CMake builds

awnawab:
A small PR to re-enable editable CMake-driven builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,11 +154,6 @@ if( NOT HAVE_NO_INSTALL )
     endif()
 
     # Install Loki python package with dependencies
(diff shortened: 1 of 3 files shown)


reuterbal (approved):
Ooops, sorry, I must have broken that. Changes look great!

Do you think we could capture the CMake output with a stdout watchdog in the Pytest to make sure the relevant pip command includes this flag? Possibly right here in the install fixture:
https://github.com/ecmwf-ifs/loki/blob/972116ab8a53ace7239a52cfafecf8108eb1f5c2/loki/tests/test_cmake.py#L150-L179
To make sure the option remains functional in the future

PR #568, opened 2025-06-19
Re-activate MacOS CI tests

reuterbal:
When upgrading to NumPy 2 and the corresponding switch to Meson, the MacOS runner has stopped working and was disabled as a consequence. This PR introduces the [setup-fortran](https://github.com/marketplace/actions/setup-fortran) action to the CI workflow, which allows to re-enable the MacOS tests.

I took this opportunity to also remove some now obsolete bits from the Actions definition


.github/workflows/tests.yml
```diff
@@ -49,18 +47,18 @@ jobs:
...
+          compiler: ${{ matrix.toolchain.compiler }}
+          version: ${{ matrix.toolchain.version }}
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
 
-      - name: Set up JDK
```
mlange05, line 58: [no action] Is this no longer needed for OMNI? Cool!

mlange05 (approved):
Excellent, many thanks! GTG from me. 👍 :shipit:

PR #569, opened 2025-06-20
Introduce `Item.transformation_ir` as entry point for transformations

reuterbal:
Pull request #563 highlighted a potential design flaw in our handling of Transformation entry points in transformations for items that define their own scope (in some sense) but are not program units, more specifically the transformation entry points for typedefs or interfaces. Currently, the scheduler “applies” the transformation to `Item.scope_ir`, which usually refers to the `Module` or `Subroutine`. But for `TypeDef` and `Interface` it points to the `Typedef` or `Interface`, which is then silently ignored in a `Transformation.apply` (i.e., even when including these in the processing, they wouldn’t incur any changes to be applied).

So far, we didn’t have any transformations that needed to transform either of them itself, thus this wasn’t a problem. But the analysis pass in #563  requires this now.

The underlying cause for this is that `scope_ir` is meant to provide the scope of an item, in which any definitions are declared - which is, e.g., used to figure out where definitions like `ProcedureBinding` are anchored. Until now it just happened to also coincide with the suitable transformation entry point in relevant cases so far.

My proposal to work around this is to introduce instead a separate `transformation_ir` property on items to explicitly declare for an item the entry point for transformations. In most cases this will be an alias for `scope_ir` but with specific overloads for `TypeDefItem`, `ProcedureBindingItem` and `InterfaceItem`. This will ensure a consist entry point behaviour.

I have slightly amended the tests for `TypeDefItem` to verify this behaviour on top of the implicit testing via asserting correct behaviour of transformations. I suspect test coverage might point out a few blind spots, though, which I will close with another commit.


awnawab (approved):
Accessing `TypeDef`s, `Interface`s and `ProcedureBinding`s in transformations via their encompassing modules also makes sense to me, so I'm in full agreement with the fix here. Many thanks, looks great!

Issue #513, opened 2025-03-18
Error when parsing FUNCTIONS declared both with a return type and a RESULT clause


remi-kazeroni:
### What happened?

When parsing and regenerating a `FUNCTION` that is declared with a return type, the return type keyword is removed from the `prefix` specifications and replaced by a declaration of the function's result. The syntax is different but the behavior of the function has not changed.

For example, parsing and regenerating this function:

```fortran
REAL FUNCTION square(a)
    IMPLICIT NONE
    REAL, INTENT(IN) :: a
    square = a * a
END FUNCTION square
```

leads to:
```fortran
FUNCTION square(a)
    IMPLICIT NONE
    REAL :: square
    REAL, INTENT(IN) :: a
    square = a * a
END FUNCTION square
```

Howerer, if a `RESULT` clause is added to the function declaration, this leads to an incorrect inference of the result and thus to a error when compiling the regenerated code. For example, parsing and regenerating this function:

```fortran
REAL FUNCTION square(a) result(b)
    IMPLICIT NONE
    REAL, INTENT(IN) :: a
    b = a * a
END FUNCTION square
```

leads to:
```fortran
FUNCTION square (a) RESULT(b)
      IMPLICIT NONE
      REAL :: square ! this should be REAL :: b
      REAL, INTENT(IN) :: a
      b = a*a
    END FUNCTION square
```
where the declared result should `b` instead of `square`.

We are not sure how the wrong result inference should be addressed. Keeping the result type in the declaration (as in other programing languages) would be nice but not necessary.

### What are the steps to reproduce the bug?

```python
fcode = """
REAL FUNCTION square(a) result(b)
  IMPLICIT NONE
  REAL, INTENT(IN) :: a
  b = a * a
END FUNCTION square
""".strip()
square = Subroutine.from_source(fcode)
print(square.result_name)   # b
print(square.return_type)    # None
print(square.to_fortran())
```

The output is:
```python
    b
    None
    FUNCTION square (a) RESULT(b)
      IMPLICIT NONE
      REAL :: square
      REAL, INTENT(IN) :: a
      
      b = a*a
    END FUNCTION square
```

### Version

0.3.1

### Platform (OS and architecture)

Red Hat Enterprise Linux 9.2 - Python 3.12.7

### Relevant log output

```shell

```

### Organisation

IPSL

PR #570, opened 2025-06-24
Function: Add test and fix for implicit return type with result name

mlange05:
This covers the corner case in which the function return type is declared implicitly as a function prefix, but the result variable is defined without explicit declaration via `result(var)`.

This corner case seems to have been missed in PR #552 and should now address issue #513. Many thanks to @remi-kazeroni for providing the test case that was added in this PR.


reuterbal (approved):
Uncontroversial. Many thanks for fixing this!
Thanks to @remi-kazeroni for reporting and providing the reproducer.

PR #572, opened 2025-06-26
RemoveCode: Perform dead code elimination in MultiConditional nodes

mlange05:
As the title implies, we're currently only performing dead code removal on IF/THEN/ELSE conditionals, but not SELECT CASE multi-conditional. This PR rectifies that and adds a test that shows the potential uses.

### DIFF
--- a/loki/transformations/tests/test_remove_code.py
+++ b/loki/transformations/tests/test_remove_code.py
@@ -367,6 +367,7 @@ end subroutine test_dead_code_multiconditional
     multiconds = FindNodes(ir.MultiConditional).visit(routine.body)
     assert len(multiconds) == 1
     assert multiconds[0].expr == 'i'
+    assert len(FindNodes(ir.Conditional).visit(routine.body)) == 1
     assigns = FindNodes(ir.Assignment).visit(routine.body)
     assert len(assigns) == 5
     assert assigns[0].lhs == 'b' and assigns[0].rhs == 'b + 2.0'


loki/transformations/tests/test_remove_code.py
```diff
@@ -311,6 +311,71 @@ def test_transform_dead_code_conditional_nested(frontend):
...
+
+    do_remove_dead_code(routine)
+
+    # Check that the first multi-conditional and the nested conditional
+    # inside the second conditional have been removed.
+    multiconds = FindNodes(ir.MultiConditional).visit(routine.body)
+    assert len(multiconds) == 1
+    assert multiconds[0].expr == 'i'
+    assigns = FindNodes(ir.Assignment).visit(routine.body)
+    assert len(assigns) == 5
```
reuterbal, line 371: I don't think this is actually testing that the second conditional has been removed? I suggest adding the following:
```suggestion
    # Check that the first multi-conditional and the nested conditional
    # inside the second conditional have been removed.
    multiconds = FindNodes(ir.MultiConditional).visit(routine.body)
    assert len(multiconds) == 1
    assert multiconds[0].expr == 'i'
    assert not FindNodes(ir.Conditional).visit(routine.body)
    assigns = FindNodes(ir.Assignment).visit(routine.body)
    assert len(assigns) == 5
```

reuterbal (changes requested):
Thanks, looks good. Just a small suggestion to improve the test, otherwise good to go!

PR #573, opened 2025-06-26
`DataOffloadDeepcopyTransformation`

awnawab:
This PR contributes the matching transformation to generate the deep-copy using the analysis. An assumption the deep-copy is based around is that data access is always via basic types, which rules out overloaded operators and copy on assignment for derived-types. The exception are memory status queries e.g. `ubound, lbound, size`. These are supported by the current implementation.

Another limitation is that access pointers to fields are assumed to be members of the same derived-type as the field pointer itself, i.e. `A%P_FIELD` is the access pointer associated with field `A%F_P`.  This is not an underlying assumption of the implementation however, and the transformation can easily be made more flexible in this regard.

### DIFF
--- a/loki/transformations/data_offload/offload_deepcopy.py
+++ b/loki/transformations/data_offload/offload_deepcopy.py
@@ -360,7 +360,51 @@ class DataOffloadDeepcopyAnalysis(Transformation):
 
 class DataOffloadDeepcopyTransformation(Transformation):
     """
-    A transformation to generate deepcopies of derived-types.
+    A transformation that generates a deepcopy of all the arguments to a
+    GPU kernel. It relies on the analysis gathered by the 
+    :any:`DataOffloadDeepcopyAnalysis` transformation, which must therefore
+    be run before this. Please note that the analysis and deepcopy are per
+    driver-loop, which must be wrapped in a `!$loki data` :any:`PragmaRegion`.
+
+    An underlying assumption of the transformation is that expressions used as 
+    lvalues and rvalues are of type :any:`BasicType`, i.e. the data
+    encompassed by a derived-type variable `a` with components `b` and `c` is
+    only ever accessed or modified via fully qualified derived-type expressions
+    `a%b` or `a%c`. The only accepted exception to this are memory status checks
+    such as `ubound`, `lbound`, `size` etc.
+
+    The encompassing `!$loki data` :any:`PragmaRegion` can be used to 
+    to pass hints to the transformation. Consider the following example:
+
+    .. code-block:: fortran
+       !$loki data present(a) write(b)
+       do ibl=1,nblks
+          call kernel(a, b, ...)
+       enddo
+       !$loki end data
+
+    Marking `a` as `present` instructs the transformation to skip the deepcopy
+    generation for it and simply place it in a `!$loki structured-data present`
+    clause. Marking `b` as `write` means the contents of the analysis are 
+    overriden and the generated deepcopy for `b` assumes write-only access. Other
+    hints that can be passed to the deepcopy generation are:
+     - read: Assume read-only access for the specified variables.
+     - readwrite: Assume read-write access for the specified variables.
+     - device_resident: Don't copy the specificied variables back to host and
+                        leave the device allocation intact.
+     - temporary: Wipe the device allocation of the specified variables but
+                  don't copy them  back to host.
+
+    The transformation supports two modes:
+     - offload: Generate device-host deepcopy for the arguments passed to the 
+                encompassed call-tree.
+     - set_pointers: Generate the FIELD_API boiler-plate to set host pointers
+                     for any argument representing a field.
+
+    Parameters
+    ----------
+    mode : str
+       Transformation mode, must be either "offload" or "set_pointers".
     """
 
     _key = 'DataOffloadDeepcopyAnalysis'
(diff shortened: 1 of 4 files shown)


loki/transformations/data_offload/offload_deepcopy.py
```diff
@@ -346,3 +356,429 @@ def transform_module(self, module, **kwargs): # pylint: disable=unused-argument
...
+        loopbody = ()
+        loop_vars = []
+        variable_map = routine.variable_map
+
+        for dim in range(len(var.type.shape)):
+            if f'j{dim+1}' in variable_map:
+                loop_vars += [variable_map[f'j{dim+1}']]
+            else:
+                loop_vars += [sym.Variable(name=f'J{dim+1}', type=SymbolAttributes(dtype=BasicType.INTEGER),
+                                           scope=routine)]
```
reuterbal, line 555: For this, you might be able to use the `get_integer_variable` utility instead? https://github.com/ecmwf-ifs/loki/blob/c9f55e53f535bca0485335569015aa8ac0c8ee29/loki/transformations/utilities.py#L542

loki/transformations/data_offload/offload_deepcopy.py
```diff
@@ -346,3 +356,429 @@ def transform_module(self, module, **kwargs): # pylint: disable=unused-argument
...
+
+                SubstitutePragmaStrings(str_map).visit(body)
+                loopbody = as_tuple(SubstituteExpressions(vmap).visit(body))
+            else:
+                # Add subsequent layers
+                vmap = {loop_vars[-2]: (loop_vars[-2], loop_vars[-1])}
+                str_map = {str(k): str(v) for k, v in vmap.items()}
+
+                SubstitutePragmaStrings(str_map).visit(loopbody)
+                loopbody = as_tuple(SubstituteExpressions(vmap).visit(loopbody))
```
reuterbal, line 578: Am I correctly understanding that this is responsible for supporting a 2-deep loop nest? Could this not be generalized?
Also, the marked lines here are not covered by tests.

loki/transformations/data_offload/offload_deepcopy.py
```diff
@@ -346,3 +356,429 @@ def transform_module(self, module, **kwargs): # pylint: disable=unused-argument
...
+        """Generate unstructured data delete instruction."""
+        return as_tuple(ir.Pragma(keyword='loki', content=f'end unstructured-data delete({var}) finalize'))
+
+    @staticmethod
+    def update_self(var):
+        """Pull back data to host."""
+        return as_tuple(ir.Pragma(keyword='loki', content=f'update host({var})'))
+
+    @staticmethod
+    def create_aliased_ptr_assignment(ptr, alias):
```
reuterbal, line 628: I can't find where this is used, seems not covered by tests either.

loki/transformations/data_offload/offload_deepcopy.py
```diff
@@ -346,3 +356,429 @@ def transform_module(self, module, **kwargs): # pylint: disable=unused-argument
...
+        elif analysis == 'readwrite':
+            access_mode = FieldAPITransferType.READ_WRITE
+        else:
+            access_mode = FieldAPITransferType.WRITE_ONLY
+
+        device = as_tuple(field_get_device_data(field_object, field_ptr, access_mode, scope))
+        device += self.enter_data_attach(field_ptr)
+        host = as_tuple(field_get_host_data(field_object, field_ptr, FieldAPITransferType.READ_WRITE, scope))
+        wipe = self.exit_data_detach(field_ptr)
+        wipe += as_tuple(field_delete_device_data(field_object, scope))
```
reuterbal, line 669: I _love_ the fact that we can simply re-use these utilities here!

loki/transformations/data_offload/offload_deepcopy.py
```diff
@@ -346,3 +356,429 @@ def transform_module(self, module, **kwargs): # pylint: disable=unused-argument
...
+           so we create a dummy typedef config here."""
+
+        if self.field_array_match_pattern.match(parent.type.dtype.typedef.name.lower()):
+            typedef_config = {
+                'field_prefix': 'F_',
+                'field_ptr_suffix': '_FIELD',
+                'field_ptr_map': {}
+            }
+            return typedef_config
+        return None
```
reuterbal, line 688: Not covered by tests.

reuterbal (changes requested):
A fantastic piece of work! Despite its complexity I found the overall control flow and logic understandable, which attests to how well this is structured. The small utilities and conceptual rather than semantical nomenclature really helps here. 

Particularly pleasing was to see the re-use of field_api concepts from the field_api transformation, which should help keep these things in sync and up-to-date!

There are a few minor but at least some of them important to fix comments about docstrings, and a few more loose remarks that I left here and there. Also 2-3 places that are not covered by tests at the moment. But overall nothing major that I could find!

loki/transformations/data_offload/offload_deepcopy.py
awnawab: I tried to make it general but it's possible I've missed something. I'll add a test for this and update as necessary.

awnawab:
Thanks a lot @reuterbal for the feedback 🙏 I've addressed all your comments, please have another look when you can 😄

reuterbal (approved):
Everything has been addressed, many thanks!

PR #575, opened 2025-06-29
Various `SCCHoist` fixes

awnawab:
A bugfix PR that contributes the following fixes:
- Variable kind specifiers are hoisted as necessary with temporaries
- `get_integer_variable` can now also return a type-bound integer
- Pointers/allocatable temporaries are no longer hoisted
- Already hoisted temporaries are now filtered by name, to avoid repetitions when bounds aliases are used
- The finalize clause is added to the device deletion of hoisted temporaries. This shouldn't be necessary, and likely points to an nvhpc openacc runtime bug.
- SCCRevector can now also used dimension configs with bounds declared in upper-case

### DIFF
--- a/loki/transformations/pragma_model.py
+++ b/loki/transformations/pragma_model.py
@@ -112,6 +112,10 @@ class OpenACCPragmaMapper(GenericPragmaMapper):
         if param_detach := parameters.get('detach'):
             content += f' detach({param_detach})'
         if content:
+            # Rather than simply decrementing the dynamic reference counter,
+            # finalize forces it to zero. This isn't needed for OpenMP, where
+            # target exit data map(delete:<>) statement already sets the
+            # dynamic reference counter to 0
             final = ' finalize' if 'finalize' in parameters else ''
             return Pragma(keyword='acc', content=f'exit data{content}{final}')
         return self.default_retval()
--- a/loki/transformations/single_column/hoist.py
+++ b/loki/transformations/single_column/hoist.py
@@ -68,7 +68,7 @@ class SCCHoistTemporaryArraysTransformation(HoistVariablesTransformation):
         vnames = ', '.join(v.name for v in variables)
         if vnames:
             pragma = ir.Pragma(keyword='loki', content=f'unstructured-data create({vnames})')
-            pragma_post = ir.Pragma(keyword='loki', content=f'end unstructured-data delete({vnames})')
+            pragma_post = ir.Pragma(keyword='loki', content=f'end unstructured-data delete({vnames}) finalize')
 
             # Add comments around standalone pragmas to avoid false attachment
             routine.body.prepend((ir.Comment(''), pragma, ir.Comment('')))
--- a/loki/transformations/single_column/hoist.py
+++ b/loki/transformations/single_column/hoist.py
@@ -68,6 +68,9 @@ class SCCHoistTemporaryArraysTransformation(HoistVariablesTransformation):
         vnames = ', '.join(v.name for v in variables)
         if vnames:
             pragma = ir.Pragma(keyword='loki', content=f'unstructured-data create({vnames})')
+            # Rather than simply decrementing the dynamic reference counter,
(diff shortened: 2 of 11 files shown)


loki/transformations/temporaries/hoist_variables.py
```diff
@@ -296,6 +305,9 @@ def transform_subroutine(self, routine, **kwargs):
         missing_imports_map = defaultdict(set)
         import_map = routine.import_map
         for module, var in item.trafo_data[self._key]["imported_sizes"]:
+            if not var.name in import_map:
+                missing_imports_map[module.name] |= {var}
+        for module, var in item.trafo_data[self._key]["imported_kinds"]:
             if not var.name in import_map:
                 missing_imports_map[module] |= {var}
```
reuterbal, line 312: Why do we have here `module.name` for sizes but `module` for kinds? It seems like there may be an incosistent representation, using `Module` as a key in one, and `str` in the other. I would suggest we try to use the same for both, with a preference for `str` if possible.

loki/transformations/single_column/hoist.py
```diff
@@ -68,7 +68,7 @@ def driver_variable_declaration(self, routine, variables):
         vnames = ', '.join(v.name for v in variables)
         if vnames:
             pragma = ir.Pragma(keyword='loki', content=f'unstructured-data create({vnames})')
-            pragma_post = ir.Pragma(keyword='loki', content=f'end unstructured-data delete({vnames})')
+            pragma_post = ir.Pragma(keyword='loki', content=f'end unstructured-data delete({vnames}) finalize')
```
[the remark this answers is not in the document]
MichaelSt98: It will only be translated/used for OpenACC: https://github.com/ecmwf-ifs/loki/pull/564

Strongly agree that we should document and explain this.

loki/transformations/temporaries/hoist_variables.py
```diff
@@ -159,20 +163,25 @@ def transform_subroutine(self, routine, **kwargs):
...
 
+            # We may call a subroutine again with aliased sizes, so we check hoisted
+            # variables in children by name before adding them
+            hoist_var_names = [v.name.lower() for v in item.trafo_data[self._key]["hoist_variables"]]
+
             arg_map = dict(call_map[child.local_name].arg_iter())
             hoist_variables = []
             for var in child.trafo_data[self._key]["hoist_variables"]:
+                if var.name.lower() in hoist_var_names:
+                    continue
```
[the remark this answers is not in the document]
awnawab: The only reason we need is to filter out temporaries already hoisted from another child. Each child itself can't have duplicate names in the temporary list since we can't have duplicated dummy arguments. I'm happy to update the `hoist_var_names` list but it shouldn't affect the correctness of the transformation.

loki/transformations/temporaries/hoist_variables.py
awnawab: The definitions containing the sizes are typically enriched, so we look up the module there via `var.type.module`. For kinds, the symbols are typically deferred, so I look up the module via the import which is a string. For consistency I can amend it so that the `imported_sizes` dict also stores the module as a string.

reuterbal: OK, that matches what I suspected. If there's no explicit reason that makes using the module itself necessary, then I think using the name as a string throughout would make sense.

loki/transformations/pragma_model.py
```diff
@@ -112,6 +112,10 @@ def pmap_end_unstructured_data(self, pragma, parameters, **kwargs):
         if param_detach := parameters.get('detach'):
             content += f' detach({param_detach})'
         if content:
+            # Rather than simply decrementing the dynamic reference counter,
+            # finalize forces it to zero. This isn't needed for OpenMP, where
+            # target exit data map(delete:<>) statement already sets the
+            # dynamic reference counter to 0
```
MichaelSt98, line 118: ❤️

mlange05 (approved):
Looks great, many thanks. GTG from me. :shipit:

PR #577, opened 2025-07-02
New more stable implementation of ``find_driver_loops`` using IR visitor

wertysas:
This PR updates the implementation ``find_driver_loops`` utility to support more complex scenarios where pragma-marked driver loops and loops with target kernels are mixed in the IR section and possibly multiply nested. The new implementation of ``find_driver_loops`` uses the following rule to determine which loops are driver loops:

*If there are nested loops, then the highest level pragma-marked driver loop will take precedence and be considered the driver loop. If there is no pragma-marked driver loop, then the highest level loop that does not contain a driver loop will be considered the driver loop.*


loki/transformations/utilities.py
```diff
@@ -623,20 +639,94 @@ def find_driver_loops(section, targets):
         List of subroutines that are to be considered as part of
         the transformation call tree.
     """
+    class FindDriverLoops(Visitor):
```
reuterbal, line 642: Since the class is defined insided the routine, you don't have to pass the `targets` value as an argument but can directly use it. Best to make sure it's an iterable first, though:
```suggestion
    targets = [str(t).lower() for t in as_tuple(targets)]

    class FindDriverLoops(Visitor):
```

loki/transformations/utilities.py
```diff
@@ -623,20 +639,94 @@ def find_driver_loops(section, targets):
...
-    driver_loops = []
-    nested_driver_loops = []
-    for loop in FindNodes(ir.Loop).visit(section):
-        if loop in nested_driver_loops:
-            continue
+        def visit_tuple(self, o, **kwargs):
+            nested_pragma_loop = False
+            for i in o:
+                nested_pragma_loop |= self.visit(i, **kwargs)
+            return nested_pragma_loop
```
reuterbal, line 660: Not a big fan of the mixing of return values via pass-by-reference lists and actual return value. I prefer returning a tuple of the multiple return values, with explicit combination of the values where required, such as here for the tuple:

```suggestion
        @classmethod
        def default_retval(cls):
            return False, [], []

        def visit_tuple(self, o, **kwargs):
            nested_pragma_loop = False
            nested_target_loops = []
            nested_target_calls = []
            for i in o:
                retval = self.visit(i, **kwargs)
                nested_pragma_loop |= retval[0]
                nested_target_loops += retval[1]
                nested_target_calls += retval[2]
            return nested_pragma_loop, nested_target_loops, nested_target_calls
```

loki/transformations/utilities.py
```diff
@@ -623,20 +639,94 @@ def find_driver_loops(section, targets):
...
-        if not is_driver_loop(loop, targets):
-            continue
+        visit_list = visit_tuple
+
+        def visit_CallStatement(self, call, **kwargs):
+            targets = kwargs.get('targets', [])
+            has_target_calls = kwargs.get('has_target_calls', None)
+            if has_target_calls is not None and call.name in targets:
+                has_target_calls.append(True)
+            return False
```
reuterbal, line 669: With the above addition to ensure `targets` is a tuple, you can use it here directly and return the triplet as suggested:
```suggestion
        def visit_CallStatement(self, call, **kwargs):
            if call.name in targets:
                return False, [], [call]
            return self.default_retval()
```

loki/transformations/utilities.py
```diff
@@ -623,20 +639,94 @@ def find_driver_loops(section, targets):
...
+
+        def visit_object(self, o, **kwargs):
+            return False
```
reuterbal, line 672: It's cleaner to re-define the default return value, as suggested above, making this redundant
```suggestion
```

loki/transformations/utilities.py
```diff
@@ -623,20 +639,94 @@ def find_driver_loops(section, targets):
...
+            nested_target_loops = kwargs.get('nested_target_loops', [])
+            if depth == 0:  # root loop
+                nested_target_loops = []
+                nested_pragma_loop = self.__visit_loop(loop, depth, targets, nested_target_loops)
+                if not nested_pragma_loop:
+                    self.driver_loops.extend(nested_target_loops)
+                return nested_pragma_loop
+
+            # else nested loop
+            return self.__visit_loop(loop, depth, targets, nested_target_loops)
```
reuterbal, line 725: This appears unnecessarily complicated to me. Since you are already introducing the `depth` as a kwarg (which is a good idea to identify whether you're inside a loop next), there is no need for an additional dispatch routine.

Moreover, the control flow could be much simplified if I'm understanding this correctly, because you're essentially declaring a loop as a driver loop whenever it has:

1. An explicit driver pragma
2. A call to a target somewhere under the current loop, unless there is an explicit pragma-marked loop nested inside

So, all you need is the information that you are already collecting, i.e., 

1. whether there was a pragma-marked loop (`nested_pragma_loop` return value)
2. whether there are any target calls nested inside the loop nest (`nested_target_calls` return value)

And using the `depth` property you can decide when the foot of the recursion is reached and amend the `driver_loops` property with the target_loops.

Here's a possible implementation that passes the tests for me:

```suggestion
        def visit_Loop(self, loop, **kwargs):
            depth = kwargs.pop('depth', 0)
            if is_pragma_driver_loop(loop):
                self.driver_loops.append(loop)
                # Propagate the presence of the pragma only if inside a loop nest
                return depth > 0, [], []

            # Recurse into the (potential) loop nest
            nested_pragma_loop, nested_target_loops, nested_target_calls = self.visit(loop.body, depth=depth+1, **kwargs)

            if nested_target_calls:
                if nested_pragma_loop:
                    warning("[Loki::find_driver_loops] Nested pragma marked driver loop inside loop"
                            f" with target call (skipping {loop}")
                else:
                    # If there is a target call nested inside the loop nest, the current loop
                    # is a potential driver loop (unless a pragma-annotated loop is present,
                    # which is why the target loops are collected into a list first and only
                    # added to self.driver_loops once we're back at depth==0)
                    nested_target_loops += [loop]

            if depth == 0:
                if nested_target_loops:
                    self.driver_loops.extend(nested_target_loops)
                return self.default_retval()

            return nested_pragma_loop, nested_target_loops, nested_target_calls
```

loki/transformations/utilities.py
```diff
@@ -623,20 +639,94 @@ def find_driver_loops(section, targets):
...
+            targets = kwargs.get('targets', [])
+            has_target_calls = kwargs.get('has_target_calls', None)
+            if has_target_calls is not None and call.name in targets:
+                has_target_calls.append(True)
+            return False
+
+        def visit_object(self, o, **kwargs):
+            return False
+
+        def __visit_loop(self, loop, depth, targets, target_loops):
```
reuterbal, line 674: This method becomes redundant with the above suggestion

wertysas:
> Many thanks, this is already a much more straightforward implementation of this lookup. The pattern is largely correct but looking at the code I felt that this must be possible in an even more straightforward fashion, which is why I played around with this myself and came up with a potential implementation. See inline comments for details but in essence this comes down to
> 
> 1. explicitly treating the return values as return values (instead of implicit pass-by-reference return valeus)
> 2. Using the depth-attribute to handle the base case, instead of introducing a second visitor entry point.
> 
> I hope this is sufficiently understandable and feel free to reach out in case of questions!

@reuterbal Thank you for looking at it and for taking the time to rewrite the algorithm more cleanly with multiple return values instead!

I agree, with your comment, I was afraid that the multiple return values would make the function more complicated and wanted to keep the implementation as close as possible to the other visitors in Loki, that generally use the "kwargs accumulator pattern" for return values. However, I think your implementation with multiple return values is cleaner and much easier to follow.

I have updated the PR with your changes after a slight modification. When using multiple return values  instead of a kwargs accumulator, then the ``nested_target_calls`` list could be replaced with a boolean ``has_target_call`` which is what the variable actually is tracking, making the algorithm even simpler to follow.

I also added one more complex loop nest to the test, just to cover all edge cases.

loki/transformations/tests/test_utilities.py
```diff
@@ -469,6 +470,128 @@ def test_transform_utilites_find_driver_loops(frontend):
...
+      do k=start,end ! driver loop 7 (loop 17)
+        do l=start,end
+          call target_kernel(arr(l))
+        end do
+      end do
+      !$loki driver-loop
+      do k=start,end       ! driver loop 7 (loop 19)
+        arr(i) = 2. * arr(i)
+      end do
+      do k=start,end ! driver loop 7 (loop 20)
```
reuterbal, line 554:
```suggestion
      do k=start,end ! driver loop 7 (loop 17)
        do l=start,end
          call target_kernel(arr(l))
        end do
      end do
      !$loki driver-loop
      do k=start,end       ! driver loop 8 (loop 19)
        arr(i) = 2. * arr(i)
      end do
      do k=start,end ! driver loop 9 (loop 20)
```

reuterbal (approved):
Thanks, looks good to me now!

mlange05 (approved):
Excellent, this looks very neat now, given the complexity of the logic in this little utility. Test looks extensive too, so GTG from me.

PR #581, opened 2025-07-09
Add test for correct driver loop annotation 

### LINKED ISSUE #246: SCC: Loop over blocks not annotated correctly when nested within another loop in driver
Hello, filing an issue by request of @reuterbal.

Here is a reproducer of the problem:
```
module mod
    implicit none
    contains
    subroutine wrapper()
        integer, parameter :: ngpblk = 2
        integer, parameter :: klon = 10
        integer, parameter :: klev = 5 
        integer, parameter :: niter = 10
        integer :: jblk, jiter
        integer :: arr(klon, klev, ngpblk)

        ! Begin relevant part.
        DO jiter=1,niter
            DO jblk=1,ngpblk
                CALL kernel(klon, klev, 1, 1, arr(:, :, jblk))
            END DO
        END DO
        ! End relevant part.

    end subroutine wrapper

    subroutine kernel(klon, klev, jlev_lower, jlon_lower, arr)
        integer, intent(in) :: klon
        integer, intent(in) :: klev
        integer, intent(in) :: jlev_lower, jlon_lower
        integer, intent(inout) :: arr(klon, klev)
        integer :: jlon, jlev

        DO jlev=jlev_lower,klev
            DO jlon=jlon_lower,klon
                arr(jlon, jlev) = jlon + jlev
            END DO
        END DO

    end subroutine kernel

end module mod
```
After transformin with SCC, the relevant part becomes:
```
!$acc parallel loop gang vector_length( KLON )
    DO jiter=1,niter
      DO jblk=1,ngpblk
        CALL kernel_SCC(klon, klev, 1, 1, arr(:, :, jblk))
      END DO
    END DO
!$acc end parallel loop
```

Although I believe this should be
```
    DO jiter=1,niter
!$acc parallel loop gang vector_length( KLON )
      DO jblk=1,ngpblk
        CALL kernel_SCC(klon, klev, 1, 1, arr(:, :, jblk))
      END DO
!$acc end parallel loop
    END DO
```
instead.

I invoke the SCC transform with: `loki-transform.py convert --mode scc --config config.cfg --path src --out-path scc`,
where the contents of "config.cfg" are:
```
[default]
role = 'kernel'
expand = true  # Automatically expand subroutine calls
strict = true # Throw exceptions during discovery
enable_imports = true

[routines.wrapper]
role = 'driver'

[dimensions.horizontal]
size = 'KLON'
index = 'JLON'
bounds = ['jlon_lower', 'klon']

[dimensions.vertical]
size = 'KLEV'
index = 'JLEV'

[dimensions.block_dim]
size = 'NGPBLK'
index = 'JBLK'
```

I believe the issue is coming from `SCCAnnotate.process_driver`. I did some digging and I found out that the method calls `SCCBaseTransformation.find_driver_loops`, which returns the outer loop in the above example.
Should this function return the _inner_ loop instead? (as the loops get passed to `annotate_driver`)?

reuterbal:
This fixes #246 via #577. Thanks to @mlange05 for pointing this out.

### DIFF
--- a/loki/transformations/single_column/tests/test_scc.py
+++ b/loki/transformations/single_column/tests/test_scc.py
@@ -5,6 +5,8 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+# pylint: disable=too-many-lines
+
 import pytest
 
 from loki import Subroutine, Sourcefile, Dimension, fgen
--- a/loki/tests/test_subroutine.py
+++ b/loki/tests/test_subroutine.py
@@ -141,7 +141,7 @@ end subroutine my_routine
 @pytest.mark.parametrize('frontend', available_frontends(
     xfail=[(OMNI, 'OMNI frontend interface does not provide interfaces')]
 ))
-def test_routine_bind(frontend):
+def test_routine_bind(frontend, tmp_path):
     """ Test matching of 'bind" suffix for subroutines in interfaces """
     fcode = """
 module my_module
@@ -164,7 +164,7 @@ contains
 end module my_module
 """
     # Note that Fparser fails here if the legal 'recursive' prefix is used!
-    module = Module.from_source(fcode, frontend=frontend)
+    module = Module.from_source(fcode, frontend=frontend, xmods=[tmp_path])
 
     routine = module['my_routine']
 
--- a/loki/transformations/single_column/tests/test_scc_vector.py
+++ b/loki/transformations/single_column/tests/test_scc_vector.py
@@ -88,7 +88,7 @@ def test_scc_revector_transformation(frontend, horizontal, revector_trafo, ignor
         q(jl, jk) = q(jl, jk-1) + t(jl, jk) * c
       END DO
     END DO
-    
+
     ! The scaling is purposefully upper-cased
     DO JL = START, END
       Q(JL, NZ) = Q(JL, NZ) * C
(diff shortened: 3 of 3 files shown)


loki/transformations/single_column/tests/test_scc.py
```diff
@@ -1407,3 +1407,92 @@ def test_scc_resolve_vector_dimension(frontend, horizontal):
...
+
+    source = Sourcefile.from_source(fcode, frontend=frontend, xmods=(tmp_path,))
+
+    horizontal = Dimension('horizontal', index='JLON', bounds=['jlon_lower', 'klon'], size='KLON')
+    block_dim = Dimension('block_dim', index='JBLK', size='NGPBLK')
+
+    # Test OpenACC annotations on non-hoisted version
+    scc_transform = (SCCDevectorTransformation(horizontal=horizontal),)
+    scc_transform += (SCCDemoteTransformation(horizontal=horizontal),)
+    scc_transform += (SCCRevectorTransformation(horizontal=horizontal),)
```
mlange05, line 1466: I appreciate this was written a while ago, but I think we can now use the respective `SCCVectorPipeline` (or equivalent) to make this a single invocation line.

mlange05 (changes requested):
I think this can be made more concise by using the new pipeline semantics. Optionally, we can probably even parametrize a cross a few of these by now. 😉

reuterbal:
Thanks for the review! I have made the requested change and also added missing tmp_path use for xmods in two places after noticing some stray artifacts after a test run.

loki/transformations/single_column/tests/test_scc.py
```diff
@@ -5,6 +5,8 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+# pylint: disable=too-many-lines
```
mlange05, line 8: 😞

mlange05 (approved):
Thanks for addressing this. GTG now. :shipit:

PR #586, opened 2025-07-24
REGEX frontend: Correctly handle nested parentheses/brackets

### LINKED ISSUE #585: Regex frontend parses call statements with brackets in arguments incorrectly
### What happened?

The regex frontend parses the name of a subroutine in a call statement wrong when an array argument defined inside brackets is part of the arguments. (see reproducer below)

I have tracked the issue down to the class method ``_remove_quoted_string_nested_parentheses`` from ``regex.py``, which in this case is called on line 982 of ``regex.py`` in ``CallPattern.match``.

### What are the steps to reproduce the bug?

The following code can be used to reproduce the bug:

```python
from loki.subroutine import Subroutine
from loki.frontend import Frontend
from loki import fgen

fcode = """
    subroutine a_function()
        call parse_me_wrong(arg1, arg2=[1,2,3], arg3)
        call parse_me_wrong2(arg1, arg2=\(1,2,3\), arg3)
    end subroutine a_function
"""

sub = Subroutine.from_source(fcode, frontend=Frontend.REGEX)
print(fgen(sub))
```

The name of the ``ProcedureSymbol`` of the first ``CallStatement`` is then parsed incorrectly to ``parse_me_wrong,arg3)``.

However, for the second ``CallStatement`` where ``\(\)`` is used for the default argument list the name is parsed correctly.

**Output when executing the reproducer above:**
```
SUBROUTINE a_function ()
  CALL parse_me_wrong2()
  CALL parse_me_wrong,arg3)()
END SUBROUTINE a_function
```

### Version

main

### Platform (OS and architecture)

Mac OS and Linux (no reason to believe this is OS-specific)

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

reuterbal:
Fixes #585, which reported an edge case in the removal of parentheses and brackets, which occurs when the last parenthesis/bracket is actually nested inside another pair. Thanks to @wertysas for reporting and providing a reproducer.

### DIFF
--- a/loki/frontend/regex.py
+++ b/loki/frontend/regex.py
@@ -340,10 +340,11 @@ class Pattern:
 
         # We should now be left with no parentheses anymore and can build the new string
         # by using everything between these parenthesis "spans"
-        new_string = string[:spans[0][0]]
+        starts, ends = zip(*spans)
+        new_string = string[:min(starts)]
         for (_, start), (end, _) in zip(spans[:-1], spans[1:]):
             new_string += string[start+1:end]
-        new_string += string[spans[-1][1]+1:]
+        new_string += string[max(ends)+1:]
         return new_string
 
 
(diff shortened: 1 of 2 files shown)


wertysas (approved):
This looks good to me. Thank you for the quick fix of the issue!

PR #590, opened 2025-08-14
HoistVariablesAnalysis and Transformation fix

MichaelSt98:
make sure call.routine is not BasicType.DEFERRED.

As discussed offline: suboptimal since I wasn't able to recreate the problem in the Loki tests, since it seems to be a very special case related to inlining a function (and not inlining it in the same module since it has array arguments ...). However, necessary for ecphys to get the seq trafos to work.


awnawab (approved):
Quite a weird edge case that this was tripping over, but the checks for routine enrichment look sound, thanks! G2G 👌

PR #593, opened 2025-08-21
New utility within 'remove_code': remove unused temporaries/locals

MichaelSt98:
optionally only arrays.

Either via `do_remove_unused_vars(routine, remove_only_arrays=True/False)` or via `RemoveCodeTransformation(remove_unused_vars=True, remove_only_arrays=True/False)`.

### DIFF
--- a/loki/transformations/remove_code.py
+++ b/loki/transformations/remove_code.py
@@ -25,7 +25,7 @@ __all__ = [
     'RemoveCodeTransformation',
     'do_remove_dead_code', 'RemoveDeadCodeTransformer',
     'do_remove_marked_regions', 'RemoveRegionTransformer',
-    'do_remove_calls', 'RemoveCallsTransformer'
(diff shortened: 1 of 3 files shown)


MichaelSt98:
> optionally only arrays.
> 
> Either via `do_remove_unused_vars(routine, remove_only_arrays=True/False)` or via `RemoveCodeTransformation(remove_unused_vars=True, remove_only_arrays=True/False)`.

Question whether to introduce `RemoveCodeTransformation(remove_unused_vars=True, remove_only_arrays=True/False)` within the SCC pipelines after revectorization and temporary trafo or how to incorporate this new utility ...

mlange05:
Code looks great and great catch to find this so quickly! Many thanks @MichaelSt98 !

As discussed offline, I don't have a strong preference on where to call this, but I'd be happy with simply adding a similar `remove_unused_arrays` option to both(!) `SCCSeq/VecRevectorTransformations` and call the utility as the last step of kernel re-vectorisation.

PR #597, opened 2025-08-28
SCC: Apply Annotate after RemoveUnusedVars

reuterbal:
The order was wrong in a few pipelines, which could lead to spurious declarations of privatised variables during the annotate step, that are then removed subsequently.


MichaelSt98 (approved):
Makes more sense like that ... thanks! 👍

PR #601, opened 2025-09-04
replace ancient "toml" with newer "tomli"/"tomli_w" couple

a-detiste:
Hi,

This replace usage of the old `toml` library from the Python2 era with the newer & maitained `tomli` & `tomli_w` couple.

https://wiki.debian.org/Python/Backports

Later when support for obsolete Python3 versions is dropped, `tomli` can be replaced with builtin ``tomllib`


a-detiste:
I'm not sure this magical `Commit Suggestion` did actualy commited something

PR #602, opened 2025-09-04
resolve vector notation: fix for inline calls

MichaelSt98:
(that are assumed to be array accesses)

Fixes the problem for CLOUDSC Loki SCC CUF variants.


reuterbal (approved):
Many thanks for this!

PR #605, opened 2025-09-19
Only privatise arrays defined in vector loops

awnawab:
Rather than privatising all local static arrays, this PR updates the `SCCAnnotateTransformation` so that only arrays defined within a vector loop are marked as private.


MichaelSt98 (approved):
Looks good to me!

PR #606, opened 2025-09-24
Don't offload driver-loop scalars scalars

awnawab:
Driver-loop level scalars should not be copied to device or pulled back, which would override the default thread-privatisation of such scalars.


mlange05 (approved):
LGTM! 👍

PR #607, opened 2025-09-25
Gang privatisation of array components of derived-types

awnawab:
If the encompassing derived-type is being declared as gang private in the driver loop, then its array components should not be privatised individually.


mlange05 (approved):
Looks good to me. 👍

PR #610, opened 2025-09-30
Modify dataflow analysis for kind specifiers

awnawab:
Consider the following snippet:
```
use iso_fortran_env, only : real64
real(kind=real64) :: a

a = 0._real64
```

Previously, `real64` would be included in `routine.body.uses_symbols` and not in `routine.spec.uses_symbols`. I think it should be the other way around, as we are only genuinely reading from the symbol `real64` in the variable declaration, and in the body it is only being used to qualify the size of the literal 0.

This PR addresses the above.

### DIFF
--- a/loki/analyse/tests/test_analyse_dataflow.py
+++ b/loki/analyse/tests/test_analyse_dataflow.py
@@ -11,6 +11,7 @@ from loki import Module, Sourcefile, Subroutine
 from loki.analyse import (
     dataflow_analysis_attached, read_after_write_vars, loop_carried_dependencies
 )
+from loki.analyse.analyse_dataflow import DataflowAnalysisAttacher, DataflowAnalysisDetacher
 from loki.backend import fgen
 from loki.expression import symbols as sym
 from loki.frontend import available_frontends, OMNI
@@ -169,7 +170,8 @@ end subroutine analyse_read_after_write_vars
 
 
 @pytest.mark.parametrize('frontend', available_frontends())
-def test_read_after_write_vars_conditionals(frontend):
+@pytest.mark.parametrize('include_literal_kinds', [True, False])
+def test_read_after_write_vars_conditionals(frontend, include_literal_kinds):
     fcode = """
 subroutine analyse_read_after_write_vars_conditionals(a, b, c, d, e, f)
   use iso_fortran_env, only : int32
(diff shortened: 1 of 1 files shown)


loki/analyse/analyse_dataflow.py
```diff
@@ -176,6 +177,11 @@ def visit_Conditional(self, o, **kwargs):
         query_args = as_tuple(flatten(FindVariables().visit(i.parameters) for i in mem_call))
         cset = set(v for v in FindVariables().visit(o.condition) if not v in query_args)
 
+        # Filter out any symbols used to qualify literals e.g. 0._JPRB
+        literals = FindLiterals().visit(o.condition)
+        literal_vars = FindVariables().visit(literals)
+        cset -= set(literal_vars)
+
```
reuterbal, line 184: I'm not sure I agree with this. After all, it is a use of the symbol and transplanting a part of code that uses this into a different region without the matching import would cause problems. That's something that one might then try to resolve by searching for all used symbols, which now isn't possible anymore.

reuterbal (commented):
I agree with the spec change but I dont' think we should remove the symbol from literals in the body

loki/analyse/analyse_dataflow.py
awnawab: Ah but we could be writing to an imported symbol, in which case we wouldn't have the declaration. Hmm. This bites me in the deepcopy since the deepcopy now thinks it has to offload `JPRB`. An easy workaround would be to exclude all parameters, but these kind specifiers are typically deferredtype symbols.

Could I make this behaviour optional, off by default?

reuterbal: You mean, having something like `include_literal_kinds=True` as an option to the dataflow analysis? I think that sounds like a good idea to me

reuterbal: You mean, having something like `include_literal_kinds=True` as an option to the dataflow analysis? I think that sounds like a good idea to me

loki/analyse/analyse_dataflow.py
```diff
@@ -176,6 +183,12 @@ def visit_Conditional(self, o, **kwargs):
         query_args = as_tuple(flatten(FindVariables().visit(i.parameters) for i in mem_call))
         cset = set(v for v in FindVariables().visit(o.condition) if not v in query_args)
 
+        if not self.include_literal_kinds:
+            # Filter out any symbols used to qualify literals e.g. 0._JPRB
+            literals = FindLiterals().visit(o.condition)
+            literal_vars = FindVariables().visit(literals)
+            cset -= set(literal_vars)
```
reuterbal, line 190: According to codecov this is untested

loki/analyse/analyse_dataflow.py
```diff
@@ -210,7 +223,13 @@ def visit_MultiConditional(self, o, **kwargs):
 
     def visit_MaskedStatement(self, o, **kwargs):
         live = kwargs.pop('live_symbols', set())
+
         conditions = self._symbols_from_expr(o.conditions)
+        if not self.include_literal_kinds:
+            # Filter out any symbols used to qualify literals e.g. 0._JPRB
```
reuterbal, line 229: According to codecov this is untested

reuterbal (changes requested):
Would it be possible to make sure the marked code regions are included in the testing, e.g., by parametrizing test(s) accordingly?

reuterbal (approved):
All good now, many thanks!

PR #613, opened 2025-10-08
fix resolve associates

MichaelSt98:
edge case with some ambiguity wasn't resolved correctly:

```fortran
  ASSOCIATE (ndim=>some_obj%ndim, nested_obj=>other_obj%nested)
    CALL some_func(nested_obj%ndim)
  END ASSOCIATE
```

do_resolve_associates ...

```fortran
CALL some_func(some_obj%ndim)
```

which is obviously wrong and fixed with this PR to produce 

```fortran
CALL some_func(other_obj%nested%ndim)
```


mlange05 (approved):
Excellent! GTG for me as is. :shipit: 

Many thanks for finding and fixing this. 🙏

reuterbal (approved):
Looks good to me, too!

PR #614, opened 2025-10-13
PragmaModel: rename closing unstructured pragmas to "!$loki exit unstructured"

awnawab:
In the `PragmaModelTransformation`, unstructured offload pragmas are specified as:
`!$loki unstructured ...
 !$loki end unstructured ...
`

This leads loki to incorrectly mark these as candidates for pragma region start and end points, and then when during pragma region attachment it fails to find matching end pragmas at the same level in the IR, a warning is thrown. Whilst this is safe, the output of some transformation pipelines is now incredibly verbose.

The current PR aims to address this by renaming the closing unstructured pragma to `!$loki exit-unstructured`.

### DIFF
--- a/docs/source/loki_pragma_model.csv
+++ b/docs/source/loki_pragma_model.csv
@@ -2,7 +2,7 @@ Loki,OpenACC,OMP-GPU
 ``create device(...)``,``declare create(...)``,``declare target(...)``
 ``update device(...) host(...)``,``update device(...) self(...)``,
 ``unstructured-data in(...) create(...) attach(...)``,``enter data copyin(...) create(...) attach(...)``,``target enter data map(to: ...) map(alloc: ...)``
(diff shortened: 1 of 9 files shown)


loki/transformations/pragma_model.py
```diff
@@ -103,7 +103,7 @@ def pmap_unstructured_data(self, pragma, parameters, **kwargs):
             return Pragma(keyword='acc', content=f'enter data{content}')
         return self.default_retval()
 
-    def pmap_end_unstructured_data(self, pragma, parameters, **kwargs):
+    def pmap_exit_unstructured_data(self, pragma, parameters, **kwargs):
```
reuterbal: Does this mean `end unstructured-data` will no longer work, i.e., breaking backwards-compatibility? I'm just wondering if we have used this somewhere that needs adapting in source.

reuterbal (approved):

PR #619, opened 2025-10-17
Pool allocator outsourcing offload to an external module

MichaelSt98:
Instead of offloading the stack/pool via pragmas, rely on an external module (which allows re-usage for multiple (subsequent) kernels).

### DIFF
--- a/loki/transformations/temporaries/pool_allocator.py
+++ b/loki/transformations/temporaries/pool_allocator.py
@@ -947,6 +947,38 @@ class EcstackPoolAllocatorTransformation(TemporariesPoolAllocatorTransformation)
     Analog to :any:`TemporariesPoolAllocatorTransformation`, however, instead of
     inserting offload pragmas use an external defined module to get a pointer to
     an offloaded chunk of memory.
+
+    The minimal interface expected from ECSTACK should look like:
+
+    .. code-block:: fortran
+
+        MODULE ECSTACK_MOD
+        
+        IMPLICIT NONE
+        
+        TYPE TECSTACK
+          ...
+        CONTAINS
+          PROCEDURE :: GET_STACK_PTR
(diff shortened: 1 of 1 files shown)


loki/transformations/temporaries/pool_allocator.py
```diff
@@ -205,6 +207,7 @@ def __init__(
         self.directive = directive
         self.check_bounds = check_bounds
         self.cray_ptr_loc_rhs = cray_ptr_loc_rhs
+        self.stack_size_var_type = stack_size_var_type or self.stack_int_type_kind
```
reuterbal: To avoid the misnomer, we could store here directly a type instead of the kind:

```suggestion
        self.stack_size_var_type = stack_size_var_type or SymbolAttributes(BasicType.INTEGER, kind=self.stack_int_type_kind)
```

loki/transformations/temporaries/tests/test_pool_allocator.py
```diff
@@ -140,12 +143,34 @@ def test_pool_allocator_temporaries(tmp_path, frontend, generate_driver_stack, b
             ylstack_l = loc(zstack(1, b))
             ylstack_u = ylstack_l + istsz * c_sizeof(real(1, kind={kind_stack}))
         """
-    fcode_stack_dealloc = "DEALLOCATE(ZSTACK)"
+    fcode_stack_dealloc = "DEALLOCATE(ZSTACK)" if trafo else ''
```
reuterbal: Shouldn't this be

```suggestion
    fcode_stack_dealloc = "DEALLOCATE(ZSTACK)" if if trafo == TemporariesPoolAllocatorTransformation else ''
```

reuterbal (approved):
Looks great, many thanks!

PR #620, opened 2025-10-23
CLI: Move "scripts" into a dedicated CLI sub-package and share common options

mlange05:
Initial housekeeping PR that internalises the previous `scripts` sub-directory to `loki.cli` and factors out common options into a common utility classes/decorators. The two common decorators combine:
* `FrontendOptions` - `--frontend` and `--cpp`, including all the necessary paths for preprocessing
* `SchedulerOptions` - `--build`, `--source` and `--header`

It also adds the use of `click-option-group` to improve the helper string in the CLI interfaces (with actual grouping).

Note, this is only the first step in an longer-standing attempt to improve the entry points, but let's take this step-by-step.

### DIFF
--- a/loki/cli/common.py
+++ b/loki/cli/common.py
@@ -61,7 +61,8 @@ def frontend_options(func):
     @optgroup.option('--cpp/--no-cpp', default=False,
                      help='Trigger C-preprocessing of source files.')
     @optgroup.option('--include', '-I', type=click.Path(), multiple=True,
-                     help='Path for additional header file(s)')
+                     help='Path for header file(s) that provide type '
+                     'information without being part of the call tree')
     @optgroup.option('--define', '-D', multiple=True,
                      help='Additional symbol definitions for the C-preprocessor')
     @optgroup.option('--xmod', '-M', type=click.Path(), multiple=True,


loki/cli/loki_transform.py
reuterbal: With the scripts internalized into the package they shouldn't be invoked directly anymore anyway - the relevant script entry points should be created by setuptools during installation. Thus we should remove these hooks (also from loki_lint.py, please).

loki/cli/common.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    @optgroup.group('Loki frontend options',
+                    help='Frontend parsing options for Loki.')
+    @optgroup.option('--frontend', default='fp', type=click.Choice(['fp', 'ofp', 'omni']),
+                     help='Frontend parser to use (default FP)')
+    @optgroup.option('--cpp/--no-cpp', default=False,
+                     help='Trigger C-preprocessing of source files.')
+    @optgroup.option('--include', '-I', type=click.Path(), multiple=True,
+                     help='Path for additional header file(s)')
```
reuterbal: I always found the "header" description not too clear. Maybe this is a good occasion to update the help message with something that says "these will be parsed first to provide interprocedural annotations but they are not included in the transformations."

reuterbal (approved):
Looks great now, thanks!

PR #625, opened 2025-10-27
Introduce multiple 'FieldAPIAccessorType's

MichaelSt98:
via `get_` type-bound and `sget`

### DIFF
--- a/loki/transformations/field_api.py
+++ b/loki/transformations/field_api.py
@@ -56,14 +56,17 @@ class FieldAPIAccessorType(Enum):
         CALL FIELD%GET_HOST/DEVICE_DATA_...()
     """
 
+    GENERIC = 'SGET'
     """
     Create FIELD_API data access calls using a generic interface that
     takes a FIELD as an argument e.g.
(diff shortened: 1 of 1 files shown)


loki/transformations/data_offload/offload_deepcopy.py
```diff
@@ -424,13 +425,16 @@ class DataOffloadDeepcopyTransformation(Transformation):
     ----------
     mode : str
        Transformation mode, must be either "offload" or "set_pointers".
+    accessor_type : str
```
reuterbal: Wouldn't it make sense to also allow a `FieldAPIAccessorType` instance to be provided here?

loki/transformations/field_api.py
```diff
@@ -45,6 +45,15 @@ def suffix(self):
         return self.value
 
 
+class FieldAPIAccessorType(Enum):
```
reuterbal: Can we get some docstring for this please?

reuterbal (changes requested):
Thanks, looks good in principle. Could we please provide documentation for the new Enum and maybe capture the future plans in an issue?

PR #627, opened 2025-11-07
Create ExternalItem also for non-module procedures that cannot be found

reuterbal:
We had an idiosyncracy around the Scheduler behaviour for procedures it cannot find. While module procedures would be added as `ExternalItem` to the scheduler graph, the same wasn't true for procedures declared via an interface. This PR should align the behaviour.


mlange05 (approved):
Looks great and works as intended. Many thanks for the quick fix! 🙏 👍 :shipit:

Issue #630, opened 2025-11-10
Scheduler: support for transient imports


reuterbal:
### Is your feature request related to a problem? Please describe.

Consider transient module imports, e.g., for procedures, derived types or global variables, such as in the following example:
```f90
module intf_mod
    use type_mod, only: derived_type
    use func_mod, only: proc
    use data_mod, only: global_var
    implicit none
end module intf_mod

module type_mod
  implicit none
  type derived_type
    integer :: var
  contains
    procedure :: set_var
  end type derived_type
contains
  subroutine set_var(self, val)
    class(derived_type), intent(inout) :: self
    integer, intent(in) :: val
    self.var = val
  end subroutine set_var
end module type_mod

module func_mod
  implicit none
contains
  subroutine proc
    print *,'hello world'
  end subroutine proc
end module func_mod

module data_mod
  implicit none
  integer :: global_var
end module data_mod

subroutine driver
  use intf_mod, only derived_type, proc, global_var
  implicit none
  type(derived_type) :: my_type
  call my_type%proc(global_var)
  call proc
end subroutine driver
```

The Scheduler will currently fail to follow the indirection via the `intf_mod` from the driver to the various implementation modules (see also #629). 

### Describe the solution you'd like

We should introduce an `ImportItem` to the Scheduler, similarly to the existing `ProcedureBindingItem`. This would allow to encode this indirections, even across multiple levels, and encode any renaming etc.
The reason for this is that, albeit not an actual executable item in a program, this is a definition of an external API in the enclosing (module) scope. More specifically, such `ImportItem` nodes would only be required for publicly imported symbols in a module.

The same should then be added for public Interfaces.

There is then the option to refactor this into a generic indirection node for all situations where an `Item` essentially encodes an edge, i.e., imports, interfaces and procedure bindings.

Issue #629, opened 2025-11-10
Scheduler: Error detecting procedure bindings via transient imports


mlange05:
### What happened?

When a derived type with explicitly defined procedure bindings is imported via a transient module import, the scheduler throws an `AttributeError`, as shown below.

The transient import is of the form:
```
driver :: 
  * use mod_a, only: my_type
  * type(my_type) :: obj
  * call obj%some_routine()

=> mod_a :: use mod_b, only: my_type

=> mod_b :: 
   * type my_type (with type-bound procedure)
   * subroutine some_routine
```

### What are the steps to reproduce the bug?

```
@pytest.mark.parametrize('frontend', available_frontends())
def test_scheduler_transient_imports(frontend, tmp_path):
    """ Test that use transiently imported typedefs succeeds. """

    fcode_mod_a = """
module mod_a
  use mod_b, only: my_type
  implicit none
end module mod_a
"""
    
    fcode_mod_b = """
module mod_b
  implicit none
  type my_type
    real(kind=4) :: a, b, x

    contains
    procedure :: add_a_b => my_type_add_a_b
  end type my_type

contains
  subroutine my_type_add_a_b(obj)
    type(my_type), intent(inout) :: obj

    obj%x = obj%a + obj%b
  end subroutine my_type_add_a_b
end module mod_b
"""

    fcode_driver = """
subroutine test_scheduler()
  use mod_a, only: my_type
  implicit none

  type(my_type) :: d

  d%a = 42.0
  d%a = 66.6
  call d%add_a_b()
end subroutine test_scheduler
"""
    src_path = tmp_path/'src'
    src_path.mkdir()
    (src_path/'mod_a.F90').write_text(fcode_mod_a)
    (src_path/'mod_b.F90').write_text(fcode_mod_b)
    (src_path/'driver.F90').write_text(fcode_driver)

    # Create the Scheduler
    config = SchedulerConfig.from_dict({
        'default': {
            'role': 'kernel', 'expand': True, 'strict': True, 'enable_imports': True
        },
        'routines': {'test_scheduler': {'role': 'driver'}}
    })
    scheduler = Scheduler(
        paths=[src_path], config=config, seed_routines='test_scheduler',
        frontend=frontend, xmods=[tmp_path]
    )
```

### Version

main

### Platform (OS and architecture)

Python 3.11.10

### Relevant log output

```shell
[...]

PR #631, opened 2025-11-10
Bugfix: Do not abort when encountering transient imports

reuterbal:
This is a fix to alleviate the immediate issue reported in #629. The longer term improvement/solution to this is outlined in #630.

The changes in here consist of 
1. testingthe three main situations where this could cause issues at the moment -- transient imports of types, variables and procedures 
2. some checks throughout the dependency derivation to continue without resolving a dependency rather than aborting with an error. Logging these situations to debug.


mlange05 (approved):
LGTM - appreciate the thorough testing! I also used it in anger and now works as intended.  🙏 👍 :shipit:

PR #633, opened 2025-11-10
Fix option to exclude explicitly named variables from block-index injection pipeline

awnawab:
This PR fixes and renames the per-subroutine configurable option to exclude explicitly named variables from the block-index injection pipeline.


reuterbal (approved):
Looks good to me!

PR #635, opened 2025-11-20
AI agents - integration of requests to ollama server

maurinl26:
Creation of 3 requests to an ollama server :
- fortran to python
- fortran to dace
- inserting OpenACC directives in Fortran

The idea is to open the discussion about integrating LLMs into loki workflows


reuterbal:
Thanks for this suggestion. I did not comment on this until now because the PR was still in draft stage and I was waiting for it to be marked ready. But since there hasn't been any activity for more than 6 months, I'm assuming this is not being developed further. The tests are currently not fully developed.

The way this integration is being done relies on backend generation of source code and re-parsing the LLM response to a Loki IR. This intermediate conversion and invocation of backend and frontend from within transformations does not really fit within the conceptual setup of transformation pipelines that Loki builds upon. In particular, since LLM responses are not deterministic, this will yield a different response every time even for the same input, which is something we are working hard to avoid for the rest of Loki. Therefore, I don't consider this suitable for inclusion in the core transformations package in Loki.

Note that you have several options of taking this forward nonetheless, leaving only two suggestions here:

1. You could structure your invocation of Loki + LLM as a two- or three-step workflow: Call a Loki transformation pipeline to preprocess input files and write out Fortran again, then feed this to the LLM, and (optionally) Call Loki again with a second transformation pipeline to carry out any follow-on tasks. That would retain deterministic processing pipelines inside Loki and combine this with LLM use.
2. You can create a standalone transformations package with these transformations and use this inside your transformation pipeline. The pipeline config allows you to specify any import location for transformations and therefore they can come from an external package as well. Take a look at [this test](https://github.com/ecmwf-ifs/loki/blob/main/loki/batch/tests/test_scheduler_config.py#L620) as an example.

Let me know if you have any questions or want further feedback on internals - happy to discuss of course!

maurinl26:
@reuterbal 

I left Météo France few month ago, thanks for the comment.

This shifted as a side project : legacy fortran transformation to JAX or, at least, cython wrapped fortran code.
It is a LLM first transformation, with loki to minimize LLM call for deterministic actions, knowing the agent performs several attempts to translate + compile + test code.

https://github.com/maurinl26/coding-agent.git

If it makes sense somehow :
- a langchain agent + loki integration
- works as python lib (uvx utility) with an Mistral endpoint 
- on prem deploy on a VPS with a GPU enabled (like ECWMF's Cloud or Azure machine)

The idea is to combine the power of loki, or compilator like parsor, with generative power of LLMs.

Happy to discuss of it, eventually a demo at ECMWF could make sense !

PR #637, opened 2025-11-25
Fix spurious dependencies in typedef items due to shared imports (fixes #634)

### LINKED ISSUE #634: Scheduler: Interface imports create erroneous dependencies in scheduler graph
### What happened?

When combining a derived type import with an import of a subroutine interface in a module, a calling routine that only depends on the nested derived type definition from that module gets an erroneous call dependencies to the subroutine implementations behind the interface routine in the scheduler graph. In the provided example, this leads to the `driver` routine depending on `routine_a` and `routine_b` despite not importing or calling them.

### What are the steps to reproduce the bug?

```
def test_scheduler_module_interface_import(frontend, tmp_path):
    """ Test module-level imports of interface routines. """

    fcode_mod_a = f"""
module mod_a
    use mod_b, only: inner_type, intfb_routine
    implicit none
    type outer_type
        type(inner_type) :: da
    end type outer_type
end module mod_a
"""

    fcode_mod_b = """
module mod_b
    implicit none
    type inner_type
        integer :: da, db
    end dtype inner_type

    interface intfb_routine
    module procedure routine_a, routine_b
    end interface intfb_routine

    contains
    subroutine routine_a
    end subroutine routine_a

    subroutine routine_b
    end subroutine routine_b
end module mod_b
"""

    fcode_driver = """
subroutine test_scheduler()
    use mod_a, only: outer_type
    implicit none
    type(outer_type) :: da

    da%da%da = 42.0
end subroutine test_scheduler
"""
    src_path = tmp_path/'src'
    src_path.mkdir()
    (src_path/'mod_a.F90').write_text(fcode_mod_a)
    (src_path/'mod_b.F90').write_text(fcode_mod_b)
    (src_path/'driver.F90').write_text(fcode_driver)

    # Create the Scheduler
    config = SchedulerConfig.from_dict({
        'default': {
            'role': 'kernel', 'expand': True, 'strict': True, 'enable_imports': True
        },
        'routines': {'test_scheduler': {'role': 'driver'}}
    })

    scheduler = Scheduler(
        paths=[src_path], config=config, seed_routines='test_scheduler',
        frontend=frontend, xmods=[tmp_path], full_parse=False
    )
```

### Version

main

### Platform (OS and architecture)

Python 3.11.10

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ECMWF

reuterbal:
This was an artifact of the way how dependencies incurred via imports are derived for typedefs.
`TypeDef.dependencies` correctly lists only `Import` nodes that contain a type name that is used inside the derived type as a dependency. However, that `Import` node is then used to spawn a dependency for each symbol in the symbol list of that import. To avoid this, the given import dependency is now trimmed to include only relevant symbols.

### DIFF
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -12,7 +12,7 @@ Collection of dataflow analysis schema routines.
 from contextlib import contextmanager
 from loki.expression import Array, ProcedureSymbol
 from loki.ir.expr_visitors import FindLiterals
-from loki.tools import as_tuple, flatten
+from loki.tools import as_tuple, flatten, OrderedSet
 from loki.types import BasicType
 from loki.ir import (
     Visitor, Transformer, FindVariables, FindInlineCalls, FindTypedSymbols
@@ -64,11 +64,11 @@ class DataflowAnalysisAttacher(Transformer):
         uses along the way.
         """
         if live is None:
-            live = set()
+            live = OrderedSet()
         if defines is None:
-            defines = set()
+            defines = OrderedSet()
         if uses is None:
-            uses = set()
+            uses = OrderedSet()
         visited = []
         for i in flatten(body):
             visited += [self.visit(i, live_symbols=live|defines, **kwargs)]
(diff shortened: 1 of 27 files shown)


mlange05 (approved):
Looks good now.

PR #643, opened 2026-01-12
Introduce OrderedSet class to achieve stable code generation (fixes #498, #642)

### LINKED ISSUE #498: Reproducibility of the ordering of expressions
### Is your feature request related to a problem? Please describe.

Loki is generating the same code with the same initial conditions/source files/...

Arguments can be reordered, in function declarations, but also in module use statements, (e.g. `USE parkind1, ONLY: JPIM, JPRB` vs `USE parkind1, ONLY: JPRB, JPIM`)

This is **not a bug** and the generated code is valid either way, however, this makes comparing files more difficult.

There are several reasons for that but mostly the usage of python sets is responsible for that.

### Describe the solution you'd like

Rewrite some parts, e.g., sort module imports via name so that the generated code always looks the same.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Organisation

_No response_

### LINKED ISSUE #642: Stable code generation
### Is your feature request related to a problem? Please describe.

Generating the same code twice tends to give different code, e.g. reordering of symbols in USE clauses or update device clauses.

Some of those differences would be easy to solve, e.g. the pragmas in transformations/data_offload/global_var.py could just create do
```
Pragma(keyword='loki', content=f'create device({", ".join(sorted(v.name for v in offload_variables))})')
```
instead of
```
Pragma(keyword='loki', content=f'create device({", ".join(v.name for v in offload_variables)})')
```
to guarantee stable results.

Or, the `backend/fgen.py` could make sure that the output is stable (somewhat a built-in pretty print).

What would remain is unstable code transformations; those might be harder to fix if they exist.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Organisation

_No response_

reuterbal:
A long-standing inconvenience was the unstable code generation in Loki, due to the use of `set` in several places - predominantly the dataflow analysis and transformations that use it. While convenient to automatically eliminate duplicates and provide native functionality for unions and intersections, this made comparisons between builds harder.

To retain the convenience and ensure stable code generation, this PR introduces an `OrderedSet` class and applies it to all relevant situations (that I managed to find). I've verified that two independent IFS builds now generate the same code (using SCC-STACK pipeline) and are bit-identical to Loki main.

This fixes #642, thanks for the suggestion @lukasm91.

### DIFF
--- a/loki/analyse/analyse_dataflow.py
+++ b/loki/analyse/analyse_dataflow.py
@@ -81,11 +81,11 @@ class DataflowAnalysisAttacher(Transformer):
         """
         Return set of symbols found in an expression.
         """
-        variables = {strip_nested_dimensions(v) for v in FindVariables().visit(expr)}
-        parents = {p for var in variables for p in var.parents}
+        variables = OrderedSet(strip_nested_dimensions(v) for v in FindVariables().visit(expr))
+        parents = OrderedSet(p for var in variables for p in var.parents)
         variables -= parents
         if condition is not None:
-            return {v for v in variables if condition(v)}
+            return OrderedSet(v for v in variables if condition(v))
         return variables
 
     @classmethod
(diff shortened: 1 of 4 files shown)


mlange05 (approved):
Looks fantastic and has been on the wishlist for ages! Many thanks for finally doing this! GTG from me. 👍 🙏 :shipit:

PR #644, opened 2026-01-13
Optionally include internal procedures in the Scheduler graph

reuterbal:
This PR contributes the ability to include internal procedures in the Scheduler's dependency graph. This can be controlled via a new Scheduler config option `ignore_internal_procedures`. By default, the option is set to `True` to maintain current behaviour. It can also be overwritten at a per-routine level for fine-grained control over the scheduler expansion behaviour.

Internal procedure items use a second hash to separate name parts, i.e., `<module>#<procedure>#<internal procedure>`, with `<module>` being empty for free subroutines or functions.

Piggy-backed is a small chance that disables DaCe by default in the install script, since Dace have not managed to add support for Numpy 2+ in almost a year while we are mandating Numpy 2+

### DIFF
--- a/loki/batch/configure.py
+++ b/loki/batch/configure.py
@@ -170,8 +170,10 @@ class SchedulerConfig:
             scope_name, local_name = '', name_parts[0]
         elif len(name_parts) == 2:
             scope_name, local_name = name_parts
+        elif len(name_parts) == 3:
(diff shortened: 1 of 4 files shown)


mlange05 (approved):
Looks great, and verified offline. GTG from me. 👍 🙏

PR #646, opened 2026-01-25
Types: Source per type and test re-organisation

mlange05:
A small housekeeping PR that move the `SymbolTable`/`SymbolAttr`, as well as the dedicated special type classes into separate source files. It also adds a few basic tests for derived-type symbol behaviour that have previously been assumed implicitly.

On top, I've consolidated a bunch of old JIT-based tests for derived types, to reduce the amount of time spend compiling when testing the type system. This is in preparation of some further refactoring of the type/scope and expression systems.

### DIFF
--- a/loki/types/tests/test_scope.py
+++ b/loki/types/tests/test_scope.py
@@ -30,6 +30,9 @@ def test_symbol_attributes():
     delattr(_type, 'foofoo')
     assert _type.foofoo is None
 
+    _type.b = None
+    assert _type.b is None
+
 
 def test_symbol_attributes_compare():
     """


loki/types/tests/test_scope.py
```diff
@@ -0,0 +1,47 @@
...
+    assert _type.a == 'a'
+    assert _type.b
+    assert _type.c is None
+    assert _type.foofoo is None
+
+    _type.foofoo = 'bar'
+    assert _type.foofoo == 'bar'
+
+    delattr(_type, 'foofoo')
+    assert _type.foofoo is None
```
reuterbal: Can you add a test for the second delete path here (by setting the value to `None`)?


```suggestion
    assert _type.foofoo is None
    
    _type.b = None
    assert _type.b is None
```

PR #647, opened 2026-01-27
Types: New Scope accessors, refactoring and LazyNodeLookup removal

mlange05:
~_Note: This sits on top of PR #646 and should only be reviewed after that is merged._~

The primary purpose of this PR is introduce a new "Scope accessor" API and use it to avoid repeated cloning and re-instantiating of symbols for the sole purpose of updating symbol/type attributes. The key idea is to make insertions and updates into the symbol-table explicitly by providing the methods `s.declare(name, **attrs)` / `s.update(name, **attrs)`, as well as `s.get_type(name)` and `s.get_dtype(name)` to `Scope` objects. This should allow us to gradually only adjust type information through this high-level API, and optionally fail if there is an unexpected state - updating an undeclared symbol or re-declaring something that's already been declared.

In the long-run this is intended to replace all manual modifications to `scope.symbol_attrs`, but in this PR I've addressed two of the main inconsistencies in the FParser frontend:
* When declaring variables, we now let the `Entity_Decl` (the first time we have the variable name) declare the base type (LHS of the `::`), before creating the corresponding symbol. We then recurse on the symbol itself to pick up RHS attributes (implicit shape/dimension or initial value) and update the declaration accordingly.
* For `StatementFunction` objects, we no longer use the `LazyNodeLookup`, but instead explicitly "re-declare" the base type to change from "array" to procedure type. 

The latter also required a more low-level intervention: Making `StatementFunction` nodes `ScopedNode`, so that they no longer get cloned on default `Transformer` passes. A bunch of tests has also been added to ensure this behaviour more explicitly.

Other small additions and cosmetids:
* Expose the enum `BasicType` values directly, so that we can do `from loki.types import DEFERRED`
* Move tests for contained/internal procedures into a separate test and chance nomenclature from "members" to "internal"
* Remove now redundant `LazyNodeLookup` entirely


loki/analyse/analyse_dataflow.py
```diff
@@ -327,6 +327,11 @@ def visit_VariableDeclaration(self, o, **kwargs):
             uses |= {o.symbols[0].type.kind}
         return self.visit_Node(o, defines_symbols=defines, uses_symbols=uses, **kwargs)
 
+    def visit_StatementFunction(self, o, **kwargs):
+        # Only the return symbol is meaningful to the outside
+        defines = self._symbols_from_expr(o.variable)
+        return self.visit_Node(o, defines_symbols=defines, **kwargs)
```
reuterbal: This seems not to be covered by tests - any chance of monkey-patching this into an existing test to make sure it behaves as expected?

loki/ir/nodes.py
```diff
@@ -1571,9 +1571,12 @@ class _StatementFunctionBase():
...
 
 @dataclass_strict(frozen=True)
-class StatementFunction(LeafNode, _StatementFunctionBase):
+class StatementFunction(ScopedNode, LeafNode, _StatementFunctionBase):
     """
-    Internal representation of Fortran statement function statements
+    Internal representation of Fortran statement function statements.
+
+    Internally, this is considered a :any:`ScopedNode` with an empty
+    body, because it may be the target of a :any:`ProcedureType`.
```
reuterbal:
```suggestion
    Internally, this is considered a :any:`ScopedNode`, because 
    it may be the target of a :any:`ProcedureType`.
```
ScopedNode doesn't imply being an `InternalNode`, which is what would introduce a body - so I would leave that reference out.

loki/transformations/inline/functions.py
```diff
@@ -222,6 +222,10 @@ def inline_statement_functions(routine):
     # Apply expression-level substitution to routine
     routine.body = SubstituteExpressions(exprmap).visit(routine.body)
 
+    # Rescope the routine body, as the recursive update call does not
+    # set scopes on the RHS of `exprmap`, and so we might miss a scope.
+    routine.rescope_symbols()
```
reuterbal, line 227: Would it not be better to combine these two into a single pass, applying the rescoping only when substitution occurs?

loki/types/tests/test_procedure_types.py
```diff
@@ -82,3 +82,47 @@ def test_procedure_type(tmp_path, frontend):
...
+    skip=[(OMNI, 'Statement functions not supported with OMNI')]
+))
+def test_procedure_type_stmt_func(frontend):
+    """ Test `ProcedureType` keeps its link to `StatementFunction` when transformed. """
+    fcode_mod = """
+subroutine test_routine(n, a)
+  implicit none
+  integer, intent(in) :: n
+  real(kind=4), intent(inout) :: a(3)
+  real(kind=4) :: pants, on, fire
```
reuterbal, line 97: 🔥

loki/types/scope.py
```diff
@@ -162,3 +163,41 @@ def _reset_parent(self, parent):
 
         if self.parent is not None:
             self.symbol_attrs.parent = self.parent.symbol_attrs
+
+    def declare(self, name, dtype, fail=True, **kwargs):
```
reuterbal, line 167: Since this method (as well as `update` and the getters) constitute a rather crucial part of the new API, would you mind adding some meaningful docstrings to explain their role, use, and rationale?

reuterbal (changes requested):
Many thanks, a very nice and important step towards the overhaul of the symbol type system! Only some remarks about minor details, no objections per se!

loki/analyse/analyse_dataflow.py
mlange05: Yes, in fact, I think I found that the current implementation was faulty, and simplified it.

loki/transformations/inline/functions.py
mlange05: As discussed offline, "yes, in theory", but the benefit is likely small and I'd like to use off-the-shelf utilities here instead of building custom transformers.

loki/types/procedure_type.py
```diff
@@ -56,23 +55,11 @@ def __init__(
...
-            self._name = self.procedure.name
-            assert is_function is None or is_function == self.procedure.is_function
-            self._is_function = self.procedure.is_function
-            # TODO: compare return type once type comparison is more robust
-            self._return_type = self.procedure.return_type if self.procedure.is_function else None
+
+        self.procedure = procedure
+        self._name = procedure.name if procedure else name
+        self._is_function = is_function or False
+        self._return_type = return_type
```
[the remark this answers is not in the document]
mlange05: I see the point, but for pickle/unpickle we should always provide the procedure to ensure re-generation of the weak-ref. I'd prefer to fail here, instead of keeping cached values, but we might need to revisit that in a follow-up. Good spot, though!

mlange05:
Ok, I believe I've addressed all comments (including very helpful suggestions from Copilot). Ready for another look, @reuterbal

reuterbal (approved):

Issue #632, opened 2025-11-10
Always use generic interface for FIELD_API data access calls in `OffloadDeepcopyTransformation`


awnawab:
### What maintenance does this project need?

Using the `SGET_` generic interface for FIELD_API data access calls is much safer, as it ensures that a valid memory address is returned even for zero-sized and uninitialised fields. Moreover, as the field initialisation checks now happen within FIELD_API, the boilerplate generated by loki can be significantly reduced.

Due to both the above, the `OffloadDeeopcopyTransformation` should be modified to always use the generic interface, rather than it being a configurable transformation option.

PR #648, opened 2026-02-02
Always use generic interface based FIELD_API accessors in `DataOffloadDeepcopyTransformation`

awnawab:
As described in #632, this PR updates the `DataOffloadDeepcopyTransformation` to always use the generic interface based FIELD_API accessors. Not only does this lead to code safer from 0-sized and uninitialised fields, but it also significantly reduces the amount of boilerplate generated. I've also included a fix requested a fix in a previous PR to make the dependency on pyyaml optional.

### DIFF
--- a/loki/transformations/data_offload/offload_deepcopy.py
+++ b/loki/transformations/data_offload/offload_deepcopy.py
@@ -9,7 +9,11 @@ from collections import defaultdict
 from pathlib import Path
 
 import re
-import yaml
+try:
+    import yaml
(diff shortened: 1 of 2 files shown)


MichaelSt98 (approved):
Looks good to me.

reuterbal (changes requested):
Thanks, this looks very good to me. However, a small change request on the trivial part in this PR ;-)

reuterbal (approved):

Issue #275, opened 2024-04-08
Re-think the DependencyTrafo


reuterbal:
The DependencyTransformation stems from early Loki days and it has accumulated some technical debt. Let's collect some of the pain points in this issue to address them in a future rewrite of the transformation.

- Remove the baked-in assumption of `.intfb` in the name of interface block C-Includes (introduced in #251)
- The C-include to module import flip could be informed with analysis data by the `ModuleWrapTransformation`
- ...

reuterbal:
Just to provide the link here: #453 provides a step in a direction that will allow making this more flexible

PR #649, opened 2026-02-04
Driver-specific Loki pipelines

MichaelSt98:
allow for multi-mode/driver-specific Loki pipelines, which includes possibly making the graph more tree-like for relevant routines/items.

E.g, starting from:

callgraph_before.pdf

with 

```py
'driver_0': {'role': 'driver', 'mode': 'm1', 'replicate': False},
'driver_1': {'role': 'driver', 'mode': 'm1', 'replicate': False},
'driver_2': {'role': 'driver', 'mode': 'm2', 'replicate': False},
'driver_3': {'role': 'driver', 'mode': 'm3', 'replicate': False},
'driver_4': {'role': 'driver', 'mode': 'm3', 'replicate': False},
```

getting to:

callgraph_after.pdf

### DIFF
--- a/loki/batch/item_factory.py
+++ b/loki/batch/item_factory.py
@@ -290,6 +290,7 @@ class ItemFactory:
 
             # Create a new FileItem for the new source
             new_source.path = item.path.with_name(f'{scope_name or local_name}{item.path.suffix}')
+            new_source.orig_path = item.path
             file_item = self.get_or_create_file_item_from_source(new_source, config=config)
 
             # Get the definition items for the FileItem and return the new item
--- a/loki/batch/item_factory.py
+++ b/loki/batch/item_factory.py
@@ -347,6 +347,13 @@ class ItemFactory:
         self.item_cache[item_name] = file_item
         return file_item
 
+    def get_file_item_from_source(self, source):
+        # Check for file item with the same source object
+        for item in self.item_cache.values():
+            if isinstance(item, FileItem) and item.source is source:
+                return item
+        return None
+
     def get_or_create_file_item_from_source(self, source, config):
         """
         Utility method to create a :any:`FileItem` corresponding to a given source object
@@ -369,9 +376,9 @@ class ItemFactory:
             The config object from which the item configuration will be derived
         """
         # Check for file item with the same source object
-        for item in self.item_cache.values():
-            if isinstance(item, FileItem) and item.source is source:
-                return item
+        item_ = self.get_file_item_from_source(source)
+        if item_ is not None:
+            return item_
 
         if not source.path:
             raise RuntimeError('Cannot create FileItem from source: Sourcefile has no path')
--- a/loki/batch/item_factory.py
+++ b/loki/batch/item_factory.py
@@ -290,7 +290,7 @@ class ItemFactory:
 
             # Create a new FileItem for the new source
             new_source.path = item.path.with_name(f'{scope_name or local_name}{item.path.suffix}')
-            new_source.orig_path = item.path
+            new_source.orig_path = item.orig_path
             file_item = self.get_or_create_file_item_from_source(new_source, config=config)
 
             # Get the definition items for the FileItem and return the new item
--- a/loki/batch/item_factory.py
+++ b/loki/batch/item_factory.py
@@ -348,6 +348,14 @@ class ItemFactory:
         return file_item
 
     def get_file_item_from_source(self, source):
+        """
+        Utility method to get a :any:`FileItem` from a given source object
+
+        Parameters
+        ----------
+        source : :any:`Sourcefile`
+            The existing sourcefile object for which to create the file item
+        """
         # Check for file item with the same source object
         for item in self.item_cache.values():
             if isinstance(item, FileItem) and item.source is source:
@@ -621,6 +629,20 @@ class ItemFactory:
                 )
                 items += [_it for _it in definition_items if _it.name[_it.name.index('#')+1:] == name.lower()]
         return tuple(items)
+
+    def get_scope_and_file_items(self, item):
+        """
+        Get corresponding module and file item of a given :data:`item`.
+
+        Parameters
+        ----------
+        item : :any:`Item`
+            The item node in the dependency graph for which to determine the successors
+        """
+        mod_item = self.item_cache.get(item.scope_name)
+        _items = (mod_item,) if mod_item is not None else ()
+        _items += (self.get_file_item_from_source(item.source),)
+        return _items
 
     @staticmethod
     def _get_imported_symbol_name(imprt, symbol_name):
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -445,9 +445,24 @@ class Scheduler:
             error('[Loki::Scheduler] Batch processing requires Transformation or Pipeline object')
             raise RuntimeError(f'Could not batch process {transformation}')
 
-    def process_pipeline(self, pipeline, proc_strategy=ProcessingStrategy.DEFAULT):
+    def process_config(self, proc_strategy=ProcessingStrategy.DEFAULT):
         """
-        Process a given :any:`Pipeline` by applying its assocaited
+        Process a given config by applying pipelines in dependence
+        of the given mode (per (driver) item).
+
+        Parameters
+        ----------
+        proc_strategy : :any:`ProcessingStrategy`
+            The processing strategy to use when applying the given
+            :data:`pipeline` to the scheduler's graph.
+        """
+        modes = {item.mode for item in self.items}
+        for mode_ in modes:
+            self.process_pipeline(pipeline=self.config.pipelines[mode_], proc_strategy=proc_strategy, mode=mode_)
+
+    def process_pipeline(self, pipeline, proc_strategy=ProcessingStrategy.DEFAULT, mode=None):
+        """
+        Process a given :any:`Pipeline` by applying its associated
         transformations in turn.
 
         Parameters
@@ -457,11 +472,14 @@ class Scheduler:
         proc_strategy : :any:`ProcessingStrategy`
             The processing strategy to use when applying the given
             :data:`pipeline` to the scheduler's graph.
+        mode : str, optional
+            Transformation mode, selecting which code transformations/pipeline on which graph to apply.
+            Default: `None`, thus mode agnostic.
         """
         for transformation in pipeline.transformations:
-            self.process_transformation(transformation, proc_strategy=proc_strategy)
+            self.process_transformation(transformation, proc_strategy=proc_strategy, mode=mode)
 
-    def process_transformation(self, transformation, proc_strategy=ProcessingStrategy.DEFAULT):
+    def process_transformation(self, transformation, proc_strategy=ProcessingStrategy.DEFAULT, mode=None):
         """
         Process all :attr:`items` in the scheduler's graph
 
@@ -489,6 +507,9 @@ class Scheduler:
         proc_strategy : :any:`ProcessingStrategy`
             The processing strategy to use when applying the given
             :data:`transformation` to the scheduler's graph.
+        mode : str, optional
+            Transformation mode, selecting which code transformations on which graph to apply.
+            Default: `None`, thus mode agnostic.
         """
         def _get_definition_items(_item, sgraph_items):
             # For backward-compatibility with the DependencyTransform and LinterTransformation
@@ -527,7 +548,8 @@ class Scheduler:
                 sgraph_items = sgraph.items
                 traversal = SFilter(
                     graph, reverse=transformation.reverse_traversal,
-                    include_external=self.config.default.get('strict', True)
+                    include_external=self.config.default.get('strict', True),
+                    mode=mode
                 )
             else:
                 graph = self.sgraph
@@ -535,7 +557,8 @@ class Scheduler:
                 traversal = SFilter(
                     graph, item_filter=item_filter, reverse=transformation.reverse_traversal,
                     exclude_ignored=not transformation.process_ignored_items,
-                    include_external=self.config.default.get('strict', True)
+                    include_external=self.config.default.get('strict', True),
+                    mode=mode
                 )
 
             # Collect common transformation arguments
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -25,6 +25,7 @@ from loki.frontend import FP, REGEX, RegexParserClass
 from loki.tools import as_tuple, CaseInsensitiveDict, flatten
 
 from loki.logging import info, perf, warning, error
+from loki.transformations.dependency import SeparateModesKernel
 
 __all__ = ['ProcessingStrategy', 'Scheduler']
 
@@ -201,6 +202,37 @@ class Scheduler:
 
             # Attach interprocedural call-tree information
             self._enrich()
+
+    def propagate_and_separate_modes(self, proc_strategy=ProcessingStrategy.DEFAULT):
+        self._propagate_modes()
+        self.process_transformation(SeparateModesKernel(), proc_strategy=proc_strategy)
+        self._propagate_modes_set()
+        modes = {item.mode for item in self.items}
+        return as_tuple(modes)
+
+
+    def _propagate_modes(self):
+        driver_items = [item for item in self.items if item.role == 'driver']
+        for item in driver_items:
+            module_file_items = self.sgraph.get_corresponding_module_and_file_item(item, self.item_factory)
+            for _item in module_file_items:
+                if _item is not None:
+                    _item.config['mode'] = item.mode
+            descendants = self.sgraph.descendants(item, self.item_factory,
+                    include_module_items=True, include_file_items=True)
+            for descendant in descendants:
+                if descendant is not None:
+                    descendant.config.setdefault('inherited_mode', set()).add(item.mode)
+
+    def _propagate_modes_set(self):
+        driver_items = [item for item in self.items if item.role == 'driver']
+        for item in driver_items:
+            descendants = self.sgraph.descendants(item, self.item_factory,
+                    include_module_items=True, include_file_items=True)
+            for descendant in descendants:
+                if descendant is not None:
+                    descendant.config['mode'] = item.mode
+                    descendant.config['inherited_mode'] = set()
 
     @Timer(logger=info, text='[Loki::Scheduler] Performed initial source scan in {:.2f}s')
     def _discover(self):
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -5,7 +5,6 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-import sys
 from enum import Enum, auto
 from os.path import commonpath
 from pathlib import Path
@@ -224,7 +223,6 @@ class Scheduler:
         self.process_transformation(SeparateModesKernel(), proc_strategy=proc_strategy)
         modes = {item.mode for item in self.items}
         return as_tuple(modes)
-
 
     def _propagate_modes(self):
         """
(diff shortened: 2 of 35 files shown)


loki/tests/sources/projMultiMode/driver_0_mod.F90
```diff
@@ -0,0 +1,12 @@
+module driver_0_mod
+  implicit none
+contains
+
+   subroutine driver_0()
+     #include "subroutine_1.intfb.h"
+     #include "subroutine_3.intfb.h"
```
reuterbal: FYI: Most Fortran compilers will refuse to compile these if the #include is not left-aligned without leading white space

loki/tests/sources/projMultiMode/subroutine_3.F90
```diff
@@ -0,0 +1,12 @@
+subroutine subroutine_3()
+  #include "nested_subroutine_1.intfb.h"
+  #include "nested_subroutine_3.intfb.h"
+
+! INTERFACE
+!   SUBROUTINE nested_subroutine_3()
+!   END SUBROUTINE nested_subroutine_3
+! END INTERFACE
+
```
reuterbal: Could be deleted?

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+                    s = '.'.join(im.module.split('.')[1:])
+                    im._update(module=f'{new_dependencies[target_symbol].local_name}.{s}')
+            else:
+                # Modify module import if it imports any call targets
+                if new_dependencies and im.symbols and any(s in new_dependencies for s in im.symbols):
+                    relevant_symbol = [s for s in im.symbols if s in new_dependencies][0]
+                    _new_symbol = new_dependencies[relevant_symbol]
+                    new_symbol = _new_symbol.scope_ir.procedure_symbol # local_name
+                    new_module_name = _new_symbol.scope_name
+                    im._update(module=new_module_name, symbols=(new_symbol,))
```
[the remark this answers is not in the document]
reuterbal: I think the comment is wrong due to the string-conversion of symbols in comparisons

loki/batch/item_factory.py
```diff
@@ -346,6 +347,13 @@ def get_or_create_file_item_from_path(self, path, config, frontend_args=None):
         self.item_cache[item_name] = file_item
         return file_item
 
+    def get_file_item_from_source(self, source):
+        # Check for file item with the same source object
+        for item in self.item_cache.values():
+            if isinstance(item, FileItem) and item.source is source:
+                return item
+        return None
```
reuterbal, line 363: Is it required to move this functionality out or could this stay in `get_or_create_file_item_from_source`? I'm a bit worried about adding more entry points to the item factory, it's already too complicated to my taste.

loki/batch/scheduler.py
```diff
@@ -202,6 +203,38 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
             # Attach interprocedural call-tree information
             self._enrich()
 
+    def propagate_and_separate_modes(self, proc_strategy=ProcessingStrategy.DEFAULT):
```
reuterbal, line 206: Could you add a docstring here please?

loki/cli/loki_transform.py
```diff
@@ -39,8 +39,10 @@
 @click.option('--log-level', '-l', default='info', envvar='LOKI_LOGGING',
               type=click.Choice(['debug', 'detail', 'perf', 'info', 'warning', 'error']),
               help='Log level to output during batch processing')
+@click.option('--multimode', '-z', is_flag=True, help="Multi-mode processing.")
```
reuterbal: Some musings on namings: `multimode` is in line with the previous use of `mode` to select the transformation pipeline - but it's also somewhat a remnant of early Loki days. Since mode is effectively equivalent to the pipeline we choose, I'm wondering if we would want to converge on a single name for that concept. In wich case "pipeline" is the more appropriate term in my opinion (and would match also your PR title ;-) )

For backward-compatibility, we should at least keep the `mode` CLI argument as an alias around, though.

That means, all "multi-mode" references should become "multi-pipeline", and the use of `mode` in item configs etc. should instead be replaced by `pipeline`. But I'd like to also hear @mlange05 's opinion on this, so I suggest discussing this before making the change.

loki/batch/sgraph.py
```diff
@@ -491,3 +491,44 @@ def export_to_file(self, dotfile_path):
...
+        file_items = []
+        if include_module_items:
+            module_items = [item_factory.item_cache.get(_item.scope_name)
+                    for _item in item_descendants if not _item.is_ignored]
+        if include_file_items:
+            file_items = [item_factory.get_file_item_from_source(_item.source)
+                    for _item in item_descendants if not _item.is_ignored]
+        return item_descendants + module_items + file_items
+
+    def get_corresponding_module_and_file_item(self, item, item_factory):
```
reuterbal: This is purely a query on the item factory, which needs to be provided anyway - so I suggest making this a method on `ItemFactory` instead.

As a more conceptual description, I suggest calling this simply `get_scope_and_file_items`

loki/batch/scheduler.py
```diff
@@ -202,6 +203,38 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
...
+        from loki.transformations.dependency import SeparateModesKernel # pylint: disable=import-outside-toplevel
+        self._propagate_modes()
+        self.process_transformation(SeparateModesKernel(), proc_strategy=proc_strategy)
+        self._propagate_modes_set()
+        modes = {item.mode for item in self.items}
+        return as_tuple(modes)
+
+
+    def _propagate_modes(self):
```
reuterbal, line 228: A short docstring here would be helpful to explain that this propagates the pipeline selection to all applicable items _as well as_ their enclosing scopes.

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+        return scope_name, local_name, new_item_name
+
+    def _get_or_create_or_rename_item(self, item, mode, item_factory, config):
+        """
+        Get, create or rename item including the scope item if there is a
+        scope.
+
+        Parameters
+        ----------
+        mode : TODO
```
reuterbal: TODO

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+
+    @staticmethod
+    def get_parent_items(item, item_factory):
+        item_cache = item_factory.item_cache
+        parent_items = ()
+        if item.scope_name in item_cache:
+            parent_items += (item_cache[item.scope_name],)
+        if str(item.source.path) in item_cache:
+            parent_items += (item_cache[str(item.source.path)],)
+        return parent_items
```
reuterbal: If the `get_scope_and_file_item` method moved to the ItemFactory, you should be able to use it instead of the custom method here, no?

loki/batch/scheduler.py
```diff
@@ -202,6 +203,38 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
...
+        for item in driver_items:
+            module_file_items = self.sgraph.get_corresponding_module_and_file_item(item, self.item_factory)
+            for _item in module_file_items:
+                if _item is not None:
+                    _item.config['mode'] = item.mode
+            descendants = self.sgraph.descendants(item, self.item_factory,
+                    include_module_items=True, include_file_items=True)
+            for descendant in descendants:
+                if descendant is not None:
+                    descendant.config.setdefault('inherited_mode', set()).add(item.mode)
```
reuterbal: I understand the concept of "inherited_mode" and "mode" is meant to distinguish between all modes that apply to a given item (which may be more than 1 if an item is shared between two driver pipelines). Two related questions to this:
1. If the module of the driver contains more than 1 driver, would we not need to make the same distinction here instead of directly applying the mode?
2. The inherited modes are filled with one ore multiple entries, which are then popped out in the SeparateModesKernel transformation to create copies with the relevant mode applied to their `mode` property instead. That makes the `inherited_mode` config property a transient attribute that is only required for the separation of the pipelines - and is in fact reset to an empty set in `propagate_modes_set` at the end. Therefore, could we not for example store this in the transformation data of the items instead, and keep the item config API slim?

loki/batch/scheduler.py
```diff
@@ -202,6 +203,38 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
...
+
+    def _propagate_modes_set(self):
+        driver_items = [item for item in self.items if item.role == 'driver']
+        for item in driver_items:
+            descendants = self.sgraph.descendants(item, self.item_factory,
+                    include_module_items=True, include_file_items=True)
+            for descendant in descendants:
+                if descendant is not None:
+                    descendant.config['mode'] = item.mode
```
reuterbal: I don't understand why this is required: Doesn't the SeparateModesKernel transformation already assign the transformation mode to all descendants?

loki/batch/sgraph.py
```diff
@@ -491,3 +491,44 @@ def export_to_file(self, dotfile_path):
             graph.render(path, view=False)
         except gviz.ExecutableNotFound as e:
             warning(f'[Loki] Failed to render callgraph due to graphviz error:\n  {e}')
+
+    def descendants(self, item, item_factory, include_module_items=False, include_file_items=False):
```
reuterbal: Since this always used with both, module and file items enabled, we could simplify the API and implementation a bit by simply having one option `include_scope_items`.

loki/batch/sgraph.py
```diff
@@ -491,3 +491,44 @@ def export_to_file(self, dotfile_path):
...
+        item_descendants = list(nx.descendants(self._graph, item))
+        module_items = []
+        file_items = []
+        if include_module_items:
+            module_items = [item_factory.item_cache.get(_item.scope_name)
+                    for _item in item_descendants if not _item.is_ignored]
+        if include_file_items:
+            file_items = [item_factory.get_file_item_from_source(_item.source)
+                    for _item in item_descendants if not _item.is_ignored]
+        return item_descendants + module_items + file_items
```
reuterbal: That should then also simplify the implementation like this:

```suggestion
        item_descendants = list(nx.descendants(self._graph, item))
        if include_scope_items:
            item_desendants += [*item_factory.get_scope_and_file_items(_item) for _item in item_descendants]
        return item_descendants
```

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+                new_item_new_items, new_item_removed_items = self._create_new_items(sub_sgraph.successors(child),
+                        item_factory, config, new_item, sub_sgraph)
+                new_item.plan_data['additional_dependencies'] += new_item_new_items
+                new_item.plan_data['removed_dependencies'] += new_item_removed_items
+                if rename_calls:
+                    new_dependencies = CaseInsensitiveDict((new_item.local_name, new_item)
+                            for new_item in new_item_new_items)
+                    new_dependencies = CaseInsensitiveDict((k.local_name, v) for k, v in
+                            new_item.trafo_data.get('SeparateModes', {}).items() if not isinstance(k, str))
+                    self._adapt_calls_imports(new_item.ir, new_dependencies)
```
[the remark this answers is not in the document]
reuterbal: Agreed. Could also be a missing `+=` on the second statement

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+            else:
```
reuterbal, line 499: This is essentially the `len(modes_to_duplicate) == 1` case, right? Is it useful to have a different MO for that, or could we simply treat this the same as for `>1`?

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+                    im._update(module=f'{new_dependencies[target_symbol].local_name}.{s}')
+            else:
+                # Modify module import if it imports any call targets
+                if new_dependencies and im.symbols and any(s in new_dependencies for s in im.symbols):
+                    relevant_symbol = [s for s in im.symbols if s in new_dependencies][0]
+                    _new_symbol = new_dependencies[relevant_symbol]
+                    new_symbol = _new_symbol.scope_ir.procedure_symbol # local_name
+                    new_module_name = _new_symbol.scope_name
+                    im._update(module=new_module_name, symbols=(new_symbol,))
+            # TODO: Deal with unqualified blanket imports
```
reuterbal: Maybe throw an error/warning if that is being encountered?

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+
+    def _adapt_calls_imports(self, routine, new_dependencies):
+        call_map = {}
+        if not new_dependencies:
+            return
+        for call in FindNodes(ir.CallStatement).visit(routine.body):
+            call_name = str(call.name).lower()
+            if call_name in new_dependencies:
+                new_item = as_tuple(new_dependencies[call_name])[0]
+                # new_call_name = (new_item.name).lower() # str(new_item.local_name).lower()
```
reuterbal:
```suggestion
```

reuterbal (changes requested):
Thank you, this is really an exciting new development that unlocks not only a fantastic new set of capabilities but also allows to finally carry out the long overdue modernization of the dependency injection mechanics. The approach looks good, as discussed offline, but I left several remarks to suggest improvements in the implementation.

I'm particularly interested in trying to expose as little additional API as possible, because dealing with the batch processing is already complex enough...

reuterbal:
Also I'd appreciate @mlange05's input on these changes to make sure this aligns with his view on the dependency injection modernization.

loki/batch/scheduler.py
mlange05: I agree with this - the superset of modes should only be handled by the Trafo and not be inherently exposed to the Scheduler. At the very least, we should move the metadata to the trafo-data and then mop up the rest in the subsequent rewrite.

mlange05 (commented):
Ok, I think the overall idea behind this PR is great, including the testing and CMake plumbing, but I wonder if the algorithmic approach is right for this. If I understood this correctly, the approach does a depth-first traversal that marks each item with all the possible modes it is used in, before then using this meta-information to duplicate items as needed. I wonder if a purely recursive breadth-first traversal that always specialises each item with its "mode" marker immediately would remove a lot of the added complexity in this PR - and would at the same time do most of the job of the `DependencyTrafo`. However, this is clearly a big re-write that should be done as a follow-up.

The other thing I noted is what @reuterbal had already flagged: the 'inherited_mode' property should be handled as trafo-specific meta-data, not scheduler config data, as it is transient. 

So, I'm happy for this PR to land once all remaining comments are addressed. However, please also file an "enhancement" issue to schedule the follow-up re-write; I'm happy to take that one.

loki/transformations/dependency.py
MichaelSt98: Not completely sure what you mean. Do you question the implementation or the logic to keep the original item for the last remaining inherited mode/pipeline?

loki/batch/item_factory.py
```diff
@@ -290,6 +290,7 @@ def get_or_create_item_from_item(self, name, item, config=None):
 
             # Create a new FileItem for the new source
             new_source.path = item.path.with_name(f'{scope_name or local_name}{item.path.suffix}')
+            new_source.orig_path = item.path
```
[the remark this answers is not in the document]
reuterbal: This might make sense @MichaelSt98 ?

loki/batch/scheduler.py
```diff
@@ -5,6 +5,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+import sys
 from enum import Enum, auto
 from os.path import commonpath
 from pathlib import Path
```
reuterbal: Import order:
```suggestion
from enum import Enum, auto
from os.path import commonpath
from pathlib import Path
import sys
```

loki/batch/sgraph.py
```diff
@@ -491,3 +491,23 @@ def export_to_file(self, dotfile_path):
...
+        Get all descendants of a given :data:`item`.
+
+        Parameters
+        ----------
+        item : :any:`Item`
+            The item node in the dependency graph for which to determine the successors
+        item_factory : :any:`ItemFactory`
+            The item factory to use.
+        include_scope_items : bool, optional
+            Whether to scope items like module and file items.
```
reuterbal:
```suggestion
            Whether to include scope items like module and file items.
```

loki/transformations/dependency.py
```diff
@@ -330,3 +331,292 @@ def plan_subroutine(self, routine, **kwargs):
...
+        Kernel name(s) to be duplicated.
+    duplicate_suffix : str, optional
+        Suffix to be used to append the original kernel name(s).
+    duplicate_module_suffix : str, optional
+        Suffix to be used to append the original module name(s),
+        if defined, otherwise `duplicate_suffix`
+    duplicate_subgraph : bool, optional
+        Whether or not duplicate the subgraph beneath the kernel(s)
+        that are duplicated.
+    """
```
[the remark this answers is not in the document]
reuterbal: This is correct, as a minimum I suggest to drop the Parameters section of the docstring.

loki/transformations/dependency.py
reuterbal: As discussed offline: simply always renaming items, even if not multiple pipelines are involved, might make the implementation simpler and shouldn't hurt.

@copilot Create a new issue suggesting that suggests this as an improvement and add a cross-link to #275, which this would be working towards.

loki/batch/scheduler.py
```diff
@@ -441,13 +483,24 @@ def process(self, transformation, proc_strategy=ProcessingStrategy.DEFAULT):
...
             self.process_pipeline(pipeline=transformation, proc_strategy=proc_strategy)
 
+        elif isinstance(transformation, dict):
+            assert all(isinstance(trafo, Pipeline) for trafo in transformation.values())
+            modes = self.propagate_and_separate_modes(proc_strategy=proc_strategy)
+            # check if all required pipelines are available
+            for mode in modes:
+                if mode not in transformation:
+                    msg = f'[Loki] ERROR: Pipeline or transformation mode {mode} not found in config file.\n'
+                    sys.exit(msg)
```
reuterbal: Since this is no longer in the CLI script, we should instead raise an error:

```suggestion
                    error('[Loki] ERROR: Pipeline or transformation mode %s not found in config file.', mode)
                    raise RuntimeError(f'Could not find pipeline or transformation mode {mode}')
```

loki/batch/scheduler.py
```diff
@@ -202,6 +203,47 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
...
+        proc_strategy : :any:`ProcessingStrategy`
+            The processing strategy to use when applying the given
+            :data:`transformation` to the scheduler's graph.
+        """
+        from loki.transformations.dependency import SeparateModesKernel # pylint: disable=import-outside-toplevel
+        self._propagate_modes()
+        self.process_transformation(SeparateModesKernel(), proc_strategy=proc_strategy)
+        modes = {item.mode for item in self.items}
+        return as_tuple(modes)
+
```
reuterbal, line 227:
```suggestion
```

reuterbal (changes requested):
Thank you for implementing the suggestions. A few details that would be great to sort out, but otherwise I'd be happy to bring this in now.

reuterbal:
Gnah, it did all the things I didn't ask it to do and didn't do the one thing I wanted it to do... @MichaelSt98 please check if you're happy with the modifications.

MichaelSt98:
> Gnah, it did all the things I didn't ask it to do and didn't do the one thing I wanted it to do... @MichaelSt98 please check if you're happy with the modifications.

Haha perfect and tests fail ... Will push the fix/fixes.

reuterbal (approved):
Thank you!

mlange05 (approved):
Looks fantastic! Very nicely done and many thanks! GTG 👍 :shipit:

Issue #650, opened 2026-02-19
Scheduler raises assertion errors when dealing with unresolved array components of derived type accesses


rommelDB:
### What happened?

Expected Results:
If there is an element that cannot be resolved due to various factors, what I would expect is that the Scheduler continues its execution as it does with other unresolved elements (see reproducer).

### What are the steps to reproduce the bug?

Just run:
```
$ python sched.py 
```

Given the following files:

sched.py
```
from loki.batch.scheduler import Scheduler
from loki.batch.configure import SchedulerConfig

config_dict = {
    "default": {}
}

config = SchedulerConfig.from_dict(config_dict)

scheduler = Scheduler(
    paths=["./index_module.f90"],
    config=config,
    full_parse=True,
    seed_routines=["init"]
)

for item in scheduler.items:
    ir = item.ir
    name = getattr(ir, "name", "<anonymous>")
    kind = type(ir).__name__

    print(f"{kind}: {name}")
```

index_module.f90
```
MODULE index_module
  IMPLICIT NONE
CONTAINS

  SUBROUTINE init

   my_index(1) = 1 !ok
   my_index%len = 1 !ok

   my_index(1)%len(1) = 1 !assert error
   my_index%len(1) = 1 !assert error

  END SUBROUTINE

END MODULE index_module
```

### Version

0.3.6.dev89+g6f5d88e8c

### Platform (OS and architecture)

Linux 6.2.0-26-generic #26~22.04.1-Ubuntu x86_64

### Relevant log output

```shell
$ python sched.py 
Traceback (most recent call last):
  File "/workspace/reproducible/sched.py", line 10, in <module>
    scheduler = Scheduler(
  File "/conda/lib/python3.10/site-packages/loki/batch/scheduler.py", line 203, in __init__
    self._enrich()
  File "/conda/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/conda/lib/python3.10/site-packages/loki/batch/scheduler.py", line 342, in _enrich
    item.ir.enrich(enrich_definitions, recurse=True)
  File "/conda/lib/python3.10/site-packages/loki/subroutine.py", line 420, in enrich
    self.body = ExpressionTransformer(inplace=True).visit(self.body)
  File "/conda/lib/python3.10/site-packages/loki/ir/expr_visitors.py", line 246, in visit
    obj = super().visit(o, *args, **kwargs)
[...]

PR #652, opened 2026-02-23
Loki: Better error message on a key assert

mlange05:
One of the key asserts in Loki that triggers on erroneous type information from an inconsistent IR without any meaningful error message. This PR adds a short message that highlights the symbol in question.

I've also added another assert on parent-symbol lookup in FParser that came from investigating the examples provided by @rommelDB in issue #650 .


reuterbal (approved):
Looks good, many thanks!

PR #653, opened 2026-02-24
resolve vector notation: refactored and extended

MichaelSt98:
Refactored and extended resolve vector notation utility.

**Done**: 
Makes a bunch of assumptions which probably should be at least be optional to allow for more conservative transformations. Further, this does not work strided array dimensions. 

**Done**:
Obviously some more rigorous testing required too.

### DIFF
--- a/loki/backend/fgen.py
+++ b/loki/backend/fgen.py
@@ -681,8 +681,7 @@ class FortranCodegen(Stringifier):
             [...body...]
           END IF [name]
         """
-        # TODO: hack ..., this is a necessary but not a sufficient criteria
-        if o.inline and len(o.body) <= 1:
+        if o.inline:
             # No indentation and only a single body node
             cond = self.visit(o.condition, **kwargs)
             d = self.depth
--- a/loki/backend/fgen.py
+++ b/loki/backend/fgen.py
@@ -681,7 +681,8 @@ class FortranCodegen(Stringifier):
             [...body...]
           END IF [name]
         """
-        if o.inline:
+        # TODO: hack ..., this is a necessary but not a sufficient criteria
+        if o.inline and len(o.body) <= 1:
             # No indentation and only a single body node
             cond = self.visit(o.condition, **kwargs)
             d = self.depth
--- a/loki/transformations/array_indexing/vector_notation.py
+++ b/loki/transformations/array_indexing/vector_notation.py
@@ -20,9 +20,7 @@ from loki.ir import (
 from loki.tools import as_tuple, dict_override, OrderedSet
 from loki.types import SymbolAttributes, BasicType
 
-from loki.transformations.utilities import (
-    get_integer_variable, get_loop_bounds
-)
+from loki.transformations.utilities import get_integer_variable
 
 if HAVE_FP:
     from fparser.two import Fortran2003
@@ -176,7 +174,7 @@ def _get_all_valid_loop_bounds(routine, lower, upper):
     valid_lower = [_ for _ in valid_lower if _ is not None]
     valid_upper = [get_valid(_upper, variable_map) for _upper in upper]
     valid_upper = [_ for _ in valid_upper if _ is not None]
-   
+
     for _lower in valid_lower:
         for _upper in valid_upper:
             bounds += ((_lower, _upper),)
@@ -207,7 +205,7 @@ def resolve_vector_dimension(routine, dimension, derive_qualified_ranges=False,
     """
     # Find the iteration index variable and bound variables
     index = get_integer_variable(routine, name=dimension.index)
-    
+
     _lower = as_tuple(dimension.lower) + ('1',)
     _upper = as_tuple(dimension.upper) + as_tuple(dimension.sizes)
     bounds = _get_all_valid_loop_bounds(routine, lower=_lower, upper=_upper)
@@ -309,7 +307,7 @@ class IterationRangeIndexMapper(LokiIdentityMapper):
                     # Skip if we're not supposed to create new indices
                     if not self.map_unknown_ranges or dim == sym.RangeIndex((None, None)):
                         continue
-                
+
                     # Create new index variable
                     vtype = SymbolAttributes(BasicType.INTEGER)
                     ivar = sym.Variable(name=f'{self.basename}_{i}', type=vtype, scope=self.scope)
@@ -681,9 +679,7 @@ class ResolveVectorNotationTransformer(Transformer):
             new_rhs_array_list.append(_array.clone(dimensions=as_tuple(new_arr_dims)))
 
         # Update the statement in-place
-        rhs_substitution = {
-            old: new for old, new in zip(rhs_arrays, new_rhs_array_list)
-        }
+        rhs_substitution = dict(zip(rhs_arrays, new_rhs_array_list))
         stmt._update(
             lhs=new_lhs_arr,
             rhs=SubstituteExpressions(rhs_substitution).visit(stmt.rhs)
--- a/loki/transformations/array_indexing/vector_notation.py
+++ b/loki/transformations/array_indexing/vector_notation.py
@@ -464,10 +464,10 @@ class ResolveVectorNotationTransformer(Transformer):
         return [i for i, dim in enumerate(dims) if isinstance(dim, sym.RangeIndex)]
 
     @staticmethod
-    def _find_qualified_range_indices(dims, range_positions):
+    def _find_qualified_range_positions(dims, range_positions):
         """
-        Return indices into ``range_positions`` excluding bare ``(:)``
-        (i.e. ``RangeIndex((None, None))``) ranges.
+        Return ordinal indices into ``range_positions`` whose corresponding
+        dimension is *not* a bare ``(:)`` (i.e. ``RangeIndex((None, None))``).
         """
         return [
             i for i, j in enumerate(range_positions)
@@ -708,18 +708,18 @@ class ResolveVectorNotationTransformer(Transformer):
         lhs_array = stmt.lhs
         lhs_dims = lhs_array.dimensions
         lhs_range_positions = self._find_range_positions(lhs_dims)
-        lhs_qualified_range_indices = self._find_qualified_range_indices(
+        lhs_qualified_positions = self._find_qualified_range_positions(
             lhs_dims, lhs_range_positions
         )
 
         # --- Step 4: Filter to resolvable dimensions ---
         if self.resolve_implicit_rhs_ranges:
-            resolvable_dim_indices = lhs_qualified_range_indices
+            resolvable_dim_indices = lhs_qualified_positions
         else:
             resolvable_dim_indices = [
-                i for i, j in enumerate(lhs_qualified_range_indices)
+                j for j in lhs_qualified_positions
                 if all(
-                    array_dims[rhs_pos[i]] != sym.RangeIndex((None, None))
+                    array_dims[rhs_pos[j]] != sym.RangeIndex((None, None))
                     for array_dims, rhs_pos in zip(rhs_dims_per_array, rhs_range_positions_per_array)
                 )
             ]
--- a/loki/transformations/array_indexing/vector_notation.py
+++ b/loki/transformations/array_indexing/vector_notation.py
@@ -108,7 +108,8 @@ def add_explicit_array_dimensions(routine):
     routine.body = SubstituteExpressions(array_map).visit(routine.body)
 
 
-def resolve_vector_notation(routine, resolve_implicit_rhs_ranges=True):
+def resolve_vector_notation(routine, resolve_implicit_rhs_ranges=True,
+                            substitute_derived_type_bounds=False):
     """
     Resolve implicit vector notation by inserting explicit loops.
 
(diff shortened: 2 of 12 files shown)


MichaelSt98:
Two things that most likely need some changes:

1. I currently add a comment `! loki resolved vector notation'` which we probably don't want to keep or make optional
2. the change in `fgen.py`

```py
    def visit_Conditional(self, o, **kwargs):
        # TODO: hack ..., this is a necessary but not a sufficient criteria
        if o.inline and len(o.body) <= 1:
            # No indentation and only a single body node
 ```

File: loki/backend/fgen.py [hunk 1]
```diff
@@ -681,7 +681,8 @@ def visit_Conditional(self, o, **kwargs):
             [...body...]
           END IF [name]
         """
-        if o.inline:
+        # TODO: hack ..., this is a necessary but not a sufficient criteria
+        if o.inline and len(o.body) <= 1:
```
reuterbal: Not sure I understand this. Do we have situations where `o.inline == True && len(o.body) != 1`? That is in my perspective an invalid IR and should be prevented to occur in the first place, not be hacked in the backend.

Probably first thing to add would be an assert like
```
if self.inline:
    assert len(self.body) == 1
```
in the post_init:
https://github.com/ecmwf-ifs/loki/blob/16e964b834018091a4b35d434742a0503b7704f5/loki/ir/nodes.py#L744

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -121,19 +173,71 @@ def resolve_vector_notation(routine):
...
+        if elem.split('%', maxsplit=1)[0] in variable_map:
+            return routine.resolve_typebound_var(elem, variable_map)
+        return None
+
+    bounds = ()
+    variable_map = routine.variable_map
+    valid_lower = [get_valid(_lower, variable_map) for _lower in lower]
+    valid_lower = [_ for _ in valid_lower if _ is not None]
+    valid_upper = [get_valid(_upper, variable_map) for _upper in upper]
+    valid_upper = [_ for _ in valid_upper if _ is not None]
```
reuterbal: To make this more concise, `variable_map` could be extracted before declaring the utility to save passing it around as an argument, and with walrus operator a single list comprehension would suffice:

```suggestion
    variable_map = routine.variable_map

    def get_valid(elem):
        if isinstance(elem, str) and elem.isnumeric():
            return sym.Literal(int(elem))
        if elem.split('%', maxsplit=1)[0] in variable_map:
            return routine.resolve_typebound_var(elem, variable_map)
        return None

    bounds = ()
    valid_lower = [valid_lower for _lower in lower if valid_lower := get_valid(_lower) is not None]
    valid_upper = [valid_upper for _upper in upper if valid_upper := get_valid(_upper) is not None]
```

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -146,18 +199,25 @@ def resolve_vector_dimension(routine, dimension, derive_qualified_ranges=False):
...
     index = get_integer_variable(routine, name=dimension.index)
-    bounds = get_loop_bounds(routine, dimension=dimension)
+
+    _lower = as_tuple(dimension.lower) + ('1',)
+    _upper = as_tuple(dimension.upper) + as_tuple(dimension.sizes)
+    bounds = _get_all_valid_loop_bounds(routine, lower=_lower, upper=_upper)
 
     # Map any range indices to the given loop index variable
-    loop_map = {sym.RangeIndex(bounds): index}
+    loop_map = {sym.RangeIndex(_bounds): index for _bounds in bounds}
```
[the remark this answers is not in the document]
reuterbal: A warning is printed instead, which is better.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
         if self.derive_qualified_ranges:
             shape_mapper = IterationRangeShapeMapper()
             stmt._update(lhs=shape_mapper(stmt.lhs), rhs=shape_mapper(stmt.rhs))
 
-        # Replace all range indices with loop indices and collect the corresponding mapping
-        index_mapper = IterationRangeIndexMapper(
-            loop_map=self.loop_map, basename=f'i_{stmt.lhs.basename}', scope=self.scope,
-            map_unknown_ranges=self.map_unknown_ranges
+        # --- Step 3: Identify range-indexed dimensions ---
+
```
reuterbal, line 644: The IterationRangeIndexMapper was introduced to unify behaviour between Assignment and MaskedStatement. Do we now diverge here again or would it be possible to bake the logic below into the mapper directly? For example, by amending this with an additional argument that provides the list of resolvable dimensions to accommodate the newly gained functionality?

_Edit:_ concrete suggestion further down

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
+            lhs_dims, lhs_range_positions
         )
-        stmt._update(lhs=index_mapper(stmt.lhs), rhs=index_mapper(stmt.rhs))
 
-        # Record all newly create loop index variables,
-        # so that we can declare them in the outer context
-        index_range_map = index_mapper.index_range_map
+        # --- Step 4: Filter to resolvable dimensions ---
+        if self.resolve_implicit_rhs_ranges:
+            resolved_dim_indices = lhs_qualified_range_indices
```
reuterbal: I would suggest calling this `resolvable_dim_indices` here, and `resolved_dim_indices` only after the `actually_resolved` evaluation.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
+        # --- Step 3: Identify range-indexed dimensions ---
+
+        # RHS arrays that have at least one RangeIndex dimension
+        rhs_vars = FindVariables(unique=False).visit(stmt.rhs)
+        rhs_arrays = [
+            var for var in rhs_vars
+            if isinstance(var, sym.Array)
+            and any(isinstance(dim, sym.RangeIndex) for dim in var.dimensions)
+        ]
+        rhs_arrays_dims = [array.dimensions for array in rhs_arrays]
```
reuterbal:
```suggestion
        rhs_dims_per_array = [array.dimensions for array in rhs_arrays]
```
for consistency?

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
-        # so that we can declare them in the outer context
-        index_range_map = index_mapper.index_range_map
+        # --- Step 4: Filter to resolvable dimensions ---
+        if self.resolve_implicit_rhs_ranges:
+            resolved_dim_indices = lhs_qualified_range_indices
+        else:
+            resolved_dim_indices = [
+                i for i, j in enumerate(lhs_qualified_range_indices)
+                if all(
+                    array_dims[rhs_pos[i]] != sym.RangeIndex((None, None))
```
reuterbal: Is the `i` correct here or should this be `j`? Because `i` is the relative position of a dimension in `lhs_qualified_range_indices` while `j` is the actual dimension's position.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
+                    for array_dims, rhs_pos in zip(rhs_arrays_dims, rhs_range_positions_per_array)
+                )
+            ]
```
reuterbal, line 679: To double-check I understand this correctly: we have a list of dimension expressions from the lhs, and (possibly multiple) lists of dimension expressions from the rhs; and we're building the list of dimensions that have not `:` in any of these lists. (And store the position of each such dimension).

If that's correct, it might be a little more accessible to make this explicit, e.g. like this?

```suggestion
            rhs_qualified_range_indices_per_array = [
                self._find_qualified_range_indices(rhs_dim, rhs_range_pos)
                for rhs_dim, rhs_range_pos in zip(rhs_arrays_dims, rhs_range_positions_per_array)
            ]
            resolved_dim_indices = [
                pos for pos, idx in enumerate(lhs_qualified_range_indices)
                if all(idx in qualified_ranges for  qualified_ranges in rhs_qualified_range_indices_per_array)
            ]
```

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
+                    # that each dimension gets its own distinct loop index.
+                    if ivar in index_range_map:
+                        if not map_unknown_ranges:
+                            continue
+                        vtype = SymbolAttributes(BasicType.INTEGER)
+                        ivar = sym.Variable(name=f'{basename}_{i}', type=vtype, scope=scope)
+                        synthesized_ivars.add(ivar)
+                else:
+                    # Skip if we're not supposed to create new indices
+                    if not map_unknown_ranges or dim == sym.RangeIndex((None, None)):
```
reuterbal, line 506: The amount of `dim == sym.RangeIndex((None, None))` comparisons makes me wonder if we shouldn't add the StrCompareMixin to RangeIndex nodes to simplify these types of checks? (Cc @mlange05 )

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
+
+        Returns
+        -------
+        expression
+            The offset-adjusted index expression, simplified.
+        """
+        return simplify(sym.Sum((loop_var, sym.Product((-1, lhs_range.lower)), rhs_range.lower)))
+
+    @staticmethod
+    def _map_ranges_to_indices(dims, loop_map, map_unknown_ranges=True, basename='i', scope=None):
```
reuterbal, line 452: This is very similar to IterationRangeIndexMapper. Could we not amend that and use it instead of replicating the functionality?

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -279,48 +408,422 @@ class ResolveVectorNotationTransformer(Transformer):
...
+        for array, resolved_rhs_ranges in zip(rhs_arrays, resolved_rhs_ranges_per_array):
+            new_rhs_dims = []
+            for lhs_range, new_lhs_dim, rhs_range in zip(
+                resolved_lhs_ranges, new_lhs_dims, resolved_rhs_ranges
+            ):
+                if (lhs_range == rhs_range
+                        or rhs_range == sym.RangeIndex((None, None))
+                        or isinstance(lhs_range, sym.RangeIndex)
+                        and isinstance(rhs_range, sym.RangeIndex)
+                        and lhs_range.lower == rhs_range.lower):
```
reuterbal: Had to reformat this for me to disect the condition. Maybe the following would be more readable?

```suggestion
                is_aligned_dim = (
                    lhs_range == rhs_range or rhs_range == sym.RangeIndex((None, None))
                ) or (
                    isinstance(lhs_range, sym.RangeIndex) and isinstance(rhs_range, sym.RangeIndex) and
                    isinstance lhs_range.lower == rhs_range.lower
                )
                
                if is_aligned_dim:
```

loki/transformations/single_column/base.py
```diff
@@ -80,27 +80,58 @@ def transform_subroutine(self, routine, **kwargs):
             Subroutine to apply this transformation to.
         role : string
             Role of the subroutine in the call tree; should be ``"kernel"``
+            or ``"driver"``.
+
+        Notes
```
reuterbal, line 77: The notes section has to appear before `Parameters`, otherwise it will not be rendered

reuterbal (changes requested):
Many thanks for this expansion of the vector notation resolver transformation utility. This is a comprehensive addition to the functionality and I had to go through the steps in detail to understand each of their purpose - hopefully successfully. In the process I've left a few remarks how some of it might be possible to simplify or base on existing utilities.

loki/backend/fgen.py
MichaelSt98: This can happen if there is e.g. a vector notation assignment in an inline conditional which is resolved. Would that be solved via `__post_init__` modifications?

loki/transformations/array_indexing/vector_notation.py
MichaelSt98: Sure but I think it makes more sense to go the opposite way which gives us some functionality for masked statements for free.

loki/backend/fgen.py
reuterbal: Thanks, understood and agreed that this is a possibility.

The post-init should only ensure it is a valid IR (and can be helpful by raising an error when it is not), but making it valid needs to be done by the transformation. For that case, I think the resolve_vector_notation needs to be amended such that it sets the `inline` property to false.

loki/transformations/array_indexing/vector_notation.py
reuterbal: Yes, that would be fine, too - removing the mapper for masked statements and replacing it with this.

reuterbal (approved):
Many thanks, this looks great now! GTG from my side

loki/transformations/array_indexing/vector_notation.py
mlange05: I'm pretty sure it is already-string comparable, the AI just doesn't use string comparison by default; `dim == ':'` should "just work" here.

MichaelSt98: I'm afraid I can't blame the LLM here since this PR was mostly written by myself and just refactored with the help of the LLM 😃 want me to change it or do a follow up PR?

mlange05: [no-action] Lolz, certainly looked AI-formatted in many places. 😜 Doesn't necessarily have to be done here, but in general, we should prefer `== ':'` where possible. I'd also think `sym.RangeIndex()` should default to the same thing, but that's for a separate follow-on.

mlange05 (approved):
Looks good - quite a substantial re-write and with good test coverage. GTG for me. 👍 :shipit:

PR #655, opened 2026-03-06
Loki: Bug fixing on _get_procedure_item

rommelDB:
- The issue on line 552 is trivial, since item_name is not declared at that point, we got a 'referenced before assignment' error.

- Then, somehow the output from get_or_create_module_definitions_from_candidates might be a list with duplicated items, so we could get something like: 'RuntimeError: Procedure XXX defined in multiple imported modules: YYY, YYY'. So, it might be better to handle the list as a set.

### DIFF
--- a/loki/batch/item_factory.py
+++ b/loki/batch/item_factory.py
@@ -542,16 +542,16 @@ class ItemFactory:
         if unqualified_imports:
             # We try to find the ProcedureItem in the unqualified module imports
             module_names = [imprt.module for imprt in unqualified_imports]
-            candidates = self.get_or_create_module_definitions_from_candidates(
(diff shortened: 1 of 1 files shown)


reuterbal (commented):
I agree on the change in l. 552, that's a bug.

However, do you have a reproducer for the multiple module names? If there's erroneous behaviour right now we ought to add a matching test to ensure such behaviour cannot be reintroduced in the future.

Looking at the implementation, the candidate lookup should only consider modules from unqualified imports, and then procedures defined in these. To me it seems like this would either require the same module to be imported twice via an unqualified import, or the same procedure to be defined multiple times in the module? Happy to be proven wrong of course!

PR #656, opened 2026-03-09
IR: Refactoring and introduction of explicitly names "statement nodes"

mlange05:
_Note: This PR technically introduces an externally API-changing change, as it renames `ir.Intrinsic` -> `ir.GenericStmt`_

This PR refactors the `loki.ir.nodes` source into a dedicated sub-package and introduces a new set of statement nodes that all inherit from `loki.ir.nodes.GenericStmt`. These replace previous `ir.Intrinsic` nodes (a misnomer from the early days 😉 ), which in turn allows us to refer to specific statement nodes explicitly by type (eg. `ir.ImplicitStmt` or `ir.ContainsStmt`). It is worth pointing out that I've only specialised `CONTAINS`, `IMPLICIT`, `CYCLE`, `RETURN`, `GO TO`, `PUBLIC`, `PRIVATE` thus far, but others could follow. 

Pleas note also that the latter two now exist as dedicated node types, but I've not changed (yet) the internal mechanics of the `Module` that use this information.

In some more detail:
* Moved IR utilities (including `dataclass_strict` to `loki.tools`
* Separated `loki.ir.nodes` into sub-package with sources for `abstract_nodes`, `internal_nodes` and `leaf_nodes`
* Renamed `loki.ir.Intrinsic` => `loki.ir.GenericStmt`
* Added source for `stmt_nodes` to `loki.ir.nodes` and implemented `ImplicitStmt`, `ContainsStmt`, `ReturnStmt`, `CycleStmt`, `GotoStmt`, `PrivateStmt` and `PublicStmt`
* Adjusted used of generic statement nodes in both code and test base
* Consolidated some statement tests in `ir.nodes.tests` to capture new node types and avoid JIT
* Some more import massaging in a test files that I had to touch anyway


mlange05:
@reuterbal I've rebased this over recent main (0.3.6) for now and added a few more (less commonly used) statement node types. Seeing this, it might make sense to wait for PR #665 and then expand the explicit frontend use of these uncommon types?

mlange05:
Ok, @reuterbal I've rebased and added some bits to the new frontend test structure. There's more to be done there, but this one was already getting big, so I'll keep it to what it is now. Ready for another look.

loki/backend/tests/test_fgen.py
```diff
@@ -69,7 +69,7 @@ def test_fgen_literal_list_linebreak(frontend, tmp_path):
     # Make sure all lines are continued correctly
     code = module.to_fortran()
     code_lines = code.splitlines()
-    assert len(code_lines) in (35, 36) # OMNI produces an extra line
+    assert len(code_lines) in (35, 37) # OMNI produces an extra line
```
reuterbal, line 72: 🤨

loki/ir/nodes/stmt_nodes.py
```diff
@@ -0,0 +1,404 @@
...
+    @classmethod
+    def ensure_tuple(cls, value):
+        return sanitize_tuple(value)
+
+    def __repr__(self):
+        return f'Save:: {truncate_string(self.text)}'
+
+
+@dataclass_strict(frozen=True)
+class PublicStmt(GenericStmt):
```
reuterbal, line 116: Not sure if I 100% like the very Fortran-string-equivalent representation of statement nodes, as it abandons the user-friendly view onto execution concepts that Loki encapsulated so far. It removes some of the higher-level approach over an AST representation that we tried to maintain so far.

Specifically, I'm concerned about `PublicStmt` vs. `PrivateStmt` - both of which could be classified as a `VisibilityStmt` with a positive vs. negative default. A transformation or lint rule that is interested in visibility of symbols will now have to look for both node types instead of searching directly for one type only.

loki/ir/nodes/stmt_nodes.py
```diff
@@ -0,0 +1,404 @@
...
+    @classmethod
+    def ensure_tuple(cls, value):
+        return sanitize_tuple(value)
+
+    def __repr__(self):
+        return f'Common:: {truncate_string(self.text)}'
+
+
+@dataclass_strict(frozen=True)
+class ContainsStmt(GenericStmt):
```
reuterbal, line 194: Similarly for `ContainsStmt`: Semantically, this does not add any additional information other than delineating the beginning of the contains part. Loki's IR already captures this through the contains property and I'm wondering if a `ContainsSection` as an overload of the internal node type `Section` wouldn't be a more intuitive way of representing this concept without the need to explicitly represent this as a single-word-string-only node?

loki/ir/nodes/stmt_nodes.py
```diff
@@ -0,0 +1,404 @@
...
+    keyword = 'GO TO'
+
+    text: str
+
+    def __repr__(self):
+        return f'Goto:: {truncate_string(self.text)}'
+
+
+@dataclass_strict(frozen=True)
+class PrintStmt(GenericStmt):
```
reuterbal, line 358: I suppose we'd want a `WriteStmt`, too, at some point, but that would constitute a larger impact due to it's use in transformations?

Also here, wondering if a generic `OutputStmt` might not be a better fit?

loki/frontend/fparser.py
```diff
@@ -3477,29 +3487,49 @@ def visit_Parenthesis(self, o, **kwargs):
...
+        return ir.GotoStmt(text=label, **kwargs)
+
+    def visit_Exit_Stmt(self, o, **kwargs):
+        return ir.ExitStmt(o.items[1], **kwargs)
+
+    def visit_Stop_Stmt(self, o, **kwargs):
+        return ir.StopStmt(**kwargs)
+
+    visit_Intrinsic_Stmt = visit_Generic_Stmt
+    visit_Format_Stmt = visit_Generic_Stmt
```
reuterbal: Did I not spot a `FormatStmt` class in `stmt_nodes.py`?

loki/frontend/fparser.py
```diff
@@ -3239,14 +3245,18 @@ def visit_Include_Stmt(self, o, **kwargs):
...
+            return ir.ImplicitStmt(text=o.items[0], **kwargs)
+        content = tuple(i if isinstance(i, str) else self.visit(i, **kwargs) for i in o.items)
+        return ir.ImplicitStmt(text=content, **kwargs)
 
     def visit_Print_Stmt(self, o, **kwargs):
         # NOTE: fparser returns None for an empty print (`PRINT *`) instead of
         #       the usual `Output_Item_List` entity.
-        return ir.Intrinsic(text=f'PRINT {", ".join(str(i) for i in o.items if i is not None)}',
-                            source=kwargs.get('source'), label=kwargs.get('label'))
+        return ir.GenericStmt(
```
reuterbal: Should this not generate a `PrintStmt`?

loki/frontend/omni.py
```diff
@@ -1286,49 +1293,49 @@ def visit_FstructConstructor(self, o, **kwargs):
...
-        return ir.Intrinsic(text=f'write({nargs}) {args}', source=kwargs['source'])
+        return ir.GenericStmt(text=f'write({nargs}) {args}', source=kwargs['source'])
 
     def visit_FprintStatement(self, o, **kwargs):
         values = [self.visit(v, **kwargs) for v in o.find('valueList')]
         args = ', '.join(f'{v}' for v in values)
         args = f", {args}" if values else ""
         fmt = o.attrib['format']
-        return ir.Intrinsic(text=f'print {fmt}{args}', source=kwargs['source'])
+        return ir.GenericStmt(text=f'print {fmt}{args}', source=kwargs['source'])
```
reuterbal: Shouldn't this be a `PrintStmt`?

loki/frontend/omni.py
```diff
@@ -1286,49 +1293,49 @@ def visit_FstructConstructor(self, o, **kwargs):
...
 
     def visit_FformatDecl(self, o, **kwargs):
         fmt = f'FORMAT{o.attrib["format"]}'
-        return ir.Intrinsic(text=fmt, source=kwargs['source'])
+        return ir.GenericStmt(text=fmt, source=kwargs['source'])
```
reuterbal: Shouldn't this be a `FormatStmt`?

reuterbal (changes requested):
Thanks! 

The refactoring of IR nodes into multiple files is a very welcome change. Unfortunately, the Github diff view doesn't make it very easy to see where things have simply moved or where other changes have been incorporated, too - but I trust that this has been executed with caution and that tests would have failed if something was out of the ordinary.

I have flagged a few cases where I think the frontends are not correctly translating to the new statement node types but instead create "old-fashioned" intrinsic/generic nodes. That should be straightforward to fix.

The only conceptual question for me is the fact, that the new statement nodes are not capturing execution/programming concepts in the way we used to do this as much as possible in the Loki IR, but are very much aligned with the specific Fortran-way of expressing such concepts. I flagged this in the respective places and would be interested in your thoughts on this. I am not fundamentally opposed to doing it in the way this is implement now, but wanted to point out the change this constitutes and ensure that it is a concious decision to design it that way.

loki/frontend/fparser.py
mlange05: Yes, it's more involved than expected, but yes - please see top commit after rebase.

loki/ir/nodes/stmt_nodes.py
mlange05: As discussed offline, in principle yes, but due to the various flavours of string and output formatting it's quite hard to generalise. So, we might look into generalising output (and file handling) intrinsics at a later stage, but defer to a future PR for this.

loki/ir/nodes/stmt_nodes.py
mlange05: As discussed offline: https://github.com/ecmwf-ifs/loki/issues/689

loki/ir/nodes/stmt_nodes.py
mlange05: As discussed offline: https://github.com/ecmwf-ifs/loki/issues/690

reuterbal (approved):
Many thanks for the discussions and actioning the suggested modifications. As discussed offline, I agree that this strikes a sensible balance between semantic representation and Fortran syntax specifics. In any case, this provides a great improvement of the IR and the representation of the catch-all `Intrinsic` node bucket. Fine tuning can come later and the spun-out issues capture the most promising avenues.

PR #657, opened 2026-03-09
Fix inclusion of internal functions to scheduler

awnawab:
This PR fixes the logic for including internal functions in the scheduler so that now they are only included upon request.


reuterbal (approved):
Apologies, silly oversight on my end. Thanks for fixing!

PR #660, opened 2026-03-29
IR: Add direct sequence semantics for Section and Associate

### LINKED ISSUE #106: Make `InternalNode` iterable
`InternalNode` encapsulates a tuple of IR nodes, which often results in awkward indirection because it requires writing `internal_node.body` quite often. Making this iterable would allow easier subscripting etc., such as `internal_node[i]` or `node in internal_node`.

_Originally posted by @mlange05 in https://github.com/ecmwf-ifs/loki/pull/95#discussion_r1235181518_

mlange05:
## Summary
- add direct sequence-style access for `ir.Section` and `ir.Associate`, including iteration, indexing, containment, and `Section.index(...)`
- preserve existing helper and transformer behavior by treating iterable IR block nodes as atomic in generic utilities such as `as_tuple` and `is_iterable`
- adopt the new API across transformation code and tests by replacing many `routine.body.body` / `routine.spec.body` access patterns with direct `Section` access

## Why
Issue #106 asked for less awkward access to IR block nodes without having to repeatedly reach into `.body`. This change makes the most common block-like nodes easier to work with while avoiding the broader compatibility problems that come from making all `InternalNode` subclasses generically iterable.

## What Changed
- `ir.Section`
  - supports direct iteration, indexing, containment, length, and `index(...)`
- `ir.Associate`
  - inherits the same direct sequence-style access as a scoped block node
  - retains truthy node semantics for scope-sensitive code paths
- `loki.tools.util`
  - keeps iterable IR block nodes atomic for helper logic that distinguishes nodes from replacement sequences
- tests
  - add regression coverage for `Section` and `Associate` sequence behavior
  - add transformer compatibility coverage to ensure iterable block nodes are still treated as atomic replacements
  - update transformation and test code to use direct section accessors

## Validation
Ran:

```bash
. ./loki-activate
python -m pytest   loki/ir/tests/test_ir_nodes.py   loki/ir/tests/test_transformer.py   loki/transformations/single_column/tests/test_scc_vector.py   loki/transformations/tests/test_utilities.py   loki/transformations/data_offload/tests/test_offload_deepcopy.py   loki/tests/test_subroutine.py   loki/tests/test_sourcefile.py   loki/batch/tests/test_transformation.py   loki/backend/tests/test_fgen.py   loki/types/tests/test_derived_types.py   loki/expression/tests/test_expression.py::test_stmt_func_heuristic[fp]   loki/frontend/tests/test_frontends.py::test_associates[fp]   loki/ir/tests/test_scoped_nodes.py::test_scoped_node_get_symbols[fp]   loki/ir/tests/test_scoped_nodes.py::test_scoped_node_parse_expr[fp]   loki/ir/tests/test_visitor.py::test_attach_scopes_associates[fp]   loki/transformations/sanitise/tests/test_associates.py
```

Results:
- focused IR / transformation cleanup suite: `374 passed, 8 skipped, 10 xfailed`
- associate/scoping validation suite: `68 passed, 16 skipped`
- transformer suite after final associate cleanup: `36 passed`

### DIFF
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -426,6 +426,11 @@ class Section(InternalNode, _SectionBase):
     def __contains__(self, item):
         return item in self.body
 
+    def index(self, item, start=0, stop=None):
+        if stop is None:
+            return self.body.index(item, start)
+        return self.body.index(item, start, stop)
+
     def append(self, node):
         """
         Append the given node(s) to the section's body.
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -499,20 +499,8 @@ class Associate(ScopedNode, Section, _AssociateBase):  # pylint: disable=too-man
 
     _traversable = ['body', 'associations']
 
-    def __iter__(self):
-        raise TypeError('Associate nodes can not be traversed. Try `body` instead.')
-
     def __bool__(self):
         return True
-
-    def __getitem__(self, index):
-        raise TypeError('Associate nodes do not support item access. Try `body` instead.')
-
-    def __len__(self):
-        raise TypeError('Associate nodes do not define a sequence length. Try `body` instead.')
-
-    def __contains__(self, item):
-        raise TypeError('Associate nodes do not support sequence membership. Try `body` instead.')
 
     def __post_init__(self, parent=None):
         super(ScopedNode, self).__post_init__(parent=parent)
--- a/loki/ir/nodes.py
+++ b/loki/ir/nodes.py
@@ -414,6 +414,18 @@ class Section(InternalNode, _SectionBase):
     Internal representation of a single code region.
     """
 
+    def __iter__(self):
+        return iter(self.body)
+
+    def __getitem__(self, index):
+        return self.body[index]
+
+    def __len__(self):
+        return len(self.body)
+
+    def __contains__(self, item):
+        return item in self.body
+
     def append(self, node):
         """
         Append the given node(s) to the section's body.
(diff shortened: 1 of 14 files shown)


mlange05:
This is an alternative implementation to fix issue #106, which restricts native iterability and indexing to `Section` and `Associates` nodes. Fully agentic, including exploration of alternative approaches and subsequent API usage cleanup.

loki/tools/util.py
```diff
@@ -45,6 +45,20 @@
 ]
 
 
+def _is_atomic_iterable_ir_node(obj):
+    """
+    Return ``True`` for iterable IR nodes that should stay atomic in helpers.
+    """
+    nodes_module = sys.modules.get('loki.ir.nodes')
```
awnawab: this is technically correct, but aren't the safety checks essentially redundant; I can't imagine a world where `Section` and `Associate` don't exist as ir nodes. And this kind of import is definitely wrong, we should just have `from loki.ir import nodes as ir` like we do throughout the rest of the code

awnawab (approved):
This is really great, a genuine improvement to loki's ease of use, many many thanks for this 👌🙏 I have one knitpicky comment but it's not a big deal so I'm approving anyway ✅

loki/transformations/single_column/scc_cuf.py
```diff
@@ -815,7 +815,7 @@ def driver_device_variables(self, routine, targets=None):
 
             allocations = FindNodes(ir.Allocation).visit(routine.body)
             if allocations:
-                insert_index = routine.body.body.index(allocations[-1]) + 1
+                insert_index = routine.body.index(allocations[-1]) + 1
             else:
                 insert_index = None
             # or: insert_index = routine.body.body.index(calls[0])
```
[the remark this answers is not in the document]
mlange05: Won't fix - this is a purely cosmetic change and comment pertains to something unrelated to this PR.

loki/tools/util.py
mlange05: Nice catch! Update now.

loki/ir/nodes.py
```diff
@@ -487,6 +504,9 @@ class Associate(ScopedNode, Section, _AssociateBase):  # pylint: disable=too-man
 
     _traversable = ['body', 'associations']
 
+    def __bool__(self):
+        return True
```
reuterbal, line 508: What's the underlying reason for this? PR summary mentions something related to scope-sensitive code paths but I don't quite understand.

Related just a random question: is there a situation where we would want to consider an Associate as falsey? E.g., if there's no body and no associations?

reuterbal (approved):
Thanks, this is great and I agree with @awnawab that this makes for a very pleasing improvement of the API by taking away the awkward daisy-chaining of body attributes. 
Nothing that stood out to me, so this should be good to go.

loki/ir/nodes.py
mlange05: Ok, I've had a quick look, and I think it comes from the fact that we check for the existence of an enclosing scope with `if scope:` in some places in the frontend, even when the body has not been populated yet. Without this override, the frontend would not recognise the existence of the enclosing scope when creating symbols, simply because its body is empty. I can add a comment to this effect, if that's preferable, but removing this creates test failures.

PR #521, opened 2025-04-01
Dataflow Analysis Refactor

Andrew-Beggs-ECMWF:
Split off of #515 and implemented suggested changes.

This PR refactors dataflow analysis into a framework that can handle multiple kinds of analysis.

PR #515, opened 2025-03-18
Constant Propagation Transform

Andrew-Beggs-ECMWF:
Putting here for visibility and getting the CI to run. Still WIP (including this description).

Adds:
1. Constant Propagation Analysis
  - `ConstantPropagationAnalysis`
3. Constant Propagation Transformation
  - `ConstantPropagationTransformer`

Changes:
1. Analysers now inherit from the abstract `AbstractDataflowAnalysis`
  - This is instead of there only being `analyse_dataflow`, which is now under `DataFlowAnalysis`
  - Introduces a standardise format for analysers

Todo:
- [ ]  Write todos

PR #663, opened 2026-04-01
Replay constant propagation on top of rebased DFA framework

mlange05:
## Summary

- replay the constant-propagation work from #515 on top of the rebased DFA framework from #662
- add a `ConstantPropagationAnalysis` and `ConstantPropagationTransformer`
- support constant propagation through assignments, conditionals, loops, and dynamic array invalidation
- include the loop-unroll optimization path needed by the replayed implementation
- bring over and validate the relevant constant-propagation test coverage

## Important

This PR is based on top of #662 and therefore requires #662 to be merged first.

## Background

This PR is a modern replay of the original constant-propagation work proposed in #515.

The original branch for #515 was stacked on top of the older DFA refactor from #521. Since that historical branch diverged, the underlying DFA implementation and related infrastructure changed substantially on `main`. Replaying the feature on top of the rebased DFA framework from #662 was therefore safer and easier to review than attempting a literal rebase of the original branch.

## What This PR Changes

- add `loki/analyse/constant_propagation_analysis.py`
- add `loki/transformations/constant_propagation.py`
- add and expand `loki/transformations/tests/test_constant_propagation.py`
- implement constant propagation through:
  - declaration initializers
  - assignment rewriting
  - conditional branch merging
  - loop handling, including no-unroll cases
  - dynamic array access invalidation
- add the optional loop-unroll fast path needed by the replayed constant-propagation implementation

## Why

The goal of this PR is to restore the constant-propagation feature originally developed in #515, but on top of the current Loki code base and the rebased DFA architecture from #662.

With #662 in place, constant propagation can build on explicit DFA extension points and a clearer analysis structure, instead of depending on the older stacked branch layout.

## Relationship To Earlier Work

This PR is a replay of the ideas and functionality from:

- #515
- built on top of the rebased DFA framework from #662
- with the architectural foundation originally motivated by #521

## Acknowledgements

This PR is based on the original constant-propagation work in #515 by @Andrew-Beggs-ECMWF.

It also depends directly on the rebased DFA framework introduced in #662, which itself replays the architectural direction of #521.

This replay does not preserve the original commit history directly, but it intentionally carries forward the core functionality and design goals of the original contribution.

Many thanks to @Andrew-Beggs-ECMWF for the original implementation and for the stacked development that motivated this replay.


Andrew-Beggs-ECMWF:
Happy to help improve this where required. Some things that come to mind are:

1. loki/analyse/constant_propagation_analysis.py should ideally have 100% test coverage because of how easy it is to introduce subtle bugs (ask me how I know)
2. There is currently a bug in this somewhere. I can't remember exactly what it was, but when I used this on cloudsc, there was some loop where it propagated something it shouldn't have. I _think_ it was ZEXPLICIT loop where it propagated ZEXPLICIT=0 inside the loop, so it never accumulated, but I'd have to check.
3. Not all Fortran constructs are supported. I'm not sure it's worth waiting on these to be implemented and to have branches diverge again. Perhaps ship it as is with unimplemented nodes raising exceptions?

mlange05:
Ok, following the first stage agentic adaptation of the original feature implementation I went back to refactor this entire thing a little and bring it more in line with our more widespread conventions in the `loki.transformation` package. Importantly, this changes a few fundamental things about this feature:
* Instead of implementing yet-another-expression-simplifier, the `ConstantPrropgationMapper` now inherits from the base `SimplifyMapper` and overrides the integer-division convention. This, however, requires the fixes in #688 to get the dedicated mapper tests to pass.
* The analysis and synthesis of this transformation have now been merged into a pure `Transformer`, meaning we do not need to attach the constant map onto IR nodes anymore - and thus we do not need to hijack the dataflow analysis either. Note, this might allow us to revert so of the recent encapsulation changes - but I leave that discussion for a follow-on PR.
* In addition to full-scale `Transformer`, I've also added an entry-point method that takes the `unroll_loops` flag, as before. This will ensure that loop unrolling is triggered after an initial constant propagation pass, and triggers a second one afterwards, to ensure that unrolled assignments also inherit the right constants, if these are known.

And final comment, I don't think this feature is entirely complete or fully tested yet - but I believe it's in a good enough condition to merge. Therefore, @reuterbal @awnawab I'd appreciate a review to get this over the line (with potential for more corner-case massaging afterwards. 

Also, many thanks to @Andrew-Beggs-ECMWF for the original implementation. I don't think a strict review is required for this, but if you have time for a quick look, comments are always welcome.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+    map_deferred_type_symbol = map_array
+
+
+class ConstantPropagationTransformer(Transformer):
+    """Apply constant-propagation analysis as a transformation driver."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        self.constants_map = {}
```
awnawab: This appears unused, and we seem to instead rely on looking up `constants_map` from the `kwargs`.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+        if isinstance(child, sym._Literal):
+            separated[0].append(child)
+        else:
+            separated[1].append(child)
+    return separated
+
+
+
+def _array_indices_to_accesses(dimensions, shape):
+    accesses = functools.partial(itertools.product)
```
awnawab: This is very fancy, and I learned a new thing, but perhaps this would be more readable:
```
  arg_lists = []
  for dimension in dimensions:
      if isinstance(dimension, sym.RangeIndex):
          arg_lists.append([sym.IntLiteral(v) for v in get_pyrange(...)])
      else:
          arg_lists.append([dimension])
  return itertools.product(*arg_lists)
```

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+    for count, dimension in enumerate(dimensions):
+        if isinstance(dimension, sym.RangeIndex):
+            start = dimension.start if dimension.start is not None else sym.IntLiteral(1)
```
awnawab: Couldn't these checks just be `if dimension.start`?

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+        return o._rebuild(lhs=new_lhs, rhs=new_rhs)
+
+    def visit_Conditional(self, o, **kwargs):
+        constants_map = kwargs.get('constants_map', {})
+        mapper = ConstantPropagationMapper()
+
+        new_condition = mapper(o.condition, constants_map=constants_map)
+
+        # Pass two copies of the constants map forward ...
+        with dict_override(kwargs, {'constants_map': deepcopy(constants_map)}):
```
awnawab, line 143: Might be worth adding a TODO here to check if we know whether the condition evaluates to true or false, in which case we can make the constant propagation a bit less conservative. Classic example would be after inlining when `if(present(arg))` checks are actually resolved.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+            _, non_literal_dimensions = _separate_literals(new_dimensions)
+            if non_literal_dimensions:
+                _pop_array_accesses(new_lhs, constants_map=constants_map)
+                return o._rebuild(lhs=new_lhs, rhs=new_rhs)
+
+        _, non_literals = _separate_literals((new_rhs,))
+        if not non_literals and not isinstance(new_lhs, sym.Array):
+            update_constants_map(new_lhs, new_rhs, constants_map)
+        else:
+            invalidate_constants_map(new_lhs, constants_map)
```
awnawab, line 132: I could be mistaken, but I don't think we are hitting this line in the case of an array with all literal dimensions and literal rhs.

loki/transformations/tests/test_constant_propagation.py
```diff
@@ -0,0 +1,321 @@
...
+def test_constant_propagation_conditional_basic(frontend):
+    fcode = """
+subroutine test_constant_propagation_conditional_basic(c)
+  integer :: a = 5
+  integer :: b = 3
+  logical :: cond = .true.
+  integer, intent(out) :: c
+
+  if (cond) then
+    c = a
```
awnawab, line 222: It would be good to add a second integer which is assigned the same value in both branches, and then checking if a `d=d` assignment is also resolved correctly after the conditional.

loki/transformations/tests/test_constant_propagation.py
```diff
@@ -0,0 +1,321 @@
...
+    assert declarations_map[('l', ())] == sym.LogicLiteral(True)
+    assert declarations_map[('b', (sym.IntLiteral(1),))] == sym.IntLiteral(2)
+    assert declarations_map[('b', (sym.IntLiteral(2),))] == sym.IntLiteral(3)
+
+
+def test_constant_propagation_mapper_folds_expressions():
+    mapper = ConstantPropagationMapper()
+
+    assert mapper(sym.Sum((sym.IntLiteral(1), sym.IntLiteral(2)))) == sym.IntLiteral(3)
+    assert mapper(sym.Quotient(sym.IntLiteral(7), sym.IntLiteral(2))) == sym.IntLiteral(3)
```
awnawab, line 41: Might be worth adding a test here for negative integer division.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+        ) > 0
+
+        if bounds_are_const:
+            if bounds_has_steps:
+                loop_constants_map = constants_map
+            else:
+                loop_constants_map = deepcopy(constants_map)
+
+            for assign in assignments:
+                if not set(FindVariables().visit(assign.rhs)).intersection(lhs_vars):
```
awnawab, line 199: Should we pop the relevant entry if `o.variable` is in the rhs? Specifically for a case like:
```
integer :: d = 0
do i = 1, 5     ! const bounds
  d = i * 2     ! loop-variant: RHS uses i which is in lhs_vars
end do
! outer map still has d = 0 here — but at runtime d = 10
```

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,285 @@
...
+                    assign_kwargs = dict(kwargs)
+                    assign_kwargs['constants_map'] = loop_constants_map
+                    self.visit_Assignment(assign, **assign_kwargs)
+        else:
+            for assign in assignments:
```
awnawab, line 204: This branch is also currently untested I believe.

loki/transformations/tests/test_constant_propagation.py
```diff
@@ -0,0 +1,321 @@
...
+    fcode = """
+subroutine test_constant_propagation_for_loop_basic(c)
+  integer :: a = 5
+  integer :: b = 3
+  integer :: i
+  integer, intent(out) :: c
+
+  c = 0
+  do i = 1, a
+    c = c + b
```
awnawab, line 258: The in loop accumutation is currently untested for the non unroll case, that would be a nice test to add.

loki/expression/symbolic.py
```diff
@@ -276,36 +276,41 @@ def flatten_expr(expr):
     return sym.Sum(as_tuple(done))
 
 
-def sum_int_literals(expr):
+def sum_int_literals(expr, integer_arithmetic=True, floating_point=False):
```
awnawab: [optional] Should this now be called `sum_literals`?

awnawab (changes requested):
Many thanks @Andrew-Beggs-ECMWF and @mlange05 for this very powerful transformation 🙏 I've left a bunch of nitpicky comments in the implementation and highlighted a few instances of missing coverage. Going by the PR description, these can probably be patched afterwards. However there is one corner case I believe where the constant propagation can be wrong for a loop with constant bounds and a loop variant assignment in the body. If I'm mistaken then please ignore and merge anyway, otherwise it would be good to fix that corner case before this is merged.

loki/transformations/constant_propagation.py
mlange05: Ah, yes, but we already have a dead-code removal transformation that would do this if applied straight after the constant propagation. Arguably this could be added as an optional step in the outer wrapper (much like the loop unrolling) to capture the cross-cutting effects (remove one if branch, all assignments in the other can be propagated). However, I think this could be left for a follow-up PR and I propose to raise an issue for this at this point?

loki/transformations/constant_propagation.py
mlange05: Good point, but again, I think this feature extension could be added in a follow-on PR. I propose parking this in an issue again?

mlange05: Edit: I think the "advanced" analysis of using the last-known value for `i` as a constant that could be added to the constant map would be a future extension - the incorrect invalidation is indeed a bug. Will fix ...

loki/transformations/tests/test_constant_propagation.py
reuterbal: could we also have one for, e.g., `7/(-2)` please?

loki/transformations/tests/test_constant_propagation.py
reuterbal: I think that' change is still pending?

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,278 @@
...
+    """ Mapper for expression-level constant replacement and folding. """
+
+    def map_array(self, expr, *args, **kwargs):
+        constants_map = kwargs.get('constants_map', {})
+        return constants_map.get((expr.basename, getattr(expr, 'dimensions', ())), expr)
+
+    def map_quotient(self, expr, *args, **kwargs):
+        """ Always force-evaluate integer-division """
+        if isinstance(expr.numerator, sym.IntLiteral) and isinstance(expr.denominator, sym.IntLiteral):
+            return sym.IntLiteral(float(expr.numerator.value) / float(expr.denominator.value))
```
reuterbal, line 96: Is there a specific reason for converting to float instead of using integer division directly: `expr.numerator.value // expr.denominator.value`?

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def update_constants_map(lhs, value, constants_map):
+    constants_map[(lhs.basename, ())] = value
+
+
+def invalidate_constants_map(lhs, constants_map):
+    if isinstance(lhs, sym.Array):
+        for access in tuple(key for key in constants_map if key[0] == lhs.basename):
+            constants_map.pop((lhs.basename, access), None)
```
reuterbal, line 56: These lines are currently not covered by tests.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,278 @@
...
+    literal_mask = [is_constant(dimension) for dimension in lhs.dimensions]
+    computable_dimension_mask = [is_constant(extent) for extent in new_shape]
+
+    masked_indices = []
+    ignore_mask = []
+    partial = False
+    for literal, computable, dimension in zip(literal_mask, computable_dimension_mask, lhs.dimensions):
+        if literal:
+            masked_indices.append(dimension)
+            ignore_mask.append(False)
```
reuterbal: Currently not covered by tests.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,278 @@
...
+    possible_accesses = _array_indices_to_accesses(masked_indices, new_shape)
+    keys = tuple(constants_map.keys())
+    for access in possible_accesses:
+        if partial:
+            for key in keys:
+                if key[0] == lhs.name and all(
+                        current == candidate or ignore
+                        for current, candidate, ignore in zip(key[1], access, ignore_mask)
+                ):
+                    constants_map.pop(key, None)
```
reuterbal: Also not covered by tests.

loki/transformations/constant_propagation.py
reuterbal: Also, just noticing this here but this is also the case for the original simplify utilities: I think we are silently dropping all kind specifiers in a few places. Might be worth noting this in an issue for future action.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,278 @@
...
+        )
+        bounds_has_steps = bounds_are_const and len(
+            get_pyrange(sym.LoopRange((new_bounds.start, new_bounds.stop, new_bounds.step)))
+        ) > 0
+
+        if bounds_are_const:
+            if bounds_has_steps:
+                loop_constants_map = constants_map
+            else:
+                loop_constants_map = deepcopy(constants_map)
```
reuterbal: Tests appear not to use this path at the moment.

loki/transformations/constant_propagation.py
```diff
@@ -0,0 +1,278 @@
...
+                        [sym.RangeIndex((None, None, None))] * len(symbol.shape), symbol.shape
+                    )
+                })
+            else:
+                declarations_map[(symbol.basename, ())] = symbol.initial
+        return declarations_map
+
+
+def do_constant_propagation(routine, unroll_loops=False):
+    """ Apply constant-propagation gover the body of a :any:`Subroutine`. """
```
reuterbal:
```suggestion
    """ Apply constant-propagation over the body of a :any:`Subroutine`. """
```

Since this is the main entry point, it would be good to slightly expand the docstring and, in particular, explain the `unroll_loops` option.

loki/transformations/transform_loop.py
```diff
@@ -5,10 +5,9 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-"""
-Collection of utility routines that provide loop transformations.
+"""Collection of utility routines that provide loop transformations."""
 
-"""
+# pylint: disable=cyclic-import
```
reuterbal: Not a fan of this. Where's the cycle coming from and could this be solved by moving a utility instead maybe?

reuterbal (changes requested):
Many thanks for this tedious rebasing and alignment work! Very much looking forward to bringing this finally in. I've left a few comments where it looks like test coverage is not complete, yet - maybe tests can be tweaked slightly to make sure we cover most relevant edge cases?

Otherwise a few minor observations, overall this is pretty mature.

I have also triggered a rebase over #696 to resolve the failing parametrisation test

loki/transformations/constant_propagation.py
mlange05: With direct integer division the negative-integer tests (see above) are failing. It's probably debatable what the correct behaviour is here (`sym.Integer(-7)` is not how we represent negative integers anyway), but it would be great if we can defer this discussion to an issue or follow-on PR.

I'll raise an issue for the dropped kind specifiers.

loki/transformations/transform_loop.py
mlange05: Redundant now - removed.

loki/transformations/constant_propagation.py
mlange05: See issue #701

mlange05:
@reuterbal Ok, I've had another pass at simplifying some of the more low-level expression logic, but I do not think 100% code coverage is possible here due to the historic nature of the PR content.

reuterbal (approved):
Many thanks for the diligent incorporation of suggestions. Testing is also very exhaustive now. To me this looks ready!

awnawab (approved):
Thanks a lot @mlange05 for addressing all the comments 🙏 G2G ✅

PR #667, opened 2026-04-13
Inline: fix for edge case mapping unbound dims

MichaelSt98:
### Description

This pull request fixes a bug in how array bounds are remapped when inlining subroutines with mismatched array bounds, and adds a regression test to ensure correct behavior. The most important changes are:

**Bug Fix: Array Bound Remapping**

* Updated the `_offset_lbound` function in `procedures.py` to correctly handle cases where the dimension is a `Range`, ensuring that both lower and upper bounds are individually adjusted rather than wrapping the range in a sum. This prevents incorrect expressions like `-1 + (1:n)` and instead generates the correct range `0:n-1`.

**Testing: Regression Test for 3-Layer Inlining**

* Added a new test, `test_inline_marked_subroutines_3layer_array_offset`, in `test_procedures.py` to verify that after inlining, array bounds are properly remapped in a three-layer call scenario, specifically checking that the bug does not reappear.
### Contributor Declaration

### DIFF
--- a/loki/transformations/inline/procedures.py
+++ b/loki/transformations/inline/procedures.py
@@ -124,7 +124,13 @@ def map_call_to_procedure_body(call, caller, callee=None):
                 else:
                     new_dimensions[indices[index]] = _offset_lbound(lower, decl_lbound, dim)
             else:
-                new_dimensions[indices[index]] = simplify(sym.Sum((dim, lbdiff)))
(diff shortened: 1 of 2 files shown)


reuterbal (commented):
Looks good to me, thanks!

reuterbal (approved):

mlange05 (approved):
Looks good. 👍 :shipit:

PR #669, opened 2026-04-22
Don't mark loops with constant bounds as sequential

awnawab:
The compiler may unroll a loop with constant bounds, which leaves dangling `!$acc loop seq` statements that lead to compiler errors. This PR amends the revector transformations so that loops with constant bounds are not annotated as sequential.

### DIFF
--- a/loki/transformations/single_column/revector.py
+++ b/loki/transformations/single_column/revector.py
@@ -8,7 +8,7 @@
 import re
 
 from loki.batch import Transformation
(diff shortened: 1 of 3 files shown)


mlange05 (approved):
Yes, looks good. GTG for me. 👍 :shipit:

reuterbal (approved):
Nice fix and well tested. Thanks, GTG!

PR #671, opened 2026-04-30
Various unused argument removal fixes and improvements

awnawab:
This PR contributes a series of fixes and improvements for the unused argument removal utility:
- Fix for edge-case when the same argument is passed to two dummy arguments (my inner Fortran enthusiast is in pain)
- Also remove unused arguments from inline calls
- Also remove unused arguments from internal routines
- Also account for internal routines when determining unused arguments

### DIFF
--- a/loki/transformations/remove_code.py
+++ b/loki/transformations/remove_code.py
@@ -225,10 +225,10 @@ def do_remove_unused_call_args(routine, unused_args_map):
         if call.routine is BasicType.DEFERRED or not unused_args_map.get(call.routine, None):
             continue
 
-        unused_args = [call.arguments[c] for c in unused_args_map[call.routine].values() if c < len(call.arguments)]
+        unused_positions = {c for c in unused_args_map[call.routine].values() if c < len(call.arguments)}
         unused_kwargs = [(kw, arg) for kw, arg in call.kwarguments if kw.lower() in unused_args_map[call.routine]]
 
-        new_args = [arg for arg in call.arguments if not arg in unused_args]
+        new_args = [arg for i, arg in enumerate(call.arguments) if i not in unused_positions]
         new_kwargs = [(kw, arg) for kw, arg in call.kwarguments if not (kw, arg) in unused_kwargs]
 
         call._update(arguments=as_tuple(new_args), kwarguments=as_tuple(new_kwargs))
--- a/loki/transformations/remove_code.py
+++ b/loki/transformations/remove_code.py
@@ -153,16 +153,54 @@ class RemoveCodeTransformation(Transformation):
         if self.remove_unused_args and (item := kwargs['item']):
             # collect unused args from successors
             successors = kwargs['sub_sgraph'].successors(item=item)
-            unused_args_map = {successor.ir: successor.trafo_data.get(self._key, {}).get('unused_args', {})
-                               for successor in successors}
+            unused_args_map = {}
+            for successor in successors:
+                succ_data = successor.trafo_data.get(self._key, {})
+                succ_ir = successor.ir
+                unused_args_map[succ_ir] = succ_data.get(succ_ir.name, {}).get('unused_args', {})
+
+            variable_map = routine.symbol_map
+            used_or_defined_symbols = OrderedSet()
+            if item.config.get('remove_unused_args', True) and kwargs['role'] == 'kernel':
+                # internal procedures are transformed after the parent routine,
+                # so determine their unused args eagerly to update call sites here
+                for child in routine.subroutines:
+                    child_symbol_map = child.symbol_map
+                    child_used_symbols = get_used_or_defined_symbols(child)
+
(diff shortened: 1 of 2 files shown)


loki/transformations/remove_code.py
```diff
@@ -153,16 +153,54 @@ def transform_subroutine(self, routine, **kwargs):
...
+            for successor in successors:
+                succ_data = successor.trafo_data.get(self._key, {})
+                succ_ir = successor.ir
+                unused_args_map[succ_ir] = succ_data.get(succ_ir.name, {}).get('unused_args', {})
+
+            variable_map = routine.symbol_map
+            used_or_defined_symbols = OrderedSet()
+            if item.config.get('remove_unused_args', True) and kwargs['role'] == 'kernel':
+                # internal procedures are transformed after the parent routine,
+                # so determine their unused args eagerly to update call sites here
```
reuterbal, line 167: since the transformation specifies `reverse_traversal=True`, I'm wondering if we shouldn't make sure that we also reverse the recursion into internal procedures in that case?
That would be a bug in the scheduler's behaviour

loki/transformations/remove_code.py
```diff
@@ -225,23 +263,70 @@ def do_remove_unused_call_args(routine, unused_args_map):
...
+        if call.routine is BasicType.DEFERRED or not unused_args_map.get(call.routine, None):
+            continue
+
+        unused_positions = {c for c in unused_args_map[call.routine].values() if c < len(call.arguments)}
+        unused_kwargs = [(kw, arg) for kw, arg in call.kwarguments if kw.lower() in unused_args_map[call.routine]]
+
+        new_args = tuple(arg for i, arg in enumerate(call.arguments) if i not in unused_positions)
+        new_kwargs = {kw: arg for kw, arg in call.kwarguments if not (kw, arg) in unused_kwargs}
+
+        inline_call_map[call] = call.clone(parameters=new_args, kw_parameters=new_kwargs)
```
reuterbal: Unless I missed something, the code is remarkably similar for CallStatements and InlineCalls, so I would suggest combining the two. 

Github's suggest feature is too stupid to handle it here and produces a really weird diff, so putting the code here like this:

```python
inline_call_map = {}
for call in chain(FindNodes(ir.CallStatement).visit(routine.body),  FindInlineCalls().visit(routine.body)):
    if call.routine is BasicType.DEFERRED or not unused_args_map.get(call.routine, None):
        continue

    unused_positions = {c for c in unused_args_map[call.routine].values() if c < len(call.arguments)}
    unused_kwargs = [(kw, arg) for kw, arg in call.kwarguments if kw.lower() in unused_args_map[call.routine]]

    new_args = tuple(arg for i, arg in enumerate(call.arguments) if i not in unused_positions)
    new_kwargs = tuple((kw, arg) for kw, arg in call.kwarguments if not (kw, arg) in unused_kwargs)

    if isinstance(call, ir.CallStatement):
        call._update(arguments=new_args, kwarguments=new_kwargs)
    else:
        inline_call_map[call] = call.clone(parameters=new_args, kw_parameters=dict(new_kwargs))
```

loki/transformations/remove_code.py
```diff
@@ -225,23 +263,70 @@ def do_remove_unused_call_args(routine, unused_args_map):
...
+    if inline_call_map:
+        routine.body = SubstituteExpressions(inline_call_map).visit(routine.body)
+
+
+def get_used_or_defined_symbols(routine):
+    """
+    Collect symbols used or defined in a routine, including array-shape references.
+    """
+    variable_map = routine.symbol_map
+    routine_arg_names = [arg.name.lower() for arg in routine.arguments]
```
reuterbal, line 288:
```suggestion
    routine_arg_names = [arg.lower() for arg in routine.argnames]
```

[No action] We should make the `routine.argnames` property lower-case by default.

reuterbal (approved):
This looks good to me. Two suggestions to make the code more concise - if you agree please feel free to push a commit to implement them.

loki/transformations/remove_code.py
awnawab: Yup that's much better indeed, I'll amend the PR accordingly. Thanks!

loki/transformations/remove_code.py
awnawab: That makes a lot of sense. Let me try doing that, and if all hell doesn't break lose in the scheduler/transformations tests then we can include it in this PR, otherwise I'll mop this up in a follow up PR.

awnawab: Having looked into this further, I realised that internal procedures can on a per item basis be included in the scheduler callgraph. However this will conflict with the `recurse_to_internal_procedures` setting applied across the entire `RemoveCodeTransformation`. Changing that will require careful testing across the IFS, so I would say that puts this development outside of the scope of the current PR and I've created #674 to track this cleanup.

reuterbal: Agreed. Thanks for this!

PR #675, opened 2026-05-05
Loki: Fixing spurious CI failures and skipping OMNI tests

mlange05:
### Description

This PR implements the key idea from PR #670 and then fixes a racy JIT issue in one of the transformation tests. I've also piggy backed a wide-ranging switch to skipping OMNI tests instead of xfailing them to reduce load on test runners. 

Please let me know if you'd rather review the second part in isolation. 

The racy JIT issue comes from the fact that we compile different modified versions of the same test in separate test directories. However, the Python entry module is cached in the dyn-loader, leading to a situation were consecutive runs try to load `parametrise.py` with different compiled signatures for `parametrise.f90`. The simplest fix for this is to add a UID for each test instance - more rigorous attempts at fixing this in the actual JIT-lib tools unfortunately just made things worse.

### Contributor Declaration

### DIFF
--- a/loki/analyse/tests/test_analyse_dataflow.py
+++ b/loki/analyse/tests/test_analyse_dataflow.py
@@ -542,7 +542,7 @@ end subroutine test
             assert assign.live_symbols == {'ia', 'ib', 'ic'}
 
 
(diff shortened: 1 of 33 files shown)


mlange05:
Turns out the agents were wrong ... 😭

PR #678, opened 2026-05-05
Loki: Mark OMNI as skip instead of xfail

mlange05:
### Description

Instead of marking OMNI tests as `xfail` we can skip them outright to speed up the CI test turnaround.

### Contributor Declaration

### DIFF
--- a/loki/backend/tests/test_fgen.py
+++ b/loki/backend/tests/test_fgen.py
@@ -143,7 +143,7 @@ end subroutine data_stmt
         assert '31*0' in spec_code
 
 
-@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Loki likes only valid code')]))
+@pytest.mark.parametrize('frontend', available_frontends(skip=[(OMNI, 'Loki likes only valid code')]))
 def test_multiline_inline_conditional(frontend):
     """
     Test correct formatting of an inline :any:`Conditional` that
@@ -177,7 +177,7 @@ end subroutine test_fgen
             assert len(line.split('&')[1]) > 60
 
 
-@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'Loki likes only valid code')]))
+@pytest.mark.parametrize('frontend', available_frontends(skip=[(OMNI, 'Loki likes only valid code')]))
 def test_multiline_inline_conditional_long(frontend):
     """
     Test correct formatting of an inline :any:`Conditional` that
@@ -320,7 +320,7 @@ END SUBROUTINE SPCMNEW
     assert 'procedure(real), pointer, intent(out) :: func' in source.to_fortran().lower()
 
 
-@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'SELECT TYPE not implemented')]))
+@pytest.mark.parametrize('frontend', available_frontends(skip=[(OMNI, 'SELECT TYPE not implemented')]))
 def test_fgen_select_type(frontend, tmp_path):
     fcode_module = """
 module select_type_mod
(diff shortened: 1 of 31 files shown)


mlange05:
Throughput improvements are only small, but logs look cleaner. Happy to drop if this is not desirable, @reuterbal

loki/backend/tests/test_fgen.py
```diff
@@ -320,7 +320,7 @@ def test_fgen_procedure_pointer(frontend, use_module, tmp_path):
     assert 'procedure(real), pointer, intent(out) :: func' in source.to_fortran().lower()
 
 
-@pytest.mark.parametrize('frontend', available_frontends(xfail=[(OMNI, 'SELECT TYPE not implemented')]))
+@pytest.mark.parametrize('frontend', available_frontends(skip=[(OMNI, 'SELECT TYPE not implemented')]))
```
reuterbal, line 323: While most others fail because OMNI is not capable itself, this particular one is failing because we haven't implemented the relevant bit in the Loki frontend.
I think it's fine to skip anyway -- I'm just pointing this out as it's not hinging on (unlikely) external fixes but on something under our control.

loki/frontend/tests/test_source_model.py
```diff
@@ -251,7 +250,7 @@ def test_sourcefile_lazy_construction(frontend, tmp_path):
         source.make_complete(frontend=frontend, xmods=[tmp_path])
     except CalledProcessError as ex:
         if frontend == OMNI and ex.returncode == -11:
-            pytest.xfail('F_Front segfault is a known issue on some platforms')
+            pytest.skip('F_Front segfault is a known issue on some platforms')
```
reuterbal: Why not skip the test entirely for OMNI?

loki/transformations/build_system/tests/test_file_write.py
```diff
@@ -147,7 +149,7 @@ def test_file_write_module_imports(frontend, tmp_path, enable_imports, import_le
             # If not all header modules appear in the dependency graph, then these
             # will not be parsed by OMNI and therefore the required xmod files will
             # not be generated, thus making modules 'c' and 'd' fail at parsing
-            pytest.xfail('Without parsing imports, OMNI does not have the xmod for imported modules')
+            pytest.skip('Without parsing imports, OMNI does not have the xmod for imported modules')
         raise e
```
reuterbal: With the test skipped for OMNI, we can remove the entire try-except handling.

loki/transformations/inline/tests/test_procedures.py
```diff
@@ -840,7 +840,7 @@ def test_inline_marked_routine_with_optionals(frontend, adjust_imports, tmp_path
 
 
 @pytest.mark.parametrize('frontend', available_frontends(
-    xfail=[(OMNI, 'OMNI has no sense of humour!')])
+    skip=[(OMNI, 'OMNI has no sense of humour!')])
```
reuterbal, line 843: 🤣

reuterbal (changes requested):
Thanks, all for it - the xfails were put in place in the hope that future improvements might fix the tests, but this looks very unlikely for the specific OMNI limitations.

I've pointed out two tests where the change was not fully applied - I think it would make sense to fix that before merging.

Otherwise, quite entertainig to read some of the xfail reasons...

loki/backend/tests/test_fgen.py
mlange05: True! I'll file an issue and see what I can do about that. 😉

mlange05: Issue filed: #681

reuterbal (approved):

PR #680, opened 2026-05-05
Batch: Split scheduler tests into themed modules

mlange05:
### Description

Another small housekeeping PR that simply splits the bloated scheduler test base into multiple themed modules.

### Contributor Declaration


mlange05:
@reuterbal This was obviously done with agents, but I've confirmed that all tests have been moved, and that comments have been ~kept~ restored (after the fact 😉 ).

loki/batch/tests/_test_scheduler_utils.py
reuterbal, line 1: I don't think we use a utilties module specific for tests elsewhere but put common utilities into the conftest instead. Is there a benefit doing it one way or another?

reuterbal (changes requested):
Thanks, this looks good to me. Admittedly I haven't vetted every test's placement into the most sensible module but the breakdown itself appears sensible.

It requires a rebase following the last merge, so marking this as requiring changes.

loki/batch/tests/_test_scheduler_utils.py
mlange05: Not that I can think of - moved now as suggested in top commit after latests rebase.

reuterbal (approved):

PR #682, opened 2026-05-08
PromoteLocalArrayTransformation - Add new transformation for promotion of local arrays

joherics:
# Add `PromoteLocalArrayTransformation` for local array promotion in SCC pipeline

_This is the first in a series of three PRs for the ecRad Loki GPU work._

This PR adds a new transformation `PromoteLocalArrayTransformation` that promotes local arrays with a horizontal dimension in the SCC pipeline. The transformation is included in `SCCVVectorPipeline` but disabled by default and can be activated with the item config key `promote_locals`.

## `PromoteLocalArrayTransformation`

The transformation identifies local arrays used inside vector sections (as marked by `
SCCDevectorTransformation`) and promotes them by appending the horizontal range dimension.
It:
- Skips arrays that already contain the horizontal dimension
- Skips arrays with entirely deferred shapes (pointers, allocatables, assumed-shape)
- By default skips arrays with only compile-time constant dimensions; this can be over
ridden per routine via the `promote_constant_arrays` item config key (default `False`)

## Changes to existing utilities

The declaration-update logic previously lived inside `demote_variables` but has been extracted into a shared `update_variable_declarations` utility in `utilities.py`. This is now used in both the demotion and promotion transformations. A new `promote_variable_declarations` in `promote.py` handles spec-only promotion and uses `single_variable_declaration` to isolate each variable before modifying shapes. A helper `_is_deferred_shape` decides whether promoted dimensions should also be deferred.

Furthermore, the following changes have been made:

- **`Dimension.size_expressions`** (`loki/dimension.py`): Now returns all lower/upper bounds pairs, not just the first bounds pair.

- **`SCCVVectorPipeline`** (`loki/transformations/single_column/scc.py`): `PromoteLocalArrayTransformation` added between `SCCDemoteTransformation` and `SCCVecRevectorTransformation`. (It is off by default but is required to sit in between the devectorization and revectorization steps).

### Contributor Declaration

### DIFF
--- a/loki/dimension.py
+++ b/loki/dimension.py
@@ -186,7 +186,8 @@ class Dimension:
         """
         exprs = self.sizes
         exprs += (f'1:{self.size}', )
-        if self.bounds:
-            exprs += (f'{self.bounds[1]} - {self.bounds[0]} + 1', )
-            exprs += (f'{self.bounds[0]}:{self.bounds[1]}', )
+        if self._lower and self._upper:
+            for lower, upper in zip(self._lower, self._upper):
+                exprs += (f'{upper} - {lower} + 1', )
+                exprs += (f'{lower}:{upper}', )
         return exprs
--- a/loki/transformations/array_indexing/demote.py
+++ b/loki/transformations/array_indexing/demote.py
@@ -7,13 +7,10 @@
 
 """ Utilities for demoting the rank of array variables. """
 
-from loki.expression import symbols as sym
-from loki.ir import (
-    nodes as ir, FindNodes, Transformer, FindVariables,
-    SubstituteExpressions
-)
+from loki.ir import FindVariables, SubstituteExpressions
 from loki.logging import info
 from loki.tools import as_tuple, CaseInsensitiveDict
+from loki.transformations.utilities import update_variable_declarations
 
 
 __all__ = ['demote_variables']
@@ -65,23 +62,6 @@ def demote_variables(routine, variable_names, dimensions):
     routine.spec = SubstituteExpressions(vmap).visit(routine.spec)
 
     # Ensure all declarations with `DIMENSION` keywords are modified too!
-    decls = tuple(
-        d for d in FindNodes(ir.VariableDeclaration).visit(routine.spec)
-        if d.dimensions and any(s.name.lower() in vnames for s in d.symbols)
-    )
-    decl_map = {}
-    for decl in decls:
-        # If all symbols have the same shape (after demotion)
-        sym_shape = tuple(s.shape if isinstance(s, sym.Array) else None for s in decl.symbols)
-        if all(d == sym_shape[0] for d in sym_shape):
-            dimensions = decl.symbols[0].shape if isinstance(decl.symbols[0], sym.Array) else None
-            decl_map[decl] = decl.clone(dimensions=dimensions)
-        else:
-            # If not, split into multiple declarations
-            sdims = tuple(s.shape if isinstance(s, sym.Array) else None for s in decl.symbols)
-            decl_map[decl] = tuple(
-                decl.clone(symbols=(s,), dimensions=d) for s, d in zip(decl.symbols, sdims)
-            )
-    routine.spec = Transformer(decl_map).visit(routine.spec)
+    routine.spec = update_variable_declarations(routine.spec, vmap.values())
 
     info(f'[Loki::Transform] Demoted variables in {routine.name}: {", ".join(variable_names)}')
--- a/loki/transformations/array_indexing/demote.py
+++ b/loki/transformations/array_indexing/demote.py
@@ -10,7 +10,7 @@
 from loki.ir import FindVariables, SubstituteExpressions
 from loki.logging import info
 from loki.tools import as_tuple, CaseInsensitiveDict
-from loki.transformations.utilities import update_variable_declarations
+from loki.transformations.utilities import update_variable_declaration_dimensions
 
 
 __all__ = ['demote_variables']
@@ -62,6 +62,6 @@ def demote_variables(routine, variable_names, dimensions):
     routine.spec = SubstituteExpressions(vmap).visit(routine.spec)
 
     # Ensure all declarations with `DIMENSION` keywords are modified too!
-    routine.spec = update_variable_declarations(routine.spec, vmap.values())
+    routine.spec = update_variable_declaration_dimensions(routine.spec, vmap.values())
 
     info(f'[Loki::Transform] Demoted variables in {routine.name}: {", ".join(variable_names)}')
--- a/loki/transformations/array_indexing/promote.py
+++ b/loki/transformations/array_indexing/promote.py
@@ -19,12 +19,68 @@ from loki.ir import (
 )
 from loki.logging import info
 from loki.tools import as_tuple, OrderedSet
-
+from loki.transformations.utilities import single_variable_declaration, update_variable_declarations, _is_deferred_shape
 
 __all__ = [
-    'promote_variables', 'promote_nonmatching_variables',
+    'promote_variables', 'promote_variable_declarations',
+    'promote_nonmatching_variables',
     'promotion_dimensions_from_loop_nest',
 ]
+
+
+def promote_variable_declarations(routine, variable_names, pos, size):
+    """
+    Promote the declared dimensions of a list of variables by inserting
+    new array dimensions of given size.
+
+    This updates only the variable declarations in ``routine.spec``,
+    not the variable uses in the routine body.
+
+    Parameters
+    ----------
+    routine : :any:`Subroutine`
+        The subroutine in which the variable declarations should be promoted.
+    variable_names : list of str
+        The names of variables to be promoted. Matching of variables against
+        names is case-insensitive.
+    pos : int
+        The position of the new array dimension using Python indexing
+        convention (i.e., count from 0 and use negative values to count from
+        the end).
+    size : :py:class:`pymbolic.Expression`
+        The size of the dimension (or tuple for multi-dimension promotion) to
+        insert at `pos`.
+    """
+    variable_names = {name.lower() for name in variable_names}
+    if not variable_names:
+        return
+
+    # Ensure that all variables to promote are in single variable decls
+    single_variable_declaration(routine, variable_names)
+
+    size = as_tuple(size)
+
+    var_list = [var for decl in FindNodes(ir.VariableDeclaration).visit(routine.spec)
+                for var in decl.symbols if var.name.lower() in variable_names]
+
+    # For deferred-shape arrays (all ':' dimensions), the promoted
+    # dimension must also be deferred — explicit sizes are invalid in
+    # declarations of assumed-shape or allocatable arrays.
+    deferred_size = tuple(sym.RangeIndex((None, None)) for _ in size)
+
+    var_shapes = [getattr(var, 'shape', ()) for var in var_list]
+    if pos < 0:
+        var_pos = [len(shape) - pos + 1 for shape in var_shapes]
+    else:
+        var_pos = [pos] * len(var_shapes)
+    var_shapes = [d[:p] + (deferred_size if _is_deferred_shape(d) else size) + d[p:]
+                  for d, p in zip(var_shapes, var_pos)]
+    var_map = {v: v.clone(type=v.type.clone(shape=shape), dimensions=shape)
+               for v, shape in zip(var_list, var_shapes)}
+    routine.spec = SubstituteExpressions(var_map).visit(routine.spec)
+
+    # Update variable declarations with updated dimensions
+    routine.spec = update_variable_declarations(routine.spec, var_map.values())
 
 
 def promote_variables(routine, variable_names, pos, index=None, size=None):
@@ -58,7 +114,6 @@ def promote_variables(routine, variable_names, pos, index=None, size=None):
         is updated accordingly.
     """
     variable_names = {name.lower() for name in variable_names}
-
     if not variable_names:
         return
 
@@ -91,6 +146,10 @@ def promote_variables(routine, variable_names, pos, index=None, size=None):
                         var_dim = var.dimensions
                     else:
                         var_dim = ()
+                    # If the variable is declared as an array but used without
+                    # subscripts, fill in the existing dimensions with ':'
+                    if not var_dim and getattr(var, 'shape', None):
+                        var_dim = tuple(sym.RangeIndex((None, None)) for _ in var.shape)
                     if pos < 0:
                         var_pos = len(var_dim) - pos + 1
                     else:
@@ -105,21 +164,7 @@ def promote_variables(routine, variable_names, pos, index=None, size=None):
 
     # Apply shape promotion
     if size is not None:
-        size = as_tuple(size)
-
-        var_list = [var for decl in FindNodes(ir.VariableDeclaration).visit(routine.spec)
-                    for var in decl.symbols if var.name.lower() in variable_names]
-
-        var_shapes = [getattr(var, 'shape', ()) for var in var_list]
-        if pos < 0:
-            var_pos = [len(shape) - pos + 1 for shape in var_shapes]
-        else:
-            var_pos = [pos] * len(var_shapes)
-        var_shapes = [d[:p] + size + d[p:] for d, p in zip(var_shapes, var_pos)]
-
-        var_map = {v: v.clone(type=v.type.clone(shape=shape), dimensions=shape)
-                   for v, shape in zip(var_list, var_shapes)}
-        routine.spec = SubstituteExpressions(var_map).visit(routine.spec)
+        promote_variable_declarations(routine, variable_names, pos, size)
 
 
 def promotion_dimensions_from_loop_nest(var_names, loops, promotion_vars_dims, promotion_vars_index):
--- a/loki/transformations/array_indexing/promote.py
+++ b/loki/transformations/array_indexing/promote.py
@@ -19,7 +19,7 @@ from loki.ir import (
 )
 from loki.logging import info
 from loki.tools import as_tuple, OrderedSet
-from loki.transformations.utilities import single_variable_declaration, update_variable_declarations, _is_deferred_shape
+from loki.transformations.utilities import update_variable_declaration_dimensions, _is_deferred_shape
 
 __all__ = [
     'promote_variables', 'promote_variable_declarations',
@@ -55,9 +55,6 @@ def promote_variable_declarations(routine, variable_names, pos, size):
     if not variable_names:
         return
 
-    # Ensure that all variables to promote are in single variable decls
-    single_variable_declaration(routine, variable_names)
-
     size = as_tuple(size)
 
     var_list = [var for decl in FindNodes(ir.VariableDeclaration).visit(routine.spec)
@@ -80,7 +77,7 @@ def promote_variable_declarations(routine, variable_names, pos, size):
     routine.spec = SubstituteExpressions(var_map).visit(routine.spec)
 
     # Update variable declarations with updated dimensions
-    routine.spec = update_variable_declarations(routine.spec, var_map.values())
+    routine.spec = update_variable_declaration_dimensions(routine.spec, var_map.values())
 
 
 def promote_variables(routine, variable_names, pos, index=None, size=None):
--- a/loki/transformations/array_indexing/promote_local_array.py
+++ b/loki/transformations/array_indexing/promote_local_array.py
@@ -1,0 +1,289 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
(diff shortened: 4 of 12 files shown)


joherics:
I have disabled the tests that are not compatible with OMNI. However,  the tests does not seem to trigger for this PR anymore, after I force pushed the branch 5 days ago, and when I tried to manually re-run the failing test the actions was using the old cached checkout instead of the new updated tests.

reuterbal:
Another rebase did not revive the tests but an additional commit on top seems to finally have done the trick... 🤷

loki/transformations/utilities.py
```diff
@@ -62,27 +63,102 @@ def single_variable_declaration(routine, variables=None, group_by_shape=False):
     """
     decl_map = {}
     for decl in FindNodes(VariableDeclaration).visit(routine.spec):
-        if len(decl.symbols) > 1:
+        if len(decl.symbols) >= 1:
             if not group_by_shape:
                 unique_symbols = [s for s in decl.symbols if variables is None or s.name in variables]
                 if unique_symbols:
-                    new_decls = tuple(decl.clone(symbols=(s,)) for s in unique_symbols)
+                    new_decls = tuple(decl.clone(symbols=(s,), dimensions=None) for s in unique_symbols)
```
reuterbal: As discussed offline, it wasn't fully clear to me why it is required to specify `dimensions=None` here - although I have understood now that this removes the `dimension` attribute from the declaration and thus forces the backend to generate declarations like
```
real :: arr(dim1,dim2)
```
instead of
```
real, dimension(dim1, dim2) :: arr
```

I think I know why this is being done but I consider this an unexpected side-effect which should be a separate modification in the transformation instead.

loki/transformations/utilities.py
```diff
@@ -62,27 +63,102 @@ def single_variable_declaration(routine, variables=None, group_by_shape=False):
     """
     decl_map = {}
     for decl in FindNodes(VariableDeclaration).visit(routine.spec):
-        if len(decl.symbols) > 1:
+        if len(decl.symbols) >= 1:
```
reuterbal: This change is likely only to force the `dimensions=None` effect on all declarations and should therefore also not be required.

loki/dimension.py
```diff
@@ -186,7 +186,8 @@ def size_expressions(self):
...
         exprs = self.sizes
         exprs += (f'1:{self.size}', )
-        if self.bounds:
-            exprs += (f'{self.bounds[1]} - {self.bounds[0]} + 1', )
-            exprs += (f'{self.bounds[0]}:{self.bounds[1]}', )
+        if self._lower and self._upper:
+            for lower, upper in zip(self._lower, self._upper):
+                exprs += (f'{upper} - {lower} + 1', )
+                exprs += (f'{lower}:{upper}', )
         return exprs
```
reuterbal, line 193: Why is this modification required? `self.lower` and `self.upper` or `self.bounds` has the benefit of including some implicit ranges, e.g., `1` as the lower bound if not specified explicitly.

loki/transformations/array_indexing/tests/test_array_promote.py
```diff
@@ -8,12 +8,16 @@
...
 from loki.expression import symbols as sym
-from loki.frontend import available_frontends
+from loki.frontend import available_frontends, OMNI
+from loki.ir import nodes as ir, FindNodes, FindVariables, Section, SubstituteExpressions
 
-from loki.transformations.array_indexing.promote import promote_variables
+from loki.transformations.array_indexing.promote import promote_variables, promote_variable_declarations
+from loki.transformations.array_indexing.promote_local_array import PromoteLocalArrayTransformation
+from loki.transformations.utilities import update_variable_declarations
+from loki.dimension import Dimension
```
reuterbal: Import order:
```suggestion
from loki import Subroutine, fgen
from loki.dimension import Dimension
from loki.expression import symbols as sym
from loki.frontend import available_frontends, OMNI
from loki.ir import nodes as ir, FindNodes, FindVariables, Section, SubstituteExpressions
from loki.jit_build import jit_compile

from loki.transformations.array_indexing.promote import promote_variables, promote_variable_declarations
from loki.transformations.array_indexing.promote_local_array import PromoteLocalArrayTransformation
from loki.transformations.utilities import update_variable_declarations
```

loki/transformations/array_indexing/demote.py
```diff
@@ -65,23 +62,6 @@ def demote_variables(routine, variable_names, dimensions):
...
-            dimensions = decl.symbols[0].shape if isinstance(decl.symbols[0], sym.Array) else None
-            decl_map[decl] = decl.clone(dimensions=dimensions)
-        else:
-            # If not, split into multiple declarations
-            sdims = tuple(s.shape if isinstance(s, sym.Array) else None for s in decl.symbols)
-            decl_map[decl] = tuple(
-                decl.clone(symbols=(s,), dimensions=d) for s, d in zip(decl.symbols, sdims)
-            )
-    routine.spec = Transformer(decl_map).visit(routine.spec)
+    routine.spec = update_variable_declarations(routine.spec, vmap.values())
```
reuterbal: Good idea to extract this feature into a utility. The name `update_variable_declarations` is a bit too arbitrary, though, it should make clear that this is to update the `dimension` attribute only. Something like `fix_variable_declaration_dimensions` or so?

loki/transformations/array_indexing/promote.py
```diff
@@ -19,14 +19,70 @@
...
+    else:
+        var_pos = [pos] * len(var_shapes)
+    var_shapes = [d[:p] + (deferred_size if _is_deferred_shape(d) else size) + d[p:]
+                  for d, p in zip(var_shapes, var_pos)]
+    var_map = {v: v.clone(type=v.type.clone(shape=shape), dimensions=shape)
+               for v, shape in zip(var_list, var_shapes)}
+    routine.spec = SubstituteExpressions(var_map).visit(routine.spec)
+
+    # Update variable declarations with updated dimensions
+    routine.spec = update_variable_declarations(routine.spec, var_map.values())
```
reuterbal: This should already take care of splitting declarations into separate declarations if the dimensions of multi-variable declarations don't match anymore. I'm not sure I understand why the explicit call to single_variable_declaration is required further up?

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+        # dimension is already present.
+        skipped = [a for a in candidates if _is_deferred_shape(a.shape)]
+        for a in skipped:
+            warning(
+                '[PromoteLocalArrayTransformation] Skipping %s in %s'
+                ' — deferred shape (pointer/allocatable), cannot determine'
+                ' if horizontal dimension is present',
+                a.name, routine.name
+            )
+        candidates = [a for a in candidates if not _is_deferred_shape(a.shape)]
```
reuterbal, line 84: Wouldn't that be more efficient to avoid re-evaluating?
```suggestion
        candidates = [a for a in candidates if a not in skipped]
```

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+        sections : list of :any:`Section`
+            The vector-section IR nodes.
+        horizontal : :any:`Dimension`
+            The horizontal dimension descriptor.
+        promote_constant_arrays : bool, optional
+            If True, also promote arrays whose dimensions are all
+            compile-time constants; default: ``False``.
+        """
+        # Create a list of local temporary arrays to filter down
+        candidates = get_local_arrays(routine, routine.spec)
```
reuterbal, line 63: Would it not make sense to combine the relevant calls to `get_local_arrays` instead of starting with all and then filtering them later?

```suggestion
        candidates = [*get_local_arrays(routine, s) for s in sections]
```

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+
+        # Create a copy of the tree and apply promotion in-place
+        routine.body = Transformer().visit(routine.body)
+
+        # Re-find vector sections from the cloned tree
+        section_label = sections[0].label
+        sections = [
+            s for s in FindNodes(ir.Section).visit(routine.body)
+            if s.label == section_label
+        ]
```
reuterbal, line 147: I don't think it's necessary to clone the tree.

Also, doesn't this find only the first section, since you are now only checking for `section_label = sections[0].label`?

Thus, would be good to have a test that uses more than 1 vector section

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+                    # var.shape includes the new dimension; subtract len(index)
+                    # to get the original number of dimensions.
+                    if not var_dim and getattr(var, 'shape', None):
+                        orig_ndim = len(var.shape) - len(index)
+                        var_dim = tuple(sym.RangeIndex((None, None)) for _ in range(orig_ndim))
+                    if pos < 0:
+                        var_pos = len(var_dim) - pos + 1
+                    else:
+                        var_pos = pos
+                    dimensions = as_tuple(var_dim[:var_pos] + index + var_dim[var_pos:])
```
reuterbal, line 171: I'm curious, is the translation to positive indices really required? This should work fine even if negative Python indexing is used:
```
>>> l = list(range(10))
>>> l
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> l[:-5] + [10, 11] + l[-5:]
[0, 1, 2, 3, 4, 10, 11, 5, 6, 7, 8, 9]
```

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+        candidates = [
+            a for a in candidates
+            if any(a.name.lower() in vs for vs in vector_sec.values())
+        ]
+
+        return candidates
+
+
+    @classmethod
+    def promote_variables_vector_sections(cls, routine, variable_names, pos, index, sections):
```
reuterbal, line 103: To avoid the multiple passes, this would be better to implement as a custom transformer. Via recursion you can mark when you're inside a vector section and when outside - thus dynamically switching between index or colon injection. Something along these lines (untested, just to give you the idea):

```
class PromoteVariablesInVectorSections(Transformer):
    def __init__(self, variable_names, pos, index, sections):
        ...

    def visit_Section(self, o, **kwargs):
        if o in self.sections:
            kwargs['in_section'] = True
        return super().visit(o, **kwargs)

    def visit_Expression(self, o, **kwargs):
        in_section = kwargs.get('in_section', False)
        ...<do the in-place expression substitution, using index or : depending on in_section>...
```

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+        role = kwargs['role']
+        item = kwargs.get('item', None)
+        if role == 'kernel':
+            promote_locals = False
+            preserve_arrays = []
+            promote_constant_arrays = False
+            if item:
+                promote_locals = item.config.get('promote_locals', False)
+                preserve_arrays = item.config.get('preserve_arrays', [])
+                promote_constant_arrays = item.config.get('promote_constant_arrays', False)
```
reuterbal: I would suggest to make the default values of `promote_locals`, `preserve_arrays` and `promote_constant_arrays` constructor arguments of the transformation. That allows overriding the config both, globally and on a per-item level:
```
def __init__(self, horizontal, promote_locals=False, preserve_arrays=None, promote_constant_arrays=False):
    self.horizontal = horizontal
    self.promote_locals = promote_locals
    self.preserve_arrays = as_tuple(preserve_arrays)
    self.promote_constant_arrays = promote_constant_arrays

...
    if item:
        promote_locals = item.config.get('promote_locals', self.promote_locals)
        preserve_arrays = item.config.get('preserve_arrays', self.preserve_arrays)
        promote_constant_arrays = item.config.get('promote_constant_arrays', self.promote_constant_arrays)
    else:
        promote_locals = self.promote_locals
        preserve_arrays = self.preserve_arrays
        promote_constant_arrays = self.promote_constant_arrays
...
```

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+            if s.label == 'vector_section'
+        ]
+
+        # Determine which local arrays to promote
+        to_promote = self.get_locals_to_promote(routine, sections, self.horizontal,
+                                                promote_constant_arrays=promote_constant_arrays)
+
+        # Filter out arrays marked explicitly for preservation
+        if preserve_arrays:
+            to_promote = [v for v in to_promote if v.name not in preserve_arrays]
```
reuterbal: I think we need to be careful about ignoring case here:

```suggestion
        if preserve_arrays:
            preserve_arrays = {arr.lower() for arr in preserve_arrays}
            to_promote = [v for v in to_promote if v.name.lower() not in preserve_arrays]
```

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+                        f"loop variables in subroutine: {routine.name}"
+                    )
+                lower = _lower
+                upper = _upper
+
+        horizontal_var = get_integer_variable(routine, self.horizontal.index)
+
+        # Step 1: Promote declarations
+        promote_variable_declarations(
+            routine, variable_names, pos=-1,
```
reuterbal: Should we not prepend the promotion dimension, i.e., `pos=0`?

loki/transformations/single_column/tests/test_scc.py
```diff
@@ -355,6 +356,96 @@ def test_scc_demote_transformation(frontend, horizontal):
...
+    for transform in scc_transform:
+        transform.apply(kernel, role='kernel', item=kernel_item)
+
+    assert isinstance(kernel.variable_map['tmp'], Array)
+    assert len(kernel.variable_map['tmp'].shape) == 2
+    assert kernel.variable_map['tmp'].shape[0] == 'nz'
+    assert kernel.variable_map['tmp'].shape[1] == 'start:end'
+
+    assigns = FindNodes(Assignment).visit(kernel.body)
+    assert 'tmp(nz, start:end)' in fgen(kernel.spec).lower()
```
reuterbal: I'm pretty sure this ought to become `tmp(start:end, nz)` to be useful

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+                if any(i in var_dim for i in index):
+                    continue
+                # If the variable is declared as an array but used without
+                # subscripts, fill in the existing dimensions with ':'.
+                # Note: declarations are already promoted at this point, so
+                # var.shape includes the new dimension; subtract len(index)
+                # to get the original number of dimensions.
+                if not var_dim and getattr(var, 'shape', None):
+                    orig_ndim = len(var.shape) - len(index)
+                    var_dim = tuple(sym.RangeIndex((None, None)) for _ in range(orig_ndim))
```
reuterbal, line 198: This is not covered by a test.

reuterbal (changes requested):
Many thanks, this looks very good but I have a few remarks how to improve the implementation - see the inline comments. In addition, there's a few things that need to be added to the tests. I've marked two inline, additionally a test that specifies `preserve_arrays` (ideally using non-matching upper/lower case) would be ideal.

awnawab:
I'll do a full review later today/tomorrow, but for now I can confirm that enabling `PromoteLocalArrayTransformation` in the sccs and sccv pipelines is safe for IFS builds.

joherics:
> I'll do a full review later today/tomorrow, but for now I can confirm that enabling `PromoteLocalArrayTransformation` in the sccs and sccv pipelines is safe for IFS builds.

Great, thank you!

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+
+        # Identify the horizontal bounds for the promoted dimension
+        lower = None
+        upper = None
+        for l, u in zip(self.horizontal.lower, self.horizontal.upper):
+            _lower = routine.variable_map.get(l, None)
+            _upper = routine.variable_map.get(u, None)
+            if _lower and _upper:
+                if lower or upper:
+                    warning(
```
awnawab, line 291: Do we really need this warning? Wouldn't it be enough to just take the first lower and upper bound present in the current symbol table? And throw an error if none of them are present?

loki/transformations/array_indexing/promote_local_array.py
awnawab: Agreed, I might be missing something, but shouldn't we always just prepend the promotion dimension? If we do, would that also remove the need for the position logic in `promote_variable_declarations`?

loki/transformations/array_indexing/promote_local_array.py
```diff
@@ -0,0 +1,304 @@
...
+        horizontal_var = get_integer_variable(routine, self.horizontal.index)
+
+        # Step 1: Promote declarations
+        promote_variable_declarations(
+            routine, variable_names, pos=-1,
+            size=RangeIndex((lower, upper))
+        )
+
+        # Step 2: Insert horizontal index into variable uses
+        self.promote_variables_vector_sections(
```
awnawab, line 307: Same question here about prepending promotion dimension rather than determining its position.

loki/transformations/single_column/revector.py
```diff
@@ -198,7 +198,7 @@ def mark_driver_loop(self, routine, loop):
         # Find a horizontal size variable to mark vector_length
         symbol_map = routine.symbol_map
         sizes = tuple(
-            routine.resolve_typebound_var(size, symbol_map) for size in self.horizontal.size_expressions
+            routine.resolve_typebound_var(size, symbol_map) for size in self.horizontal.sizes
```
awnawab, line 201: Considering that the new promotion transformation is only applied to kernels, I'm wondering why this seemingly unrelated change is needed for the driver?

awnawab (changes requested):
Thanks a lot @wertysas for this very useful and well implemented new transformation 🙏 I've left a few minor comments, which don't necessarily require code changes and clarifications might be enough.

loki/transformations/array_indexing/promote_local_array.py
joherics: Yes, you are likely right, but this is following the same logic as was already present in the promote.py. I have not changed this yet, but I see no obvious reason why this wouldn't work with negative indices. Then we could simplify both this snippet and the similar snipper in `promote_variable_declarations` in `promote.py`

loki/dimension.py
joherics: In ecRad I ran into issues due to multiple variables being used for the same iteration space. This is supported in the lower/upper bounds pairs, but wasn't handled properly in the `size_expressions` property, so I had to extend it from only returning the size expression for the lower/upper bound pair, to all of them.

loki/transformations/array_indexing/tests/test_array_promote.py
joherics: Updated this now.

loki/transformations/utilities.py
joherics: Yes, this is not required anymore, and I reverted this to the way it was before.

loki/transformations/utilities.py
joherics: Yes, I have reverted to keep the declaration using the same style for declaring dimensions as it had before.

loki/transformations/array_indexing/tests/test_array_promote.py
joherics: I have updated this to follow the alphabetical import-style you suggested.

loki/transformations/array_indexing/demote.py
joherics: Agreed, I renamed this `update_variable_declaration_dimensions`.

loki/transformations/array_indexing/promote.py
joherics: Yes, I have removed that call to `single_variable_declarations`.

loki/transformations/array_indexing/promote_local_array.py
joherics: Yes, your suggestion is exactly what it should be like.

loki/transformations/array_indexing/promote_local_array.py
joherics: Yes, I agree that would be a more a beautiful way to write it. I have not changed it from the current two-pass structure, but this is definitely worth to consider for the future.

loki/transformations/array_indexing/promote_local_array.py
joherics: Originally had something like this but then removed the ctor args in favour of only using the items (after looking at some of the other trafos). However, I agree with you and it is now possible to set default in the ctor and I have added tests to cover this.

loki/transformations/array_indexing/promote_local_array.py
joherics: Yes, my mistake, my brain automatically promoted this the CPU way and not for coalesced GPU accesses. I have changed this to `pos=0` now.

loki/transformations/array_indexing/promote_local_array.py
joherics: This is fixed now.

Issue #668, opened 2026-04-15
Multipipeline/SeparateModesKernel refactoring and consolidation with DependencyTrafo


MichaelSt98:
In SeparateModesKernel._create_new_items, the current logic has a special else branch that keeps the original item for the last remaining inherited mode/pipeline (i.e., when only one pipeline applies to an item). This conditional behaviour could be altered so that **all the items create/duplicate/rename the item depending on its pipeline** and with this the `else` branch could be eliminated.
Further, it probably makes sense to **combine the Multipipeline/SeparateModesKernel logic with the `DependencyTrafo` to simplify both.**

PR #685, opened 2026-05-19
Refactor SeparateModesKernel transformation

reuterbal:
This PR refactors the SeparateModesKernel transformation towards #668 

- Always rename scheduler items to include the mode, not only for multi-mode items
- Eliminate the hidden graph traversal and recursion by converting to a reverse transformation that only updates the current item
- Adapted more sophisticated rename methods from DependencyInjection  for use in SeparateModesKernel transformation.

The intention is that this is a first step towards the overhaul of the dependency injection, putting the mode separation at the start of the transformation pipeline.

### Contributor Declaration

### DIFF
--- a/loki/transformations/dependency.py
+++ b/loki/transformations/dependency.py
@@ -489,7 +489,7 @@ class SeparateModesKernel(Transformation):
 
     def _create_new_items(self, item, successors, item_factory, config, ignore=None):
         """
-        Duplicate :data:`item` for each transformation mode and update teh plan
+        Duplicate :data:`item` for each transformation mode and update the plan
         dependencies accordingly.
 
         Driver items are not duplicated but their dependencies are still updated to point
--- a/loki/transformations/dependency.py
+++ b/loki/transformations/dependency.py
@@ -5,6 +5,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+from collections import defaultdict
 from itertools import chain
 
 from loki.batch.transformation import Transformation
@@ -389,8 +390,8 @@ class SeparateModesKernel(Transformation):
         scope_name = item.scope_name
         local_name = f'{item.local_name}{mode_rename}'
         if scope_name:
-            if "_mod" in scope_name:
-                scope_name = scope_name.replace('_mod', f'_{mode_rename}_mod').replace('__', '_')
+            if scope_name.endswith('_mod'):
+                scope_name = f'{scope_name[:-4]}{mode_rename}_mod'
             else:
                 scope_name = f'{scope_name}{mode_rename}'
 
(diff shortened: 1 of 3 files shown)


mlange05 (approved):
I cannot say that I fully understand all the finer intricacies of this trafo, but from high-level this looks good to me. I defer to @MichaelSt98 for a more detailed assessment - but GTG from me.

loki/transformations/dependency.py
```diff
@@ -425,181 +433,269 @@ def _get_or_create_or_rename_item(self, item, mode, item_factory, config):
...
-        if call_map:
-            routine.body = Transformer(call_map).visit(routine.body)
+                if import_.symbols:
+                    if not any(s.name in imports_to_rename for s in import_.symbols):
+                        continue
+
+                    symbols_modules = {
+                        s.clone(name=f'{imports_to_rename[s.name].local_name}'): imports_to_rename[s.name].scope_name
+                        if s.name in imports_to_rename else (s, import_.module) for s in import_.symbols
+                    }
```
MichaelSt98: That won't work for mixed imports like `use callee_mod, only kernel, some_constant` since the `if s.name in imports_to_rename` only applies to value part of the dict comprehension and not to the key part which leads to a key error in `imports_to_rename`. Further, this shows a gap in the testing.

Further the question is whether we want this to be transformed to:

```fortran
use callee_loki_m1_mod, only: kernel_loki_m1, some_constant
```

or 

```fortran
use callee_mod, only: some_constant
use callee_loki_m1_mod, only: kernel_loki_m1
```

loki/transformations/dependency.py
```diff
@@ -425,181 +433,269 @@ def _get_or_create_or_rename_item(self, item, mode, item_factory, config):
...
+                    f'[Loki] SeparateModesKernel: Could not find new item {new_child_name} for '
+                    f'child {child.name} of item {item.name} in the item cache.'
+                ))
+
+        return new_dependencies
+
+    def _create_new_items(self, item, successors, item_factory, config, ignore=None):
+        """
+        Duplicate :data:`item` for each transformation mode and update teh plan
+        dependencies accordingly.
```
MichaelSt98:
```suggestion
        Duplicate :data:`item` for each transformation mode and update the plan
        dependencies accordingly.
```

MichaelSt98 (changes requested):
**Looks good to me!**

**However:** problem on renaming imports which needs a test and fix and/or potentially a discussion.

Moreover:

* multiple docstrings missing
* nondeterministic mode-copy creation which is not a problem since all items/routines are renamed but could lead to different order of plan entries which is again not a problem and nothing we cared about in the past

loki/transformations/dependency.py
reuterbal: Nice spot, many thanks for this! That was indeed handled insufficiently.
Right now the expected level of support should be the second case in my opinion: Rename imports that are touched in the dependency graph and retain the rest as before. I have added a test to validate this behaviour

MichaelSt98 (approved):
Thanks!
Some functions still missing docstrings.

PR #687, opened 2026-05-22
Add warnings for unresolved vector notation

awnawab:
Considering how painful unresolved vector notation has proven to be, this PR contributes warnings for the cases where loki is unable to resolve them itself.


MichaelSt98 (approved):
Agreed. Will most likely produce warnings for unproblematic assignments too but I guess having warnings which easily allows to find them makes sense instead of just silently ignoring them,

reuterbal (approved):
LGTM! :shipit:

PR #688, opened 2026-05-26
Expression: Further enhance and fix `simplify`

mlange05:
### Description

This PR adds floating-point arithmetic simplification, as well as a range of smaller fixes to the `simplify(expr)` utility in `loki.expression`. These corner cases where found by checking against the testing in PRs #663 and #515, which were doing this as part of the constant propagation feature. With these in place, we can now re-use the underlying mapper and get full test coverage.

The smaller fixes in detail:
- Make `sym.StringConcat` use the `StrCompare` mixing for quick string-based test checks
- Extend simplify to cover string concatenation reduction
- Extend simplify to cover logical `not` reduction on constants
- Extend simplify to correctly reduce `sym.Power` expressions
- Extend simplify to do full floating-point arithmetic if the `Simplification.FloatingPointArithmetic` enum flag is set.    

### Contributor Declaration

### DIFF
--- a/loki/expression/symbolic.py
+++ b/loki/expression/symbolic.py
@@ -276,36 +276,41 @@ def flatten_expr(expr):
     return sym.Sum(as_tuple(done))
 
 
-def sum_int_literals(expr):
+def sum_int_literals(expr, integer_arithmetic=True, floating_point=False):
     """
-    Sum up the values of all `IntLiteral` in the sum and return the reduced sum.
+    Sum up the values of all numeric literals in the sum and return the reduced sum.
     """
     def _process(child):
-        if isinstance(child, sym.IntLiteral):
-            return child.value, None
+        if (integer_arithmetic or floating_point) and isinstance(child, sym.IntLiteral):
+            return child.value, False, None
+        if floating_point and isinstance(child, sym.FloatLiteral):
+            return float(child.value), True, None
         if is_minus_prefix(child):
-            value, stripped_child = _process(strip_minus_prefix(child))
+            value, has_float, stripped_child = _process(strip_minus_prefix(child))
             if value != 0:
-                return -value, stripped_child
-        return 0, child
+                return -value, has_float, stripped_child
+        return 0, False, child
 
     if not isinstance(expr, sym.Sum):
         return expr
 
     transformed_components = list(zip(*[_process(child) for child in expr.children]))
     value = sum(transformed_components[0])
-    remaining_components = [ch for ch in transformed_components[1] if ch is not None]
+    has_float = any(transformed_components[1])
+    if not integer_arithmetic and not has_float:
+        return expr
+    remaining_components = [ch for ch in transformed_components[2] if ch is not None]
     if value != 0:
-        remaining_components = [sym.IntLiteral(value)] + remaining_components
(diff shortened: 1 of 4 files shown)


loki/expression/symbolic.py
```diff
@@ -276,36 +276,41 @@ def flatten_expr(expr):
     return sym.Sum(as_tuple(done))
 
 
-def sum_int_literals(expr):
+def sum_int_literals(expr, integer_arithmetic=True, floating_point=False):
```
reuterbal: for consistency, the options should maybe be called `integer_arithmetic` and `floating_point_arithmetic` or, to not be overly verbose `int_arithmetic` and `fp_arithmetic`?

loki/expression/symbolic.py
```diff
@@ -315,47 +320,60 @@ def separate_coefficients(expr):
     :rtype: (int, list)
     """
     def _process(child):
-        if isinstance(child, (int, np.integer)):
-            return child, None
-        if isinstance(child, sym.IntLiteral):
-            return child.value, None
+        if (integer_arithmetic or floating_point) and isinstance(child, (int, np.integer)):
```
reuterbal: Curious, why do we include `np.integer` here?

loki/expression/symbolic.py
```diff
@@ -315,47 +320,60 @@ def separate_coefficients(expr):
...
+    value, has_float, remaining_components = separate_coefficients(
+        expr, integer_arithmetic=integer_arithmetic, floating_point=floating_point
+    )
     if value == 0:
-        return sym.IntLiteral(0)
+        return sym.Literal(0.0) if has_float else sym.IntLiteral(0)
     if abs(value) != 1:
-        remaining_components = [sym.IntLiteral(abs(value))] + remaining_components
+        remaining_components = [
+            sym.Literal(abs(value)) if has_float else sym.IntLiteral(abs(value))
```
reuterbal: Just from reading through the code I noticed that the `sym.Literal(...) if has_float else sym.IntLiteral(...)` is a recurring pattern. Could be worthwhile to consider a LiteralFactory at some point or so instead, but no need to mop this up in the current PR.

reuterbal (changes requested):
Many thanks for this great expansion of the symbolic expression engine! I don't have much to say, the only thing that would be good to change is to maybe pick slightly better names for the arguments that enable integer and floating point arithmetic - the current naming is too inconsistent for me to ignore ;-)

loki/expression/symbolic.py
mlange05: Done - I've also updated some of the docstrings to reflect that.

loki/expression/symbolic.py
mlange05: AI-foo-bar probably - removed now.

loki/expression/symbolic.py
mlange05: We already have - it's `sym.Literal`. However, there's a few instances where we do `... = `sym.Literal(1.0) if has_float else `sym.IntLiteral(0)` where the factory has no value to determine the right type from. I tried to streamline where possible.

reuterbal (approved):
Thanks, looks great now!

PR #692, opened 2026-06-01
New transformation: `ReplaceKernels`

MichaelSt98:
Replace configured kernel calls with alternative routines.

Conceptually, the transformation walks each call site in the processed routine, looks for names listed in ``replace_kernels_map``, resolves the replacement routine, and reconstructs the actual argument list against the replacement signature. When the original call target is unresolved in the IR, the transformation also tries to resolve the original routine from source so that remapping can still be done against the original dummy argument names instead of the raw actual expressions. Once the replacement call has been rebuilt, the transformation updates imports/includes to match the new callee and optionally marks the replacement routine as blocked for further scheduler traversal.

### Description

This pull request introduces comprehensive tests for the new `ReplaceKernels` transformation and ensures it is imported as part of the main transformation package. The main focus is on verifying the correct remapping of kernel routine arguments, handling of optional arguments, import rewriting, plan generation, and pipeline integration.

### Testing and Validation of `ReplaceKernels` Transformation

**New test suite for kernel replacement:**
* Added `test_replace_kernel.py` containing multiple tests that cover argument remapping, handling of optional arguments, include/import rewriting, plan output, and integration in a pipeline. This ensures the `ReplaceKernels` transformation works as expected across a variety of scenarios.

**Integration with transformation package:**
* Exposed `replace_kernel` transformation in the main `loki.transformations` package by importing it in `__init__.py`, making it available for general use.

### Contributor Declaration

### DIFF
--- a/loki/transformations/replace_kernel.py
+++ b/loki/transformations/replace_kernel.py
@@ -1,0 +1,532 @@
+# (C) Copyright 2018- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from pathlib import Path
+
+from loki.batch import ModuleItem, ProcedureItem, Transformation
+from loki.ir import FindNodes, Transformer, nodes as ir
+from loki.logging import warning
+from loki.tools import CaseInsensitiveDict, as_tuple
+from loki.types import BasicType
+
+__all__ = ['ReplaceKernels']
+
+
+class ReplaceKernels(Transformation):
+    """
+    Replace configured kernel calls with alternative routines.
+
+    This transformation supports:
+
+    - replacing selected call targets by routine name;
+    - remapping arguments by direct rename;
+    - remapping arguments by extracting a member from a derived-type actual;
+    - remapping arguments by formatting a custom expression template;
+    - overriding an argument with a positional value;
+    - overriding an argument with a literal or expression string;
+    - rewriting ``USE`` imports and ``.intfb.h`` include-style imports;
+    - optionally keeping the replacement routine out of scheduler traversal and
+      CMake planning via ``block=True``.
+
+    Conceptually, the transformation walks each call site in the processed
+    routine, looks for names listed in ``replace_kernels_map``, resolves the
+    replacement routine, and reconstructs the actual argument list against the
+    replacement signature. When the original call target is unresolved in the
+    IR, the transformation also tries to resolve the original routine from
+    source so that remapping can still be done against the original dummy
+    argument names instead of the raw actual expressions. Once the replacement
+    call has been rebuilt, the transformation updates imports/includes to match
+    the new callee and optionally marks the replacement routine as blocked for
+    further scheduler traversal.
+
+    The replacement map accepts entries of the form::
+
+        {
+            'old_kernel': {
+                'routine': 'new_kernel',
+                'args': {
+                    'old_dummy': 'new_dummy',
+                    'struct_arg': {'map_to': 'new_dummy', 'member': 'MEMBER'},
+                    'old_end': {
+                        'map_to': 'new_end',
+                        'expr': 'MOD({geom}%TOTAL%NGPTOT, {geom}%DIM%NPROMA)',
+                        'placeholders': {'geom': 'geom'},
+                    },
+                    'old_start': {'position': 1},
+                    'new_flag': '.true.',
+                },
+                'block': True,
+            }
+        }
+
+    Parameters
+    ----------
+    replace_kernels_map : dict, optional
+        Case-insensitive mapping from original routine name to replacement
+        configuration. Each configuration entry supports the following keys:
+
+        ``routine``
+            Name of the replacement routine.
+
+        ``args``
+            Optional mapping that controls how actual arguments are rebuilt for
+            the replacement call. Supported forms are:
+
+            ``'old_dummy': 'new_dummy'``
+                Map the actual argument of ``old_dummy`` onto the replacement
+                dummy ``new_dummy``.
+
+            ``'old_dummy': {'position': 1}``
+                Pass a positional literal or expression to the replacement
+                dummy named ``old_dummy``.
+
+            ``'old_dummy': {'map_to': 'new_dummy', 'member': 'MEMBER'}``
+                Pass ``actual%MEMBER`` from the original argument ``old_dummy``
+                to the replacement dummy ``new_dummy``.
+
+            ``'old_dummy': {'map_to': 'new_dummy', 'expr': '...',
+            'placeholders': {'name': 'old_dummy_name'}}``
+                Build an expression by formatting the template string with the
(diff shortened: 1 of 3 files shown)


loki/transformations/tests/test_replace_kernel.py
```diff
@@ -0,0 +1,365 @@
...
+    # These fixtures mirror the relevant call-shape patterns we care about for
+    # PR 1 without pulling in real IFS names or source files.
+    sources = {
+        'geom_types_mod.F90': """
+module geom_types_mod
+  implicit none
+  type dim_type
+    integer :: nproma
+    integer :: ngpblks
+  end type dim_type
```
mlange05, line 60: Linebreaks make code much more readable for humans... 😜

loki/transformations/tests/test_replace_kernel.py
```diff
@@ -0,0 +1,365 @@
...
+            'strict': False,
+        },
+        'routines': {name: {'role': 'driver'} for name in driver_names},
+    })
+
+
+def _write_replace_sources(tmp_path):
+    """Write anonymized Fortran sources that cover the replacement-call patterns under test."""
+    # These fixtures mirror the relevant call-shape patterns we care about for
+    # PR 1 without pulling in real IFS names or source files.
```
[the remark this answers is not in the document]
mlange05: Agreed, this looks like a dev leftover.

loki/transformations/tests/test_replace_kernel.py
```diff
@@ -0,0 +1,365 @@
...
+        frontend=frontend,
+        xmods=[tmp_path],
+        includes=[replace_source_dir],
+    )
+    scheduler.process(transformation=ReplaceKernels({
+        'leaf': {
+            'routine': 'leaf_replacement',
+            'args': {
+                'flag2': 'flag2_renamed',
+                'geom': {'map_to': 'block_count', 'member': 'DIM%NGPBLKS'},
```
mlange05, line 226: Ok, this nomenclature seems quite cumbersome to me - and though I appreciate the complexity of the remapping, but given that we have the `expr` notation below, I'm wondering why we cannot use this here instead of explicit encoding of "member" lookup? 

In general, I'm wondering why we cannot use the expression parser here to make all the argument replacements use the same `"arg": {map_to:..., expr: "expression"}` notation, where the call context provides the scope, instead of "member" and "placeholder" semantics?

loki/transformations/tests/test_replace_kernel.py
```diff
@@ -0,0 +1,365 @@
...
+    """Skip unmapped optional replacement arguments and emit a warning for the omission."""
+    scheduler = Scheduler(
+        paths=[replace_source_dir],
+        config=_scheduler_config(('include_driver',)),
+        frontend=frontend,
+        xmods=[tmp_path],
+        includes=[replace_source_dir],
+    )
+
+    with caplog.at_level('WARNING'):
```
mlange05, line 267: I think we want to check that the correct warning was generated, not just any warning.

loki/transformations/replace_kernel.py
```diff
@@ -0,0 +1,548 @@
...
+            return item_cache[replacement_name]
+
+        qualified_matches = [
+            item for item in item_cache.values()
+            if isinstance(item, ProcedureItem) and item.local_name.lower() == replacement_name
+        ]
+        if len(qualified_matches) == 1:
+            return qualified_matches[0]
+
+        for key, item in tuple(item_cache.items()):
```
mlange05, line 128: Do we really need to have this fallback in the transformation? If the replacement cannot be found, should we not error and fail here, and improve the item lookup in the scheduler? Fixes ti general functionality should always be preferred to local workarounds (maybe add this to AGENTS.md?).

loki/transformations/replace_kernel.py
```diff
@@ -0,0 +1,548 @@
...
+                    if (
+                        isinstance(definition_item, ProcedureItem)
+                        and definition_item.local_name.lower() == replacement_name
+                    ):
+                        return definition_item
+
+        return None
+
+    @staticmethod
+    def _complete_replacement_source(replacement_item, scheduler_config, build_args):
```
mlange05, line 147: Is this added complexity truly necessary? Should we not just force the replacement to be in the scheduler's item cache (can be a literal file path entry under `sources`) instead of adding all the complexity of these workarounds for a missing replacement file? Or am I missing something here?

loki/transformations/replace_kernel.py
```diff
@@ -0,0 +1,548 @@
...
+            )
+        if routine_item is None or routine_item.source is None:
+            return None
+
+        self._complete_replacement_source(routine_item, scheduler_config, build_args)
+        return routine_item
+
+    @staticmethod
+    def _mark_replacement_as_blocked(item, replacement_item):
+        """Add the replacement routine and module names to the current scheduler block list."""
```
mlange05, line 229: [Optional] Should adding this to "blocked" items be a config option rather than a default?

loki/transformations/replace_kernel.py
```diff
@@ -0,0 +1,548 @@
...
+            candidates.append(replacement_item.scope_name)
+
+        for candidate in candidates:
+            if candidate not in block_entries:
+                block_entries.append(candidate)
+
+        item.config['block'] = tuple(block_entries)
+
+    @staticmethod
+    def _split_argument_rules(replacement_args, new_routine):
```
mlange05, line 242: [no action] Again, do we really need this special-casing, or could we simply make use of the expression-evaluation utility to derive the replacement symbols. I'd leave the `map_to` argument to toggle between keyword-mapped and positional argument, but "expr" seems to me is the overarching way to derive any of these cases, no? Again, I'm happy to be corrected, if I'm missing some special-cases or context here, but I'm not seeing it atm.

loki/transformations/replace_kernel.py
```diff
@@ -0,0 +1,548 @@
...
+    def _complete_replacement_source(replacement_item, scheduler_config, build_args):
+        """Complete the parsed source for a replacement item with scheduler frontend settings."""
+        frontend_args = {
+            key: value for key, value in build_args.items()
+            if key in ('definitions', 'preprocess', 'includes', 'defines', 'xmods', 'omni_includes', 'frontend')
+        }
+        frontend_args = scheduler_config.create_frontend_args(replacement_item.source.path, frontend_args)
+        replacement_item.source.make_complete(**frontend_args)
+
+    def _load_replacement_item_from_source(self, replacement_name, item_factory, scheduler_config, build_args):
```
mlange05, line 156: This entire method looks to be untested - do we really need this code path at all (see also later comment). If not, I suggest removing it and making sure we error early and clearly when we cannot load the replacement item from cache.

mlange05 (changes requested):
First of all, I really like this functionality and I think it's very powerful and nicely implemented and tested. I have left a few remarks though on some of the seemingly redundant boilerplate code paths for deriving source objects from the scheduler, and some of the replacement semantics and use of expressions. Otherwise I think this approach is great for a first implementation - many, many thanks! 🙏 

One note for the future: The per-call replacement semanticsalmost seems like something that would benefit from the ability to annotate in source, rather than treat all calls equivalently from config. In particular if we consider that certain utility routines can be called with very different arguments in two places within the same source tree (ie. CUADJTQ in EC-physics). However, extending this is far beyond the scope of this initial PR and can be left for later - we might put this into a wishlist issue once merged.

PR #696, opened 2026-06-03
Transformations: Isolated CI JIT test

mlange05:
### Description

This PR aims to address a long-standing issue with racy failure in the transformation test base. In particular, `test_parametrise.py` creates multiple variations of the same module procedures with varying signatures and no renaming for the various test variations, leading to a racy situation where the assignment of two subsequent variants in the same thread can cause symbol aliasing and thus trigger mem corruption and sigsevs. To prevent this, simple de-aliasing techniques are not sufficient, as the compiled fortran symbols are still loaded dynamically into the Python runtime within the same process.

To fix this, this PR adds a small `run_isolated` utility that executes a helper method in a short-lived child process, allowing the `compile-and-check` utility in the problematic test to be forced onto a separate process and thus avoid dynamic symbol aliasing.

### Contributor Declaration

### DIFF
--- a/loki/jit_build/jit.py
+++ b/loki/jit_build/jit.py
@@ -24,7 +24,7 @@ from loki.subroutine import Subroutine
 from loki.tools import as_tuple, gettempdir, filehash
 
 
-__all__ = ['jit_compile', 'jit_compile_lib', 'run_isolated', 'clean_test']
+__all__ = ['jit_compile', 'jit_compile_and_run', 'jit_compile_lib', 'run_isolated', 'clean_test']
 
 
 _f90wrap_kind_map = Path(__file__).parent.parent/'tests/kind_map'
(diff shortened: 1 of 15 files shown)


mlange05:
@reuterbal Ok, as discussed offline, the final problem seems to have been that other JIT tests had already exported symbols for the crashing test to inherit. To prevent this, I added a `jit_compile_and_run` utility that by default executes the JIT-and-run on an isolated process. This mechanism has now been rolled out to most simple JIT-based tests - and after several re-runs this seems to have reduced the number of spurious fails. I think this is ready for another look.

loki/jit_build/jit.py
```diff
@@ -21,12 +24,148 @@
...
+
+def jit_compile_and_run(source, *args, filepath=None, objname=None, isolated=True,
+                        multiprocessing_context='fork', exit_after_result=False,
+                        timeout=10, **kwargs):
+    """
+    JIT-compile a source item and execute the selected entry point.
+
+    The complete compile, load, and execution cycle can optionally run in a
+    short-lived subprocess to isolate native extension module state from the
+    parent test process.
```
reuterbal: Because this will be the outward facing routine and I was just looking up their meaning, I'd suggest adding something like this:

```suggestion
    parent test process. See :any:`run_isolated` for available options for arguments.
```

loki/transformations/tests/test_transform_loop.py
```diff
@@ -1089,13 +1083,12 @@ def test_transform_loop_fission_promote_multiple_read_after_write(tmp_path, fron
...
     routine = Subroutine.from_source(fcode, frontend=frontend)
     filepath = tmp_path/(f'{routine.name}_{frontend}.f90')
-    function = jit_compile(routine, filepath=filepath, objname=routine.name)
 
     # Test the reference solution
     klon, klev, nclv = 32, 100, 5
     a = np.zeros(shape=(klon, klev), order='F', dtype=np.int32)
     b = np.zeros(shape=(klon, klev, nclv), order='F', dtype=np.int32)
-    function(a=a, b=b, klon=klon, klev=klev, nclv=nclv)
+    jit_compile_and_run(routine, a=a, b=b, klon=klon, klev=klev, nclv=nclv, filepath=filepath, isolated=False)
```
reuterbal: This appears to be the only test that requires `isolated=False`, but I can also not spot anything that would explain that this is necessary. The Fortran code is missing an `implicit none` in the spec but all used symbols appear to have been declared, so it is unlikely to be related.

loki/jit_build/jit.py
```diff
@@ -21,12 +24,148 @@
...
+        kind = payload = None
+    parent_conn.close()
+
+    if kind == 'exception':
+        raise RuntimeError(payload)
+    if process.exitcode != 0:
+        raise RuntimeError(f'Isolated process failed with exit code {process.exitcode}')
+    if kind == 'result':
+        return payload
+    return None
```
reuterbal: I don't think anything other than `'exception'` or `'result'` for `kind` can be expected - so I'm contemplating whether an exception might not be more appropriate here?

reuterbal (approved):
Great stuff, a really useful addition and clean solution to the recurring issues associated with JIT testing. Just some minor comments but overall happy to include this as is!

loki/transformations/tests/test_transform_loop.py
mlange05: Good catch, I finally figured out why that is. Effectively the argument arrays used were so big that they would saturate the process pipe buffers and thus cause the isolated process to hang. Simply reducing the array dimensions seems to fix this - allowing us to use process isolation now.

loki/transformations/tests/test_transform_loop.py
```diff
@@ -1089,13 +1084,12 @@ def test_transform_loop_fission_promote_multiple_read_after_write(tmp_path, fron
 """
     routine = Subroutine.from_source(fcode, frontend=frontend)
     filepath = tmp_path/(f'{routine.name}_{frontend}.f90')
-    function = jit_compile(routine, filepath=filepath, objname=routine.name)
 
     # Test the reference solution
-    klon, klev, nclv = 32, 100, 5
+    klon, klev, nclv = 16, 16, 5
```
MichaelSt98: Makes sense to make the values smaller, however, I think I'd prefer `klon` and `klev` to not be the same value?!

MichaelSt98 (approved):
**Very nice!**

* regarding timeout
  * seems like `jit_compile_and_run`: 10 seconds
  * `run_isolated`: no timeout by default -> can hang forever?
* copy back mechanism: I think this only works for 
  * seems to be safe enough for NumPy-style array in/out args
  * however, not universally safe/effective, maybe adding a warning when copying back is not possible?

loki/transformations/tests/test_transform_loop.py
mlange05: Fair point - I've put it at `(16, 37, 5)` and locally it still works. Will let CI confirm

PR #698, opened 2026-06-10
Scheduler handling of seed routines

MichaelSt98:
### Description

This pull request introduces several improvements to the scheduler's handling of seed routines and multi-mode planning, alongside updates to tests to reflect the new logic. The main changes ensure that implicit seeds are correctly inferred from the configuration, requiring at least one driver or routine marked with `seed_routine = true`, and that explicit routine entries alone do not become seeds unless marked or provided. Test cases and assertions have been updated to document and verify the new behaviors, especially for multi-mode and shared kernel scenarios.

### Scheduler seed routine logic and multi-mode planning

* Added `_get_implicit_seeds` static method to `Scheduler` to derive seed routines from the config, defaulting to routines with `role = 'driver'` or `seed_routine = true`, and raising an error if none are found. (`loki/batch/scheduler.py`)
* Updated `Scheduler.__init__` to use the new implicit seed logic when `seed_routines` is not provided. (`loki/batch/scheduler.py`)
* Added comprehensive tests for multi-mode planning and shared kernel behavior, including new helper functions and test cases to document and verify routine selection, plan file generation, and call/import rewrites. (`loki/batch/tests/test_scheduler_config.py`)
* Ensured that explicit routine entries in the config do not become seeds unless marked with `seed_routine = true` or specified in `seed_routines`. (`loki/batch/tests/test_scheduler_config.py`)

### Attribute propagation and dependency handling

* Improved propagation of the `lib` attribute in the scheduler's dependency graph, so child routines only inherit the parent's library unless they have an explicit override. (`loki/batch/sgraph.py`)

### Test updates and consistency

* Updated and added tests to use the new seed routine logic, ensuring correct routine selection and plan consistency in various transformation, linting, and build scenarios. This includes passing `seed_routines` explicitly where needed and marking routines with `seed_routine = true` in test configs. [[1]](diffhunk://#diff-9d0b832c88c1e413fedcd6ae7b80d7134f622d381993bdea06dd3d8439768889R490) [[2]](diffhunk://#diff-a999af48b93ac47f0c05921f255e19c790f6e7fd1a22a9927a6addf758d93400L1195-R1197) [[3]](diffhunk://#diff-812f3ad989b8ca17c314162292fd78847e34e5dd810d41603882339d66a3de76R235) [[4]](diffhunk://#diff-99686d8aa1d59b5b7109fde85cfcf15dd5b4ee9255215d4748d91a88669da7b1L515-R515) [[5]](diffhunk://#diff-99686d8aa1d59b5b7109fde85cfcf15dd5b4ee9255215d4748d91a88669da7b1L587-R587) [[6]](diffhunk://#diff-769ca0b4b448441e2ba4c099b7492416d4c72ae238880f19aacdc8488bfc376fL272-R274)

### Minor test and assertion corrections

* Adjusted test assertions to access scheduler items by name rather than index, and made minor corrections to expected results in plan file and dependency tests. [[1]](diffhunk://#diff-812f3ad989b8ca17c314162292fd78847e34e5dd810d41603882339d66a3de76L291-R295) [[2]](diffhunk://#diff-21642120a43170785a63e36d1a42cf8972aec83febda51ce8105536fd9b0e978R660-R664) [[3]](diffhunk://#diff-9ab4f0e7b1dd8ae49f97393292b89e4e4c70e19938dbdada4d4a36690fa9ba18L827) [[4]](diffhunk://#diff-9ab4f0e7b1dd8ae49f97393292b89e4e4c70e19938dbdada4d4a36690fa9ba18L893-R892)

### Contributor Declaration

### DIFF
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -160,6 +160,27 @@ class Scheduler:
     # TODO: Should be user-definable!
     source_suffixes = ['.f90', '.F90', '.f', '.F']
 
+    @staticmethod
+    def _get_implicit_seeds(config):
+        """
+        Derive implicit seed routines from the scheduler config.
+
+        By default, only driver routines and routines explicitly marked with
+        ``seed_routine = true`` become graph roots.
+        """
+        implicit_seeds = tuple(
+            name.lower() for name in config.routines
+            if config.create_item_config(name).get('role') == 'driver'
+            or config.create_item_config(name).get('seed_routine', False)
+        )
+        if not implicit_seeds:
+            raise RuntimeError(
+                'No seed routines provided and no implicit seeds found in the scheduler config. '
+                'Set `seed_routines`, mark routines with `role = "driver"`, or set '
+                '`seed_routine = true`.'
+            )
+        return implicit_seeds
+
     def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
                  includes=None, defines=None, definitions=None, xmods=None,
                  omni_includes=None, full_parse=True, frontend=FP, output_dir=None):
@@ -175,10 +196,8 @@ class Scheduler:
 
         # Build-related arguments to pass to the sources
         self.paths = [Path(p) for p in as_tuple(paths)]
-        self.seeds = tuple(
-            seed.lower()
-            for seed in as_tuple(seed_routines) or self.config.routines.keys()
-        )
+        self.seeds = tuple(seed.lower() for seed in as_tuple(seed_routines)) if seed_routines else \
+            self._get_implicit_seeds(self.config)
 
         # Accumulate all build arguments to pass to `Sourcefile` constructors
         self.build_args = {
--- a/loki/batch/scheduler.py
+++ b/loki/batch/scheduler.py
@@ -174,11 +174,13 @@ class Scheduler:
             or config.create_item_config(name).get('seed_routine', False)
         )
         if not implicit_seeds:
-            raise RuntimeError(
-                'No seed routines provided and no implicit seeds found in the scheduler config. '
-                'Set `seed_routines`, mark routines with `role = "driver"`, or set '
-                '`seed_routine = true`.'
-            )
+            implicit_seeds = tuple(seed.lower() for seed in config.routines.keys())
+            warning("[Loki]: "
+                "Using implicit or unspecified seed routines is deprecated and will be removed in a future release. "
+                "Currently, no seed routines were provided and none were discovered in the scheduler config. "
+                "Please update your configuration by setting `seed_routines`, "
+                "marking routines with `role = \"driver\"`, "
+                "or setting `seed_routine = true`.")
         return implicit_seeds
 
     def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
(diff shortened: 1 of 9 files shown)


loki/batch/scheduler.py
```diff
@@ -160,6 +160,27 @@ class Scheduler:
...
+
+        By default, only driver routines and routines explicitly marked with
+        ``seed_routine = true`` become graph roots.
+        """
+        implicit_seeds = tuple(
+            name.lower() for name in config.routines
+            if config.create_item_config(name).get('role') == 'driver'
+            or config.create_item_config(name).get('seed_routine', False)
+        )
+        if not implicit_seeds:
```
reuterbal, line 176: To retain backwards compatibility, we could fall back to the original behaviour of using all mentioned routines as seeds and print only a warning for now, with a deprecation notice.

loki/batch/scheduler.py
```diff
@@ -175,10 +196,8 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
 
         # Build-related arguments to pass to the sources
         self.paths = [Path(p) for p in as_tuple(paths)]
-        self.seeds = tuple(
-            seed.lower()
-            for seed in as_tuple(seed_routines) or self.config.routines.keys()
-        )
+        self.seeds = tuple(seed.lower() for seed in as_tuple(seed_routines)) if seed_routines else \
+            self._get_implicit_seeds(self.config)
```
reuterbal: I would suggest being a bit more verbose here to help users understand how the dependecy graph is being initialized - as this is likely something that may leave someone puzzled if they start using this on a new piece of code. Something along these lines:


```suggestion
        if seed_routines:
            info('Initializing Scheduler graph from list of seed routines provided in the config')
        self.seeds = tuple(seed.lower() for seed in as_tuple(seed_routines)) 
        else:
            info('Initializing Scheduler graph from driver routines')
            self.seeds = self._get_implicit_seeds(self.config)
```

Paired with then the warning (later error) about missing driver routines this may give users a good understanding if their transformations fail due to improper seeding.

reuterbal (changes requested):
Many thanks, the implementation looks sound and the proposed seeding of the scheduler graph is very nice, with both an implicit derivation via driver routines and the preferred use of an explicit seed list.

I left two suggestions to make this change less disruptive (without impacting our use cases) and provide more guidance to users - please have a look if that looks sensible!

MichaelSt98:
I incorporated your suggestions @reuterbal. However, be aware that this only restores some backwards compatibility. If some config defines `driver` routine(s) and additionally relies on config `routine` keys this won't work anymore and requires additional `is_seed=true`.

loki/batch/scheduler.py
```diff
@@ -175,10 +198,14 @@ def __init__(self, paths, config=None, seed_routines=None, preprocess=False,
 
         # Build-related arguments to pass to the sources
         self.paths = [Path(p) for p in as_tuple(paths)]
-        self.seeds = tuple(
-            seed.lower()
-            for seed in as_tuple(seed_routines) or self.config.routines.keys()
-        )
+        # self.seeds = tuple(seed.lower() for seed in as_tuple(seed_routines)) if seed_routines else \
+        #     self._get_implicit_seeds(self.config)
```
reuterbal:
```suggestion
```

reuterbal (commented):
Yes, that's a fair point. We can include this with the other breaking changes for 0.4.x - depending on ecwam integration for the config.

loki/batch/tests/test_scheduler_config.py
```diff
@@ -1059,3 +1059,327 @@ def test_scheduler_multi_modes_imports(tmp_path, proc_strategy, check_multimodes
             item_imports = item_ir.imports
             imported_modules = {imp.module for imp in item_imports}
             assert imported_modules == imports
+
+
+def _write_shared_kernel_multi_mode_sources(src_dir):
```
mlange05, line 1064: [no action] I think we need to at some point separate the multi-mode test into its own sub-package to avoid ending up with bloated test clusters again. But happy to leave this for a later stage.

mlange05 (approved):
Ok, changes themselves look good and API change is unfortunately necessary for this. 

On a side note, I think this highlights that some of the fundamentals of the `Scheduler` need to be re-designed to cater for these types of uses - but for now is is ok.

reuterbal (approved):
Many thanks for the changes! This looks good to me now - I will mark this as a breaking change in the next Loki tag release notes.

reuterbal:
ecWAM regression tests are expected to fail until https://github.com/ecmwf-ifs/ecwam/pull/148/ has been merged.

PR #703, opened 2026-06-17
Add optional error codes to `StopStmt`

awnawab:
A small PR that adds optional error codes to stop statements, and a matching small fix to the `RemoveCodeTransformation` to stringify them as necessary.


reuterbal (approved):
Great find, many thanks! Looks good to me

mlange05 (approved):
Looks great. Many thanks for mopping this up!

PR #704, opened 2026-06-23
Release 0.4.0

reuterbal:
Suggested Release notes:

**This release introduces breaking changes** that may require adaptations in transformation configs or user transformations:

### IR: Refactoring and introduction of explicitly named "statement nodes" by @mlange05 in https://github.com/ecmwf-ifs/loki/pull/656
The `Intrinsic` IR node type has been removed in favour of several, more targeted statement node types that inherit from a `GenericStmt` node. Any code that uses `Intrinsic` node type will have to be adapted. See the PR description for more details and the subsequent bugfix PR by @awnawab in https://github.com/ecmwf-ifs/loki/pull/703.

### Make Subroutine.argnames return lower-case names by @reuterbal in https://github.com/ecmwf-ifs/loki/pull/686
Due to Fortran's case-insensitive syntax, Loki tries to be case agnostic as much as possible (but allows for case-sensitivity in some key parts if required, e.g., for C or Python transpilation). In most places, the API returns therefore lower-case names or uses case-insensitive string comparisons. The `Subroutine.argnames` property did not follow this pattern, leading to manually lower-casing in many places where it is used. To homogenise the API, dummy argument names are now returned as lower case strings.

Previous use of `Subroutine.argnames` where the case-sensitivity matters should be replacing by `[arg.name for arg in routine.arguments]`.

### Scheduler handling of seed routines by @MichaelSt98 in https://github.com/ecmwf-ifs/loki/pull/698
The Scheduler graph in Loki is being built starting from seed nodes and adding dependency nodes recursively until all branches of the graph end in leaf nodes. While it had already been possible to specify the list of seed routines explicitly using the Scheduler's `seed_routines` parameter, an often used alternative was to simply list any routines that need to become seed nodes in the Scheduler config's `routines` section without any additional properties, which would implicitly add them to the list of seed routines. 

A common use case is when using the InlineTransformation, where some routines "disappear" during the transformation pipeline, but are still expected in the CMake plan. To force their generation, they could be added to the config, e.g., an example from ecWAM:
```
# we add loki inlined routines here to force them to be created
[routines.femeanws]
[routines.scosfl]
[routines.frcutindex]
...
```

While the explicit specification via the `seed_routines` argument continues to behave as before, the implicit derivation of `seed_routines` has changed now: only routines with `role='driver'` or an explicit `seed_routine=True` attribute will now be included as a seed. The above config entries need therefore adapting as follows:
```
# we add loki inlined routines here to force them to be created
[routines.femeanws]
  seed_routine = true
[routines.scosfl]
  seed_routine = true
[routines.frcutindex]
  seed_routine = true
...
```

**Full Changelog**: https://github.com/ecmwf-ifs/loki/compare/0.3.7...0.4.0

### Contributor Declaration


mlange05 (approved):
Looks great, many thanks!

PR #706, opened 2026-06-26
Also apply `RemoveCode` to driver loop bodies

awnawab:
This PR updes the `RemoveCode` trafo to also operate on driver loop bodies if `kernel_only=True`. I still kept the option name as is, but I'm happy to update it to `kernel_and_driver_loop_only`, but this will be a non-backwards compatible change.

`do_remove_calls` is also updated so that pattern-matching is applied to the specified `call_names`, rather than removing only those calls that match exactly.

### DIFF
--- a/loki/transformations/remove_code.py
+++ b/loki/transformations/remove_code.py
@@ -11,17 +11,22 @@ section and to perform Dead Code Elimination.
 """
 
 import operator as op
+from fnmatch import fnmatchcase
 from itertools import chain
 
 from loki.analyse import dataflow_analysis_attached
 from loki.batch import Transformation
 from loki.expression import simplify, symbols as sym, symbolic_op
-from loki.ir import nodes as ir, Transformer, FindNodes, FindVariables, FindInlineCalls, SubstituteExpressions
+from loki.ir import (
+    nodes as ir, Transformer, FindNodes, FindVariables, FindInlineCalls,
+    SubstituteExpressions, pragmas_attached
+)
 from loki.ir.pragma_utils import (
     is_loki_pragma, pragma_regions_attached, get_pragma_parameters
 )
 from loki.program_unit import ProgramUnit
 from loki.tools import flatten, as_tuple, OrderedSet
+from loki.transformations.utilities import find_driver_loops
 __all__ = [
     'RemoveCodeTransformation',
     'do_remove_dead_code', 'RemoveDeadCodeTransformer',
@@ -67,8 +72,8 @@ class RemoveCodeTransformation(Transformation):
         Use :any:`simplify` when branch pruning in during
         :meth:`remove_dead_code`.
     call_names : list of str
-        List of subroutine names against which to match
-        :any:`CallStatement` nodes during :meth:`remove_calls`.
+        List of subroutine names or glob-style name patterns against which
+        to match :any:`CallStatement` nodes during :meth:`remove_calls`.
     intrinsic_names : list of str
         List of module names against which to match :any:`GenericStmt`
         nodes during :meth:`remove_calls`.
@@ -77,7 +82,8 @@ class RemoveCodeTransformation(Transformation):
         objects during :meth:`remove_calls`; default: ``True``
     kernel_only : boolean
         Only apply the configured removal to subroutines marked as
-        "kernel"; default: ``False``
+        "kernel" and to driver-loop bodies in routines marked as
+        "driver"; default: ``False``
     remove_unused_args : boolean
         Remove unused dummy arguments from routines.
     remove_unused_vars : boolean
@@ -123,12 +129,25 @@ class RemoveCodeTransformation(Transformation):
 
     def transform_subroutine(self, routine, **kwargs):
 
-        if kwargs.get('role') == 'kernel' or not self.kernel_only:
+        role = kwargs.get('role')
+
+        if role == 'driver' and self.kernel_only:
+            # Apply named node removal to driver loops, while preserving the
+            # non-transformed setup/teardown parts of driver routines.
+            if self.call_names:
+                with pragmas_attached(routine, ir.Loop):
+                    driver_loops = find_driver_loops(section=routine.body, targets=kwargs.get('targets', ()))
+                    for loop in driver_loops:
+                        do_remove_calls(
+                            body=loop, call_names=self.call_names, remove_imports=False
+                        )
+
+        if role == 'kernel' or not self.kernel_only:
             # Apply named node removal to strip specific calls
             if self.call_names or self.intrinsic_names:
                 do_remove_calls(
-                    routine, call_names=self.call_names,
-                    intrinsic_names=self.intrinsic_names,
+                    body=routine.body, spec=routine.spec,
+                    call_names=self.call_names, intrinsic_names=self.intrinsic_names,
                     remove_imports=self.remove_imports
                 )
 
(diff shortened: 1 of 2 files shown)


loki/transformations/remove_code.py
```diff
@@ -123,19 +130,42 @@ def __init__(
...
+        role = kwargs.get('role')
+
+        if role == 'driver' and self.kernel_only:
+            # Apply code removal to driver loops, while preserving the
+            # non-transformed setup/teardown parts of driver routines.
+            if self.call_names or self.remove_marked_regions:
+                with pragmas_attached(routine, ir.Loop):
+                    driver_loops = find_driver_loops(section=routine.body, targets=kwargs.get('targets', ()))
+                    for loop in driver_loops:
+                        if self.call_names:
```
reuterbal: Should this not include `self.intrinsic_names`?

loki/transformations/remove_code.py
```diff
@@ -123,19 +130,42 @@ def __init__(
...
+                                mark_with_comment=self.mark_with_comment,
+                                replacement_call=self.replacement_call,
+                                replacement_msg=self.replacement_msg,
+                                replacement_module=self.replacement_module
+                            )
+
+        if role == 'kernel' or not self.kernel_only:
             # Apply named node removal to strip specific calls
             if self.call_names or self.intrinsic_names:
                 do_remove_calls(
```
reuterbal: Just to point out that you could create a utility `self._apply_to_ir(body=<ir>, spec=routine.spec)` that encapsulates the two calls to the removal utilities to avoid the duplication here.

loki/transformations/remove_code.py
```diff
@@ -409,12 +439,12 @@ def visit_MultiConditional(self, o, **kwargs):
...
-        routine, mark_with_comment=True, replacement_call=None,
+        body, scope=None, mark_with_comment=True, replacement_call=None,
         replacement_msg=None, replacement_module=None
 ):
     """
     Utility routine to remove code regions marked with
-    ``!$loki remove`` pragmas from a subroutine's body.
+    ``!$loki remove`` pragmas from a given IR body.
 
     Optionally, any removed region might be marked with a
```
reuterbal: I'd suggest calling this `internal_node` instead of `body` because you're not passing the body but actually the IR node that has a body.

reuterbal (changes requested):

reuterbal:
Regarding the renaming of the option I'm a little undecided. It feels a bit over-the-top to introduce a breaking API change just for this.

loki/transformations/remove_code.py
awnawab: Yes true, I should add a test for that and include it.

reuterbal (approved):
Many thanks, this looks very neat now!

loki/transformations/remove_code.py
```diff
@@ -121,25 +128,43 @@ def __init__(
         self.remove_unused_vars = remove_unused_vars
         self.remove_only_arrays = remove_only_arrays
 
+    def _apply_to_ir(self, internal_node, scope, spec=None, remove_imports=False):
```
mlange05, line 130: Does this really require a private method?

loki/transformations/remove_code.py
```diff
@@ -121,25 +128,43 @@ def __init__(
...
+                replacement_msg=self.replacement_msg,
+                replacement_module=self.replacement_module
+            )
+
     def transform_subroutine(self, routine, **kwargs):
 
-        if kwargs.get('role') == 'kernel' or not self.kernel_only:
+        role = kwargs.get('role')
+
+        if role == 'driver' and self.kernel_only:
```
mlange05, line 154: [no action] For my own context-window, do we still require/use the `kernel_only` flag, or is this now superfluous with these changes?

loki/transformations/remove_code.py
```diff
@@ -448,22 +476,26 @@ def do_remove_marked_regions(
         mark_with_comment=mark_with_comment,
         replacement_call=replacement_call,
         replacement_msg=replacement_msg,
+        inplace=True
     )
 
-    with pragma_regions_attached(routine, keyword='loki'):
-        routine.body = transformer.visit(routine.body, scope=routine)
+    attach_pragma_regions(internal_node, keyword='loki')
+    try:
```
mlange05: Are truly expecting this to fail and continue safely? Or am I missing the rationale here for removing the context-handling?

loki/transformations/remove_code.py
```diff
@@ -606,13 +647,18 @@ def __init__(
...
         super().__init__(**kwargs)
 
-        self.call_names = as_tuple(call_names)
+        self.call_names = tuple(str(name).lower() for name in as_tuple(call_names))
         self.intrinsic_names = as_tuple(intrinsic_names)
         self.remove_imports = remove_imports
 
+    def matches_call_name(self, name):
+        """Return ``True`` if ``name`` matches any configured call name pattern."""
+        name = str(name).lower()
```
mlange05, line 652: Is explicit lower-case handling needed here, if we're ultimately going against `o.symbols`? These should be case-gnostic, or am I missing something?

mlange05 (commented):
All in all looks good, but I picked up a few cosmetic (agentic) oddities that I'd like to understand the need for. Otherwise looks safe.

loki/transformations/remove_code.py
awnawab: No this doesn't need to be private, it also doesn't have to be a class method. I made it such because it felt like an operation bespoke to this class. Outside this trafo it would make more sense to call the generic `do_remove_calls` and `do_remove_marked_regions` utilities explicitly.

loki/transformations/remove_code.py
awnawab: Yes the `kernel_only` flag will still protect from removing stuff from non loki transformed parts of the driver layer, e.g. we would want to preserve DR_HOOK labels in driver layers.

loki/transformations/remove_code.py
awnawab: `pragma_regions_attached` documents that it expects a module or subroutine. The way it's implemented it can also handle an `internal_node`, but I didn't want to use undocumented functionality. I'm happy to reinstate the context manager, but then I would also update the documentation and variable names there.

loki/transformations/remove_code.py
awnawab: No I think you're right, we can omit the lower-casing of the call name. Please let me know if you would like me to fix this.

loki/transformations/remove_code.py
reuterbal: I think the restriction to module or subroutine is historic baggage and not an actual limitation we want to imply - particularly with program unit nodes moving closer to becoming "just an internal node" in the IR. So, if we can restore the context manager that might indeed be a more consistent use of the API.

loki/transformations/remove_code.py
mlange05: Yes, I'm trying to really get rid of all unnecessary string case-handling, so ideally we should use symbol-matchin where possible.

loki/transformations/remove_code.py
mlange05: Ok 🤷

loki/transformations/remove_code.py
mlange05: Agreed, I prefer updating outdated docs over trafo-level workarounds.

loki/transformations/remove_code.py
awnawab: Since we are now trying to match globs rather than exact matches, the matching is done by `fnmatchcase`, which is a case-sensitive string comparison so we do in fact need the explicit lower-casing here.

loki/transformations/remove_code.py
awnawab: Turns out the docs aren't outdated and because an `InternalNode` is a frozen dataclass we can't assign to `internal_node.body` the same way we can to `subroutine.body`. To restore the use of the context manager here, I will have to change `pragma_regions_attached` to something like this:
```
    if isinstance(ir_or_internal_node, InternalNode):
        ir_or_internal_node = attach_pragma_regions(ir_or_internal_node, keyword=keyword)
    else:
        if hasattr(ir_or_internal_node, 'spec'):
            ir_or_internal_node.spec = attach_pragma_regions(ir_or_internal_node.spec, keyword=keyword)
        if hasattr(ir_or_internal_node, 'body'):
            ir_or_internal_node.body = attach_pragma_regions(ir_or_internal_node.body, keyword=keyword)
```

Please let me know which solution you prefer.

awnawab: After discussion offline with @mlange05 I'll file a small PR to update `pragma_regions_attached` to also accept internal nodes, and then rebase this once that is merged.

edit: PR #711 filed.

reuterbal: Thanks. #711 is merged.

awnawab: Thanks. I've rebased and restored the use of the context manager.

mlange05 (approved):
Ok, looks great now. Thanks for addressing the comments.

PR #707, opened 2026-06-26
Adapt pool allocator for nested driver loops

awnawab:
This PR adds support to the pool allocator transformation for nested driver loops, i.e. the following pattern:

```
do  nt=1,nouter_loop
   !.. driver loop
    do jkglo=...
    enddo
enddo
```

### DIFF
--- a/loki/transformations/temporaries/pool_allocator.py
+++ b/loki/transformations/temporaries/pool_allocator.py
@@ -19,13 +19,13 @@ from loki.ir import (
     FindNodes, FindVariables, FindInlineCalls, Transformer, GenericStmt,
     Assignment, Conditional, CallStatement, Import, Allocation,
     Deallocation, Loop, Pragma, Interface, get_pragma_parameters,
-    SubstituteExpressions
+    SubstituteExpressions, pragmas_attached
 )
 from loki.logging import warning, debug
 from loki.tools import as_tuple, OrderedSet
 from loki.types import SymbolAttributes, BasicType, DerivedType
 
-from loki.transformations.utilities import recursive_expression_map_update
+from loki.transformations.utilities import recursive_expression_map_update, find_driver_loops
 
 
 __all__ = ['TemporariesPoolAllocatorTransformation', 'EcstackPoolAllocatorTransformation']
@@ -244,7 +244,7 @@ class TemporariesPoolAllocatorTransformation(Transformation):
             if item:
                 # import variable type specifiers used in stack allocations
                 self.import_allocation_types(routine, item)
-            self.create_pool_allocator(routine, stack_size)
+            self.create_pool_allocator(routine, stack_size, targets)
 
         self.inject_pool_allocator_into_calls(routine, targets, ignore, driver=role=='driver')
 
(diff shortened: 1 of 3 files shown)


loki/transformations/temporaries/pool_allocator.py
```diff
@@ -815,9 +829,20 @@ def create_pool_allocator(self, routine, stack_size):
...
+        with pragmas_attached(routine, Loop):
+            driver_loops = find_driver_loops(routine.body, targets)
+            annotated_loops = [
+                loop for loop in FindNodes(Loop).visit(routine.body)
+                if any(
+                    self._is_driver_loop_pragma(pragma)
+                    for pragma in as_tuple(loop.pragma)
+                )
+            ]
+            loops = OrderedSet(as_tuple(driver_loops) + as_tuple(annotated_loops))
```
reuterbal: Just as a suggestion: If we allowed to optionally specify an additional callback in `find_driver_loops` that receives a loop and returns `True` or `False` then we could simplify this bit to a single invocation of that utility.

It would require moving the `for pragma in as_tuple(loop.pragma)` loop into the `self._is_driver_loop_pragma` utility and then calling this as something like `loops = OrderedSet(find_driver_loops(routine.body, targets, self._is_driver_loop))`.

reuterbal (approved):
This looks good to me, thanks! Just one small suggestion as an alternative way of incorporating the additional identifier pattern.

loki/transformations/temporaries/pool_allocator.py
```diff
@@ -771,7 +771,22 @@ def apply_pool_allocator_to_temporaries(self, routine, item=None):
...
+    @staticmethod
+    def _is_driver_loop(loop):
+        """
+        Check whether an attached pragma identifies a driver loop that needs
+        pool allocator stack-pointer initialisation.
+        """
+        for pragma in as_tuple(loop.pragma):
+            keyword = pragma.keyword.lower()
+            content = pragma.content.lower()
+            if 'loop gang' in content or 'teams distribute' in content:
```
mlange05, line 783: I'm not sure I understand why we are picking this up via OpenMP and OpenACC pragma semantics here (and the PR description is very terse, given this is not obvious). Why do we need to rely on OpenMP/OpenACC pragmas to identify driver loops when the pool allocator already know the "block loop" bounds/dimensions (`b` and `nb` in the test example)?

awnawab: To identify driver loops, we either look for loops containing target calls or for loops explicitly marked via loki pragmas. If we were to instead identify them via loop bounds/variable, it would be hard to distinguish in the same subroutine between offloaded driver loops and those intended to run on host (and therefore should not be transformed by loki).

By the time we get to the pool allocator, we've resolved loki pragmas and so we have to look for the different flavours of parallel pragmas to identify driver loops.

reuterbal: Thanks - I think this points to the urgency to finally refactor 
- the transformation pipeline to break it up into more composable blocks, allowing to flexibly choose temporary handling before programming model injection
- the pool allocator transformation to insert loki data directives instead of openacc
- a generic identification of driver loops early-on that mark them accordingly for subsequent use by transformations (to avoid the patch-up flurry across transformations you had to go through now); this was an offline suggestion by @mlange05

awnawab: Agreed on all the above. The second suggestion alone should be enough to alleviate the workaround here, so I've created issue #710 to track that, and once that's fixed this workaround can be dropped.

mlange05 (approved):
Ok, all looks good. Thanks for the explanation.

PR #709, opened 2026-06-30
Testing workaround for the OMNI frontend on macos

awnawab:
A subtle change to a code removal test to workaround a weird quirk of OMNI on macos as compared to on linux.


reuterbal (approved):
Thanks!

PR #712, opened 2026-07-13
Unroll array initialisers

awnawab:
This PR updates the `ResolveVectorNotationTransformation` so that array initialisers are unrolled. 

Some array initialisation statements proved to be very costly in the GPU runs on jupiter. It's hard to determine a priori which ones will bite, so we unroll simple 1D constant bound array initialiasation statements automatically with this PR. A warning is printed for more complex unresolvable ones so that they are easily identifiable and can be addressed in source if necessary.

### DIFF
--- a/loki/transformations/array_indexing/tests/test_vector_notation.py
+++ b/loki/transformations/array_indexing/tests/test_vector_notation.py
@@ -940,7 +940,7 @@ end subroutine test_literal_list
         "Unqualifiable ':' should remain a RangeIndex on the LHS"
 
     literal_warnings = [r for r in caplog.records
-                        if 'Literal list on RHS' in r.message
+                        if 'prevents literal-list unrolling' in r.message
                         and 'ResolveVectorNotationTransformer' in r.message]
     assert literal_warnings
     assert any('test_literal_list' in r.message for r in literal_warnings)
--- a/loki/transformations/array_indexing/tests/test_vector_notation.py
+++ b/loki/transformations/array_indexing/tests/test_vector_notation.py
@@ -947,6 +947,53 @@ end subroutine test_literal_list
 
 
 @pytest.mark.parametrize('frontend', available_frontends())
+def test_resolve_vector_notation_literal_list_unsupported(frontend, caplog):
+    """
+    Only literal-list elements that are scalar variables or literals are
+    unrolled.  Any other element leaves the assignment unchanged, with a
+    warning.
+    """
+    fcode = """
+subroutine test_literal_list_unsupported(arr, brr)
+  implicit none
+  integer, intent(inout) :: arr(3), brr(3)
+  integer :: i
+
+  ! Implied-do array constructor: a LiteralList with a single InlineDo element
+  arr(1:3) = [(i, i=1,3)]
+
+  ! Arithmetic expression elements are not scalar variables or literals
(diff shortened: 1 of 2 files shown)


loki/transformations/array_indexing/vector_notation.py
```diff
@@ -645,6 +685,21 @@ def visit_Assignment(self, stmt, **kwargs):  # pylint: disable=unused-argument
...
+        # --- Resolve literal-list RHS by unrolling ---
+        # This runs after Step 2 so the LHS range has explicit bounds.
+        # Falls back to the previous warn-and-bail behaviour when unrolling is
+        # not possible (mixed RHS, unknown ranges).
+        if FindLiteralLists().visit(stmt.rhs):
+            resolved = self._resolve_literal_list(stmt)
+            if resolved:
+                return resolved
+            scope_str = f' in routine "{self.scope.name}"' if self.scope is not None else ''
+            warning(
```
MichaelSt98: This warning is not really up-to-date anymore. It will get here either because the RHS is a mixed expression or because the LHS array has bar `:`. I'd move the warning inside `_resolve_literal_list`

MichaelSt98 (changes requested):
There are more funky Literal lists in Fortran ...

For example:

```fortran
arr(1:3) = [(i, i=1,3)]
```

gets transformed to:

```fortran
arr(1) = ( i, i = 1,3 )
```

which is clearly wrong.
Especially those literal lists/ array initialisers are dangerous regarding performance...

MichaelSt98 (approved):
Looks great!
The only thing we could think of is making this optional.

loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -493,16 +494,27 @@ def test_resolve_vector_dimension_extended(frontend):
...
-    # B2: Literal list assignment should be unchanged (no loop wrapping)
+    # B2: Literal list assignment should be unrolled into one scalar
+    # assignment per element; no LiteralList should survive, and the
+    # unrolled assignments should not be wrapped in a generated loop.
     generated_loop_bodies = []
     for l in loops:
         if l.variable.name.startswith('i_'):
             generated_loop_bodies.extend(FindNodes(ir.Assignment).visit(l.body))
+    literal_assigns = [a for a in assigns
+                       if hasattr(a.rhs, 'elements') or 'LiteralList' in type(a.rhs).__name__]
```
mlange05: Why not `isinstance(a.rhs, sym.LiteralList)`? This is a very odd way of checking type, no?

mlange05 (approved):
Looks good to me - only picked on a minor cosmetic nitpick (optional). Otherwise GTG and very neatly implemented. 👍 :shipit:

loki/transformations/array_indexing/tests/test_vector_notation.py
awnawab: That was an LLM-ism I didn't catch, fixed now. Thanks!

awnawab:
The latest release of fastjsonschema introduced a bug (see https://github.com/horejsek/python-fastjsonschema/pull/213) which affects python < 3.10 builds, so I've also updated loki's `pyproject.toml` to skip that version.

EDIT: a patched version of the package was released, so I've removed the workaround commit.

PR #714, opened 2026-07-17
Further improving resolve vector notation

MichaelSt98:
#### Summary
This PR improves the robustness of vector-notation resolution by making it more conservative where transformations are ambiguous, while still preserving important scalarizable cases that appear in real physics code.

#### Key Changes

- Improves handling of inline function calls so elemental cases are recognized more reliably during vector-notation resolution.
- Refines how slice bounds are analyzed to distinguish safe scalar bounds from unsafe vector-valued expressions.
- Restores valid transformations for scalar derived-type bounds while keeping unsafe unresolved cases visible through warnings.
- Prevents incorrect resolution when array subscripts or bounds are effectively vector-valued.
- Expands regression coverage for representative IFS and ecRad-style patterns.
- Cleans up the related test setup so frontend-generated artifacts do not leak into the repository root.

#### Outcome

Overall, the diff makes resolve_vector_notation safer and more predictable: it avoids wrong-code generation in ambiguous cases, fixes missed transformations in valid cases, and adds coverage for the boundary cases that triggered the regressions.

### Contributor Declaration

### DIFF
--- a/loki/transformations/array_indexing/tests/test_vector_notation.py
+++ b/loki/transformations/array_indexing/tests/test_vector_notation.py
@@ -2377,3 +2377,97 @@ end subroutine test_dup_dims
         f"indicating only one loop variable was used for two distinct dimensions; "
         f"expected two distinct loop variables"
     )
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_resolve_vector_notation_scalar_plus_range_duplicate(frontend):
+    """
+    Regression test: when a loop variable already appears as a fixed scalar
+    subscript in the same array reference, a range that maps to the same loop
+    variable must get a fresh index variable instead.
+
+    Example:  ``field(i,j,1:n)`` inside ``DO j=1,n`` must NOT produce
+    ``field(i,j,j)`` — the third subscript must be a distinct
+    synthesized loop variable.
+    """
+    fcode = """
+subroutine test_scalar_range_dup(nouter, ninner, field)
+  implicit none
+  integer, intent(in) :: nouter, ninner
+  real, intent(inout) :: field(nouter, ninner, ninner)
+  integer :: i, j
+
+  do i = 1, nouter
+    do j = 1, ninner
+      field(i, j, 1:ninner) = 0.0
+    enddo
+  enddo
+
+end subroutine test_scalar_range_dup
+    """.strip()
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+    resolve_vector_notation(routine)
+
+    assigns = FindNodes(ir.Assignment).visit(routine.body)
+
+    # Find the resolved field assignment (no more RangeIndex dimensions)
+    resolved_assigns = [
+        a for a in assigns
+        if isinstance(a.lhs, sym.Array)
+        and 'field' in a.lhs.name.lower()
+        and not any(isinstance(d, sym.RangeIndex) for d in a.lhs.dimensions)
+        and len(a.lhs.dimensions) == 3
+    ]
+    assert len(resolved_assigns) == 1
+
+    assign = resolved_assigns[0]
+    dims = assign.lhs.dimensions
+    # dims[1] should be the original J; dims[2] must be different
+    assert dims[1] == 'j'
+    assert str(dims[2]).lower() != 'j', (
+        f"Third subscript of field is '{dims[2]}' which aliases the existing "
+        f"scalar subscript J; expected a distinct synthesized loop variable"
+    )
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_resolve_vector_notation_nested_loop_variable_shadowing(frontend):
+    """
+    Regression test: resolving a range inside an existing loop must not reuse
+    the active outer loop variable for the new inner loop, or scalar references
+    to that outer variable on the RHS become shadowed.
+    """
+    fcode = """
+subroutine test_nested_loop_shadowing(nouter, ninner, work, edge_hi, edge_lo)
+  implicit none
+  integer, intent(in) :: nouter, ninner
+  real, intent(inout) :: work(nouter, ninner)
+  real, intent(in) :: edge_hi(nouter, 0:ninner), edge_lo(nouter, 0:ninner)
+  integer :: i, j
+
+  do i = 1, nouter
+    do j = 1, ninner
+      work(i, 1:ninner) = max(edge_hi(i, j-1), edge_lo(i, 0:ninner-1))
+    enddo
+  enddo
+
+end subroutine test_nested_loop_shadowing
+    """.strip()
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+    resolve_vector_notation(routine)
+
+    loops = FindNodes(ir.Loop).visit(routine.body)
+    assert len(loops) == 3
+
+    outer_j_loop = next(loop for loop in loops if loop.variable == 'j' and loop.bounds == '1:ninner')
+    inner_loops = FindNodes(ir.Loop).visit(outer_j_loop.body)
+    assert len(inner_loops) == 1
+    assert str(inner_loops[0].variable).lower() != 'j'
+
+    assign = FindNodes(ir.Assignment).visit(inner_loops[0].body)[0]
+    lhs_dims = assign.lhs.dimensions
+    assert lhs_dims[1] == inner_loops[0].variable
+
+    rhs_text = str(assign.rhs).replace(' ', '').lower()
+    assert 'edge_hi(i,j-1)' in rhs_text
+    assert f'edge_lo(i,-1+{str(inner_loops[0].variable).lower()})' in rhs_text
--- a/loki/transformations/array_indexing/tests/test_vector_notation.py
+++ b/loki/transformations/array_indexing/tests/test_vector_notation.py
@@ -1958,6 +1958,40 @@ end module test_shifted_multi_mod
             f'[{label}] Expected one frac dim to have +1 offset, got: {frac_dims}'
 
 
+@pytest.mark.parametrize('frontend', available_frontends(
+    skip=[(OMNI, 'OMNI cannot parse unresolved external function some_func')]
+))
+def test_resolve_vector_notation_unresolved_inline_call_is_conservative(frontend):
+    """
+    Do not scalarize array arguments to unresolved inline calls, even if the
+    LHS uses vector notation.
+    """
+
+    fcode = """
+subroutine test_unresolved_inline(jl, n, lhs, arr)
+  implicit none
+  integer, intent(in) :: jl, n
+  real, intent(inout) :: lhs(n)
+  real, intent(in)    :: arr(n)
+
+  lhs(1:n) = some_func(arr(:), n)
+end subroutine test_unresolved_inline
+    """.strip()
+
+    routine = Subroutine.from_source(fcode, frontend=frontend)
+    resolve_vector_notation(routine)
+
+    assigns = FindNodes(ir.Assignment).visit(routine.body)
+    loops = FindNodes(ir.Loop).visit(routine.body)
+    inline_calls = FindInlineCalls().visit(routine.body)
+
+    assert len(assigns) == 1
+    assert not loops
+    assert assigns[0].lhs == 'lhs(1:n)'
+    assert not inline_calls
+    assert assigns[0].rhs == 'some_func(arr(:), n)'
+
+
 @pytest.mark.parametrize('frontend', available_frontends())
 def test_resolve_vector_notation_non_elemental_inline_call_is_conservative(frontend, tmp_path):
     """
--- a/loki/transformations/array_indexing/tests/test_vector_notation.py
+++ b/loki/transformations/array_indexing/tests/test_vector_notation.py
@@ -1536,16 +1536,21 @@ def test_resolve_vector_notation_no_implicit_rhs(frontend):
     Only assignments where all RHS ranges are explicit should be resolved.
     """
     fcode = """
-subroutine test_no_implicit_rhs(start, end, n, a, b, c)
+subroutine test_no_implicit_rhs(start, end, n, a, b, c, scalar)
   implicit none
   integer, intent(in) :: start, end, n
   real, intent(inout) :: a(n), b(n), c(n)
+  real, intent(in) :: scalar
 
   ! RHS uses bare ':' -- should NOT be resolved with resolve_implicit_rhs_ranges=False
   a(start:end) = b(:)
 
   ! Both sides use explicit range -- SHOULD be resolved
   a(start:end) = c(start:end)
+
+  ! Scalar RHS should still resolve
+  a(start:end) = 0.0
+  a(start:end) = scalar
 
 end subroutine test_no_implicit_rhs
     """.strip()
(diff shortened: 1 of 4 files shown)


loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -1680,6 +1838,437 @@ def test_resolve_vector_notation_shifted_range_multi_arg(frontend):
...
+    integer :: band(3)
+  end type config_type
+contains
+  subroutine test_vector_slice_bounds(config, arr, out)
+    implicit none
+    type(config_type), intent(in) :: config
+    real, intent(in) :: arr(5)
+    real, intent(out) :: out(5)
+
+    out(config%band:5) = arr(config%band:5)
```
awnawab: Is this even valid Fortran?

loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -1680,6 +1838,437 @@ def test_resolve_vector_notation_shifted_range_multi_arg(frontend):
...
+    """
+    from loki.logging import WARNING  # pylint: disable=import-outside-toplevel
+
+    fcode = """
+subroutine test_unresolved_scalar_bound_warning(arr, out)
+  implicit none
+  real, intent(in) :: arr(5)
+  real, intent(out) :: out(5)
+
+  out(1:dims%klon) = arr(1:dims%klon)
```
awnawab: If `dims%klon` were anything but a scalar, this wouldn't be valid fortran, so if we are slicing a dimension, the assumption of a scalar is a safe one.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -313,6 +313,18 @@ def _shape_to_range(s):
...
+    @staticmethod
+    def _shape_lower(s):
+        return s.lower if isinstance(s, sym.Range) else sym.IntLiteral(1)
+
+    @staticmethod
+    def _shape_upper(s):
+        return s.upper if isinstance(s, sym.Range) else s
+
+    @staticmethod
+    def _shape_step(s):
```
awnawab: A dimension can have a step, but can a shape even have a step? for example, I think this declaration would be invalid:
```
real :: a(1:5:2)
```

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -419,10 +441,320 @@ def _find_qualified_range_positions(dims, range_positions):
         Return ordinal indices into ``range_positions`` whose corresponding
         dimension is *not* a bare ``(:)`` (i.e. ``RangeIndex((None, None))``).
         """
+        return ResolveVectorNotationTransformer._find_resolvable_range_positions(dims, range_positions)
```
awnawab: Is there a point in making this static if we then have to access a method via the class name?

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -419,10 +441,320 @@ def _find_qualified_range_positions(dims, range_positions):
...
+
+    @staticmethod
+    def _has_explicit_range_bounds(dim):
+        return isinstance(dim, sym.RangeIndex) and dim.lower is not None and dim.upper is not None
+
+    @staticmethod
+    def _warn_on_unresolved_scalarizable_bound(bound):
+        if any(isinstance(var, sym.DeferredTypeSymbol) and var.parent is not None
```
awnawab: If the bound itself is a deferred type symbol, isn't that all we need to know? So is the presence of changing the situation at all? consider:
```
a(c) = ...
a(b%c) = ...
```

Whether `c` is deferredtype or `b%c` is deferredtype, the risk here would be the same.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -419,10 +441,320 @@ def _find_qualified_range_positions(dims, range_positions):
...
+        wrapped_body = body
+        for ivar, irange in index_range_map.items():
+            if isinstance(irange, sym.RangeIndex):
+                bounds = sym.LoopRange(irange.children)
+            else:
+                bounds = sym.LoopRange((sym.Literal(1), irange, sym.Literal(1)))
+            loop = ir.Loop(variable=ivar, body=as_tuple(wrapped_body), bounds=bounds)
+            wrapped_body = loop
+
+        if insert_comments and loop is not None:
```
awnawab: A consistent style here would make it more readable:
`if insert_comments and loop`
or
`if insert_comments is not None and loop is not None`

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -419,10 +441,320 @@ def _find_qualified_range_positions(dims, range_positions):
...
+            if scope is not None and hasattr(scope, 'subroutines'):
+                return next(
+                    (routine for routine in scope.subroutines
+                     if routine.name.lower() == call.name.lower()),
+                    None
+                )
+            return None
+
+        procedure_dtype = getattr(call.function.type, 'dtype', None)
+        if procedure_dtype is None or procedure_dtype is BasicType.DEFERRED:
```
awnawab: `BasicType.DEFERRED` was recently made falsey to facilitate these kinds of checks, so this could just be:
`if not procedure_dtype`.

awnawab (changes requested):
Thanks a lot for this great contribution @MichaelSt98 🙏 This makes the vector notation resolution much more powerful and will require fewer interventions in the source itself 👌 In fact, I'll hold my ifs-source PR until this is merged, as possibly now loki can handle some of  the cases I resolved manually.

Overall the changes look very good. I've just a few comments where the implementation can perhaps be tidied up a little, or where the warnings or tests can expose the underlying problem a little more clearly.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -693,12 +1024,21 @@ def visit_Assignment(self, stmt, **kwargs):  # pylint: disable=unused-argument
...
-        # --- Step 5: Map LHS ranges to loop index variables ---
+        # --- Step 6: Map LHS ranges to loop index variables ---
         resolved_lhs_ranges = [
             lhs_dims[lhs_range_positions[i]] for i in resolvable_dim_indices
         ]
+        # Collect loop variables already present as fixed scalar subscripts so
+        # that _map_ranges_to_indices does not alias them with a resolved range.
+        reserved_ivars = {
+            d for i, d in enumerate(lhs_dims)
+            if i not in lhs_range_positions and isinstance(d, sym.Scalar) and d in self.loop_map.values()
```
awnawab, line 1111: Since we haven't resolved any vector notation yet in this assignment, isn't `d in self.loop_map.values()` in itself sufficient? by definition, none of the loop vars will be in lhs_range_positions. It can't hurt to leave that in, but it makes that check more readable if we only check the contents of the loop_map imo.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -693,12 +1024,21 @@ def visit_Assignment(self, stmt, **kwargs):  # pylint: disable=unused-argument
...
+        }
+        # Keep existing scalar subscripts and active surrounding loop indices
+        # distinct from any new loop variable chosen for resolved ranges.
+        reserved_ivars.update(self.active_loop_vars)
         new_lhs_dims, index_range_map, synthesized_ivars = self._map_ranges_to_indices(
-            resolved_lhs_ranges, self.loop_map,
+            resolved_lhs_ranges, self.loop_map, reserved_ivars=reserved_ivars,
```
awnawab, line 1117: `map_unknown_ranges` is a little misleading, since we only resolve fully qualified ranges anyway. A better name might be `create_new_loop_indices`?

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -413,16 +432,325 @@ def _find_range_positions(dims):
...
+        if not resolvable_dim_indices:
+            return expr, {}, False
+
+        resolved_ranges = [
+            cond_dims_per_array[0][lhs_range_positions[i]] for i in resolvable_dim_indices
+        ]
+        cond_all_dims = cond_arrays[0].dimensions
+        reserved_ivars = {
+            d for i, d in enumerate(cond_all_dims)
+            if i not in lhs_range_positions and isinstance(d, sym.Scalar) and d in self.loop_map.values()
```
awnawab, line 729: Same comment here about loop indices, `i not in lhs_range_positions` is possibly redundant.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -831,59 +1166,28 @@ def visit_Conditional(self, o, **kwargs):
...
+            return masked
 
-        conditions = SubstituteExpressions(subst_map).visit(conditions)
+        condition, index_range_map, resolved = self._resolve_mask_expr(masked.conditions[0])
+        if not resolved:
+            return masked
 
         with dict_override(kwargs, {'create_loops': False}):
-            bodies = self.visit(masked.bodies, **kwargs)
+            body = self.visit(masked.bodies[0], **kwargs)
```
awnawab: Because the expression substitution is different for the condition and body, we will end up using different loop variables unless the dims to be resolved already happen to be in self.loop_map (in most cases these would be the horiztonal dim for example). For example:
```
where (a(1:n) ..)
 b(1:n)
end where
```
if we don't have any loops over n in that subrouine and it's not the dim over which we call `resolve_vector_dimension`, then n won't be in self.loop_map, and the transformation will substitute a different loop index in the mask and body assignment. We could fix this by adding the `MaskedStatement`s index_range_map to self.loop_map before visiting the bodies.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -413,16 +432,325 @@ def _find_range_positions(dims):
...
+            if routine is not None:
+                return routine.procedure_type.is_elemental
+            return False
+        if hasattr(procedure_dtype, 'is_elemental'):
+            if getattr(procedure_dtype, 'procedure', BasicType.DEFERRED) is BasicType.DEFERRED:
+                routine = resolve_routine_from_scope()
+                if routine is not None:
+                    return routine.procedure_type.is_elemental
+            return procedure_dtype.is_elemental
+        procedure_type = call.procedure_type
```
awnawab: This last check is redundant. call.procedure_type is the same as `call.function.type.dtype`, and we've already checked whether that's deffered or not and whether it's elemental.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -413,16 +432,325 @@ def _find_range_positions(dims):
...
+                return next(
+                    (routine for routine in scope.subroutines
+                     if routine.name.lower() == call.name.lower()),
+                    None
+                )
+            return None
+
+        procedure_dtype = getattr(call.function.type, 'dtype', None)
+        if not procedure_dtype:
+            routine = resolve_routine_from_scope()
```
awnawab, line 599: We'll end up here for an unenriched inline call, but this is currently uncovered, so it would be good to add a test for that.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -413,16 +432,325 @@ def _find_range_positions(dims):
...
+                if cls._expr_contains_vector_array(dim):
+                    return True
+        return False
+
+    def _has_unsafe_inline_call(self, expr):
+        """Return True if expr contains a non-scalarizable inline call with vector args."""
+        for call in FindInlineCalls().visit(expr):
+            if self._is_scalarizable_inline_call(call):
+                continue
+            if any(isinstance(arg, sym.Array) for arg in FindVariables(unique=False).visit(call)):
```
awnawab, line 675: That's an edge case we might actually hit, an inline call with an array arg. Could you please add a test for this?

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -413,16 +432,325 @@ def _find_range_positions(dims):
...
+    @staticmethod
+    def _build_loop_nest(index_range_map, body, insert_comments=False):
+        """Wrap ``body`` in loops for all ranges from ``index_range_map``."""
+        loop = None
+        wrapped_body = body
+        for ivar, irange in index_range_map.items():
+            if isinstance(irange, sym.RangeIndex):
+                bounds = sym.LoopRange(irange.children)
+            else:
+                bounds = sym.LoopRange((sym.Literal(1), irange, sym.Literal(1)))
```
awnawab: We always seem to construct a `RangeIndex` by the time we get here, so this branch can either be removed or replaced by an assert.

loki/transformations/array_indexing/vector_notation.py
```diff
@@ -413,16 +432,325 @@ def _find_range_positions(dims):
...
+            loop = ir.Loop(variable=ivar, body=as_tuple(wrapped_body), bounds=bounds)
+            wrapped_body = loop
+
+        if insert_comments and loop:
+            return (ir.Comment('! loki resolved vector notation'), loop)
+        if loop:
+            return (loop,)
+        return body
```
awnawab: I think we are guaranteed to always have `index_range_map` defined, so we'll always create a loop and this `return body` is redundant.

awnawab (changes requested):
Many thanks for improving this already very powerful utility 🙏 I'm sorry for doing this review in a piecemeal fashion, but this deserved a closer look than I had given it first time round.

On the whole the changes are great, I just found one potential bug and a couple of minor implementaiton details that can be tidied up.

loki/transformations/array_indexing/vector_notation.py
MichaelSt98: `lhs_range_positions` is needed to select only fixed scalar subscripts. Without it, a range dimension equal to a loop-map variable could incorrectly reserve that variable.

loki/transformations/array_indexing/vector_notation.py
MichaelSt98: See other comment.

loki/transformations/array_indexing/vector_notation.py
MichaelSt98: Should be covered by: `test_resolve_vector_notation_non_elemental_inline_call_is_conservative`

loki/transformations/array_indexing/vector_notation.py
awnawab: Got it, thanks for the clarification.

awnawab (approved):
Thanks a lot @MichaelSt98 for addressing all the feedback 🙏 This is now G2G ✅

loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -804,6 +805,95 @@ def test_resolve_vector_notation_ifs_patterns(frontend):
...
+    first_assign = FindNodes(ir.Assignment).visit(loops[0].body)[0]
+    assert first_assign.lhs == 'ssa_total(jl)'
+    assert first_assign.rhs == '(ssa(jl) + cloud(jl))/od_total(jl)'
+
+    second_assign = FindNodes(ir.Assignment).visit(loops[1].body)[0]
+    assert second_assign.lhs == 'g_total(jl)'
+    assert second_assign.rhs == '(g(jl) + cloud(jl))/ssa_total(jl)'
+
+    second_inner_cond = next(c for c in FindNodes(ir.Conditional).visit(loops[1].body))
+    second_cond = str(second_inner_cond.condition).replace(' ', '').lower()
```
mlange05: String fuffing should not be needed here - does native expression comparison not work for comparison operators?

loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -1467,37 +1562,47 @@ def test_resolve_vector_notation_no_implicit_rhs(frontend):
...
+        "Only the assignment with bare RHS ':' should remain unresolved"
+    assert 'b(:)' in str(unresolved[0].rhs), \
+        f"Unexpected unresolved assignment: {unresolved[0]}"
+
+    # The resolved loops should cover explicit-range and scalar RHS assignments.
+    loop_assigns = [FindNodes(ir.Assignment).visit(loop.body)[0] for loop in loops]
+    rhs_texts = {str(assign.rhs).replace(' ', '') for assign in loop_assigns}
+    assert any('c(' in rhs for rhs in rhs_texts), \
+        f"Expected resolved c(...) assignment, got: {rhs_texts}"
+    assert '0.0' in rhs_texts, f"Expected scalar literal assignment, got: {rhs_texts}"
```
mlange05, line 1595: Direct symbol-comparison should definitely work here.

loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -1507,21 +1612,95 @@ def test_resolve_vector_notation_no_implicit_rhs(frontend):
...
+        f"Unexpected unresolved assignment: {unresolved2[0]}"
+
+    # The resolved loop bodies should reference explicit-range and scalar RHS cases.
+    loop_assigns2 = [FindNodes(ir.Assignment).visit(loop.body)[0] for loop in loops2]
+    rhs_texts2 = {str(assign.rhs).replace(' ', '') for assign in loop_assigns2}
+    assert any('c(' in rhs for rhs in rhs_texts2), \
+        f"Expected resolved c(...) assignment, got: {rhs_texts2}"
+    assert '0.0' in rhs_texts2, f"Expected scalar literal assignment, got: {rhs_texts2}"
+    assert 'scalar' in rhs_texts2, f"Expected scalar variable assignment, got: {rhs_texts2}"
+    assert all(str(assign.lhs) == 'a(1, jl)' for assign in loop_assigns2)
```
mlange05: Please prefer direct comparison over string-casting

loki/transformations/array_indexing/tests/test_vector_notation.py
```diff
@@ -1507,21 +1612,95 @@ def test_resolve_vector_notation_no_implicit_rhs(frontend):
...
+
+
+@pytest.mark.parametrize('frontend', available_frontends())
+def test_resolve_masked_statements_scalar_rhs_with_no_implicit_rhs(frontend):
```
mlange05: Isn't this the same test code as in `test_resolve_masked_statements_fallback_unresolved_rhs`? Could these two be merged and the differing flag-output pairs be parametrised?

loki/transformations/single_column/tests/test_scc.py
```diff
@@ -204,6 +204,151 @@ def test_scc_base_resolve_vector_notation_config(frontend, horizontal):
...
+    assert 'jl' in kernel2.variables
+    assert not FindNodes(MaskedStatement).visit(kernel2.body)
+    loops = FindNodes(Loop).visit(kernel2.body)
+    assert len(loops) == 1
+    assert loops[0].variable == 'jl'
+    assert loops[0].bounds == 'start:end'
+
+    conds = FindNodes(Conditional).visit(kernel2.body)
+    assert len(conds) == 1
+    assert conds[0].condition == 'q(jl) > 0.'
```
mlange05, line 263: Should we also test the correct `Assignment` adjustment here?

loki/transformations/single_column/tests/test_scc.py
```diff
@@ -204,6 +204,151 @@ def test_scc_base_resolve_vector_notation_config(frontend, horizontal):
...
+    fcode_driver = """
+  SUBROUTINE drive_scalar_rhs(start, end, nlon, q, scalar)
+    INTEGER, INTENT(IN) :: start, end, nlon
+    REAL, INTENT(INOUT) :: q(nlon)
+    REAL, INTENT(IN) :: scalar
+
+    q(start:end) = 0.
+    q(start:end) = scalar
+  END SUBROUTINE drive_scalar_rhs
+"""
```
mlange05, line 287: Is this the same code block as above? Could we use test parametrisation instead here?

mlange05 (commented):
The changes in this PR look great and the increased test coverage is very much appreciate. I've flagged a bunch of cosmetics and LLM-isms that would be great to address, as the increased coverage balloons the associated test quite a bit. I'll see if this can be improved before merging, but content-wise it looks good.

mlange05 (approved):
With the cosmetic changes in this is now GTG for me. A very nice contribution overall - many thanks!

PR #715, opened 2026-07-21
constant propagation fixes/additions

MichaelSt98:
### Description

Some changes to constant propagation to make it work for CLOUDSC.

### Contributor Declaration


loki/expression/symbolic.py
```diff
@@ -434,7 +434,10 @@ def div_literals(expr, fp_arithmetic=False):
     if isinstance(expr.numerator, sym.FloatLiteral) or isinstance(expr.denominator, sym.FloatLiteral):
         if not fp_arithmetic:
             return expr
-        return sym.Literal(float(expr.numerator.value) / float(expr.denominator.value))
+        literal_types = (sym.IntLiteral, sym.FloatLiteral)
+        if isinstance(expr.numerator, literal_types) and isinstance(expr.denominator, literal_types):
```
mlange05: Why are we not adjusting the outer `isinstance(numerator/denominator. sym.FloatLiteral)`? Seems odd to specialise the check inside the outer check...

loki/transformations/tests/test_constant_propagation.py
```diff
@@ -82,6 +82,25 @@ def test_constant_propagation_analysis_dynamic_array_invalidation():
...
+  integer :: a = 1
+
+  a = a + state%value
+end subroutine const_prop_symbols_without_initial
+    """.strip()
+    routine = Subroutine.from_source(fcode)
+
+    declarations_map = ConstantPropagationTransformer().generate_declarations_map(routine)
+
+    assert declarations_map[('a', ())] == sym.IntLiteral(1)
```
mlange05, line 101: How is this different to the test above in lines 22 and 31 in `test_constant_propagation_analysis_declarations_map`?

loki/expression/mappers.py
```diff
@@ -172,7 +172,7 @@ def map_quotient(self, expr, enclosing_prec, *args, **kwargs):
         # quotients for the numerator
         kwargs['force_parens_around'] = (pmbl.FloorDiv, pmbl.Remainder)
         numerator = self.rec_with_force_parens_around(expr.numerator, PREC_PRODUCT, *args, **kwargs)
-        kwargs['force_parens_around'] = self.multiplicative_primitives
+        kwargs['force_parens_around'] = self.multiplicative_primitives + (pmbl.Product, pmbl.Quotient)
```
mlange05: I believe both are already part of `self.multiplicative_primitives` and do not need to be added again:
```
In [3]: pmbl.Product in self.multiplicative_primitives
Out[3]: True

In [4]: pmbl.Quotient in self.multiplicative_primitives
Out[4]: True
```

mlange05 (changes requested):
Looks good in general - I've picked out a few oddities, but nothing drastic, so should be GTG once addressed.

loki/transformations/constant_propagation.py
```diff
@@ -196,10 +196,13 @@ def visit_Loop(self, o, **kwargs):
...
-                    assign_kwargs = dict(kwargs)
-                    assign_kwargs['constants_map'] = loop_constants_map
-                    self.visit_Assignment(assign, **assign_kwargs)
+                if set(FindVariables().visit(assign.rhs)).intersection(lhs_vars):
+                    invalidate_constants_map(assign.lhs, loop_constants_map)
+                    continue
+
+                assign_kwargs = dict(kwargs)
+                assign_kwargs['constants_map'] = loop_constants_map
+                self.visit_Assignment(assign, **assign_kwargs)
```
Andrew-Beggs-ECMWF, line 205: I was having a think about this, and while it definitely fixes a bug, I think there's another subtlety with the RHS.

I think you should also visit the RHS, as well as invalidating the constant map for the LHS, as you might have side effects. E.g. a function that mutates its inputs.

Fortran saves us from that specific example with subroutines vs functions, but only if people follow the standards ; )

PR #718, opened 2026-08-24
Transformations: Offload nested driver region and data offload trafo bugfix

mlange05:
### Description

This PR exposes and fixes a bug in the `DataOffloadDeepCopy` analysis/transformation, where nested Loki regions can cause erroneous lookup failures. The fundamental problem is that we use loops as dict indices when passing info from analysis to transformation, but those loops will change their internal hash if `PragmaRegion` objects are inserted (the trafo uses `with_pragma_regions_attached`, and IR objects derive their hash recursively). To prevent this, a new utility `DriverLoopRegion` that enumerates loops-per-region and associated tools `driver_loop_region_attached`, which makes the traversal and lookup of the objects safe.

I've also tried to roll this out across other transformations, hoping to eventually unify the access to the "driver loop" concept across all trafos - but I only regard this as a first step. Comments on further generalisation are welcome. 

### Contributor Declaration

### DIFF
--- a/loki/transformations/data_offload/offload_deepcopy.py
+++ b/loki/transformations/data_offload/offload_deepcopy.py
@@ -21,8 +21,7 @@ from loki.batch import Transformation, TypeDefItem, ProcedureItem
 from loki.expression import symbols as sym
 from loki.ir import (
         nodes as ir, FindNodes, SubstituteExpressions, Transformer,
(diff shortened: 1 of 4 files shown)


awnawab (approved):
A very neat and clever fix for what was a brittle design, many thanks! I've tested this in an ifs gpu offload build and I can confirm it works as intended. G2G ✅

PR #722, opened 2026-08-28
Only test scc pipelines in ecwam regression test

awnawab:
A small fix to only test the scc pipelines in the ecwam regression test, given that the idem pipelines no longer had any really use in ecwam and have therefore been removed.


mlange05 (approved):
Looks great, and should address regression test failures. Many thanks for the prompt fix. 🙏

PR #723, opened 2026-08-28
Lint: Removal of in-situ linting rules

mlange05:
### Description

This PR removes the `lint_rules` sub-project in favour of the externalised linter rule repository [ial-lint](https://github.com/ecmwf-ifs/ial-lint). It also adds an ial-lint regression test setup that executes the IAL-lint test base. 

TO make this work, two small changes are needed to the rule detection mechanism:
- No longer assumer the `lint_rules` prefix in the target module
- Using `.__dir__()` plus a filter by class-type instead of `.__all__` on the target module. This is to enable sub-package modules that do not explicitly propagate `__all__` to the parent package if individual rules are kept in separate files.

### Contributor Declaration


.github/workflows/lint_tests.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -U .[tests]
+          python -m pip install --no-deps ${{ github.workspace }}/ial-lint
```
reuterbal, line 50: To verify that the tests run fine purely with a pip-installed ial-lint (and thus validating there is no hidden dependency on an ial-lint worktree), you could consider skipping the checkout above and simply installing it here as

```suggestion
          python -m pip install --no-deps git+https://github.com/ecmwf-ifs/ial-lint@main
```

reuterbal (changes requested):
Thanks, this is an exciting milestone!

Just a small comment to further decouple Loki and ial-lint in the testing, but this is really an implementation detail to treat ial-lint more like an external dependency.

The more important update should probably be to the docs, which are now outdated. Technically, this could have been done before the removal already, but with this it would definitely be incorrect: https://github.com/ecmwf-ifs/loki/blob/main/docs/source/loki_lint.rst

.github/workflows/lint_tests.yml
mlange05: I'm not sure this would work. `ial_lint` uses a tagged release by default - so would this not fallback to the tagged release, rather than using the local checkout that is being tested for compatibility with the upstream package?

mlange05:
@reuterbal Many thanks for the review and apologies for letting this hang again. I've now updated the documentation accordingly (and slightly restructured it). Please have a look and see if this is ok.

reuterbal (approved):
Thanks, this looks very good to me now!

.github/workflows/lint_tests.yml
mlange05: Sorry, misunderstood before - just pushed a fix for this; will merge once green.

awnawab (approved):
loki-lint is shaping up really nicely, thanks a lot 🙏 can't wait to start rolling this out across some of our repos.

